PR #10, opened 2013-05-04
Fix malformed JSON

martijnvermaat:
Keys are strings and thus must be quoted, as per JSON spec.


steveklabnik:
Thanks! So easy to make this mistake.

Issue #7, opened 2013-05-04
Errors?


steveklabnik:
We should address how errors are returned.

This replaces https://github.com/rails-api/active_model_serializers/issues/214 and https://github.com/rails-api/active_model_serializers/issues/199

wrightling:
I've been scouring the internet a bit as I work on my first JSON API (using rails-api).  Two references I came across return errors in some varation of { error: "message" } or { errors: ["message1", "message2"] }.  Is there a more standard way?

Referencing sample app from a rails 3 book sample app [ticketee](https://github.com/rails3book/ticketee/blob/master/app/controllers/api/v2/projects_controller.rb) and a [Travis Jeffery blog post](http://travisjeffery.com/b/2012/04/rendering-errors-in-json-with-rails/).

steveklabnik:
I feel like there is some sort of interent-draft for an errors spec, but I can't find it at the moment.

wrightling:
Here are a few more examples:
- [github API](http://developer.github.com/v3/#client-errors)
- [google JSON style guide](http://google-styleguide.googlecode.com/svn/trunk/jsoncstyleguide.xml#error)

Looks like there is nothing if not variety in how folks are returning errors.  Personally, I'm not sure I see a point in Google's recommendation to include the code in the response body - seems like duplication.

bcardarella:
We've discussed this in the Ember CF quite a bit. The errors object should be a key/value of arrays. So `{ 'errors': { 'first_name': [] } }`

What is not as clear is if full messages should be retured in the array or references to messages that will be determined by the client. The discussion seemed to be in favor of returning message keys that can then be rendered by the client's `i18n` message processor.

wrightling:
The format you suggest would be pretty darn easy to support in rails or rails-api as well, as shown in [this gist](https://gist.github.com/wrwright/5519521)

trek:
Please, for all that is good in the world, provide as detailed data as possible. 

Working with APIs that return _just_ a string (when I've specifically said "I'll only accept `application/json') or only string messages inside JSON {`errors: ["good luck figuring out what property is wrong", "hope you like regexps"] }` when the design calls for highlighting where on the page the user erred has got to be on of the most frustrating experiences as UI engineering.

For maximum usefulness, I usually at minimum want something like:

``` json
{
  "errors": {
    "firstName": ["cannot be blank", "must be 10+ characters"],
    "lastName": ["cannot contain symbols"]
  }
}
```

[...]

PR #11, opened 2013-05-05
Error response section for ID-based JSON API

wrightling:
Relates to json-api/json-api#7

One question I have here is whether, for compound documents, we need to allow for multiple error objects, and if one part of a compound document can contain normal results while other parts contain errors.

I'm happy to revise this or scrap it, per the collective wisdom of you more experienced API developers.

### DIFF
--- a/source/index.md
+++ b/source/index.md
@@ -204,6 +204,19 @@ The related documents are provided as an additional top-level document or docume
 
 The linkage between the key under `"rels"` and the top-level keys is hardcoded into the client.
 
+### Error Responses
+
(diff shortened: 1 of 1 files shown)


wycats:
It's important to cover how error responses for compound requests should work. I think compound requests are pretty vague at the moment, but let's take the opportunity to discuss how we expect it to work.

bcardarella:
I believe strongly that the errors should not return full messages but keys that the client application can use to look up the error message

wycats:
This seems like a dup of #7.

PR #15, opened 2013-05-05
cleanup URL-Based JSON API relationships example

soulware:
use comments?ids=1,2,3,4 rather than comments/1,2,3,4
or alternatively posts/1/comments

### DIFF
--- a/source/index.md
+++ b/source/index.md
@@ -276,7 +276,8 @@ The value of the `"rels"` key is a JSON object that represents related documents
     "title": "Rails is Omakase",
     "rels": {
       "author": "http://example.com/people/1",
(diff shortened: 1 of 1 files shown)


wycats:
This example is showing how URL templates work. The URL templates provided with the example is:

``` js
  "rels": {
  }
```

This means that the URL for the comments is `http://example.com/comments/1,2,3,4`. It's intentionally different from the default URL template for the ID-based approach.

soulware:
Ah. I had assumed the same rules applied to both ID and URL style and that the server must respond to ?id=x,y,z style requests.
It makes sense that the URL style would be more flexible.

PR #17, opened 2013-05-05
fix copypasta

indirect:
pretty sure this should be photos and not posts


steveklabnik:
thanks. there are some others on this page too...

Issue #4, opened 2013-05-04
allow PUT


steveklabnik:
```
> On creation, if I accept the pain of generating an ID on the client and can construct the URI using the template, why can't I use PUT instead of POST?

Good point. That seems fine.
```

https://news.ycombinator.com/item?id=5655592

lausobo:
:+1: for idempotency

b:
And a 201 and a Location header MUST be returned, per HTTP/1.1. Thanks.

lausobo:
@b yep.

Note that, by allowing PUT, creation and update of a resource will be available via the same verb and URL

`PUT /post/asda-asdasd-asdasd-asdasdasd` could either create a post of replace an existing post (assuming that the semantics of the service allow both operations). In that particular case it should only return a 201 it the requests resulted the creation of the post but not if it was an update.

b:
Right, just like HTTP/1.1 says.

lausobo:
:relieved:

Issue #2, opened 2013-05-04
Consider relaxing some MUSTs


steveklabnik:
One from HN:

```
That's great to hear. One particular example which I found needlessly strict is this:

    The request MUST contain a Content-Type header whose value is
    application/json.  It MUST also include application/json as the only or
    highest quality factor.

It makes sense for fully compliant implementation to have those headers, but they way I understand MUST here is that a server would reject any request without them.
```

https://news.ycombinator.com/item?id=5656097

b:
Every MUST is one less opportunity for implementations to be compliant and incompatible. Please keep the MUST.

reinh:
The MUST will not be a barrier to compliant implementations because they will do what the spec says. It will also not be a barrier to non-compliant implementations because they _won't do_ what the spec says.

:+1: for keeping the MUST

mamund:
note that each MUST is a promise _forever_. no future mod, update, or bug fix would be allowed if it ran counter to an existing MUST. aim for SHOULD and MAY, limit MUST to bare minimum.  adding MUSTs puts limits on an unknown future - that's rarely a smart idea.

to encourage implementations to follow you specs closely it is common to use the following in your documentation:

_An implementation is not compliant if it fails to satisfy one or more of the MUST or REQUIRED level requirements. An implementation that satisfies all the MUST or REQUIRED level and all the SHOULD level requirements is said to be "unconditionally compliant"; one that satisfies all the MUST level requirements but not all the SHOULD level requirements is said to be "conditionally compliant."_

b:
"adding MUSTs puts limits on an unknown future - that's rarely a smart idea."

No, adding MUSTs make the lives of implementors easier and eliminates opportunities for incompatible implementations. That unknown future is properly handled in a protocol through versioning and extension mechanisms that allow _entirely optional_ additional behavior, not through a proliferation of MAYs.

For example, look at how few MAYs are in RFC2616 and how most of them relate to client behavior in the face of MUST and SHOULD behaviors on the server. That is the result of careful protocol design and painful experience. Postel's Law in action: http://en.wikipedia.org/wiki/Robustness_principle

mamund:
My count of RFC2119 words in RFC2616 (including initial description of the words):
- MUST/SHALL/REQUIRED = 353
- SHOULD/RECOMMENDED = 256
- MAY/OPTIONAL = 187

[...]

PR #25, opened 2013-05-06
Promote '201 Created' HTTP status

lukfugl:
The HTTP spec includes the '201 Created' HTTP status and its
accompanying Location header specifically for situation such as
described in the "Creating" section of jsonapi.org.
- Retains the existing '204 No Content' and '200 OK' responses for
  compatibility with existing frameworks, but adds the '201' response
  and promote it as the preferred response.
- Restricts the availability of the '204 No Content' response to
  situations where the client can accurately determine the server
  location of the newly created resource.
- Fixes some 'posts' vs. 'photos' typos in the examples.

### DIFF
--- a/source/write.md
+++ b/source/write.md
@@ -110,31 +110,82 @@ Accept: application/json
 }
 ```
 
-#### 204 Responses
+### Responses
 
-A server **MAY** respond to a `POST` request with a `204 No Content`
-response. If it does so, they client **MUST** assume that the server has
+A server **MUST** respond to a successful `POST` request with one of the
+following HTTP statuses:
+
+* `201 Created` (preferred)
+* `204 No Content`
+* `200 OK`
+
+The `204 No Content` and `200 OK` statuses are **NOT RECOMMENDED**,
+preferring the `201 Created` status, but are included for compatibility
+with a large number of existing server frameworks.
+
+#### `201 Created`
+
+A server **SHOULD** respond to a `POST` request with the `201 Created`
+status. The response **MUST** include a `Location` HTTP response header
+containing the server location of the newly created document. The server
+**MAY** include a response body.
+
+If the response includes a response body, the response body **MUST** be
+a valid response to a JSON API `GET` request of the URI in the
+`Location` header.
+
+If the response includes a response body, it **MAY** include an `href`
+key in the attributes section; if the server is using the URL-based JSON
+API, this `href` attribute is **REQUIRED**. If present, the value of the
+`href` attribute **MUST** match the URI in the `Location` header.
+
+Example:
+
+```text
+HTTP/1.1 201 Created
+Location: http://example.com/photos/12
+Content-Type: application/json
+
+{
+  "photos": {
+    "id": "550e8400-e29b-41d4-a716-446655440000",
+    "href": "http://example.com/photos/12",
+    "title": "Ember Hamster",
+    "src": "http://example.com/images/productivity.png"
+  }
+}
+```
+
+#### `204 No Content`
+
+A server **MAY** respond to a `POST` request with the `204 No Content`
+status. If it does so, the client **MUST** assume that the server has
 successfully created the document, and accepted all of the attributes as
 is.
 
-Note that if the server is using client-generated IDs, it is possible
-for the client to determine the URL for the newly generated record if
-the collection response included a URL template:
+The server **MUST NOT** respond with `204 No Content` unless it both
+requests client-generated IDs and publishes a URI template in the JSON
+API `GET` response of the collection.
+
+The client **SHOULD** determine the URL of the newly created document by
+combining the URI template of the collection with the `id` attribute
+included in the `POST` request.
+
+For example, given the following response to a JSON API `GET` on the
+collection:
 
 ```text
-GET /photos
-
 HTTP/1.1 200 OK
 Content-Type: application/json
 
 {
-  "posts": [{
+  "photos": [{
     "id": "550e8400-e29b-41d4-a716-446655440000",
     "title": "Mustaches on a Stick",
-    "href": "http://example.com/images/mustaches.png"
+    "src": "http://example.com/images/mustaches.png"
   }],
   "rels": {
-    "posts": "http://example.com/posts/{posts.id}"
+    "photos": "http://example.com/photos/{photos.id}"
   },
   "meta": {
     "client-ids": true
@@ -142,27 +193,32 @@ Content-Type: application/json
 }
 ```
 
-If the client generated a UUID of
-"550e8400-e29b-41d4-a716-446655440001", and it receives a 204 response,
-it knows that the URL for the newly created document is
-`http://example.com/posts/550e8400-e29b-41d4-a716-446655440001`.
+If the client submits a `POST` request to the collection with a UUID of
+`550e8400-e29b-41d4-a716-446655440001` and receives `204 No Content`,
+the expected URL for the newly created document is
+`http://example.com/photos/550e8400-e29b-41d4-a716-446655440001`.
 
-#### 200 Responses
+#### `200 OK`
 
-A server **MAY** respond to a `POST` request with a `200 OK` response.
-If it does so, it **MUST** include a response body that is a valid
-response to a JSON API `GET` request.
+A server **MAY** respond to a `POST` request with the `200 OK` status.
+If it does so, it **MUST** include a response body.
 
-If the server is using the URL-based JSON API, it **MUST** include a
-`href` key in the attributes section that is the server location of the
-newly created document.
+The response body **MUST** be a valid response to a JSON API `GET`
+request of the server location of the newly created document.
+
+The response body **MAY** include an `href` key in the attributes
+section; if the server is using the URL-based JSON API, this `href`
+attribute is **REQUIRED**. If present, the value of the `href` attribute
+**MUST** be the server location of the newly created document.
+
+Example:
 
 ```text
 HTTP/1.1 200 OK
 Content-Type: application/json
 
 {
-  "posts": {
+  "photos": {
     "id": "550e8400-e29b-41d4-a716-446655440000",
     "href": "http://example.com/photos/12",
     "title": "Ember Hamster",
--- a/source/write.md
+++ b/source/write.md
@@ -112,32 +112,19 @@ Accept: application/json
 
 ### Responses
 
-A server **MUST** respond to a successful `POST` request with one of the
-following HTTP statuses:
-
-* `201 Created` (preferred)
-* `204 No Content`
-* `200 OK`
-
-The `204 No Content` and `200 OK` statuses are **NOT RECOMMENDED**,
-preferring the `201 Created` status, but are included for compatibility
-with a large number of existing server frameworks.
-
 #### `201 Created`
 
-A server **SHOULD** respond to a `POST` request with the `201 Created`
-status. The response **MUST** include a `Location` HTTP response header
-containing the server location of the newly created document. The server
-**MAY** include a response body.
+A server **SHOULD** respond to a `POST` request with the
+[`201 Created`][2] status, including a request body and `Location` header 
+(see [section 9.5][3] of the HTTP/1.1 specification).
 
-If the response includes a response body, the response body **MUST** be
-a valid response to a JSON API `GET` request of the URI in the
-`Location` header.
+[2]: http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.2.2
+[3]: http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.5
 
-If the response includes a response body, it **MAY** include an `href`
-key in the attributes section; if the server is using the URL-based JSON
-API, this `href` attribute is **REQUIRED**. If present, the value of the
-`href` attribute **MUST** match the URI in the `Location` header.
+The response body **MAY** include an `href` key in the attributes
+section; if the server is using the URL-based JSON API, this `href`
+attribute is **REQUIRED**. If present, the value of the `href` attribute
+**MUST** match the URI in the `Location` header.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


mahmoudimus:
+1, why would we deviate from this behavior? POST should always return a 201.

source/write.md
```diff
@@ -110,31 +110,82 @@ Accept: application/json
 }
 ```
 
-#### 204 Responses
+### Responses
 
-A server **MAY** respond to a `POST` request with a `204 No Content`
-response. If it does so, they client **MUST** assume that the server has
+A server **MUST** respond to a successful `POST` request with one of the
```
steveklabnik: I always give pause with `MUST`s. Why restrict another 2xx response?

source/write.md
```diff
@@ -110,31 +110,82 @@ Accept: application/json
...
+following HTTP statuses:
+
+* `201 Created` (preferred)
+* `204 No Content`
+* `200 OK`
+
+The `204 No Content` and `200 OK` statuses are **NOT RECOMMENDED**,
+preferring the `201 Created` status, but are included for compatibility
+with a large number of existing server frameworks.
```
steveklabnik: I'm not sure that I can get behind this either.

source/write.md
```diff
@@ -110,31 +110,82 @@ Accept: application/json
...
+
+#### `201 Created`
+
+A server **SHOULD** respond to a `POST` request with the `201 Created`
+status. The response **MUST** include a `Location` HTTP response header
+containing the server location of the newly created document. The server
+**MAY** include a response body.
```
steveklabnik: I'm not sure that we need this MUST or MAY. Given what HTTP says about the 201:

```
 The request has been fulfilled and resulted in a new resource being created.
The newly created resource can be referenced by the URI(s) returned in the
entity of the response, with the most specific URI for the resource given by a
Location header field. The response SHOULD include an entity containing a list
of resource characteristics and location(s) from which the user or user agent
can choose the one most appropriate. The entity format is specified by the
media type given in the Content-Type header field. The origin server MUST
create the resource before returning the 201 status code. If the action cannot
be carried out immediately, the server SHOULD respond with 202 (Accepted)
response instead.

A 201 response MAY contain an ETag response header field indicating the current
value of the entity tag for the requested variant just created, see section
14.19. 

```

TL;DR:, they're both already covered.

source/write.md
```diff
@@ -110,31 +110,82 @@ Accept: application/json
...
+
+If the response includes a response body, the response body **MUST** be
+a valid response to a JSON API `GET` request of the URI in the
+`Location` header.
```
steveklabnik: Same here as above.

source/write.md
```diff
@@ -142,27 +193,32 @@ Content-Type: application/json
...
+If the client submits a `POST` request to the collection with a UUID of
+`550e8400-e29b-41d4-a716-446655440001` and receives `204 No Content`,
+the expected URL for the newly created document is
+`http://example.com/photos/550e8400-e29b-41d4-a716-446655440001`.
+
+#### `200 OK`
 
-#### 200 Responses
+A server **MAY** respond to a `POST` request with the `200 OK` status.
+If it does so, it **MUST** include a response body.
```
steveklabnik: This also seems overly restrictive, but I'm less sure about it.

steveklabnik:
Added some commentary. I love me some `201`s, and I agree that it should be covered in some form, but I feel that parts of this go too far.

lukfugl:
Understood, will be looking over your feedback and replying.

source/write.md
lukfugl: My reasoning behind **MUST** was primarily that none of the other documented `2xx` responses seemed applicable to an API "create document" action (I'll admit overlooking `202 Accepted`, which should be allowed).
- `203 Non-Authoritative Information` doesn't seem appropriate as we're asking this server to create the document; we'd expect it to then be the authoritative source (at least from the client's perspective) about that document.
- `205 Reset Content` is not intended (as I read it) for API traffic.
- `206 Partial Content` is intended for partial `GET` requests, which is incompatible with our `POST` (or `PUT`) assumptions.

But you're right, if the server sees a reason to respond with `203 Non-Authoritative Information` or `205 Reset Content` or even a custom `2xx` status, it should be allowed. We just need to make specific that the default conforming-client behavior should be to treat them the same as a `200 OK`. The client **MAY** then take additional action based on what the server may publish about its intentions in using those other `2xx` statuses.

I'll reword this as a **SHOULD** and clarify the default handling of alternate `2xx` responses. I'll also add a section about `202 Accepted` handling.

steveklabnik: :+1: 

Also, if HTTP gets extended later, it locks out any future codes which may make sense.

source/write.md
lukfugl: Understood. I took a pretty strong stand there, intentionally, to spark discussion. :) I was really hoping there'd be a RFC word with strength between **SHOULD** and **MAY** but there's not. I'd be okay with rephrasing the preference without normative strength, if necessary.

steveklabnik: Yeah, I'd phrase this more as "201 created is a more common response, and is likely to be used most often" or something with that informative sentiment rather than the normative one. :)

source/write.md
lukfugl: This was actually adapted from the original line 253:

> A server **MAY** respond to a `POST` request with a `200 OK` response. If it does so, it **MUST** include a response body that is a valid response to a JSON API `GET` request.

I agree that it seems overly restrictive, and only put it in to minimize change. I'd be happy with a **SHOULD** here.

lukfugl: Or actually, were you referring to "already covered"? Because yeah, that makes sense since this is the `Location` header in a `201 Created` we're talking about. Will reword to be informative rather than normative.

source/write.md
lukfugl: Yup, I added this with intent to be informative, made normative to match HTTP, but should just let HTTP cover the normative aspects.

source/write.md
lukfugl: Just a carry through of the **MUST** on original line 153 below. I'm fine with relaxing it.

source/write.md
steveklabnik: Roger. One of the tricky bits here is that I think a lot of new developers don't know how to use specs together, and so they'll be some education involved. See questions about pagination where people dont know the link relations already exist.

Maybe it'd be better to explicitly call out the section of the HTTP sepc, rather than duplicate it?

source/write.md
steveklabnik: Yeah, already covered was what I meant.

source/write.md
lukfugl: Sounds good.

source/write.md
steveklabnik: Yeah, I'd like to review all MUSTs to make sure we're not being overly restrictive. :)

source/write.md
wycats: Agreed :)

b:
There are actually 2 problems:

1) HTTP 1.1 is clear that documents created with POST SHOULD (not MAY) return a 201 and a Location header. 200 or 204 should only be returned when a document was NOT created.
2) If the client knows the id of the new document, as is the case in the example, then the proper verb is PUT and the proper response is a 201 (MUST) and a Location header.

I don't understand the desire to keep the specification sloppy. That is a common recipe for incompatible but compliant implementations. Keep it at least as strict as, and aligned with, HTTP 1.1.

steveklabnik:
Thanks @b,

First of all, aligned with HTTP 1.1 is 100% correct. I don't want to deviate from it, that'd be bad. Hence the draft status of this :)

Secondly, it's not so much 'sloppy' as it is "let's not lock things down until we know for sure we want it to be that way, always and forever. Too many MUSTs does make it easy to comply, but also leaves you no wiggle room in the future.

b:
We are talking about HTTP semantics. These are already defined. You don't need wiggle room unless you intend to deviate from the HTTP spec.

steveklabnik:
Absolutely, I'm making a general statement about the media type, not just this specific case.

b:
Probably want to respond over here then? https://github.com/json-api/json-api/issues/2

lukfugl:
Incorporated @steveklabnik's feedback tempered by @b's pointing out [HTTP/1.1 section 9.5](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.5)

b:
Again, please just do what it says in HTTP/1.1, not something that warms the cockles of our hearts while we all sit around the campfire under the avocado tree singing Kumbaya.

lukfugl:
@b: can you elaborate on what portion of the current pull request is contrary to the HTTP/1.1 spec. I tried to make it conformant.

Perhaps your issue is in the normative **MAY** preceding the descriptions of `200 OK` and `204 No Content`? In my mind those **MAY**s are subservient to the existing **SHOULD** on the `201 Created` status. Basically I'm trying to phrase it such that it comes across "you shouldn't do this, but if you do, it should look like this". I think removing the **MAY**s and making those sections purely descriptive robs the other prescriptions in those sections of their weight, which only loosens the conformance restrictions on those frameworks that choose to continue using `200 OK` or `204 No Content` responses.

b:
"If a new resource is created, the origin server MUST inform the user agent via the 201 (Created) response. If an existing resource is modified, either the 200 (OK) or 204 (No Content) response codes SHOULD be sent to indicate successful completion of the request. If the resource could not be created or modified with the Request-URI, an appropriate error response SHOULD be given that reflects the nature of the problem. The recipient of the entity MUST NOT ignore any Content-\* (e.g. Content-Range) headers that it does not understand or implement and MUST return a 501 (Not Implemented) response in such cases."

The 201 on creation is not SHOULD. It is MUST. The 200 or 204 codes SHOULD, not MAY, be returned if this is NOT creation and the request succeeds - 200 when a body is returned, 204 when there is no body. You aren't "relaxing" anything. You are not complying with HTTP/1.1.

In all cases, when JSON API relies on another protocol, whether that is HTTP, PATCH, etc, the JSON API behavior MUST (MUST MUST MUST) exactly match the language and behavior specified in the other protocol unless there is a specific and compelling reason not to do so. "Flexibility" for future, unspecified changes are neither and, in fact, should merit a change to the (currently nonexistent but badly needed) JSON API protocol version.

steveklabnik:
Agreed 100% with the MUST MUST MUST.
—

lukfugl:
@b: In my defense, that's from [section 9.6 (PUT)](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.6). I was drawing from [section 9.5 (POST)](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.5), which states:

> The action performed by the POST method might not result in a resource that can be identified by a URI. In this case, either 200 (OK) or 204 (No Content) is the appropriate response status, depending on whether or not the response includes an entity that describes the result.
> 
> If a resource has been created on the origin server, the response **SHOULD** be 201 (Created) and contain an entity which describes the status of the request and refers to the new resource, and a Location header (see section 14.30).

(emphasis added)

I was writing this pull request in the context of `POST` requests only, since #4 hasn't been resolved.

All that said, I agree we should unify the responses between `POST` and `PUT` requests that taret document creation, and in that context, **MUST** becomes appropriate. I think we're all on the same page now.

Any objections to make the `201 Created` response a **MUST**, speak now (I think that's mostly you now, @wycats, since @steveklabnik seems to be on our side now as well :smiley:). Barring any, I'll push the commit for it tomorrow afternoon.

b:
Most of the things for which POST would be used are covered by the use of PATCH. The rest should be PUTs. I don't see a need for POST in the spec except for compatibility with proxies that don't support PATCH. It is even more important that POST behavior is strictly specified for those edge cases.

lukfugl:
If we were only supporting client-generated URIs (such that the client knows where to PUT a resource it is creating), I'd agree with you. But as I understand it, the intent is to also allow server-generated URIs where the client POSTs to a collection resource and the server responds with the location of the new subsidiary resource. In that situation the POST verb is entirely appropriate.

b:
You're exactly right. I was mistaken in only considering the single resource case. I was also reminded by @mnot that HTTPbis (http://tools.ietf.org/wg/httpbis/) is the right document to work from, rather than RFC2616. The semantics draft covers this topic http://tools.ietf.org/html/draft-ietf-httpbis-p2-semantics-22 . The most relevant paragraph is:

If one or more resources has been created on the origin server as a
   result of successfully processing a POST request, the origin server
   SHOULD send a 201 (Created) response containing a Location header
   field that provides an identifier for the primary resource created
   (Section 7.1.2) and a representation that describes the status of the
   request while referring to the new resource(s).

steveklabnik:
This needs a rebase.

Pointing at httpbis is better than 2616, @mnot and @b are right. It's so easy to forget about httpbis. o_O

lukfugl:
dang, bad timing. I was just finishing this up. @brixen: did you have a reason for closing other than apparent abandonment? (newborn had me held up yesterday and earlier today).

or maybe the close is the answer to my next question, which was going to be how to rebase in github-land :) I can't push (without a --force) to lukfugl:201-created because of the rebase. do we typically just do a new branch in my repo with the rebased commits and start a new pull request?

brixen:
@lukfugl sorry! I opened an issue for this here: https://github.com/json-api/json-api/issues/61

I can manually merge your changes if you'd like me to.

brixen:
@lukfugl oh my, just re-read your comment about your newborn. Congratulations! Our daughter is 6-weeks old and my sleep deprived brain has been working on updating this site. If your PR doesn't need content changes, I can just manually apply. If you need to change content, could you open a new PR?

lukfugl:
Yeah, I've got a third commit added to those two, and the rebase (minor conflict resolutions). I'll open a new PR. would you prefer I squash the three to one commit since it'll be a new branch anyways?

lukfugl:
And thanks on the congrats :) Same to you. She's 8 days right now; so json-api popping into my radar was just right for not being too distracted with work etc., but just wrong for having plannable hours to get anything done. :)

brixen:
@lukfugl you can squash or not, whatever is easiest. Sorry again for the hassle.

lukfugl:
Ok, opened at https://github.com/json-api/json-api/pull/62.

Issue #26, opened 2013-05-06
PATCH definition is not RFC compliant


seletskiy:
Sorry if I'm duplicating existing question.

What I talking about is the reference on the page http://jsonapi.org/write/#toc_updating-a-document-code-patch-code to JSON Patch draft document. Link on JSON Patch is leading to outdated version of document.

If you look at latest version (http://tools.ietf.org/html/rfc6902#section-4) of proposal, you will see major change in format:

```
{ "add": "foo", "path": "/a/b/c" } # outdated version
{ "op": "add", "path": "/a/b/c", "value": "foo" } # latest version
```

Should it be updated? I could make PR if needed.

lukfugl:
Agree it should be updated; a PR would be excellent (speaking as an involved bystander, not the project owner). If you don't get to it tonight I can make one tomorrow.

PR #27, opened 2013-05-06
Update JSON Patch specs to the latest RFC 6902

seletskiy:
See issue #26.


steveklabnik:
Thank you for this!

It was this way because we haven't updated our code to use the latest drafts, but we should.

seletskiy:
@steveklabnik Thx for fast reply!

PR #30, opened 2013-05-06
Deletions should not mandate a particular Accept header

celkins:
It has no bearing on the semantics of the DELETE method, and the protocol's implied response message contains no body. Cases where the Accept header would control the response message clearly fall under "Other Responses".


steveklabnik:
Yes, this is an oversight. Thank you.

PR #53, opened 2013-05-08
Removed the extending link under 'see also' on the extending page

ghost:


steveklabnik:
Nice catch with my copy/paste, thank you.

PR #54, opened 2013-05-09
Stringify ids.

### LINKED ISSUE #39: id always as a string?
Most of the examples use integers to represent ids. ids should be opaque, a token that we use to check for equality and we don't need an order relation between ids.

Integers smell like auto-increment ids but that won't always be the implementation server-side. Seems like strings are a more flexible and sane way of encoding any type used server-side as the primary key.

Client-side generated ids are often uuids, which are most commonly represented as strings. Also, ids will be part of URL and used in URL templates. With strings we won't need to convert the ids back and forth between two types.

I think that strings are more suitable for the job. Thoughts?

Related: note some inconsistencies between the type of different ids in this example:

``` json
{
  "posts": {
    "id": "1",
    "title": "Rails is Omakase",
    "rels": {
      "author": 9
    }
  },
  "people": [{
    "id": "9",
    "name": "@d2h"
  }]
}
```

oriolgual:
Fixes #39.

### DIFF
--- a/source/index.md
+++ b/source/index.md
@@ -152,8 +152,8 @@ The value of the `"links"` key is a JSON object that represents related document
     "id": "1",
     "title": "Rails is Omakase",
     "links": {
-      "author": 9,
-      "comments": [ 5, 12, 17, 20 ]
+      "author": "9",
+      "comments": [ "5", "12", "17", "20" ]
     }
   }]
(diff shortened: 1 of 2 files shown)


steveklabnik:
Thank you for this! Great!

PR #55, opened 2013-05-09
Fix href for comments 4,5,6 in JSON read example

chrishunt:
These looks like remnants of a copy/paste. Updating to match the first three examples in the same response.


steveklabnik:
Thanks.

PR #56, opened 2013-05-09
Remove invalid href in write example

chrishunt:
This href looks to be from a previous example where we were looking at photos. This href doesn't make sense for a `post` and we don't need it anyway because this example is showing the use of a url template.


steveklabnik:
Yep! Nice catch.

PR #57, opened 2013-05-09
Change 'posts' to 'photos' to match response attrs

chrishunt:
This example looks like a photo, not a post. Changing the name of the key to match.


steveklabnik:
Exactly. Thank you!

PR #58, opened 2013-05-09
Even singular resources must be presented as list

chrishunt:
From [`index.md`](https://github.com/json-api/json-api/blob/master/source/index.md#singular-resources)

> Singular resources are represented as JSON objects. However, they are still wrapped inside an array

This updates the responses in `write.md` and `extending.md` to match this description.


steveklabnik:
Seems good. Must have missed these. Thanks.

PR #62, opened 2013-05-10
Use '201 Created' HTTP status

lukfugl:
Continued from https://github.com/json-api/json-api/pull/25 with rebase and new commit to incorporate the final comments from that PR.

### DIFF
--- a/source/write.md
+++ b/source/write.md
@@ -117,14 +117,9 @@ Accept: application/json
 ### Response
 
 A server **MUST** respond to a successful document creation request with
-the [`201 Created`][2] status. (While [section 4.3.3][3] of the HTTPbis
-specification allows other `2xx` series responses to a `POST` request
-that successfully creates a resource, this specification utilizes `PUT`
-semantics from [section 4.3.4][4] for document creation responses.)
+the [`201 Created`][2] status.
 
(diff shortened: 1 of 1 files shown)


lukfugl:
@brixen: don't merge this quite yet; it's a new version from what was being discussed on #25 and I'd like @steveklabnik's and @b's feedback

brixen:
@lukfugl I shall tread carefully. :) Thanks for opening the new PR.

steveklabnik:
@lukfugl is this good now?

lukfugl:
Good as in "I'm happy with it myself", yes. Just asked @brixen to hold off on merge in case you or @b had feedback/editing to suggest.

b:
This looks much better! My only suggestion is that it not say 'PUT
semantics'. The behavior defined is fully compliant with POST. Calling out
PUT makes it sound as if this behavior is abnormal or that POST is not the
right verb to use.

lukfugl:
Should I just drop the parenthetical on sections 4.3.3 vs 4.3.4? If it becomes a common concern of implementers we can elaborate with more space in the FAQ.

b:
I like that better.

steveklabnik:
Awesome. Thanks so much for both of your work on hammering this out.

PR #64, opened 2013-05-11
made top level's `links` description more explicit.

mikaa123:
at first read, I thought this description:

> links: in compound resources, information about relationships that would otherwise need to be repeated

was a little confusing. So I changed it to something more specific:

> links: URL templates to be used for expanding resources' relationships


steveklabnik:
Thanks. <3

PR #70, opened 2013-05-13
fix links to id/url based json api docs

ralph:
It seems like inserting a tags with an id attribute is the way to go in markdown. I tried configuring redcarpet with @with_toc_data@, but it would name all the H\* tags sequentially, breaking existing links once you add more h tags.


steveklabnik:
Yeah, that's how it works. Thanks for fixing this, we've been re-organizing the pages and so there's been some breakage.

PR #76, opened 2013-05-18
Removes wrongfully pasted sentence

ayrton:


steveklabnik:
Nice catch, thanks :cry:

PR #78, opened 2013-05-19
Fixes link to URL Template specification.

rcillo:
Just fixing a link to the URL template specification. The correct link was correct in the .md file, but was overridden by a ref of the same number.


steveklabnik:
Nice catch, thank you!

PR #80, opened 2013-05-19
Examples page typo

xtian:
I think this is a typo; I could be wrong, though.


steveklabnik:
It is, I suck at words. Thank you.

Issue #85, opened 2013-05-21
Make meta a reserved word?


rosscooperman:
Seems that if the standard is going to say:

> The top-level of a JSON API document MAY have the following keys:
> 
> meta: meta-information about a resource, such as pagination

it would make sense to include meta as a reserved (but optional) key.

steveklabnik:
Isn't that exactly what this means?

rosscooperman:
Kinda but above that section it says:

> There are three reserved attribute names in JSON API:
> - id
> - href
> - links
> 
> Each of these names has a special meaning when included in the attributes section and should not be used as attribute names.

It just seemed to me like meta should be included in that section. I may be misreading though.

steveklabnik:
Those are resevered _attribute_ names, to go inside of the 'other resource names' (which exist at the top level.)

steveklabnik:
That said, I agree that moving that section could make this more clear.

rosscooperman:
Ah, okay, that makes sense.  Yeah, that wasn't totally clear to me on my first read.  Perhaps the section on reserved attribute names should fall after the description of the top level of a document?  Maybe put `Top Level` and `Reserved Attributes` at the same level (h3) and reverse the order?

Happy to submit a pull request if that seems like a good idea.

steveklabnik:
Sounds great. Please do.

PR #86, opened 2013-05-21
Swap the 'Top Level' and 'Reserved Attributes' section on the format page

rosscooperman:
To make it clearer that the 'Top Level' section refers to attributes at
the top level of a document while 'Reserved Attributes' refers to
attributes of objects inside of a resource array.


rosscooperman:
This would resolve #85

steveklabnik:
Thanks so much.

rosscooperman:
My pleasure

PR #89, opened 2013-05-22
In creating if body is absent must respond with 204

spastorino:


steveklabnik:
Why lock it down to only one code? I'm just asking, I need to re-read the section this is around when I'm not on a taxi-ing plane.

spastorino:
What do you mean by one code?

steveklabnik:
Why specify a 204 and only a 204? Is it important that it is always one? What does that buy us?

spastorino:
Yeah, I just added that because of the context. Perhaps we should remove what is already defined in HTTP semantics?

steveklabnik:
Yup, that's the route I'd rather go. For example, a 202 is really useful to create resources that take a long time to process. 

We want to make sure to remain compliant with HTTP, and re-iterating things for clarity is good, but we don't want to lock things down in a super rigid way unless there's good reason.

spastorino:
@steveklabnik :+1:

spastorino:
@steveklabnik what about this 

A server **MUST** respond to a successful document creation request
according to [`HTTP semantics`](http://tools.ietf.org/html/draft-ietf-httpbis-p2-semantics-22#section-6.3)

The response **MUST** include a `Location` header identifying the primary
document created by the request. It **SHOULD** also include a request body
describing that document. If absent, the client **SHOULD** treat the
transmitted document as accepted without modification.

spastorino:
I've force pushed with that change in case you like it :heart: :heart: :heart:

steveklabnik:
:metal:

PR #91, opened 2013-05-23
Fix JSON Patch operation paths

guillec:
I found this error in all the Patch operation examples:

If the target document is:
{
  "photos": [{
    "id": "1",
    "title": "Productivity",
    "src": "http://example.com/productivity.png"
  }]
}

And the JSON Patch document is this
[
  { "op": "replace", "path": "/src", "value": "http://example.com/hamster.png" }
]

"/src" wont be found.

We need to parse from "photos" no?
The paths should be something like so /photos/0/keep-drilling-down

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -670,7 +670,7 @@ PATCH /photos/1
 Content-Type: application/json-patch+json
 
 [
-  { "op": "replace", "path": "/src", "value": "http://example.com/hamster.png" }
+  { "op": "replace", "path": "/photos/0/src", "value": "http://example.com/hamster.png" }
 ]
 ```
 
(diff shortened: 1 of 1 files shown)


steveklabnik:
Thank you!

gabrielgrant:
As I mentioned in #99, I'm not sure whether this is the best approach: referencing the document within the collection by ID implies that collections are ordered, rather than "set-like"

I do understand that the way things stood before this patch was applied, simply applying the patch to the GETed document wouldn't work, but I also think it isn't all that unreasonable that the patch should be interpreted as being against the actual document contents rather than the returned representation that wraps it in a collection.

PR #92, opened 2013-05-23
Fix URL Template expansion example

kjwierenga:
The third example on the /format page about "URL Template Shorthands" has a full URL for the template. The line after the example states that it expands to a relative URL. This made me think. It's a small issue, but the correct statement is:

"In this example, the author URL for all three posts is `http://example.com/people/12`."

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -380,7 +380,7 @@ Here is another example that uses a has-one relationship:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


spastorino:
:+1: same happens a bit above

spastorino:
Added a PR for the other ones https://github.com/json-api/json-api/pull/94

PR #93, opened 2013-05-23
Add missing href

spastorino:


spastorino:
Well `href` is not on the list because we are talking about id based but ... hmmm I think it's confusing to not having there that. So or we can rephrase this or we can just apply this PR.

@steveklabnik thoughts?

steveklabnik:
@wycats has been talking about unifying the two styles by making the 'href' a default in the ID style. I don't think we lose anything by making it indicated here as well; if you're using hrefs, you're using the URL style.

spastorino:
@steveklabnik exactly :) :heart:

Issue #101, opened 2013-05-27
What to return if a collection is empty?


gabrielgrant:
What should be returned when an empty collection is queried?

a hash with an empty array?

example 1

```
GET /photos

HTTP/1.1 200 OK
Content-Type: application/json

{
  "photos": []
}
```

example 2

```
GET /photos

HTTP/1.1 200 OK
Content-Type: application/json

{
  "photos": [],
  "meta": {
    "client-ids": true
  }
}
```

or an empty hash?

example 3:

```
HTTP/1.1 200 OK
Content-Type: application/json

{}
```

example 4:

```
HTTP/1.1 200 OK
Content-Type: application/json

{
  "meta": {
    "client-ids": true
  }
}
```

I think a hash with an empty array is probably less likely to cause surprises and would be happy to provide a patch if you agree.

steveklabnik:
I'm about to get on a plane and don't have time to properly review the relevant part of the spec, but will come back to it later tonight.

My understanding at current would be that it's example 1, as per the spec right now. Why would it not be?

gabrielgrant:
sorry, just updated the issue to number my examples sanely -- i assume you were referring to the first example 1 (not the second one that is now example 3)?

like i said, my assumption is that example 1 & 2 are correct, but thought it should be spelled out explicitly. Right now the spec covers the cases of "singular resources" and says when "the length of an array at a resource key is _greater than one_, the value represents a list of documents." But it doesn't specify anything for the case when a resource collection is empty.

sorry if I'm being too pedantic... but that's kind of what specs are all about, right? :)

steveklabnik:
Specs are about being pedantic, but there's also a level of explicitness vs. implicitness.

I don't think that this should belong in the spec proper, but I wouldn't mind it being in the examples. (and i meant 1)

gabrielgrant:
also: have a good flight!

gabrielgrant:
woops, deleted my actual comment :P 

Yes, I agree: at a certain point, overly-explicit documents just become basically unreadable. [RFC 6902](http://tools.ietf.org/html/rfc6902#appendix-A) has an appendix of specific examples to keep the core spec more lean. Are you alright with me adding one of those to JSON API?

steveklabnik:
Yes, I wanted to do it myself, but if you're willing, please do.

ahx:
When using the full-url style in the links object, what would you return for an empty relation? 
I think to just omit the attribute might be problematic, so we should return `null` instead, right?

[...]

PR #102, opened 2013-05-27
Add "Examples" appendix with empty collection example

gabrielgrant:
See #101


brixen:
Shouldn't an appendix be its own page? Also, RE #106, couldn't the examples page cover both formats and implementations?

steveklabnik:
Yes, this should go against the Examples page.

steveklabnik:
See, for example: http://amundsen.com/media-types/collection/format/

PR #114, opened 2013-05-30
json based spec for implementors to use

spastorino:
**NOTE: THIS IS WORK IN PROGRESS.**

I'm sending this mostly for you guys to review.
I'm implementing json-api in https://github.com/rails-api/active_model_serializers

Still not sure about many things but I guess it's better to start discussing from this point better than waiting to have a perfect thing.

The idea is to provide json files for implementors to load them, there you have the use cases and you can build tests from there.
Again these files lacks of a lot of use cases, but anyway provide all the feedback you can.

Three files are provided:

1) fixtures.json

Defines the data that servers must load to test reading and updating.

2) read.json

Provides reading use cases.
Each use case is composed of config, expectedRequest and expectedResponse
config defines which fixture should be used, and other server settings
like if the server ask for client-ids, uses url based style and/or
compounds related objects.

3) write.json

Similar to read.json

Let me know your thoughts.

:heart: :heart: :heart:

### DIFF
--- a/json_spec/fixtures.json
+++ b/json_spec/fixtures.json
@@ -1,0 +1,50 @@
+{
+  "only_posts": {
+    "posts": [{
+      "id": "1",
+      "title": "Rails is Omakase",
+      "body":  "body text here"
+    }, {
+      "id": "2",
+      "title": "Rails is not Omakase",
+      "body":  "body text here"
+    }]
+  },
+  "posts_and_one_author": {
+    "posts": [{
+      "id": "1",
+      "title": "Rails is Omakase",
+      "body":  "body text here"
+    }, {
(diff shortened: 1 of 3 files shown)


brixen:
My opinion would be to put this into a separate repository (eg json-api/json-spec).

steveklabnik:
That's the idea, after it's working.

spastorino:
@brixen yes I agree. Let's discuss this first here.

steveklabnik:
This is really old, and @spastorino, i'm not sure you're interested in updating it, given our recent discussions. Let's just give this a close for now.

I'm not sure this problem has been figured out adequately yet.

Issue #117, opened 2013-06-18
Proposal to always wrap relationship IDs in an array


andrewsardone:
I'd like to propose a change to how the `"links"` object's values are represented.

Currently the [ID-based JSON API](https://github.com/json-api/json-api/blob/gh-pages/format/index.md#id-based-json-api) specifies that the `"links"` object represents related documents, linking to those related documents by their ID. A [to-one relationship](https://github.com/json-api/json-api/blob/4f494182c5fe66f29f62aa13d2bc1e08b68a2b0c/format/index.md#to-one-relationships) is a single string or number value, whereas a [to-many relationship](https://github.com/json-api/json-api/blob/4f494182c5fe66f29f62aa13d2bc1e08b68a2b0c/format/index.md#to-many-relationships) is a JSON array value of IDs:

``` json
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author": "9",
      "comments": [ "5", "12", "17", "20" ]
    }
  }]
}
```

With the [URL-Based JSON API](https://github.com/json-api/json-api/blob/4f494182c5fe66f29f62aa13d2bc1e08b68a2b0c/format/index.md#url-based-json-api), relationships can be expressed with string values representing URLs, or [URL template shorthands](https://github.com/json-api/json-api/blob/gh-pages/format/index.md#url-template-shorthands) to reduce duplication and/or bridge between an ID-based JSON API:

``` json
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author": "http://example.com/people/1",
      "comments": "http://example.com/comments/5,12,17,20"
    }
  }]
}
```

``` json
{
  "links": {
    "posts.comments": "http://example.com/comments/{posts.comments}"
  },
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "comments": [ "1", "2", "3", "4" ]
    }
  }]
}
```

This is very slick, but it can also create ambiguity. Within a `"links"` object, a string value can be:
1. A single ID for a to-one relationship, e.g., `{ "links" : { "author": "1" } }`
2. A URL to a to-one relationship, e.g., `{ "links": { "author": "http://example.com/people/1" } }`
3. A URL to a to-many relationship, e.g., `{ "links": { "comments": "http://example.com/comments/5,12,17,20" } }`.

I propose the following:
[...]

PR #120, opened 2013-07-06
Proposal: expose difference between id-based and url-based API

imanel:
With current specification both id-based and url-based API have the same syntax and you need to guess which one should be applied. From my perspective this might be confusing and probably should be avoided. Because of that I would like to suggest adding type-specific suffixes both of them as shown below.
### ID-based API

**Current version**

```
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author": "9",
      "comments": [ "5", "12", "17", "20" ]
    }
  }]
}
```

**Proposal:**

```
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author_di": "9",
      "comment_ids": [ "5", "12", "17", "20" ]
    }
  }]
}
```
### URL-based API

**Current version**

```
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author": "http://example.com/people/1",
      "comments": "http://example.com/comments/5,12,17,20"
    }
  }]
}
```

**Proposal:**

```
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author_url": "http://example.com/people/1",
      "comments_url": "http://example.com/comments/5,12,17,20"
    }
  }]
}
```
### URL Template Shorthands

**Current version**

```
{
  "links": {
  },
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "comments": [ "1", "2", "3", "4" ]
    }
  }]
}
```

**Proposal:**

```
{
  "links": {
    "posts.comments": "http://example.com/comments/{posts.comment_ids}"
  },
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "comment_ids": [ "1", "2", "3", "4" ]
    }
  }]
}
```

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -96,14 +96,14 @@ The value of the `"links"` key is a JSON object that represents related document
     "id": "1",
     "title": "Rails is Omakase",
     "links": {
-      "author": "9",
-      "comments": [ "5", "12", "17", "20" ]
+      "author_id": "9",
+      "comment_ids": [ "5", "12", "17", "20" ]
     }
   }]
 }
 ```
 
-NOTE: Given that `"links"` references `"author"` and the url is `"/people/9"`
+NOTE: Given that `"links"` references `"author_id"` and the url is `"/people/9"`
 clients should hardcode how to relate that reference to that url.
 
 #### To-Many Relationships
@@ -116,7 +116,7 @@ A to-many relationship is represented as a JSON array of IDs.
     "id": "1",
     "title": "Rails is Omakase",
     "links": {
-      "comments": [ "5", "12", "17", "20" ]
+      "comment_ids": [ "5", "12", "17", "20" ]
     }
   }]
 }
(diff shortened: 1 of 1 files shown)


steveklabnik:
@wycats was discussing ways of _unifying_ the two styles, rather than making them more separate... though this could be seen as doing that.

imanel:
From my perspective this is unifying - currently jsonapi has two
simultaleous specs for ID and URL-based, and from my perspective they
should coexist or even be used at the same time in one backend. But I would
like to hear @wycats opinion - should such case be allowed?

Dnia 6 lip 2013 o godz. 16:21 Steve Klabnik notifications@github.com
napisał(a):

@wycats https://github.com/wycats was discussing ways of _unifying_ the
two styles, rather than making them more separate... though this could be
seen as doing that.

—
Reply to this email directly or view it on
GitHubhttps://github.com/json-api/json-api/pull/120#issuecomment-20555045
.

andyfowler:
I think I prefer the suggestion from @andrewsardone in #117, instead of this proposal for a suffix. The distinction could be inferred by clients inspecting whether the link is a string or an array, which feels cleaner than examining part of a string.

dgeb:
Thanks for this proposal. The URL and ID styles were unified in 0759f00451087279d011eebf97346c30430db711 via a different declarative approach.

PR #123, opened 2013-08-15
Add HAL comparison

mikepmunroe:
wycats answer on 'Why not use HAL', taken directly from https://news.ycombinator.com/item?id=5654212

### DIFF
--- a/faq/index.md
+++ b/faq/index.md
@@ -8,3 +8,30 @@ title: "JSON API: Frequently Asked Questions"
 Once JSON API is stable, it will always be backwards compatible using a _never
 remove, only add_ strategy.
 [#46](https://github.com/json-api/json-api/issues/46)
+
+### Why not use the HAL specification?
+
+There are several reasons:
+* HAL embeds child documents recursively, while JSON API flattens the entire
+graph of objects at the top level. This means that if the same "people" are
+referenced from different kinds of objects (say, the author of both posts and
+comments), this format ensures that there is only a single representation of
+each person document in the payload.  
+* Similarly, JSON API uses IDs for linkage, which makes it possible to cache
+documents from compound responses and then limit subsequent requests to only 
+the documents that aren't already present locally. If you're lucky, this can 
+even completely eliminate HTTP requests.  
+* HAL is a serialization format, but says nothing about how to update
+documents. JSON API thinks through how to update existing records (leaning on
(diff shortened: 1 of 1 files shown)


steveklabnik:
Hey thanks for this.

Can you wrap to 80 cols and then I'll merge? Thanks!

mikepmunroe:
Ouch, yes. I actually know better and should have done so in the first place. New submission coming soon.

mikepmunroe:
FAQ update wrapped at 80 characters now.

steveklabnik:
It's all good! Thanks!

faq/index.md
```diff
@@ -8,3 +8,30 @@ title: "JSON API: Frequently Asked Questions"
...
+There are several reasons:
+* HAL embeds child documents recursively, while JSON API flattens the entire
+graph of objects at the top level. This means that if the same "people" are
+referenced from different kinds of objects (say, the author of both posts and
+comments), this format ensures that there is only a single representation of
+each person document in the payload.  
+* Similarly, JSON API uses IDs for linkage, which makes it possible to cache
+documents from compound responses and then limit subsequent requests to only 
+the documents that aren't already present locally. If you're lucky, this can 
+even completely eliminate HTTP requests.
```
machty, line 23: The above point/paragraph is unclear to me; "JSON API uses IDs for linkage" as opposed to HAL's what?

dgeb, line 23: HAL's embedding of child documents to establish linkage.

dgeb, line 23: Perhaps we could clarify this by changing "Similarly, JSON API uses IDs for linkage..." to "Instead of document embedding, JSON API uses IDs for linkage..."

Would that be more clear?

machty, line 23: I think what's missing is a definition of `linkage`... is linkage the `_links` hash of related resources/actions, or is `linkage` in reference to how, for example, arrays of ids are used to refer to flat-graph top level resources in the payload?

dgeb, line 23: "Linkage" refers to the representation of links (i.e. associations) between resources.

In HAL, links are represented by embedding child documents in parent documents.

In JSON API, links are represented by listing the `id(s)`, `type(s)`, and/or `href(s)` of related resources in the `links` hash of a document. These related resources are now (optionally) represented in the `linked` hash at the top level of compound documents (the text should be changed from saying "JSON API flattens the entire graph of objects at the top level").

PR #130, opened 2013-09-12
add examples: fortune.js and ember-json-api

gr0uch:
Add Fortune.js to the list of server-side implementations, modify ember-data description to include json-api adapter.


steveklabnik:
Thanks!

PR #132, opened 2013-09-25
Fixed paths for PATCH requests on format/index.md

n00ge:
I found an issue in the examples where the PATCH requests were using the wrong id for /photos. I believe i have fixed all the examples that had this issue. 

Thanks!

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -677,7 +677,7 @@ PATCH /photos/1
 Content-Type: application/json-patch+json
 
 [
-  { "op": "replace", "path": "/photos/0/src", "value": "http://example.com/hamster.png" }
+  { "op": "replace", "path": "/photos/1/src", "value": "http://example.com/hamster.png" }
 ]
 ```
 
@@ -726,7 +726,7 @@ Content-Type: application/json-patch+json
 Accept: application/json
 
 [
-  { "op": "replace", "path": "/photos/0/links/author", "value": 2 }
+  { "op": "replace", "path": "/photos/1/links/author", "value": 2 }
 ]
 ```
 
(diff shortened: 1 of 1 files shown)


steveklabnik:
Thank you so much!

guillec:
Hey guys not sure if this is correct. 

Based on application/json-patch+json it would be 0. 

That example is dealing with a JSON Patch document not a PATCH request.

The number is a index for an array of photos not a id of a photo.

n00ge:
That makes sense. Sorry for the confusion. 
I can reverse the commit and push up another pull request if you'd like.

gr0uch:
I think @guillec is correct. On this note, is there official documentation for bulk PATCHing? For example, you might need to use different indices for doing this:

``` js
PATCH /posts/1,3

{
  {"op": "replace", "path": "/posts/0/links/author", "value": 2},
  {"op": "replace", "path": "/posts/1/links/author", "value": 2}
}
```

In this case, indices 0 and 1 correspond to post IDs 1 and 3 respectively.

steveklabnik:
Ahhhh. I was confused by this as well, that makes sense. Yes, please send another PR with a revert.

n00ge:
I posted up a [pull request with the revert](https://github.com/json-api/json-api/pull/133).

I'd love to hear what the consensus is on bulk PATCH requests (as well as DELETE /photos/1,3 and potentially GET /photos/1,3 ?).

n00ge:
After rereading the docs, it looks like rather than using the syntax @daliwali suggested, we should be doing a 
PATCH /posts?ids=1,3

In the same respect, I assume to delete multiple posts we would do a 
DELETE /posts?ids=1,3

These aren't stated explicitly but I think it follows the patterns more closely.

gr0uch:
I think that the routes `/posts/1,3` and `/posts?ids=1,3` are equivalent. I would prefer the former, since then the client doesnt have to switch to using a query string for multiple IDs.

PR #136, opened 2013-10-09
fix #135 - replace hypermedia-type with hypermedia media type

### LINKED ISSUE #135: First sentence of index.md is confusing
I was confused trying to read the first sentence of `index.md`. I'd submit a pull request if I understood what it was trying to say; maybe I can just explain the problem in grammatical terms.

The first sentence lacks an object:

> "JSON API" is a JSON-based read/write hypermedia-type designed to support a smart client who wishes to build a data-store of information.

"JSON API" is a hypermedia-type ____. A hypermedia-type _what_? A hypermedia-type API? I'm not sure what that means, but at least it makes grammatical sense. Or, is it a hypermedia **type** (no hyphen), 'type' being the subject?

It also uses "who" instead of "what" when describing a non-human thing; using "who" for the client is a little confusing when the "client" you're talking about is generally a client-side web application, the pronoun for which should be "that". But then it's also a little weird that your rich client is _wishing_ to do something. And now that I look at it, "data-store of information" seems redundant, or vague. Isn't every data store a data store of information?

Anyway, if I understood what you were trying to say, I'd try to rewrite it, but I'm not sure I understand what the spec is about.

mikepmunroe:


brixen:
@mikepmunroe thanks!

PR #138, opened 2013-10-19
Fix related document key to match link type

derrickreimer:
Perhaps I'm misinterpreting the spec, but I'm fairly certain this key is supposed to match the `type` specified in the link metadata.


dgeb:
@djreimer - #143 fixes the same example and also introduces a new top-level `linked` object for containing related documents. Sorry I didn't see your PR first, or I would have included it in mine.

Issue #139, opened 2013-10-22
Resource names vs relationship names


jonah-williams:
I would like to clarify how a JSON API is expected to return related resources, particularly in a compound document.

I think this was alluded to in #23 as a "name-aliased relationship" but may have been read as applying to server-side business object types.

In a compound document related resources are included under top level keys which match the name of their relationship rather than their resource type. I think this poses problems for server implementations attempting to avoid sending duplicate data and for client implementations attempting to minimize the cost of parsing responses.

Let's consider the following response (adapted from the compound document example on http://jsonapi.org/format/#url-based-json-api) where multiple relationships exist to the same resource type and references to the same resource appear under multiple relationships.

```
{
  "links": {
    "posts.author": {
      "href": "http://example.com/people/{posts.author}",
      "type": "people"
    },
    "posts.subject": {
      "href": "http://example.com/people/{posts.subject}",
      "type": "people"
    }
  },
  "posts": [{
    "id": "1",
    "title": "Steve seems to have opinions about computers",
    "links": {
      "author": "127",
      "subject": "4"
    }},
    {
    "id": "2",
    "title": "Jonah sounds confused",
    "links": {
      "author": "4",
      "subject": "127"
    }}
  ],
  "author": [{
    "id": "127",
    "name": "@zalambar"
  },
  {
    "id": "4",
    "name": "@steveklabnik"
  }],
  "subject": [{
    "id": "4",
    "name": "@steveklabnik"
  },
  {
    "id": "127",
    "name": "@zalambar"
  }]
}
```

This is not great but seems to me to be what the spec currently calls for. On the server side we're paying the cost of serializing the same object multiple times. On the client side we're downloading an unnecessarily large payload and parsing more objects than should be necessary (and building relationships between them which can be expensive).
Additionally it is not trivial for the client to maintain a list of all known "people" resources. The client may have already received some of these `subject` or `author` resources as `people` but it is not easy to identify these duplicates.

I think it would be preferable for compound documents to include a collection per-resource rather than per-relationship:

```
{
  "links": {
    "posts.author": {
      "href": "http://example.com/people/{posts.author}",
      "type": "people"
    },
    "posts.subject": {
      "href": "http://example.com/people/{posts.subject}",
      "type": "people"
    }
  },
  "posts": [{
    "id": "1",
    "title": "Steve seems to have opinions about computers",
    "links": {
      "author": "127",
      "subject": "4"
    }},
    {
    "id": "2",
[...]

PR #140, opened 2013-10-22
Amend example to use resource name instead of relationship name

gr0uch:
Addresses issue #139.


samwgoldman:
This is a duplicate of #138, but I agree this should be the correct format.

gr0uch:
I even +1'd the other PR. I have the memory of a goldfish...

PR #143, opened 2013-11-01
Separate related resources in a top-level `linked` object.

### LINKED ISSUE #74: Requesting resource that includes self-referential compound documents
Say each person has friends that are also other people.  If I `GET /person/1`, and I'm returned a compound document, I'll get something like this:

``` json
{
  "people": [
    {
      "id": "1",
      "name": "Tammer",
      "links": { "friends": ["9"] }
    }, 
    {
      "id": "9",
      "name": "DHH"
    }
  ]
}
```

I can no longer assume that the first document in the "people" array is the one I want.  I must instead loop through them comparing `id` values to get the root document I originally requested.  Two solutions come to mind:
1. Require an order, so that it's assumed that requesting a singular resource will return an array where the _first_ document is that resource.  I don't care much for this solution.
2. Nest all related resources in another array.  Fairly heavy handed:

``` json
{
  "people": [{
    "id": "1",
    "name": "Tammer",
    "links": { "friends": ["9"] }
  }],
  "linked": [
    "people": [{
      "id": "9",
      "name": "DHH"
    }]
  ]
}
```

This affects all responses, of course:

``` json
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author": "9"
    }
  }],
  "linked": [
    "people": [{
      "id": "9",
      "name": "@d2h"
    }]
  ]
}
```

...and begs the question of what to return if I am my own friend.

So:  Thoughts?

dgeb:
It's important to separate the primary resource from related resources because:
- It allows the client to easily identify the primary resource among related resources of the same type. This is especially important when creating a resource without prior knowledge of its server-assigned id.
- It clarifies issues like pagination that are specific to the primary resource.

This can be achieved by nesting related resources in a `linked` object. For instance:

``` javascript
{
  "posts": [{
    "id": "1",
    "name": "Rails is Omakase",
    "links": {
      "author": "9",
      "related": ["2"]
    }
  }],
  "linked": {
    "posts": [{
      "id": "2",
      "name": "The Parley Letter",
      "links": {
        "author": "9",
        "related": ["1"]
      }
    }],
    "people": [{
      "id": "9",
      "name": "DHH"
    }]
  }
}
```

It's easy to imagine a scenario in which resources that are members of the primary array are secondary resources for other members of the primary array (e.g. people + friends). In those scenarios, primary resources should not be duplicated as secondary resources in `linked`. In other words, a single canonical representation of a resource should continue to be returned with each response.

[Closes #74]

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -24,7 +24,8 @@ attributes section and should not be used as attribute names.
 The top-level of a JSON API document **MAY** have the following keys:
 
 * `meta`: meta-information about a resource, such as pagination
-* Other resource names (`posts`, `comments`, `people`, etc.)
+* `linked`: a collection of documents, grouped by type, that are related to the
+   primary document and/or each other
 
 ### Reserved Attributes
 
@@ -197,7 +198,8 @@ The top-level of a JSON API document **MAY** have the following keys:
 
 * `meta`: meta-information about a resource, such as pagination
 * `links`: URL templates to be used for expanding resources' relationships URLs
-* Other resource names (`posts`, `comments`, `people`, etc.)
+* `linked`: a collection of documents, grouped by type, that are related to the
+   primary document and/or each other
 
 ### Singular Resources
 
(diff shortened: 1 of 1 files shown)


dgeb:
I've discussed this PR with @wycats.

BTW, I've tried to be follow the patterns in place regarding usage of the terms "resource" vs. "document". Perhaps we should more clearly favor one over the other?

gr0uch:
Just as a side note, I noticed here that the key names for resources is consistent with the type of the resource and not by the relationship name, as discussed in issue #139.

dgeb:
@daliwali - Good catch. I'll comment over on that issue.

PR #147, opened 2013-11-02
Clarify when default fieldsets should be returned.

dgeb:
Also, changed "A server" to "The server" in a couple places for
consistency with the "Inclusion of Related Documents" section.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -568,6 +568,10 @@ with a `fields[DOCUMENT_TYPE]` parameter.
 GET /posts?include=authors&fields[posts]=id,title&fields[people]=id,name
 ```
 
+An endpoint SHOULD return a default set of fields for a document if no fields
+have been specified for its type, or if the endpoint does not support use of
+either `fields` or `fields[DOCUMENT_TYPE]`.
+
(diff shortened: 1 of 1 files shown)


steveklabnik:
I'd actually prefer "a server" everywhere.

I'm not 100% sure this is clarifying; what does 'unsorted' mean? If it happens to be sorted via some metric, is that bad?

dgeb:
@steveklabnik - I've just removed the "a" / "the" changes. My point was to be consistent with the other "Inclusion of Related Documents" section that I wrote, but I will address these reference changes in a separate PR. I'm not crazy about the way a/the server is referenced repeatedly.

You mention "unsorted", which is not in this change. Are you just providing an example of a default? I actually touch on sorting defaults #148.

As for this particular change: @wycats suggested I add something about defaults, but didn't give specifics. Let's review with him to see if this covers his intentions.

steveklabnik:
Sorry, I read 'unfiltered' and then typed 'unsorted.'

steveklabnik:
(and I agree they should be consistent, thank you for pointing it out!)

dgeb:
#152 is my attempt to make references to "a server" more readable and consistent.

PR #148, opened 2013-11-02
Introduce support for specifying sort order.

dgeb:
Describes how a server may optionally support requests for
primary and related documents to be sorted according to
specific criteria.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -571,6 +571,51 @@ GET /posts?include=authors&fields[posts]=id,title&fields[people]=id,name
 Note: `fields` and `fields[DOCUMENT_TYPE]` can not be mixed. If the latter
 format is used, then it must be used for the primary document type as well.
 
+### Sorting
+
+A server **MAY** choose to support requests to sort documents according to
+one or more criteria.
+
+An endpoint **MAY** support requests to sort the primary document type with a
+`sort` parameter.
+
+```text
+GET /people?sort=age
+```
+
+An endpoint **MAY** support multiple sort criteria by allowing comma-separated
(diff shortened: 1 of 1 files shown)


steveklabnik:
I'm not sure this needs to be in the main spec, this is an example of something that should probably be in a profile.

That said, I can _also_ see it going in the main spec, so I'm not going to close this just yet.

dgeb:
@steveklabnik Let me back up and provide some background on this and my other recent PRs.

The goal of JSON API is "to support a smart client who wishes to build a data-store of information". Without providing guidelines for common use cases such as including related records, customizing fieldsets, sorting, filtering, and even pagination, there is a limit to how "smart" we can build client and server frameworks that consume and produce this media type. I'm in favor of providing these guidelines, while ensuring that each is optional by specifying them very liberally with MAYs.

I agree that a server's specific implementation of these guidelines belongs in a profile. For instance, some servers may not support sorting at all, while others may support sorting for only the primary document, while still others may support sorting for only specific fields. As long as client and server frameworks know how to comply with each of these conditions, they can still have the intelligence to work with them rather than leaving all these concerns to ad hoc implementations.

wycats:
I'm in favor of these as **MAY**s that have good defaults if an endpoint does not support them.

steveklabnik:
Absolutely, it's a balance.

dgeb:
@steveklabnik - how are you feeling about this sorting PR now that **MAY** has been used extensively?

steveklabnik:
Feeling good.

dgeb:
Woot - the PR backlog is clear :)

steveklabnik:
:heart: thanks for all the recent work

PR #152, opened 2013-11-07
Ensure consistency in references to "a server"

dgeb:
This changes the structure of several paragraphs to avoid redundant references
to either "a" or "the" server.


dgeb:
Going to tweak this further to remove the passive voice...

dgeb:
Ok @wycats - removed the passive voice and switched to `endpoint` as suggested.

PR #153, opened 2013-11-08
Fix plurality of `author` paths.

dgeb:
In a few instances, `authors` was used instead of `author`, even though it's a
 singular relationship. This change brings the "Fetching" section in agreement with
 the rest of the spec.


rkh:
:+1:

PR #169, opened 2013-11-30
Split code snippet with an HTTP session into two snippets

isagalaev:
I split the code sample with an HTTP session into two: one with only the request line, and the other with the response. This way highlight.js is able to recognize and correctly highlight both snippets, including markdown in the HTTP response body.

### DIFF
--- a/extending/index.md
+++ b/extending/index.md
@@ -47,9 +47,11 @@ GET http://api.example.com/
 That document will de-reference to explain your link relations:
 
 ```text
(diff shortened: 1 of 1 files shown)


steveklabnik:
Thank you!!!

isagalaev:
btw, it looks like you don't have Markdown included in highlight.pack.js on the production site, so it's not highlighted in that snippet. It's probably not a big deal but just to let you know :-).

PR #171, opened 2013-12-05
fix page alignment in some browsers (firefox)

gr0uch:
In Firefox, it seems that the page is left aligned:
screenshot from 2013-12-05 14 59 32

By setting the horizontal margin to `auto`, that fixes it so that it's centered, consistent with how Chrome renders the page:
screenshot from 2013-12-05 14 59 46


steveklabnik:
Thank you!

PR #178, opened 2013-12-29
Align example of URL style with ID style

nwillems:
Clarifying examples.
The ID style example used the ID 9 whereas the URL style used 1.


steveklabnik:
Thank you!

PR #179, opened 2014-01-17
Change ID value from number to URL in the URL-Based examples

guillec:
- The value for ID in the URL-Based format should be a URL. The current examples show a number.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -215,7 +215,7 @@ wrapped inside an array:
 ```javascript
 {
   "posts": [{
-    "id": 1
+    "id": "http://example.com/posts/1"
     // an individual post document
   }]
 }
@@ -234,10 +234,10 @@ represents a list of documents.
 ```javascript
 {
   "posts": [{
-    "id": 1
+    "id": "http://example.com/posts/1"
     // an individual post document
   }, {
-    "id": 2
+    "id": "http://example.com/posts/2"
     // an individual post document
   }]
 }
@@ -256,7 +256,7 @@ Other than the `"links"`, `"href"` and `"id"` keys, every key in a document repr
 ```javascript
 {
   "posts": [{
-    "id": "1",
+    "id": "http://example.com/posts/1",
     "title": "Rails is Omakase"
   }]
 }
@@ -269,7 +269,7 @@ The value of the `"links"` key is a JSON object that represents related document
 ```javascript
 {
   "posts": [{
-    "id": "1",
+    "id": "http://example.com/posts/1",
     "title": "Rails is Omakase",
     "links": {
       "author": "http://example.com/people/1",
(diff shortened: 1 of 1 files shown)


BRMatt:
I'm not sure if `id` attributes should contain the URL to the entity, a `self` link seems better suited to this. The id subsection of the URL format also states (emphasis mine):

> The "id" key in a document represents a unique identifier for the document, scoped to the document's type. It **can be used with URL templates to fetch related records**, as described below.

Which implies that it's some value that can be embedded in URLs. If it's a URL then you have to parse the id from it in order to generate a URL from a different template.

guillec:
Thanks @BRMatt 

When I first studied the specs I had the same thought but then I saw this video by @steveklabnik demoing json-api 

http://www.youtube.com/watch?v=FpS_E90-6O8 around 15:40.

He states that the id should be a URL.

Maybe we need to clarify this?

andrewsardone:
I second @BRMatt in that `id` should _not_ contain a URL for the document – the `href` attribute or something else (like a `self` attribute) would be much more appropriate. It would be nice to settle this in the spec.

JSON API is designed to give clients a way to work with the remote API in conjunction with a local datastore. I think using URLs as an `id` defeats the purpose.

Consider this [good point](https://groups.google.com/forum/m/#!msg/api-craft/8YOgtN_E69Y/iYz9pAIUNNUJ) from @steveklabnik over on the _API Craft_ mailing list, re: URLs:

> It's much easier to just use a full URL rather than reconstructing one from some base. Also, in theory, you lose the flexibility to point to another domain; imagine a CDN in this case, where the responses point to servers in different regions depending on where they're requested from.

As noted, _URLs can change_. This is okay with the URL-based/hypermedia-style APIs – they give clients a way to continue resolving links or fetch additional resources because the URLs (or templates) are provided.

But if a URL changes (for a bunch of legit reasons) and we use that URL as the `id` of the document, then any local datastore representation of that resource has been completely invalidated. I think that's unnecessary – the identity of the resource hasn't changed, but just some attribute.

We want the flexibility of changing URLs at the API-level, but I think local datastores will require a constant ID for their resources.

I think in both the ID-based and URL-based scenarios, all documents MUST have an `id` attribute that acts as a unique and opaque string identifier that's unique to the type. I think overloading the contents of this string value as a possible URL goes against the data synchronization aspect of JSON API.

This is the strategy we have used at @nutshellcrm. We use the URL-based JSON-API format, but still use the `id` attribute to be a constant identifier that can be used within some identity-map on the client-side of things.

dgeb:
I agree that while `id` fields could be overloaded as URLs (which are unique string identifiers after all), it is not generally recommended.

@guillec - I re-listened to @steveklabnik's talk and can definitely understand your confusion on this point.

@steveklabnik - ok to close this one?

guillec:
Thanks for the clarification everyone!

steveklabnik:
Thanks! This discussion is exactly why I left this here for a bit, I didn't have the time to type out a full justification ;)

steveklabnik:
(I wouldn't mind a pr that puts this explanation into the actual format, as a footnote)

PR #180, opened 2014-01-22
Clarify the role of id keys within the URL-based JSON API

andrewsardone:
This comes off of a discussion in json-api/json-api#179 and @steveklabnik's [call for a PR](https://github.com/json-api/json-api/pull/179#issuecomment-32909054).

I'm not sure if the additions align with what the team had in mind, but I wrote down [my earlier thoughts](https://github.com/json-api/json-api/pull/179#issuecomment-32897319) that are based on some experiences implementing a URL-based JSON API.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -251,6 +251,13 @@ The `"id"` key in a document represents a unique identifier for the document,
 scoped to the document's type. It can be used with URL templates to fetch
 related records, as described below.
 
+NOTE: While an implementation could use the values of `"id"` keys as URLs
+(which are unique string identifiers, after all), it is not generally
+recommended. URLs can change, so they are unreliable for mapping a document to
+any client-side models that represent the same resource. It is recommended that
+URL values be left to the task of linking documents while `"id"` values remain
+opaque to solely provide a unique identity within some type.
+
 ### Attributes
 
 Other than the `"links"`, `"href"` and `"id"` keys, every key in a document represents an attribute. An attribute's value may be any JSON value.
--- a/format/index.md
+++ b/format/index.md
@@ -247,7 +247,9 @@ Each document in the list **MAY** contain an `id` key.
 
 ### IDs
 
-The `"id"` key in a document represents a unique identifier for the document, scoped to the document's type. It can be used with URL templates to fetch related records, as described below.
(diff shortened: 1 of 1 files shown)


steveklabnik:
Thanks for putting this together!

I do like this, unsure about the MUST, though.

andrewsardone:
I'd be interested to hear your reservations on the MUST.

I think all documents need to have _some_ kind of identity, if only for a consumer to map it to some in-memory model, or some row in a database somewhere. I like that this identify is explicit, and qualified with some kind of rule (unique to the document type, or more preferably, a GUID).

Given this common case, I think it makes the most sense to just set up the convention (across both ID-based and URL-based specs) to _always_ have an `id`.

dgeb:
This seems good in general but I would not change MAY to MUST either.

Examples of read-only data that should not require assignment of an arbitrary server-side id include: auto-suggest data (e.g. names starting with "Ab..."), enum-like data (e.g. U.S. states), and raw data (e.g. poll statistics).

andrewsardone:
Then why the [MUST in the ID-based format](https://github.com/json-api/json-api/blob/b4a65c1ba685b86c5ddf9587a7866012aec81622/format/index.md#singular-resources)? Is this an oversight?

I think your examples are good, but if they were treated as singular resources I'd still give them an ID. How else would I link an enum-like value as a to-one relationship off of some resource? For raw statistics, they'd probably be wrapped within some parent resource, like a `pollResults` (which I would give an ID).

dgeb:
@andrewsardone Good catch. We should be consistent about MUSTs. And I don't think we should make that decision until we clarify how raw data should be returned alongside (or embedded in) "normal" documents.

andrewsardone:
@dgeb, so for the time being you would like me to revert the MUST back to a MAY, but leave the note?

dgeb:
@andrewsardone that would be great, and I'll merge right away. It would also be great if you could open another issue regarding the inconsistency between MUST and MAY in the two formats so that we can resolve it separately. Thanks much!

andrewsardone:
Okay, I have rebased-out the MUST.

dgeb:
Perfect - thank you!

Issue #183, opened 2014-01-28
Proposed Merger of URL and ID Styles


lgebhardt:
@dgeb and I are working on tools to generate client and server implementations of JSON API and have been tripped up by incompatibilities in the URL and ID styles.

We'd like to propose a merger of these two styles so that they can co-exist in the same spec without conflict. This proposal slightly modifies the way links and related IDs are handled.
## Current Spec

Currently the same keys can be used for either IDs or URLs for a document's `links` with no indication of what type of data the key contains:

```
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author": "9",
      "comments": [ "5", "12", "17", "20" ]
    }
  }]
}
```

or

```
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author": "http://example.com/people/1",
      "comments": "http://example.com/comments/5,12,17,20"
    }
  }]
}
```
### Problems with the Current Spec

Using the same keys to store either IDs or URLs creates confusion for producers and consumers of the spec (as shown by the number of issues here related to IDs and links). Given that not all IDs are going to be integer numbers (UUIDs in particular, but there are other key systems in use, some of which may have a passing resemblance to urls), it is non trivial to programmatically determine whether an ID or a URL was provided by an API in a general way.

Using the same keys to store either IDs or URLs in the current manner also precludes the option of providing both IDs and URLs. If related documents are specified by URLs in the primary document _and_ included as linked documents in a compound document, then there is no path to get from a primary document to its included linked documents. For example, a client would not be able to easily determine which comments go with which post in the following example:

```
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author": "http://example.com/people/9",
      "comments": "http://example.com/comments/5,12,17"
    }
  }, {
    "id": "2",
    "title": "The Parley Letter",
    "links": {
      "author": "http://example.com/people/9",
      "comments": "http://example.com/comments/20"
    } 
  }],
  "linked": {
    "people": [{
      "id": "9",
      "name": "@d2h"
    }],
    "comments": [{
      "id": "5",
      "body": "Mmmmmakase"
    }, {
      "id": "12",
      "body": "I prefer unagi"
    }, {
      "id": "17",
      "body": "What's Omakase?"
    }, {
      "id": "20",
      "body": "Parley is a discussion, especially one between enemies"
    }]
  }
}
```

[...]

Issue #164, opened 2013-11-12
Can normal links include meta data, too?


rkh:
According to the examples in the spec, top level links (using templates) can include meta data:

``` json
{
  "links": {
    "posts.author": {
      "href": "http://example.com/people/{posts.author}",
      "type": "people"
    }
  }
}
```

Can links in the document include this, too?

``` json
{
  "posts": [{
    "links": {
      "author": {
        "href": "http://example.com/people/dhh",
        "type": "people"
      }
    }
  }]
}
```

dgeb:
:+1: I think this makes sense and should be explicitly called out in the spec.

BRMatt:
It would also help avoid this caveat mentioned in the spec:

> NOTE: The linkage between the key under "links" ("author" in this example) and the associated document type ("people" in this example) is hardcoded into the client. The client must also hardcode how to relate that reference to a url.

dbettin:
+1 on this request. This is the only thing making us diverge from the spec. We love to see this land.

bencates:
+1.  I'd also like to propose the following variations:

``` json
{
  "posts": [{
    "links": {
      "author": {"id": "dhh", "type": "person"},
      "commenters": [
        {"id": "wycats", "type": "person"},
        {"id": "37signals", "type": "company"}
      ]
    }
  }]
} 
```

steveklabnik:
Adding to 1.0.

dgeb:
Meta data in links is now explicitly allowed via #237. See the [resource relationships](http://jsonapi.org/format/#document-structure-resource-relationships) section.

PR #187, opened 2014-02-02
Add to-many relationship format as an array of link objects

andrewsardone:
Initially [proposed](https://github.com/json-api/json-api/issues/183#issuecomment-33840493) by @mikelinington in json-api/json-api#183, this expands the to-many relationship format to accomodate an array of `link` objects.

``` json
{
  "posts": [
    {
      "id": "1",
      "title": "One Type Purr Author",
      "links": {
        "authors": [
          {
            "href": "http://example.com/people/9",
            "id": "9",
            "type": "people"
          },
          {
            "href": "http://example.com/cats/1",
            "id": "1",
            "type": "cats"
          }
        ]
      }
    }
  ]
}
```

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -139,6 +139,48 @@ supported by the rigid structure of a template. In those cases, it may also be
 necessary to include either `"id"` or `"ids"` for the related documents in a
 compound document.
 
+#### To-One Relationships
+
+A to-one relationship **MAY** be represented as a string or number value that
+corresponds to the ID of a related resource.
+
+```javascript
+{
+  "posts": [{
+    "id": "1",
+    "title": "Rails is Omakase",
+    "links": {
+      "author": "17"
+    }
+  }]
+}
+```
+
+It **MAY** alternatively be represented with a `"link"` object that contains one
+or more of the attributes: `"id"`, `"href"` and `"type"`.
+
+```javascript
+{
+  "posts": [{
+    "id": "1",
+    "title": "Rails is Omakase",
+    "links": {
+      "author": {
+        "href": "http://example.com/people/17",
+        "id": "17",
+        "type": "people"
+      }
+    }
+  }]
+}
+```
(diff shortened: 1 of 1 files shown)


dgeb:
Reviewing issues this morning, I was reminded that @bencates [originally proposed](https://github.com/json-api/json-api/issues/164#issuecomment-31991224) this in #164. 

Regardless of origin, I think it's a good idea that rounds out the possible methods for specifying links, and it does so in a consistent format.

We should probably define "link" objects higher up in the "Relationships" section and expand the possibilities to mention an array of "link" objects. For example:

```
The link to each related resource MUST be one of the following:

* a string or number - to represent a single ID.
* an array of strings or numbers - to represent multiple IDs.
* a "link" object that contains one or more of the attributes:
  `"id"`, `"ids"`, `"href"` and `"type"`. Note that `"id"` and `"ids"` should
  never be present together.
* an array of "link" objects
```

Since the term "link" won't actually appear in a JSON document, I believe it should be written "link" instead of `"link"`.

Anyway, these are minor nits. First things first: I'm :+1: on this change.

andrewsardone:
Good notes, @dgeb – thanks! I pushed [two commits](https://github.com/json-api/json-api/compare/andrewsardone:8f475c1...andrewsardone:ee6ce8a) with the suggested updates.

dgeb:
@andrewsardone looks good to me - thanks!

andrewsardone:
Do we have anything major remaining here, or is this good to merge?

dgeb:
I'm :+1: on this - just waiting for feedback from @steveklabnik and @wycats

steveklabnik:
Makes sense to me. @wycats ?

dgeb:
Just chatted with @wycats, who gave his ok to merge.

Thanks @andrewsardone :shipit:

PR #189, opened 2014-02-03
FAQ: New Q/A: How discover resource capabilities?

diosney:
Add a new Q/A to let implementors know how they can provide resource discovery capabilities.

Maybe this could be moved to standard itself?

### DIFF
--- a/faq/index.md
+++ b/faq/index.md
@@ -47,9 +47,6 @@ For instance, if `"GET,POST"` is the response to an OPTIONS request to an URL,
 then you can get information about the resource and also create new resources.
 
 If you want to know what you can do with a specific resource attribute then
-you will have to use:
-
-- an application level profile to define the attribute meaning and capabilities,
-
-- and use the errors response to let users know. This feature is still pending
-to be included in the standard since is still in discussion.
+you will have to use an application level profile to define the attribute meaning
+and capabilities and use the errors response to let users know. This feature is
+still pending to be included in the standard since is still in discussion.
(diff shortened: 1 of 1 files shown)


steveklabnik:
This is certainly a FAQ-level thing, as it just repeats what's in RFC 2616.

diosney:
I agreed. Nice I pushed to the correct section :)

So, what do you think? It worth to be included?

When I checked the issues I found one guy that requested something like resource discovery capabilities but maybe is so simple that it doesn't worth the space.

steveklabnik:
I like having it, but I'm not sure why the bottom two sentence parts are in bullet points. Maybe just as a regular sentence? Also, linking to the errors issue would be good :)

diosney:
The bullet points were just to standout them. I removed them in a new commit.

The **Errors** debate is in #7 (already linked in Q/A).

steveklabnik:
Looks great, thanks for putting up with the bikeshed. :ship: :heart: :confetti_ball:

diosney:
Awesome!

Bikeshed (and  anti-bikeshedding at the site home) are referring to [Parkinson's law of triviality](http://en.wikipedia.org/wiki/Bikeshed)?

diosney:
Ha! I understand now after reading the law lol!

Maybe I can link those words in the site home so anybody can look quickly at the term in Wikipedia?

steveklabnik:
Naw, we don't need a definition of it here. heh.

diosney:
Hehe, seems that I got caught bikeshedding :D ha!!

Okey, I won't continue spamming here :)

PR #194, opened 2014-02-11
DELETE: Added new semantic: resource collection deletion.

diosney:
a) Added new Node.js to examples page.

b) If a client wants to remove every resource in a collection, say `/posts` or `/logs`, there is no current defined semantics for doing a bulk delete.

Of course, the client can make multiple _DELETE_ for each resource in the collection, or even make a DELETE to an URL like `/logs/1,2,3,7,8,9,10`, but those solutions don't work right when one have zillions of items in a collection.

This commit add semantics to _DELETE_ an entire collection by using the collection URL.

Hope you like it.

pd. hope you find the duck so you only can bikeshed the commit and not cancel it completelly :D (just joking)

### DIFF
--- a/examples/index.md
+++ b/examples/index.md
@@ -59,3 +59,7 @@ has page describing how to emit conformant JSON.
 * [json-patch](https://github.com/guillec/json-patch) implementation of JSON Patch (rfc6902) 
 
 * [hana](https://github.com/tenderlove/hana) implementation of the JSON Patch and JSON pointer spec 
+
+### Node.js
+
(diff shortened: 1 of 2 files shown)


diosney:
Added new commit: A Nodejs package to work with JSON-PATCH

steveklabnik:
I would love to just merge 32016e9, but we need more discussion about deletion of an entire collection. If you'd be willing to submit just that commit as a separate PR, I would merge it right away.

diosney:
Okey, I will separate those PR now.

PR #196, opened 2014-02-11
Add Hy Python library

kalasjocke:
I decided to start building a tiny Python library for creating json-api responses:

https://github.com/kalasjocke/hy

It's built upon an existing serialization library called Schematics, fantastic piece of software by the way, for handling serialization. So everything Hy does is to glue Schematics together with something that creates the actual responses.

What do you think? Is this something worth adding to the project just yet?

### DIFF
--- a/examples/index.md
+++ b/examples/index.md
@@ -44,6 +44,10 @@ has page describing how to emit conformant JSON.
 
 * [Oat](https://github.com/ismasan/oat#adapters) ships with a JSON API adapter.
 
+### Python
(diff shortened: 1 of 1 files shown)


steveklabnik:
Absolutely! Thank you!

(and I appreciate the :scissors: in a separate commit :heart: )

kalasjocke:
:smile: 

You're welcome, and thanks for doing an awesome job with the json-api specification!

Issue #193, opened 2014-02-10
JSON Patch equivalent structures


guillec:
On issue #176 there is a discussion about how to possibly map JSON Patch paths. 

I've added some examples here for further discussion. 
# 

How to replace the title of a post:

<pre>
PATCH /posts/1

# JSON Patch Document:
[
  { "op": "replace", "path": "/posts/1/title", "value": "New Title" }
]

# JSON API resource:
{
  "posts": [{
    "id": "1",
    "title": "Old Title"
  }]
}

# JSON Patch equivalent structure:
{
  "posts": {
    "1": {
      "id": "1",
      "title": "The title"
    }
  }
}
</pre>

How to remove comment 5:

<pre>
PATCH /posts/1
 
# JSON Patch Document: 
[
  { "op": "remove", "path": "/posts/1/comments/5" }
]
 
# JSON API resource:
{
  "posts": [{
    "id": "1",
    "links": {
      "comments": [ "1", "3", "5", "7" ]
    }
  }]
}
 
# JSON Patch equivalent structure:
{
  "posts": {
    "1": {
      "id": "1",
      "links": {
        "comments": {
          "1": "1",
          "3": "3",
          "5": "5",
          "7": "7"
        }
      }
    }
  }
}
</pre>

How to replace author type with a new value of "human":

<pre>
PATCH /posts/1
 
# JSON Patch Document: 
[
  { "op": "replace", "path": "/posts/1/author/type", "value": "human" }
]
 
# JSON API resource:
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author": {
        "href": "http://example.com/people/17",
        "id": "17",
        "type": "people"
      }
    }
  }]
}

# JSON Patch equivalent structure: 
{
  "posts": {
    "1": {
      "id": "1",
      "links": {
        "author": {
          "href": "http://example.com/people/17",
          "id": "17",
          "type": "people"
        }
      }
    }
  }
}
</pre>

How to replace author type with id of 1 to "not-dog":

<pre>
PATCH /posts/1
 
# JSON Patch Document: 
[
  { "op": "replace", "path": "/posts/1/authors/1", "value": "not-dog" }
]
 
# JSON API resource:
{
  "posts": [
    {
      "id": "1",
      "title": "One Type Purr Author",
      "links": {
        "authors": [
          {
            "href": "http://example.com/people/9",
            "id": "9",
            "type": "people"
          },
          {
[...]

PR #198, opened 2014-02-12
DELETE: Added new semantic: resource collection deletion.

diosney:
If a client wants to remove every resource in a collection, say `/posts` or `/logs`, there is no current defined semantics for doing a bulk delete.

Of course, the client can make multiple _DELETE_ for each resource in the collection, or even make a DELETE to an URL like `/logs/1,2,3,7,8,9,10`, but those solutions don't work right when one have zillions of items in a collection.

This commit add semantics to _DELETE_ an entire collection by using the collection URL.

Hope you like it.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -926,6 +926,16 @@ document's URL.
 DELETE /photos/1
 ```
 
+A server **MAY** choose to support resource collection *deletion*.
+
+An endpoint that supports it **MUST** respond to a `DELETE` request
+to the resource collection URL by removing every document listed as
+a resource by its corresponding `GET` request.
+
+```text
+DELETE /photos
+```
+
 ### 204 Responses
 
 If a server returns a `204 No Content` in response to a `DELETE`


diosney:
@steveklabnik What do you think about this new semantic action?

steveklabnik:
I am 75% okay with it. DELETE is almost never what you actually want, but this is something people do implement.

Off-topic: you don't need to close and make new pull requests to edit them: they're tied to branches. Removing that other commit from that branch and force pushing would have just edited the original PR. In this case it's not a big deal, but just so you know for the future!

diosney:
Okey, so we will wait until more people gives her toughts.

RE:Off-topic: Yes, sorry :( Just a Github newbie here that didn't know how to do that. From now I will make separate branches for unrelated additions and push them separatelly.

diosney:
I think this can wait until what is being discussed at https://github.com/json-api/json-api/issues/193 is finished, since if it is applied as a whole this PR have no purpose.

diosney:
@steveklabnik Any updates on this?

steveklabnik:
I'm still not opposed, but right now, I'm focusing on getting what we NEED to get done in 1.0 in. This is an addition, so by necessity, I'm waiting until after we resolve all of those possibly-backwards-incompatible issues first.

steveklabnik:
We are focusing on backwards compatible issues right now, and so I cannot merge this PR for now. We're not opposed to adding something like this in the future, but it needs discussion. Can you please open an issue about this instead? Thanks.

PR #209, opened 2014-03-15
[Fix 190] Type of Ids

MajorBreakfast:
- Id must be a string
- Can contain arbitrary characters, sticking to alphnumeric chars, dashes and underscores is recommended.
- Also highlight that rfc6570 mentions URL encoding of the components

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -39,7 +39,7 @@ and keyed by the plural form of the resource type:
 ```javascript
 {
   "posts": [{
-    "id": 1
+    "id": "1"
     // an individual post document
   }]
 }
@@ -58,10 +58,10 @@ and keyed by the plural form of the resource type:
 ```javascript
 {
   "posts": [{
-    "id": 1
+    "id": "1"
     // an individual post document
   }, {
-    "id": 2
+    "id": "2"
     // an individual post document
   }]
 }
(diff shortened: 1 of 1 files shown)


MajorBreakfast:
Open points:
- I didn't read RFC6570 completely. Is calling it "URL encode" alright? The RFC mentions "pct-encode".
- Should we recommend libraries for doing the encoding? Getting it right seems to be tricky.

wycats:
1. The correct terminology is "percent encoding"
2. I don't think recommending libraries is necessary. The correct API in JavaScript is built in (`encodeURIComponent`, which should be used for URL parameters, like this situation). It works both in the browser and in Node.

MajorBreakfast:
I changed that. I also mention the JavaScript method. It's not a long sentence and it makes it completly clear what "percent encode" means to those who never heard about it and that there's not much to it.

MajorBreakfast:
Bump.

steveklabnik:
EmberConf is going on right now, so it's gonna be a few days until we move any discussions forward. This isn't being ignored, there's just only so much time in the day.

MajorBreakfast:
Yah, alright

steveklabnik:
Thank you! Sorry again about the wait.

MajorBreakfast:
thx! :)

PR #218, opened 2014-04-04
Supply a json-schema of the json-api

kjg:
pulled from https://github.com/json-api/parser-ruby/blob/master/lib/json/api/schema.json

### DIFF
--- a/faq/index.md
+++ b/faq/index.md
@@ -51,3 +51,10 @@ you will have to use an application level profile to define the attribute meanin
 and capabilities and use the errors response to let users know. This error feature
 is still pending to be included in the standard since is still in
 [discussion](https://github.com/json-api/json-api/issues/7).
+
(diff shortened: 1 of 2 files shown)


steveklabnik:
Why duplicate this across the two repositories?

Also, since JSON Schema isn't powerful enough to represent all of JSON API, I'm a little averse to actually making it so prominent.

kjg:
If this makes it into jsonapi.org I would expect the other repo to migrate
to using this one.

What can't json schema represent about json api? This pr already improves
upon the other's coverage and there is still way more we can add. I hope to
do just that if I can get the buy in.

Making it canonical means there is a sanctioned, maintained schema people
can validate responses against instead of some githubraw url.

steveklabnik:
Thanks! I'd love to merge this, but it's now out of date. Could you rebase, please?

kjg:
Sorry about that. Rebased.

steveklabnik:
No worries. Thanks so much!

PR #219, opened 2014-04-11
Enabling header links via with_toc_data and bbalter solution

meatballhat:
which is almost OK, but I'm not happy about the crappy anchor names redcarpet assigns.  Overriding this with a custom TOC handler _would_ be possible if this didn't have to be `jekyll --safe` compatible, but that's kinda the deal with github pages.  One option would be to drop the `safe` option, add a li'l plugin to make sluggy anchor ids a la `#singular-resources` and require all compiled resources to be committed :scream_cat:.  Opinions wanted!

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -15,3 +15,6 @@ exclude:
 markdown: redcarpet
 pygments: false
 port: 9876
+redcarpet:
+  extensions:
+    - with_toc_data
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -2,19 +2,22 @@
 <html>
   <head>
     <meta charset="utf-8">
-    
+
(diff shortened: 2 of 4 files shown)


steveklabnik:
What do the anchor names look like?

meatballhat:
`toc_01`  :scream_cat:

steveklabnik:
Not great, but not the worst either.

What's the risks in dropping --safe?

meatballhat:
Dropping `--safe` would mean becoming incompatible with github page
generation, meaning generated sources would have to be committed, meaning
every contributor would have to commit their source and generated changes.
Not great, but not horrible considering the number of page sources.

meatballhat:
Nah, maybe try again when there's a better solution available.

PR #220, opened 2014-04-22
Adds header links to all headers

gcollazo:
screencast

With this change when a user hovers over a header a little ¶ symbol will appear. This symbol links to #slug-of-header-text which is placed next to the header text.

This little fix makes it easier to share specific sections of the spec.


steveklabnik:
Amazing, thank you. I've been wanting to do this for a while.

PR #223, opened 2014-04-27
Removed mention of a URL-based JSON API

guillec:
Remove mention of a URL-based format since the formats were merged.


steveklabnik:
Thanks!

Issue #190, opened 2014-02-03
type of ids


MajorBreakfast:
Currently the format specification doesn't say which type ids have. I'd say it's a string. (But there are currently some examples that use a number, though) Is there a benefit in allowing Numbers? When the URL is built this information gets lost anyway.

steveklabnik:
One of the problems with allowing non-string types is that JSON basically doesn't say what type its numbers are, though floating-point only is a good bet.

guillec:
:+1: for string

mrec:
:+1: for string. Given the importance of ids I'd actually go further than that and include some recommendations on choosing good id strings.

Partly that means character restrictions, especially since all the examples assume a `/type/id,id?queryargs`-style URI template. So `,`, `/`, `?` and `#` are right out. IIS by default always 404s any paths containing `%` or `.`. And so on. Something like "alphanumerics and hyphen only" sounds like a reasonable SHOULD.

There's also the question of how strictly to enforce the opacity of ids. The spec seems (?) to lean toward ids having no meaning outside navigation via the API, but for cases like countries or languages or currencies which have well-defined and widely-used ISO ids defined, refusing to use them on a point of principle is arguably forcing unnecessary chattiness on clients wanting to do any kind of interop.

EDIT: the character-restriction part looks like it overlaps with #69

MajorBreakfast:
@mrec This restriction is not needed due to URI encoding ([give it a try here](http://meyerweb.com/eric/tools/dencoder/))

mrec:
@MajorBreakfast No, URI encoding doesn't help for the path portion. The problem is that IIS [disallows certain characters there by default](http://blogs.iis.net/nazim/archive/2011/04/18/use-of-special-characters-like-in-an-iis-url.aspx) _even if urlencoded_, and I don't think it's the only one.

Yes, most issues can be probably worked around in config, but it'd be better to dodge it in the first place.

EDIT: better ref link for IIS behaviour.

MajorBreakfast:
Okay, then: MUST be a string which SHOULD contain only alphanumeric characters. Urlencoding MUST be used (for security).

That alright? SHOULD/MUST is of course debatable but I chose MUST because url de/encoding only works well if both sides apply it once, plus it's easy to implement and very cheap.

mrec:
[...]

Issue #176, opened 2013-12-19
Proposal: Use a combination of `test` and `remove` JSON Patch operations instead of treating to-many relationships like sets


fancyremarker:
Issue #100 (and the discussion on #99 also) reveal an unresolved question in the spec, around how to properly form JSON Patch operations that remove an item from a to-many relationship. Currently, the spec says:

> While to-many relationships are represented as a JSON array in a GET response, they are updated as if they were a set. [...] To remove an element from a to-many relationship, use a `remove` operation on `links/<name>/<id>`.

This is problematic because JSON Pointer ([RFC 6901](http://tools.ietf.org/html/rfc6901)), the address format for JSON Patch's ([RFC 6902](http://tools.ietf.org/html/rfc6902)) `path`, makes no mention of sets, only arrays. Consider the following example resource, a post with 4 comments:

``` json
{
  "posts": [{
    "id": "1",
    "links": {
      "comments": [ "1", "3", "5", "7" ]
    }
  }]
}
```

The current JSON API spec says that if we want to remove the comment with ID 3, we'd make the following `PATCH` request:

```
PATCH /photos/1

[
  { "op": "remove", "path": "/posts/0/links/comments/3" }
]
```

This is a valid JSON Patch document, but its standard semantics are different from what the JSON API spec would have it mean. A naive client built on a compliant JSON  Patch library could very easily make the above request with the intention of removing the element with index 3 in the `comments` array (i.e., comment 7), but instead, our JSON API-compliant server would remove the wrong comment, comment 3.

I propose that in order to remove an element from a to-many relationship, a client must use a `remove` operation on `links/<name>/<arrayIndex>`, along with a `test` that `links/<name>/<arrayIndex>` has the value we expect. With our example from above, this would look like:

```
PATCH /photos/1

[
  { "op": "test", "path": "/posts/0/links/comments/1", "value": "3" }
  { "op": "remove", "path": "/posts/0/links/comments/1" }
]
```

This is guaranteed to always be safe, and also complies with both the JSON Patch and JSON Pointer specs. For more info on the `test` operation see [RFC 6902, §4.6](http://tools.ietf.org/html/rfc6902#section-4.6).

dgeb:
I think we should clear something up in the spec:

_Arrays of documents are functionally equivalent to sets that are keyed by document ids._

In RFC 6901 / 6902 terms, a set is simply an object with "members" and "member values".

So, even though responses are structured as:

```
{
  "posts": [{
    "id": "1",
[...]

PR #234, opened 2014-06-09
Refactoring in preparation for JSON API v1.0rc1

dgeb:
_Note: see the [revised format page](https://github.com/dgeb/json-api/blob/v1rc1/format/index.md) to review the heart of this PR._

This is a fairly thorough rewrite of the spec that formalizes its structure and addresses many outstanding issues.

The goal of this rewrite is to prepare JSON API for 1.0 versioning with a release candidate phase. We would like to shake out any breaking changes during this RC phase to provide implementers the degree of confidence they deserve from this spec.

Any and all feedback is welcome, either in this PR or in the specific issues that it addresses.

Changes include:
- New introduction explains the basics requirements and goals of JSON API. It
  also broadly explains the new optional PATCH support and how "JSON API
  represents all of a domain's resources as a single JSON document that can act
  as the target for operations".
- New "Conventions" section explains SHOULD, MUST, etc. keywords
- New "Document Structure" section describes the JSON API media type
  (application/vnd.api+json). This media type is used for both request and
  response documents (except for PATCH requests/responses).
- Introduces the option to key the primary resource by the generic `data` key.
  This key SHOULD be used for variable type (i.e. heterogenous) resource
  collections and MAY be used for constant type (i.e. homogenous) resource
  collections.
- Clarify different representations allowed for singular and plural resources.
- BREAKING CHANGE: Singular resource objects SHOULD now be be represented with
  JSON objects instead of arrays. This allows for symmetrical representations in
  request and response documents, as well as PUT/POST requests and PATCH
  operations. It also simplifies implementations which do not support batch
  operations (i.e. they can allow an object and not an array).
- Define URLs for resources, collections of resources and resource
  relationships. Allow for alternative URL definitions to be specified in
  responses.
- BREAKING CHANGE: Allow the baseline implementation of JSON API to operate via
  POST, PUT and DELETE alone (no PATCH required). This introduces brand new
  specs for updating resources via PUT and updating relationships via POST and
  DELETE requests to newly specified relationship URLs. It also specifies how
  resources can be created, updated and deleted in bulk (but only per type).
- Introduce alternative JSON PATCH syntax for all operations. This builds off
  the current spec's approach to updating relationships. JSON PATCH bulk
  operations are discussed.
- Introduce a "Filtering" section in "Fetching". Since we encourage keeping all
  resources accessible at the root level, it corresponds that root level
  filtering should be encouraged instead of filtering via nested routes.
  Furthermore, root level filtering is more flexible because it allows for more
  than one filter to be applied to a collection.
- Introduce a `clientid` key that can be used to correlate a resource on the
  client with a newly created resource on the server.
- Introduce error objects, which are specialized resource objects that MAY be
  returned in a response to provide additional information about problems
  encountered while performing an operation.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -589,8 +589,9 @@ For example, a collection of resources of type "photos" will have the URL:
 /photos
 ```
 
-The URL for an individual resource **SHOULD** be formed by appending the
-resource's ID to the collection URL.
+Collections of resources **SHOULD** be treated as sets keyed by resource ID. The
+URL for an individual resource **SHOULD** be formed by appending the resource's
+ID to the collection URL.
 
 For example, a photo with an ID of `"1"` will have the URL:
 
--- a/format/index.md
+++ b/format/index.md
@@ -5,166 +5,330 @@ title: "Format"
 
 {% include status.md %}
 
-## Document <a href="#document" id="document" class="headerlink">¶</a>
+## Introduction <a href="#introduction" id="introduction" class="headerlink"></a>
 
-In this specification, the term "document" refers to a single object with a
-set of attributes and relationships.
+JSON API is a specification for how a client should request that resources be
+fetched or modified and how a server should respond to those requests. JSON API
+is designed to minimize both the number of requests and the amount of data
+transmitted between clients and servers.
 
-A JSON response may include multiple documents, as described in this
-specification.
+JSON API requires use of the JSON API media type
+([`application/vnd.api+json`](http://www.iana.org/assignments/media-types/application/vnd.api+json)) 
+for exchanging data.
 
-### Top Level <a href="#document-top-level" id="document-top-level" class="headerlink">¶</a>
+A JSON API server supports fetching of resources through the HTTP method GET. 
+In order to support creating, updating and deleting resources, it must support 
+use of the HTTP methods POST, PUT and DELETE, respectively.
 
-The top-level of a JSON response will contain the primary document(s)
-keyed by the plural form of the primary resource type.
+A JSON API server may also optionally support modification of resources with
+the HTTP PATCH method [[RFC5789](http://tools.ietf.org/html/rfc5789)] and the
+JSON Patch format [[RFC6902](http://tools.ietf.org/html/rfc6902)]. JSON Patch
+support is possible because, conceptually, JSON API represents all of a
+domain's resources as a single JSON document that can act as the target for
+operations. Resources are grouped at the top level of this document according
+to their type. Each resource can be identified at a unique path within this
+document.
 
-The top-level of the JSON response **MAY** also have the following keys:
+## Conventions <a href="#conventions" id="conventions" class="headerlink"></a>
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in RFC 2119
+[[RFC2119](http://tools.ietf.org/html/rfc2119)].
+
+## Document Structure <a href="#document-structure" id="document-structure" class="headerlink"></a>
+
+This section describes the structure of a JSON API document, which is identified
+by the media type [`application/vnd.api+json`](http://www.iana.org/assignments
+/media-types/application/vnd.api+json). JSON API documents are defined in
+JavaScript Object Notation (JSON)
+[[RFC4627](http://tools.ietf.org/html/rfc4627)].
+
+Although the same media type is used for both request and response documents,
+certain aspects are only applicable to one or the other. These differences are
+called out below.
+
+### Top Level <a href="#document-structure-top-level" id="document-structure-top-level" class="headerlink"></a>
+
+A JSON object **MUST** be at the root of every JSON API document. This object
+defines a document's "top level".
+
+A document's top level **SHOULD** contain a representation of the resource or
+collection of resources primarily targeted by a request (i.e. the "primary
+resource(s)").
+
+The primary resource(s) **SHOULD** be keyed either by their resource type or the
+generic key `"data"`.
+
+A document's top level **MAY** also have the following members:
 
 * `"meta"`: meta-information about a resource, such as pagination.
 * `"links"`: URL templates to be used for expanding resources' relationships
   URLs.
-* `"linked"`: a collection of documents, grouped by type, that are related to
-  the primary document(s) and/or each other.
+* `"linked"`: a collection of resource objects, grouped by type, that are linked
+  to the primary resource(s) and/or each other (i.e. "linked resource(s)").
 
-Each of these keys has a special meaning when included in the top level and
-should not be used as a resource type in order to avoid conflicts.
+No other members should be present at the top level of a document.
 
-No other keys should be present at the top level of the JSON response.
+### Resource Representations <a href="#document-structure-resource-representations" id="document-structure-resource-representations" class="headerlink"></a>
 
-### Singular Resources <a href="#document-sigular-resources" id="document-sigular-resources" class="headerlink">¶</a>
+This section describes how resources can be represented throughout a JSON API
+document. It applies to primary as well as linked resources.
 
-Documents that represent singular resources are wrapped inside an array
-and keyed by the plural form of the resource type:
+#### Singular Resource Representations <a href="#document-structure-singular-resource-representations" id="document-structure-singular-resource-representations" class="headerlink"></a>
+
+A singular resource **SHOULD** be represented as a single "resource object"
+(described below) or a string value containing its ID (also described below).
+
+The following post is represented as a resource object:
 
 ```javascript
 {
-  "posts": [{
-    "id": "1"
-    // an individual post document
-  }]
+  "posts": {
+    "id": "1",
+    // ... attributes of this post
+  }
 }
 ```
 
-This simplifies processing, as you can know that documents will always be
-wrapped in arrays.
+This post is represented simply by its ID:
 
-The document **SHOULD** contain an `"id"` key.
+```javascript
+{
+  "posts": "1"
+}
+```
 
-### Resource Collections <a href="#document-resource-collections" id="document-resource-collections" class="headerlink">¶</a>
+#### Plural Resource Representations <a href="#document-structure-plural-resource-representations" id="document-structure-plural-resource-representations" class="headerlink"></a>
 
-Documents that represent resource collections are also wrapped inside an array
-and keyed by the plural form of the resource type:
+A collection of resources **SHOULD** be represented as an array of resource
+objects or IDs, or as a single "collection object" (described below).
+
+The following posts are represented as an array of resource objects:
 
 ```javascript
 {
   "posts": [{
     "id": "1"
-    // an individual post document
+    // ... attributes of this post
   }, {
     "id": "2"
-    // an individual post document
+    // ... attributes of this post
   }]
 }
 ```
 
-Each document in the array **SHOULD** contain an `"id"` key.
-
-### IDs <a href="#document-ids" id="document-ids" class="headerlink">¶</a>
-
-The `"id"` key in a document represents a unique identifier for the underlying
-resource, scoped to its type. It **MUST** be a string which **SHOULD** only
-contain alphanumeric characters, dashes and underscores. It can be used with URL
-templates to fetch related resources, as described below.
-
-
-In scenarios where uniquely identifying information between client and server
-is unnecessary (e.g., read-only, transient entities), JSON API allows for
-omitting the `"id"` key.
-
-NOTE: While an implementation could use the values of `"id"` keys as URLs
-(which are unique string identifiers, after all), it is not generally
-recommended. URLs can change, so they are unreliable for mapping a document to
-any client-side models that represent the same resource. It is recommended that
-URL values be left to the task of linking documents while `"id"` values remain
-opaque to solely provide a unique identity within some type.
-
-### Attributes <a href="#document-attributes" id="document-attributes" class="headerlink">¶</a>
-
-There are three reserved attribute names in JSON API:
-
-* `id`
-* `href`
-* `links`
-
-Every other key in a document represents an attribute. An attribute's value may
-be any JSON value.
+These posts are represented as an array of IDs:
 
 ```javascript
 {
-  "posts": [{
+  "posts": ["1", "2"]
+}
+```
+
+These comments are represented by a single "collection" object:
+
+```javascript
+{
+  "comments": {
+    "href": "http://example.com/comments/5,12,17,20",
+    "ids": [ "5", "12", "17", "20" ],
+    "type": "comments"
+  }
+}
+```
+
+#### Variable Type Resource Representations <a href="#document-structure-variable-type-resource-representations" id="document-structure-variable-type-resource-representations" class="headerlink"></a>
+
+Variable type (i.e. "heterogenous") resources can not be keyed by type, so the
+type of each resource **SHOULD** be specified in an alternative manner.
+
+Variable type primary resources **SHOULD** be keyed by `"data"` at the top level
+of a document.
+
+Variable type resources **SHOULD** be represented as objects that contain both a
+type and an ID:
+
+```javascript
+{
+  "data": [{
+    "id": "9",
+    "type": "people",
+    "name": "@tenderlove"
+  }, {
+    "id": "1",
+    "type": "cats",
+    "name": "@gorbypuff"
+  }]
+}
+```
+
+Variable type resources **SHOULD NOT** be represented by string values alone.
+
+### Resource Objects <a href="#document-structure-resource-objects" id="document-structure-resource-objects" class="headerlink"></a>
+
+Resource objects have the same internal structure, regardless of whether they
+represent primary or linked resources.
+
+Here's how a post (i.e. a resource of type "posts") might appear in a document:
+
+```javascript
+{
+  "posts": {
     "id": "1",
     "title": "Rails is Omakase"
-  }]
+  }
 }
 ```
 
-### Relationships <a href="#document-relationships" id="document-relationships" class="headerlink">¶</a>
-
-The value of the `"links"` key is a JSON object that represents related
-resources.
+In the example above, the post's resource object is simply:
 
 ```javascript
-{
-  "posts": [{
+//...
+  {
+    "id": "1",
+    "title": "Rails is Omakase"
+  }
+//...
+```
+
+This section will focus exclusively on resource objects, outside of the context
+of a full JSON API document.
+
+#### Resource Attributes <a href="#document-structure-resource-object-attributes" id="document-structure-resource-object-attributes" class="headerlink"></a>
+
+There are five reserved keys in resource objects:
+
+* `"id"`
+* `"clientid"`
+* `"type"`
+* `"href"`
+* `"links"`
+
+Every other key in a resource object represents an "attribute". An attribute's
+value may be any JSON value.
+
+#### Resource IDs <a href="#document-structure-resource-object-ids" id="document-structure-resource-object-ids" class="headerlink"></a>
+
+Each resource object **SHOULD** contain an ID, which may be represented by an
+`"id"` key, a `"clientid"` key, or both.
+
+The `"id"` key in a resource object represents a unique identifier for the
+underlying resource, scoped to its type. Its value **MUST** be a string which
+**SHOULD** only contain alphanumeric characters, dashes and underscores. It can
+be used with URL templates to fetch related resources, as described below.
+
+Servers **MAY** optionally support use of a `"clientid"` to correlate a resource
+on the client with a newly created resource on the server. The `"clientid"`
+value has the same requirements as the `"id"` value, with the exception that its
+uniqueness is only scoped to a particular client and resource type. The role of
+`"clientid"` in creating resources is described below.
+
+In scenarios where uniquely identifying information between client and server
+is unnecessary (e.g. read-only, transient entities), JSON API allows for
+omitting IDs.
+
+#### Resource Types <a href="#document-structure-resource-types" id="document-structure-resource-types" class="headerlink"></a>
+
+The type of each resource object can usually be determined from the context in
+which it is contained. As discussed above, resource objects are typically keyed
+by their type in a document.
+
+Each resource object **MAY** contain a `"type"` key to explicitly designate its
+type.
+
+The `"type"` key is **REQUIRED** when the type of a resource is not otherwise
+specified in a document.
+
+For example, here's an array of resource objects that might be part of a has-
+many polymorphic relationship:
+
+```javascript
+//...
+  [{
+    "id": "9",
+    "type": "people",
+    "name": "@tenderlove"
+  }, {
+    "id": "1",
+    "type": "cats",
+    "name": "@gorbypuff"
+  }]
+//...
+```
+
+#### Resource URLs <a href="#document-structure-resource-urls" id="document-structure-resource-urls" class="headerlink"></a>
+
+The URL of each resource object **MAY** be specified with the `"href"` key.
+Resource URLs **SHOULD** only be specified by the server and therefore are
+typically only included in response documents.
+
+```javascript
+//...
+  [{
+    "id": "1",
+    "href": "http://example.com/comments/1",
+    "body": "Mmmmmakase"
+  }, {
+    "id": "2",
+    "href": "http://example.com/comments/2",
+    "body": "I prefer unagi"
+  }]
+//...
+```
+
+A server **MUST** respond to a `GET` request to the specified URL with a
+response that includes the resource.
+
+It is generally more efficient to specify URL templates at the root level of a
+response document rather than to specify individual URLs per resource.
+
+#### Resource Relationships <a href="#document-structure-resource-relationships" id="document-structure-resource-relationships" class="headerlink"></a>
+
+The value of the `"links"` key is a JSON object that represents linked
+resources, keyed by the name of each association.
+
+For example, the following post is associated with a single `author` and a
+collection of `comments`:
+
+```javascript
+//...
+  {
     "id": "1",
     "title": "Rails is Omakase",
     "links": {
       "author": "9",
       "comments": [ "5", "12", "17", "20" ]
     }
-  }]
-}
+  }
+//...
 ```
 
-The link to each related resource **MUST** be one of the following:
+##### To-One Relationships <a href="#document-structure-resource-relationships-to-one" id="document-structure-resource-relationships-to-one" class="headerlink"></a>
 
-* a string or number - to represent a single ID.
-* an array of strings or numbers - to represent multiple IDs.
-* a "link" object that contains one or more of the attributes:
-  `"id"`, `"ids"`, `"href"` and `"type"`. Note that `"id"` and `"ids"` should
-  never be present together.
-* an array of "link" objects
+To-one relationships **MUST** be represented with one of the formats for
+singular resources described above.
 
-NOTE: Use of a document level `"links"` object is generally discouraged because
-root level URL Templates can usually provide the same data more concisely.
-However, there may be situations where each document will have a URL that isn't
-supported by the rigid structure of a template. In those cases, it may also be
-necessary to include either `"id"` or `"ids"` for the related documents in a
-compound document.
-
-#### To-One Relationships <a href="#document-relationships-to-one-relationships" id="document-relationships-to-one-relationships" class="headerlink">¶</a>
-
-A to-one relationship **MAY** be represented as a string or number value that
-corresponds to the ID of a related resource.
+For example, the following post is associated with a single author, identified
+by ID:
 
 ```javascript
-{
-  "posts": [{
+//...
+  {
     "id": "1",
     "title": "Rails is Omakase",
     "links": {
       "author": "17"
     }
-  }]
-}
+  }
+//...
 ```
 
-It **MAY** alternatively be represented with a "link" object that contains one
-or more of the attributes: `"id"`, `"href"` and `"type"`.
+And here's an example of a linked author represented as a resource object:
 
 ```javascript
-{
-  "posts": [{
+//...
+  {
     "id": "1",
     "title": "Rails is Omakase",
     "links": {
@@ -174,39 +338,35 @@ or more of the attributes: `"id"`, `"href"` and `"type"`.
         "type": "people"
       }
     }
-  }]
-}
+  }
+//...
 ```
 
-An API that provides a to-one relationship as a URL **MUST** respond to a `GET`
-request with the specified document with the specified URL.
+##### To-Many Relationships <a href="#document-structure-resource-relationships-to-many" id="document-structure-resource-relationships-to-many" class="headerlink"></a>
 
-In the above example, a `GET` request to `http://example.com/people/17` returns
-a document containing the specified author.
+To-many relationships **MUST** be represented with one of the formats for plural
+resources described above.
 
-#### To-Many Relationships <a href="#document-relationships-to-many-relationships" id="document-relationships-to-many-relationships" class="headerlink">¶</a>
-
-A to-many relationship **MAY** be represented as an array of strings or numbers
-corresponding to IDs of related resources.
+For example, the following post is associated with several comments, identified
+by their IDs:
 
 ```javascript
-{
-  "posts": [{
+//...
+  {
     "id": "1",
     "title": "Rails is Omakase",
     "links": {
       "comments": [ "5", "12", "17", "20" ]
     }
-  }]
-}
+  }
+//...
 ```
 
-It **MAY** alternatively be represented with a "link" object that contains one
-or more of the attributes: `"ids"`, `"href"` and `"type"`.
+And here's an example of an array of comments linked as a collection object:
 
 ```javascript
-{
-  "posts": [{
+//...
+  {
     "id": "1",
     "title": "Rails is Omakase",
     "links": {
@@ -216,76 +376,30 @@ or more of the attributes: `"ids"`, `"href"` and `"type"`.
         "type": "comments"
       }
     }
-  }]
-}
+  }
+//...
 ```
 
-An API that provides a to-many relationship as a URL **MUST** respond to a
-`GET` request with a list of the specified documents with the specified URL.
+### Collection Objects <a href="#document-structure-collection-objects" id="document-structure-collection-objects" class="headerlink"></a>
 
-In the above example, a `GET` request to
-`http://example.com/comments/5,12,17,20` returns a document containing the four
-specified comments.
+A "collection object" contains one or more of the members: 
 
-As another alternative, a to-many relationship **MAY** be represented as an
-array of "link" objects that contain one or more of the attributes: `"id"`,
-`"href"`, and `"type"`.
+* `"ids"` - an array of IDs for the referenced resources.
+* `"type"` - the resource type.
+* `"href"` - the URL of the referenced resources (applicable to response
+  documents).
 
-```javascript
-{
-  "posts": [{
-    "id": "1",
-    "title": "Rails is Omakase",
-    "links": {
-      "comments": [{
-        "href": "http://example.com/comments/5",
-        "id": "5",
-        "type": "comments"
-      },
-      {
-        "href": "http://example.com/comments/12",
-        "id": "12",
-        "type": "comments"
-      }]
-    }
-  }]
-}
-```
+A server that provides a collection object that contains an `"href"` **MUST**
+respond to a `GET` request to the specified URL with a response that includes
+the referenced objects as a collection of resource objects.
 
-In the above example, `GET` requests to `http://example.com/comments/5` and
-`http://example.com/comments/12` return the respective comments.
 
-NOTE: Given its verbosity, this third format should be used sparingly, but it
-is helpful when the related resources have a variable `"type"`:
+### URL Templates <a href="#document-structure-url-templates" id="document-structure-url-templates" class="headerlink"></a>
 
-```javascript
-{
-  "posts": [{
-    "id": "1",
-    "title": "One Type Purr Author",
-    "links": {
-      "authors": [{
-        "href": "http://example.com/people/9",
-        "id": "9",
-        "type": "people"
-      },
-      {
-        "href": "http://example.com/cats/1",
-        "id": "1",
-        "type": "cats"
-      }]
-    }
-  }]
-}
-```
+A top-level `"links"` object **MAY** be used to specify URL templates that can
+be used to formulate URLs for resources according to their type.
 
-### URL Template Shorthands <a href="#document-url-template-shorthands" id="document-url-template-shorthands" class="headerlink">¶</a>
-
-When returning a list of documents from a response, a top-level `"links"`
-object **MAY** be used to specify a URL template that should be used for all
-documents.
-
-Example:
+For example:
 
 ```javascript
 {
@@ -302,9 +416,12 @@ Example:
 }
 ```
 
-In this example, fetching `http://example.com/posts/1/comments` will fetch
-the comments for `"Rails is Omakase"` and fetching `http://example.com/posts/2/comments`
-will fetch the comments for `"The Parley Letter"`.
+In this example, fetching `http://example.com/posts/1/comments` will fetch the
+comments for `"Rails is Omakase"` and fetching
+`http://example.com/posts/2/comments` will fetch the comments for `"The Parley
+Letter"`.
+
+Here's another example:
 
 ```javascript
 {
@@ -321,25 +438,22 @@ will fetch the comments for `"The Parley Letter"`.
 }
 ```
 
-In this example, the `posts.comments` variable is expanded by
-"exploding" the array specified in the `"links"` section of each post.
-The [URL template specification][3] specifies that the default explosion is to
-percent encode the array members (e.g. via `encodeURIComponent()` in JavaScript)
-and join them by a comma, so in this example, fetching
-`http://example.com/comments/1,2,3,4` will return a list of all comments.
+In this example, the `posts.comments` variable is expanded by "exploding" the
+array specified in the `"links"` section of each post. The URI template
+specification [[RFC6570](https://tools.ietf.org/html/rfc6570)] specifies that
+the default explosion is to percent encode the array members (e.g. via
+`encodeURIComponent()` in JavaScript) and join them by a comma. In this example,
+fetching `http://example.com/comments/1,2,3,4` will return a list of all
+comments.
 
-[3]: https://tools.ietf.org/html/rfc6570
+The top-level `"links"` object has the following behavior:
 
-This example shows how you can start with a list of IDs and then upgrade to
-specifying a different URL pattern than the default.
-
-The top-level `"links"` key has the following behavior:
-
-* Each key is a dot-separated path that points at a repeated relationship.
-  For example `"posts.comments"` points at the `"comments"` relationship in
-  each repeated document under `"posts"`.
+* Each key is a dot-separated path that points at a repeated relationship. Paths
+  start with a particular resource type and can traverse related resources. For
+  example `"posts.comments"` points at the `"comments"` relationship in each
+  resource of type `"posts"`.
 * The value of each key is interpreted as a URL template.
-* For each document that the path points to, act as if it specified a
+* For each resource that the path points to, act as if it specified a
   relationship formed by expanding the URL template with the non-URL value
   actually specified.
 
@@ -372,25 +486,28 @@ Here is another example that uses a has-one relationship:
 }
 ```
 
-In this example, the author URL for all three posts is
+In this example, the URL for the author of all three posts is
 `http://example.com/people/12`.
 
 Top-level URL templates allow you to specify relationships as IDs, but without
 requiring that clients hard-code information about how to form the URLs.
 
-NOTE: In case of conflict, an individual document's `links` object will take
-precedence over a top-level `links` object.
+NOTE: In case of conflict, an individual resource object's `links` object will
+take precedence over a top-level `links` object.
 
-### Compound Documents <a href="#document-compound-documents" id="document-compound-documents" class="headerlink">¶</a>
+### Compound Documents <a href="#document-structure-compound-documents" id="document-structure-compound-documents" class="headerlink"></a>
 
-To save HTTP requests, it may be convenient to send related documents along
-with the requested documents.
+To save HTTP requests, responses may optionally allow for the inclusion of
+linked resources along with the requested primary resources. Such response
+documents are called "compound documents".
 
-Related documents **MUST** be included in a top level `"linked"` object, in
-which they are grouped together in arrays according to their type.
+In a compound document, linked resources **MUST** be included as resource
+objects in a top level `"linked"` object, in which they are grouped together in
+arrays according to their type.
 
-The type of each relationship **MAY** be specified in the `"links"` object with
-the `"type"` key. This facilitates lookups of related documents by the client.
+The type of each relationship **MAY** be specified in a resource-level or top-
+level `"links"` object with the `"type"` key. This facilitates lookups of linked
+resource objects by the client.
 
 ```javascript
 {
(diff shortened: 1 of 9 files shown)


steveklabnik:
I merged one or two of the other PRs, can you rebase this real quick :sob:

steveklabnik:
Also, @dgeb maybe put a link to https://github.com/dgeb/json-api/blob/v1rc1/format/index.md at the top so we can see the rendered version more easily?

krainboltgreene:
_Disclaimer: I came here specifically because I'm invested in the error API._

> "id" - A unique identifier for this particular occurrence of the problem.
> "href" - A URI that MAY yield further details about this particular occurrence of the problem.
> "status" - The HTTP status code applicable to this problem, expressed as a string value.
> "code" - An application-specific error code, expressed as a string value.

I am super happy with the work here except for the above. I seriously believe these are 4 ways of saying the same thing. Though I do admit `href`'s similarity is arguable, but a uri is a string to an error documentation and by definition of being a uri it's an _identifier_.

`id` and `code` are literally the same thing: Both application specific (though id isn't documented as such for some reason), both probably going to be strings in the real world, both specific to the error at hand.

And I've just never liked repeating HTTP Status Codes in the body of my message. It's in the header already.

steveklabnik:
I can certainly see that. In my mind, `id` is what your code checks, and `code` is what the programmer shows, to help them understand.

```
if id != 9001
   throw code
```

Does that make sense? Then again, you're right, the client could

```
if id != 9001
    throw http_client.get(href)
```

But that seems excessive.

> And I've just never liked repeating HTTP Status Codes in the body of my message. It's in the header already.

Right, you'd have something more specific. '409 conflict' comes with a body that says 'you forgot the required value foo'. 

Does that make sense? I can certainly see it your way, just laying out some of the why.

steveklabnik:
Oh, also cc @mjallday and crew (it won't let me @balanced/engineers here for some reason), since Balanced is one of the bigger deployments of JSON API. AFAIK, they should still be compatible, just using some `MAY`s rather than some `SHOULD`s.

krainboltgreene:
So it's HTTP status codes for individual leaves (fields in Rails case)? Bleergh, like most status codes don't apply for that sort of thing. Seems like shoe horning, but alright, maybe that's a thing that needs to happen.

That said, throwing an MAY on the `code` and `href` is a priority, or `id` and `href`. See? I firmly believe `id` and `code` are super interchangeable OR `code` gets subverted in your example (developer specific notes) to be the way the server sends error messages (`"Field needs to have 3 to 5 characters"`).

Eh, maybe it's no big deal, but I err on the side of cut these days.

steveklabnik:
I'm using the HTTP codes as an example, not that they should be HTTP status codes. They're application-specific codes, as I think the spec mentions.

> Eh, maybe it's no big deal, but I err on the side of cut these days.

+1, I wouldn't mind MAYing them.

krainboltgreene:
Yeah, okay, then I revert to my original `status`, `code`, `id`, (maybe) `href` are all ways of saying "Hey, this is the type of problem". MAYing all but one would be ideal IMHO. Thanks for listening.

kjg:
I'm a little concerned with the change that a singular resource still would have a plural "key". This seems inconsistent with the fact that one to one relationships have a singular key.

It also seems inconsistent that an error resource _should_ be an array always where as normal resources _should_ be an object.

steveklabnik:
> I'm a little concerned with the change that a singular resource still would have a plural "key".

This is basically for backwards compatibility with previous versions. You `SHOULD` use `data` as a key now instead.

> It also seems inconsistent that an error resource should be an array always where as normal resources should be an object.

Hmmmmm.

kjg:
> This is basically for backwards compatibility with previous versions.

Backwards compatibility as part of a _BREAKING CHANGE_ of having the singular resource be an object? Can you please expand on that future as to what compatibility exactly this maintains?

> You SHOULD use data as a key now instead.

For all resources? My reading made it sound like only variable type resources should have the "data" as a key. Singular type resources SHOULD have either the type as the key OR data as the key (with type based key listed as the first option)

dgeb:
@krainboltgreene Thanks for your feedback re: errors.

All of the members of an error object have a unique purpose and all are optional.

`id` and `code` differ in that `id` represents this particular _occurrence_ of the problem
(if your app wants to track _each and every_ error that occurs), while `code` represents
a particular error _type_. It's like the difference between a class and an instance.

`href` and `id` could in theory be identical, but it's not required. `href` should be used for lookups.

`status` is available primarily for bulk operations, as described in the "Other responses" section:

> When a server encounters multiple problems from a single request, the most
> generally applicable HTTP error code should be specified in the response. For
> instance, `400 Bad Request` might be appropriate for multiple 4xx errors or `500
> Internal Server Error` might be appropriate for multiple 5xx errors.

While a general `400` code might be returned when more than one operation fails, more specific status codes can be included in each individual error object.

dgeb:
@steveklabnik - @wycats and I backtracked on **SHOULD** for using `data` in general. The only time it's an absolute requirement is for polymorphic (i.e. multi-type) collections, in which case each resource **SHOULD** include its `type`.

dgeb:
> I'm a little concerned with the change that a singular resource still would have a plural "key". This seems inconsistent with the fact that one to one relationships have a singular key.

@kjg oh, the heartburn that this _minor_ issue has given me over the past few months.... We tried every manner of approach to resolving the singular / plural inflection issue and every solution had a problem with it (which I should write up thoroughly). We came to the conclusion that this is not about singulars and plurals as it is a string that represents a "resource type". For some APIs this might be singular; for some plural. Because most URLs reference plural collections, all the examples use the plural form.

One of the solutions was to simply use `"data"` everywhere, which has the distinct drawback in that it no longer identifies the type of the primary resource. For this reason, any use of `"data"` **MUST** be coupled with a type identifier in the resource object. The goal here is to ensure that a document can be interpreted on its own, without the context of a URL, and therefore needs to identify the type of each resource contained within.

andrewsardone:
> For this reason, any use of "data" MUST be coupled with a type identifier in the resource object. The goal here is to ensure that a document can be interpreted on its own, without the context of a URL, and therefore needs to identify the type of each resource contained within.

Independent of the use of the `"data"` key, I’d still push that JSON API should highly recommend people include a `"type"` key in each resource. In my experience, it’s very convenienent to rely on `"id"`, `"type"`, and `"href"` residing within a resource’s object body.

krainboltgreene:
@dgeb Alright, those explanations worked for me, though I'm not sure I'm cool about `status` just yet. Time will tell I suppose.

krainboltgreene:
@andrewsardone Wouldn't this be `rel` in web speak?

andrewsardone:
@krainboltgreene: just like in [_Variable Type Resource Representations_](https://github.com/dgeb/json-api/blob/v1rc1/format/index.md#variable-type-resource-representations-), I think plain ‘ol [resource objects](https://github.com/dgeb/json-api/blob/v1rc1/format/index.md#resource-objects-) should include a `"type"` key:

``` json
{
  "posts": {
    "id": "1",
    "title": "Rails is Omakase",
    "type": "posts"
  }
}
```

(I would also always include the `"href"` key)

format/index.md
```diff
@@ -5,166 +5,330 @@ title: "Format"
...
-(which are unique string identifiers, after all), it is not generally
-recommended. URLs can change, so they are unreliable for mapping a document to
-any client-side models that represent the same resource. It is recommended that
-URL values be left to the task of linking documents while `"id"` values remain
-opaque to solely provide a unique identity within some type.
+```javascript
+{
+  "data": [{
+    "id": "9",
+    "type": "people",
```
igorT, line 158: I might not be looking in a correct place, but it seems like object types are named in the plural? Is that a JSON Api convention?

dgeb, line 158: The resource type should match the name of the collection to which it belongs as defined by its URL. It's conventional to use the plural name of resources in URLs, which is why all the examples in this document follow that convention. However, if you accessed "people" at the `/person` URL, then the resource type would be "person".

steveklabnik:
I'm philosophically opposed to forcing people to include a type unless it's required to resolve an ambiguity. HTML doesn't require you to type its responses, nor does regular old JSON.

If we have to, we have to, but I think it's best if it's optional.

dgeb:
@steveklabnik I agree. 

To be clear: in this PR, `type` is only required in individual resource objects when it can't be determined from elsewhere in the document (i.e. collection keys, top-level links object, etc.).

Of course, individual implementations may choose to include `type` in every resource object for convenience, even if it's redundant.

paddycarver:
Eh, really disappointed to see reversion back to polymorphism. Is there rationale besides "It matches requests"?

Strongly-typed APIs are really nice to work with.

As for the errors, I note that we've stripped the ability to say "this part of your request is bad", leaving us just with "your request is bad and you should feel bad." Kind of disappointed about that, too. Not sure why we wouldn't point to a specific field or parameter using JSON pointers to say "here's what's wrong".

Otherwise, has anyone tried actually building--for example--a simple registration form with these error constraints? How do you display to the user that the "email address" field is not a valid email format? How do you display that the username has unacceptable characters? Are we really suggesting that codes should be unique to errors _and_ fields?

I've tuned out a bit, so I'm probably missing a lot, but we seem to have some compromises here that just... don't make sense. The errors object specifically looks like design-by-committee instead of an actual purpose-driven design.

steveklabnik:
@paddyforan this is _exactly_ why this PR is open. Let's work through that stuff! The previous stab at errors was too simple, and instead of not causing bikeshedding, caused bikeshedding. Nothing here is sacred!

paddycarver:
After a bit more consideration, my reservations on errors is that it tries to do a great many things and does none of them well.

The data being returned as "an error", as far as I can tell, is meant to:
- tell the application developer what went precisely went wrong at runtime (the error code). Unfortunately, this information isn't actionable at all unless a preposterous number of codes are created. How many different types of invalid user input does your API have? Because under this scheme, each of them needs its own unique code. Yikes.
- tell the user how to refer to their problem when contacting support (the id). Unfortunately, including this alongside the code without some very well-done explaining is going to confuse people... as we've seen above.
- tell the user and/or the client library developer where to seek more understanding about the error (the href). Unfortunately, without a clear indication of _what that href links to_, it's essentially useless. Does accessing that href give me more machine-readable information to present to the user? Does it give me a direct link to the API reference I violated so I can fix my mistake? Who is it meant for, and in what context?
- provide an error string, with the same problem: without knowing the audience of these, client libraries will have no idea whether to pass them on to the user or log it.

Has anyone actually tried _building_ an application that consumes an API that would meet these specifications? It seems very wishy-washy to me and hard to get right at this point, but we're going to pin a 1.0 on it?

If I had to make a suggestion for the errors, I'd say let's pick one or at most two purposes for the error returned and make sure it does a great job of them. Which means not everyone will get what they want, I imagine, but making errors good at _something_ is probably better than making them good at _nothing_.

dgeb:
@paddyforan - thanks for the feedback!

> Eh, really disappointed to see reversion back to polymorphism. Is there rationale besides "It matches requests"?

The default approach in this spec is to use strongly typed resources. However, I see polymorphic collections and relationships as a reality for many APIs, so it's important for the spec to address them.

A common example is an activity stream that comprises [multiple types of activities](https://developer.github.com/v3/activity/events/types/).

> As for the errors, I note that we've stripped the ability to say "this part of your request is bad", leaving us just with "your request is bad and you should feel bad."

Actually, each error resource can provide specificity to the developer and the end user regarding the error: 

The developer can use the `links` member to discern exactly which resources in the request are related to this error (e.g. `"links": {"posts": "123"}`)

The `path` member can be used to specify the relevant attribute for this error (e.g. `"path": "/title"`)

The `code` can be used by the developer when responding to this error (e.g. `"code": "1234" - client code might know that this means that an attribute can't be blank). It wasn't my intention for this to be presented to the user.

The `title` can be displayed to the user (e.g. "Title can't be blank") as can `detail` (useful for more complex errors).

The `id` could be referenced by the user when contacting support or by the developer when logging problems.

I _really_ should expand this section with examples, which I think could make these choices more clear. I'll try to come up with some soon - I think they'd be worthwhile in the spec.

> Has anyone actually tried building an application that consumes an API that would meet these specifications? It seems very wishy-washy to me and hard to get right at this point, but we're going to pin a 1.0 on it?

I share your reservations about hardening this spec immediately after so many changes. My preference would be to treat this as a release candidate. If this looks like it's 90% of the way there, let's merge it, kick the tires over a set period, and incorporate feedback. Of course, any feedback before a merge is welcome too :)

paddycarver:
Thanks for the in-depth answers!

I'm clearly doing nobody any favours by trying to comment on this while focusing on something else. I'm tied up for the night, but is it cool if I get back to you tomorrow about your responses above? Some are great, one I'd like to contest (because nobody has given me a good answer for it yet, and you seem like you might), and some I'd like to take a bit further, if they aren't explained when I take a deeper look.

In the meantime, I won't waste your time. Sound fair?

dgeb:
@paddyforan absolutely! Looking forward to your reply.

steveklabnik:
In general, specs should move slowly, for exactly the reasons mentioned. I'm not in a hurry to change some things and then rubber stamp it. Especially those with existing deployments should give some feedback before merge, and as Dan said, the merge would be like an RC1.

krainboltgreene:
Heh, it doesn't help I guess that this spec looks so well formed that it
feels like a final decision :)

After my initial reservations I've been pretty convinced of all the major
points in this spec update. I'll be passing it around at Hacknight to see
what my fellow developers think.

On Mon, Jun 9, 2014 at 9:24 PM, Steve Klabnik notifications@github.com
wrote:

> In general, specs should move slowly, for exactly the reasons mentioned.
> I'm not in a hurry to change some things and then rubber stamp it.
> Especially those with existing deployments should give some feedback before
> merge, and as Dan said, the merge would be like an RC1.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/json-api/json-api/pull/234#issuecomment-45567239.

## 

Kurtis Rainbolt-Greene, Hacker
Software Developer
1631 8th St.
New Orleans, LA
70115

format/index.md
```diff
@@ -5,166 +5,330 @@ title: "Format"
...
 }
 ```
 
-This simplifies processing, as you can know that documents will always be
-wrapped in arrays.
+This post is represented simply by its ID:
+
+```javascript
+{
+  "posts": "1"
```
kjg, line 100: What's the use case for representing a resource only by its id?

dgeb, line 100: This form can be used to [update to-one relationships](https://github.com/dgeb/json-api/blob/v1rc1/format/index.md#updating-to-one-relationships-)

paddycarver:
> The default approach in this spec is to use strongly typed resources. However, I see polymorphic collections and relationships as a reality for many APIs, so it's important for the spec to address them.

I'd contest this. I've yet to hear of a single polymorphic API that could not have been strongly typed. The example you link to is simply solved by creating an Event or Update resource that contains an update ID, a link to the resource updated, a timestamp for the event or update, and optionally the type of resource and/or type of event/update.

E.g.

``` json
{
    "updates": [
        {"id": "a2db423c", "resource": "http://api.yoursite.com/posts/12badc113", "action": "like", "timestamp": "2012-05-29T01:23:45Z"}
    ]
}
```

You still have the ability to display multiple types of resources, but you still hold on to your strong typing. It also contributes to a simpler and better understanding of the endpoint: it's meant to tie together a bunch of other resources, as evidenced by the fact that its most useful data is links.

> JSON API is designed to minimize both the number of requests

Not sure this should be a design goal. Shouldn't JSON API be purpose-built endpoints that are meant to serve the needs of a single page, then? That would be the most efficient in terms of the number of requests (a single request). Obviously the logical maximum is a strawman, but this seems diametrically opposed to hypermedia APIs.

> The following posts are represented as an array of resource objects:
> These posts are represented as an array of IDs:

These two alone making parsing a response annoying. But when you throw in

> These comments are represented by a single "collection" object:

I'm not sure how we're saying this isn't polymorphic. Parsing this in a non-dynamic language sounds like a nightmare to me, and this would probably be the point where I stop writing against an API, because it refuses to be consistent.

Whatever happened to the [Robustness Principle](http://en.wikipedia.org/wiki/Robustness_principle)? What do we need three different ways to express things?

> Variable type (i.e. "heterogenous") resources can not be keyed by type, so the type of each resource SHOULD be specified in an alternative manner.

Make that four different ways to express things.

> The "id" key in a resource object represents a unique identifier for the underlying resource, scoped to its type. Its value MUST be a string which SHOULD only contain alphanumeric characters, dashes and underscores.

Why are we limiting the characters allowed in an ID? For URL safety? Can't we just say that IDs need to be URL-encoded? This seems like a weird, arbitrary restriction.

In our top level, we have

> "links": URL templates to be used for expanding resources' relationships URLs.
> "linked": a collection of resource objects, grouped by type, that are linked to the primary resource(s) and/or each other (i.e. "linked resource(s)").

In our resource, we have

> The value of the "links" key is a JSON object that represents linked resources, keyed by the name of each association.

So if I understand correctly, we have a top level "links" property that defines links templates, a top level "linked" property that contains related resources, and a resource level "links" property that contains relayed resources. Can you guess what my criticism is? Let's be consistent. What about "links" and "href_template" or something?

> For example, the following post is associated with a single author, identified by ID:
> And here's an example of a linked author represented as a resource object:
> For example, the following post is associated with several comments, identified by their IDs:
> And here's an example of an array of comments linked as a collection object:

Again, is there any reason we can't just pick one and stick with it? We're trading ease of parsing a response for syntactic sugar that is keyed towards people writing responses by hand. I'm afraid I don't understand that trade-off.

> The URL for individual resources within a heterogenous collection SHOULD be formed by appending the resource's type and ID to the collection URL in the format: <type>:<id>. For example:
> 
> /favorites/cats:1

We're suggesting custom string parsing? Is there a reason we're not just using

> /favorites?cats=1

which is standards-based, semantically clean, and requires no new parsing? We want to filter the favourites down to the resource with a type of "cats" and ID of "1".

We could even do

> /favorites?type=cats&id=1

I'm not sure why we wouldn't take advantage of the query string parsing tools we already have. Unless we're suggesting that collection items can have sub resources... in which case that's yet another reason to not have heterogeneous collections

> /favorites?type=cats&id=1
> /cats/1/selfies

That example doesn't make much sense, but it should be indicative. It's hard to come up with an example of this that does make sense, which suggests the query string approach may not be an issue at all.

> A resource's relationship MAY be accessible at a URL formed by appending /links/<relationship-name> to the resource's URL. 

What if I have a resource named "links"?

How do I know whether `/users/paddy/links` refers to my links or relationships?

I've made it to the fetching resources section, but let's leave things there for now in an effort to not be overly critical. There's a lot of great stuff here, I'm just a bit concerned about some of the decisions being made that are markedly different from the last discussion I saw.

dgeb:
@paddyforan Thanks very much for the critique. I am open to trying to simplify this spec, as long as we are clear-eyed about the tradeoffs we are making.

Most of your critique is related to polymorphic collections and the implications of supporting them, so let's start there.

> I've yet to hear of a single polymorphic API that could not have been strongly typed.

Let's take your example:

```
{
    "updates": [
        {"id": "a2db423c", "resource": "http://api.yoursite.com/posts/12badc113", "action": "like", "timestamp": "2012-05-29T01:23:45Z"}
    ]
}
```

You've essentially translated a heterogenous collection into a collection of objects with has-one polymorphic relationships, represented by an `href`. To make this JSON API compliant, this could be represented as:

```
{
    "updates": [
        {"id": "a2db423c", "action": "like", "timestamp": "2012-05-29T01:23:45Z",
         "links": {"resource": {"type": "posts", "href": "http://api.yoursite.com/posts/12badc113"}}}
    ]
}
```

I've got no issue with the second representation after adding a `"type"` to the `"resource"`.

If you want to be able to leave off `"type"` from the linked resource, then you're giving up the ability to include that resource as a linked resource. This is a very big deal: your client will have to make at least two API calls serially in order to fully parse this collection. You're also giving up the ability for the client to know what to expect in response when following the `"href"`. To me, this makes it a non-starter to leave off `"type"`.

I guess there's still the question of whether polymorphism should be allowed for has-one relationships only vs has-many relationships. To me, this feels arbitrary and would need strong justification.

> JSON API is designed to minimize both the number of requests

I believe this is a design goal, and its important in this era of rich clients. Compound documents can provide so many efficiency benefits by reducing the number of round trips needed between a client and server.

There is definitely tension between providing efficiency and discoverability, but they are both worthy design goals and not mutually exclusive.

> What do we need three different ways to express things?

The alternative is to require the most complex and inclusive form. Instead of an array of ids as strings `["1", "2"]`, you would need an array of objects such as `[{"id": "1"}, {"id": "2"}]`.

The collection object is needed to represent a unique `"href"` for a collection of resources. We can't always provide arrays of members for very large collections. And URL templates can't always be generalized enough.

> Why are we limiting the characters allowed in an ID? For URL safety? Can't we just say that IDs need to be URL-encoded? This seems like a weird, arbitrary restriction.

See https://github.com/json-api/json-api/issues/190 for discussion.

> So if I understand correctly, we have a top level "links" property that defines links templates, a top level "linked" property that contains related resources, and a resource level "links" property that contains relayed resources.

The top level `links` object is very close in form to the resource level `links` object: both can provide hrefs and types. However, their defaults differ in terms of context. At the top level, a string will be assumed to be a URL template, while at the resource level, it will be assumed to be an id. I'm not sure that this discrepancy is worth changing the name at the top level, since the goal of both `links` objects is the same: to provide specificity and discoverability to links.

> /favorites/cats:1

Ok, I'm not crazy about this either. I admit it. However, if we accept polymorphic collections into this spec then we need a way to operate on their members. I appreciate your suggestions, but they don't cover deleting multiple members: 

```
DELETE /favorites/cats:1,dogs:3,dishwashers:5
```

Alternative suggestions are welcome :)

> What if I have a resource named "links"?

Links is (and has been) a reserved keyword at the top level. Technically, you could define a `links` resource keyed by `data`.

---

I hope that covers all your points, but please let me know if I've missed any. I do share your concerns about complexity and am very open to suggestions for simplification / clarification.

BRMatt:
First off, nice work on rewording the spec, it's great to have things explicitly spelled out.

After reading the section on errors I'm a little concerned by the fact that someone needs to loop through the errors collection to find out what kind of error is being returned. One of the things I liked about the [Nottingham media type](http://tools.ietf.org/html/draft-nottingham-http-problem-06), and APIs [such as Stripe's](https://stripe.com/docs/api#errors) is that there's a top level key that classifies the error, and implicitly, its severity.

HTTP status codes can only go so far in telling you the severity of the problem. For example, imagine I send a malformed json string as a POST body. The server would respond with a `400` status code and the error message would tell me the body was incorrect. In a separate request I send a correctly formatted payload, but some of the fields in the payload are incorrect - e.g. an email address supplied in the payload fails validation. In this case the server would also respond with a `400` status code.

In the first case this is a problem that I as a programmer have to solve, and the only thing I can tell the end user is that "we're having trouble talking to the API, we'll fix it ASAP". In the second case, this is an error caused by the end user, and it's something only they can fix.

Aside from deciding what kind of message to show the user it's also helpful for determining how an error from the API should be logged in the application. e.g. If I receive a `malformed_payload` response from the server that's something I want to log in my error tracking service so that I can fix it, but I don't want to get an alert every time someone tries to use an incorrect email address.

As the spec currently stands I'd need to loop through the errors and infer the overall problem/severity, or at the very least, inspect the severity of the first error in the array and assume that the rest of them are the same severity. A side effect of this approach is that I'd have to maintain a mapping of error codes to the "top level" kind of error, which I'd have to update every time a new error code is added to the API. For example `{"required": "validation_error", "malformed_payload": "malformed_payload", "invalid_email": "validation_error"}`.

If you squint at the two examples you could say that there are only two errors - `malformed_payload`, and `validation_failed`. In the case of the validation error, it happens that extra details are needed to work out what went wrong and help the user fix the problems.

As mentioned above, this is how the Nottingham media type works - it only allows one "error" per response. Unfortunately the spec doesn't provide a convention for specifying multiple problems related to the error, and this appears to have been one of the main concerns people have had with json-api adopting it.

Maybe we could compromise and have a single error for the entire response, then allow the API to provide a collection that drills down into the problems that caused the error:

``` json
{
  "error": {
    "id": "XXXXXXXX",
    "code": "validation_errors",
    "href": "http://dashboard.site.com/api-logs/XXXXXXXX",
    "details": [
      {
        "code": "required",
        "title": "The request headers implied the body was json, but we could not parse it as such",
        "path": "title"
      }
    ]
  }
}
```

On the other hand, the `malformed_payload` doesn't need any details to help identify what's wrong, so it could omit the `"details"` key:

``` json
{
  "error": {
    "id": "XXXXXXXX",
    "href": "http://dashboard.site.com/api-logs/XXXXXXXX",
    "code": "malformed_payload",
  }
}
```

WRT to the keys that could be allowed at the "top level", I'd suggest allowing all of the current keys, bar `"links"` and `"path"`, as they are only necessary if you're drilling down into an error. I'd also argue that we could remove `"id"` and `"href"` from the details in the collection, as based on the above explanation for these fields:

> id and code differ in that id represents this particular occurrence of the problem
> (if your app wants to track each and every error that occurs), while code represents
> a particular error type. It's like the difference between a class and an instance.
> href and id could in theory be identical, but it's not required. href should be used for lookups.

It's unlikely that you'd provide endpoints for each individual "detail" of an error, instead you'd provide an endpoint or an id that could describe the error for the entire request, which would include all the details.

Thoughts?

dgeb:
@BRMatt Thanks very much for your thoughts on errors.

> HTTP status codes can only go so far in telling you the severity of the problem.

The HTTP status code can be specified as `"status"`, while an application-specific error code can be specified as `"code"`. This code should be used by client applications to determine an appropriate response, whether it be displaying a message to the user or alerting the developer.

I appreciate your attempt to merge Nottingham with this proposal, but I'm concerned about coalescing any number and variety of errors into a single error object. It would seem necessary to pick the most severe code for that object, which could feel pretty arbitrary for implementers. Furthermore, having potentially unrelated details nested under this error seems inappropriate.

This wouldn't be an issue if it was always appropriate for a server to stop processing as soon as it hit an error. However, there are cases when a server may want to continue.

Take a request for multiple posts:

```
GET /posts/1,2,3,4
```

It's quite possible for posts 1 and 4 to be not found (404) while post 3 is forbidden (403). I'd like to be able to return these errors in a single response while maintaining their individual identity.

BRMatt:
> The HTTP status code can be specified as "status", while an application-specific error code can be specified as "code". This code should be used by client applications to determine an appropriate response, whether it be displaying a message to the user or alerting the developer.

@dgeb thanks for clarifying this. However, doesn't this still require the developer to maintain a mapping of `"code" -> "severity"`? For example, if I add a new error code `end_time_is_before_start_time` to the API, my client app wouldn't be able to recognise that this is a validation error without a top-level error, or at the very least, some severity tag in the error that identified it as a validation problem.

The developer may need to tweak the client to add better contextual feedback, but the app should still be able to identify this as a problem with the data the user entered and use the other error attributes to display some some feedback, without requiring developer intervention.

> I'm concerned about coalescing any number and variety of errors into a single error object. It would seem necessary to pick the most severe code for that object, which could feel pretty arbitrary for implementers.

This is a reasonable point. My first reaction was to suggest using a generic top-level error such as `some_resources_not_available` to indicate that only some of the pages could be returned, then use the details array to dive into the specifics. That said having to create many "partial success" errors to coalesce these errors would become tedious.

Maybe the best option would be to add a `"severity"` tag to each error?

I'm still not completely sold on including `"id"` and "`href`" in every error if they're likely to be the same in each error, but it makes more sense to do it that way and be consistent than to force folks to nest the errors in an wrapper object:

``` json
{
  "error": {
    "id": "XXXX",
    "href": "http://foo.com",
    "errors": [ ... ]
  }
}
```

Especially when the service offering the API may not offer the ability to inspect API logs, making `"id"`/`"href"` redundant:

``` json
{
  "error": {
    "errors": [ ... ]
  }
}
```

dgeb:
> However, doesn't this still require the developer to maintain a mapping of "code" -> "severity"?

Yes, that's true. Since these are completely custom codes, it's possible to include a prefix in the code to indicate a severity level for easier parsing.

> Maybe the best option would be to add a "severity" tag to each error?

I'm not sure about putting this in the spec, since it amounts to yet another application-specific "code". Of course, there's no problem including this in your particular API. If this becomes a general enough practice, we could always add it to the spec later (and it would be non-breaking). We're always open to feedback from implementers.

> I'm still not completely sold on including "id" and "href" in every error if they're likely to be the same in each error

`"id"` should be unique for each _occurrence_ of an error and can be used for tracking purposes. The intention was for each error in a single `"errors"` array to have its own id. Behind the scenes, they could be tied together to a common request, but that doesn't need to be part of the spec.

krainboltgreene:
Didn't HAL decide that reserved words were prefixed by a character that is very unlikely to appear in a column/model name? AKA "_href" or "_links"?

paddycarver:
@dgeb more replies! In case I haven't said it yet, your work on this is great and I apologise that you're getting the brunt of objections from me for no reason other than (as far as I can tell) you happened to formalize the discussions I missed. I appreciate you taking the time to answer my concerns.

That said, I still want to rip the polymorphism and parsing complexity out of this, as it is far past the level of "dealbreaker" for me.

> I've got no issue with the second representation after adding a "type" to the "resource".
> 
> If you want to be able to leave off "type" from the linked resource, then you're giving up the ability to include that resource as a linked resource. This is a very big deal: your client will have to make at least two API calls serially in order to fully parse this collection. You're also giving up the ability for the client to know what to expect in response when following the "href". To me, this makes it a non-starter to leave off "type".

I'm cool with tweaking the syntax like that. My point was "Look, here's that concept modeled without polymorphism!" So again: is there an actual need someone can show me for using polymorphism? To this day, I've yet to find one.

> I guess there's still the question of whether polymorphism should be allowed for has-one relationships only vs has-many relationships. To me, this feels arbitrary and would need strong justification.

In case it's not clear, I'm arguing polymorphism should not be allowed anywhere. I've still yet to see a reason that justifies its inclusion.

> I believe this is a design goal, and its important in this era of rich clients. Compound documents can provide so many efficiency benefits by reducing the number of round trips needed between a client and server.
> 
> There is definitely tension between providing efficiency and discoverability, but they are both worthy design goals and not mutually exclusive.

I can understand that, and even support it. i think my objection is more grounded in the fact that efficiency and discoverability (at least) are identified as design goals with tension, but they're not identified as such in the spec. Only efficiency is. Considering the tension, this seems... wrong.

Not to mention that "readability" and "ease of compliance" and "orthogonality" should all be design goals as well, and are also all in tension. Just listing "efficiency" seems to start off on the wrong foot.

> The alternative is to require the most complex and inclusive form. Instead of an array of ids as strings ["1", "2"], you would need an array of objects such as [{"id": "1"}, {"id": "2"}].
> 
> The collection object is needed to represent a unique "href" for a collection of resources. We can't always provide arrays of members for very large collections. And URL templates can't always be generalized enough.

Yes. I'm basically saying "Why did we decide against the array of objects?"

And there's nothing with the collection object per se. What I _am_ saying is it should not be keyed the same as a resource object. "comments_collection"? Sure. "comments"? No. That makes parsing this so much harder. The object itself isn't in question, just how we're returning it.

> Ok, I'm not crazy about this either. I admit it. However, if we accept polymorphic collections into this spec then we need a way to operate on their members. I appreciate your suggestions, but they don't cover deleting multiple members:

Which seems like a _very good reason_ to not accept polymorphic collections, as we still haven't identified a scenario in which they're necessary.

Should we decide to do that anyways, there are still ways to delete:

?cats=1,2&dogs=3

?type1=cats&id1=1&type2=cats&id2=2&type3=dogs&id3=3

I'm still not seeing the need to drift away from predefined standards.

> Links is (and has been) a reserved keyword at the top level. Technically, you could define a links resource keyed by data.

Noooo, kill data. Kill it with fire.

I'd agree with @krainboltgreene that perhaps we should be prefixing this reserved stuff. I've actually used "Links" as a resource in the past.

@BRMatt for errors, I've found it's helpful to use the code as an application-specific "type" of error. "invalid format" "invalid value" "overflow" "insufficient" are all different codes I use on a general basis, and make it really easy to lump errors into "my fault" and "user's fault".

krainboltgreene:
> ?type1=cats&id1=1&type2=cats&id2=2&type3=dogs&id3=3

Looks terrible.

On Wed, Jun 11, 2014 at 8:35 PM, Paddy Foran notifications@github.com
wrote:

> @dgeb https://github.com/dgeb more replies! In case I haven't said it
> yet, your work on this is great and I apologise that you're getting the
> brunt of objections from me for no reason other than (as far as I can tell)
> you happened to formalize the discussions I missed. I appreciate you taking
> the time to answer my concerns.
> 
> That said, I still want to rip the polymorphism and parsing complexity out
> of this, as it is far past the level of "dealbreaker" for me.
> 
> I've got no issue with the second representation after adding a "type" to
> the "resource".
> 
> If you want to be able to leave off "type" from the linked resource, then
> you're giving up the ability to include that resource as a linked resource.
> This is a very big deal: your client will have to make at least two API
> calls serially in order to fully parse this collection. You're also giving
> up the ability for the client to know what to expect in response when
> following the "href". To me, this makes it a non-starter to leave off
> "type".
> 
> I'm cool with tweaking the syntax like that. My point was "Look, here's
> that concept modeled without polymorphism!" So again: is there an actual
> need someone can show me for using polymorphism? To this day, I've yet to
> find one.
> 
> I guess there's still the question of whether polymorphism should be
> allowed for has-one relationships only vs has-many relationships. To me,
> this feels arbitrary and would need strong justification.
> 
> In case it's not clear, I'm arguing polymorphism should not be allowed
> anywhere. I've still yet to see a reason that justifies its inclusion.
> 
> I believe this is a design goal, and its important in this era of rich
> clients. Compound documents can provide so many efficiency benefits by
> reducing the number of round trips needed between a client and server.
> 
> There is definitely tension between providing efficiency and
> discoverability, but they are both worthy design goals and not mutually
> exclusive.
> 
> I can understand that, and even support it. i think my objection is more
> grounded in the fact that efficiency and discoverability (at least) are
> identified as design goals with tension, but they're not identified as such
> in the spec. Only efficiency is. Considering the tension, this seems...
> wrong.
> 
> Not to mention that "readability" and "ease of compliance" and
> "orthogonality" should all be design goals as well, and are also all in
> tension. Just listing "efficiency" seems to start off on the wrong foot.
> 
> The alternative is to require the most complex and inclusive form. Instead
> of an array of ids as strings ["1", "2"], you would need an array of
> objects such as [{"id": "1"}, {"id": "2"}].
> 
> The collection object is needed to represent a unique "href" for a
> collection of resources. We can't always provide arrays of members for very
> large collections. And URL templates can't always be generalized enough.
> 
> Yes. I'm basically saying "Why did we decide against the array of objects?"
> 
> And there's nothing with the collection object per se. What I _am_ saying
> is it should not be keyed the same as a resource object.
> "comments_collection"? Sure. "comments"? No. That makes parsing this so
> much harder. The object itself isn't in question, just how we're returning
> it.
> 
> Ok, I'm not crazy about this either. I admit it. However, if we accept
> polymorphic collections into this spec then we need a way to operate on
> their members. I appreciate your suggestions, but they don't cover deleting
> multiple members:
> 
> Which seems like a _very good reason_ to not accept polymorphic
> collections, as we still haven't identified a scenario in which they're
> necessary.
> 
> Should we decide to do that anyways, there are still ways to delete:
> 
> ?cats=1,2&dogs=3
> 
> ?type1=cats&id1=1&type2=cats&id2=2&type3=dogs&id3=3
> 
> I'm still not seeing the need to drift away from predefined standards.
> 
> Links is (and has been) a reserved keyword at the top level. Technically,
> you could define a links resource keyed by data.
> 
> Noooo, kill data. Kill it with fire.
> 
> I'd agree with @krainboltgreene https://github.com/krainboltgreene that
> perhaps we should be prefixing this reserved stuff. I've actually used
> "Links" as a resource in the past.
> 
> @BRMatt https://github.com/BRMatt for errors, I've found it's helpful
> to use the code as an application-specific "type" of error. "invalid
> format" "invalid value" "overflow" "insufficient" are all different codes I
> use on a general basis, and make it really easy to lump errors into "my
> fault" and "user's fault".
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/json-api/json-api/pull/234#issuecomment-45820817.

## 

Kurtis Rainbolt-Greene, Hacker
Software Developer
1631 8th St.
New Orleans, LA
70115

ahacking:
Just looking at PATCH and JSON Pointer, hmmm...

JSON Pointer refers to array items by their array index so the premise that JSON-API can represent all of a domains resources is not really going to work as the array index can change and move about as changes occur.  Imagine a million customer records ordered by whatever. We put invariant 'ids' in objects for a reason, whether they be v4 UUID's or monotonically increasing integers and then JSON Pointer doesn't care about that, and PATCH based on RFC 6902 also can't care about your ids unless you get rid of arrays completely and make collections objects with their property names being the actual ids of the members, then you can't easily add an item without prescribing the id on the client (which actually goes in favour of v4 UUIDs).

When RFCs are not useful for a given purpose, don't use them. The basic claim that you can make updates of multiple resources with PATCH cannot be realised because you cannot know what the ordinal number of a resource is in the total array space without the server telling you first and blink it changes again because someone changed an existing record, added a few or deleted some. JSON Pointer and JSON PATCH is useless used this way and is practically limited to paths _within_ a resource which ideally should not be subject to concurrent changes that could affect the array ordering and hence the indexes, this really seems to rule out updating multiple resources.  I can also see that array indexes are very context specific since access controls will limit what is visible to each user/client differently from request to request, heaven forbid someone change the 'shared' flag on a resource as everyone's resource indexes will change!

dgeb:
@ahacking Thank you for bringing this up. I should update this spec to include mention that all collections should be treated as sets, keyed by id, when forming JSON Pointers.

Please see the discussion [here](https://github.com/json-api/json-api/issues/176#issuecomment-33819540) and [here](https://github.com/json-api/json-api/issues/193) regarding PATCH equivalent structures.

dgeb:
@paddyforan Your point is well taken regarding the tension between these design goals. I've added [a clarification](https://github.com/dgeb/json-api/commit/88852c310a24bd4084e4a3d388d84e4e6572c109).

@ahacking I've also added [a clarification regarding resource URLs](https://github.com/dgeb/json-api/commit/e0ef720087ee57fdb87785cd8edd4678b95d1219) as well. The same rules apply to determining URLs and PATCH paths - they are complementary when forming JSON Pointers to identify a resource.

ahacking:
Yes those discussions are exactly what I was getting at. You need patch compatible structures.

This seems to require client assigned ids such as v4 uuids since the client can't invent a server assigned id unless a special client id value space is reserved which hopefully does not clash with existing ids. If anyone has ids based on say name or title this can never be workable without an escaping/encoding mechanism eg prefix client ids with '_' and if any id client or server has leading underscores then use double underscore for all leading underscores. This always means an odd number of leading underscores for client ids.

The use of an api which allows client assigned ids has different usage than one that uses server assigned ids. With v4 uuids its the cleanest possible solution, with cids you have an additional mapping for whatever the server assigns and it can make web clients that have browser urls for resources behave in unexpected ways as cid's are typically ephemeral and unlikley to be persisted.

So where does JSON API draw the line?

dgeb:
@ahacking please see the sections on [IDs](https://github.com/dgeb/json-api/blob/v1rc1/format/index.md#resource-ids-) and [client-generated IDs](https://github.com/dgeb/json-api/blob/v1rc1/format/index.md#client-generated-ids-).

ahacking:
@dgeb yes that confirms the issue for me. 

If collections are represented in a patch compatible fashion then both 'clientid' and 'id' share the same value space as the property names on the object representing the collection.

There needs to be a way to distinguish and separate clientids from server ids so clashes never occur when using patch compatible structures.

dgeb:
@ahacking This seems to be an pretty narrow problem that should only come up when performing a bulk patch operation in which a resource is both created and then updated or deleted in the same request. 

The creation / deletion scenario seems really rare and slightly nuts:

```
PATCH /
Content-Type: application/json-patch+json

[
  { "op": "add", "path": "/photos/-", "value": {"title": "beach", "clientid": "a"} },
  // ...
  { "op": "remove", "path": "/photos/a"} }
]
```

The creation / update scenario seems more likely as part of a bulk sequential update:

```
PATCH /
Content-Type: application/json-patch+json

[
  { "op": "add", "path": "/photos/-", "value": {"title": "beach", "clientid": "a"} },
  // ...
  { "op": "replace", "path": "/photos/a/title", "value": "I need a vacation"} }
]
```

The far more common scenario is that multiple resources are created and linked in the same request:

```
PATCH /
Content-Type: application/json-patch+json

[
  { "op": "add", "path": "/photos/-", "value": {"title": "beach", "clientid": "a"} },
  { "op": "add", "path": "/notes/-", "value": {"clientid": "b", "text": "I need a vacation", "links": {"photos": {"clientid": "a"}} }}
]
```

In the last scenario, `"clientid"` can still be used to establish links without needing to use it in a URL / path.

So, what do we do about the first two scenarios?
- A) Not allow them in the first place by not allowing clientids to be used in URLs / paths. These ops would only be allowed with UUIDs, identified as `"id"`.
- B) Defer to clientid when identifying any id, since a map must already be maintained by servers that support clientids.
- C) Use a prefix, such as 'clientid:a' when referencing clientids in URLs / paths (I know, this is done already for heterogenous collections).

I do think that this is a pretty narrow problem best solved in its own issue and not a showstopper.

dgeb:
This thread is growing quite long. My preference would be to merge this sooner than later, treat it as an RC1, and commence discussions in more focused issues. I have a very busy week coming up, starting now, and won't be able to be as active here as I have been.

I'm fine putting this on ice for a bit as well if everyone wants to give it more time / consideration. I still want to consider some of the finer points myself, like the clientid issue above, which I don't see as a blocker.

@steveklabnik @wycats - thoughts?

ahacking:
@gdeb I am thinking the simplest use case of adding a resource with a related resource using PATCH with patch compatible structures (ie no collection arrays, just objects with ids as keys).

```
[
  { "op": "add", "path": "/photos/_a", "value": {"title": "beach", "clientid": "a"} },
  { "op": "add", "path": "/notes/_b", "value": {"clientid": "b", "text": "I need a vacation", "links": {"photos": ["/photos_/a"]}}}
]
```

Note the use of JSON Pointers with a patch compatible collection approach using clientid values as keys, instead of a collection array.  Also the use of links with a JSON pointer.

When you use the 'id' or 'clientid' value as a key in the JSON Pointer then you have to worry about clashes between real ids and client ids and need a mechanism to separate and distinguish them. I suggested an example prefixing (somewhat private) client ids with underscore to distinguish from real ids in JSON Pointers. However you need to handle cases where ids actually have a leading underscore, hence suggesting use of double underscores on leading underscores as an escape mechanism. By this logic a client id will always have an odd number of leading underscores.  I don't particularly care what the escaping mechanism is, just that you need one if you want to use JSON Pointers with combinations of real ids and clientids.

I don't like any of this really, but JSON Pointer and JSON Patch are what they are and using array indexes to reference collection items is unworkable outside of a single resource and is still a problem if the resource has changed.

PATCH needs to work against a reference model where collections are objects, which means collections should be objects and not arrays, and links should be JSON Pointers too.

The other option is borrow what you can where sensible from those RFC's but do what actually works best and report findings into the RFC process.  From where I stand JSON Patch and JSON Pointer have some severe assumptions and limitations and we are trying to work around it but is not exactly clean.

I am witnessing all the XML things being repeated in JSON... here we go again...

dgeb:
@ahacking You raise good points.

> The other option is borrow what you can where sensible from those RFC's but do what actually works best and report findings into the RFC process. From where I stand JSON Patch and JSON Pointer have some severe assumptions and limitations and we are trying to work around it but is not exactly clean.

Yes, we are bending JSON Patch and Pointer to our purposes but aren't completely compliant if we are adding to `/posts/-` (an array path) instead of `/posts/_a` (an object path) and then accessing that same resource at `/posts/_a`. We should be clear-eyed about whether we want to break compliance like this.

> I don't particularly care what the escaping mechanism is, just that you need one if you want to use JSON Pointers with combinations of real ids and clientids.

Yep - the [three options above](https://github.com/json-api/json-api/pull/234#issuecomment-45884870)

ahacking:
@dgeb Another thing that I wasn't clear on was at what point does the `clientid` get replaced with the servers id?

If the response returns objects, do those objects use the `clientid` or the server assigned `id` as the key when a JSON Pointer compliant structure is used?  If the collection key on a response always represents the objects `id` does the `id` also need to be carried in the actual object as well? (it just seems like extra overhead and more chance to get an inconsistency). Obviously `clientid` should always be included in the object.

My thinking is that if I do a POST/PUT/PATCH and the created/updated objects are returned in the response and then I later do a GET, I would like the response to be consistent aside from carrying a `clientid` if one was given in the POST/PUT/PATCH request.  Said another way, always use the real id as the collection key when the object has one.  It begs the question what to do when objects don't have ids, at that point we need something to use as a collection key, perhaps an index/position number?

dgeb:
@ahacking Any time `clientid` is passed to a server, the server should return it in the resource object. That's the only place it should appear in the response.

> It begs the question what to do when objects don't have ids, at that point we need something to use as a collection key, perhaps an index/position number?

Objects that don't have ids should never be referenced as members of a collection.

gdeb:
@ahacking @dgeb 

Hmm, just to let you know, gdeb != dgeb...  Also, hello dgeb, nice nickname!

dgeb:
hey @gdeb - sorry that you got dragged into this :)

ahacking:
The more I look at this the more I am tempted to say EVERY object has a v4 UUID and go and play somewhere else if you want to make life hard. But I realise that such a hard nosed approach requires the database to use v4 UUIDs as their primary keys or to carry (v4) UUIDs.  You can't easily add JSON-API support to existing datasets but if JSON-API is setting a good example maybe that is a sensible line in the sand.

If I was to profile JSON-API without a doubt I would mandate (v4) UUIDs everywhere because of the clear advantages at all levels of the stack, they are invariant, clients can build complex object graphs, servers can easily shard for scalability.

I think JSON-API should either say all resources have a (v4) UUIDs, there are no client ids, OR we bear the cost of maintaining two profiles, JSON-API and JSON-API-LIFE-IS-HARD.  I think it is important if you want to support the life-is-hard approach then structure the spec as two separate profiles to keep the easy way clean and the all the ifs buts and maybes in the life-is-hard spec.

ahacking:
@gdeb sorry... was messaging on my phone without the nice auto-complete.

dgeb:
@ahacking I agree that v4 UUIDs are simpler conceptually and provide many advantages. I was trying to provide a bridge for implementations that don't support UUIDs, but this approach does have its own costs for implementation.

If this non-UUID clientid solution is seen by you and others as too half-baked and/or complicated for inclusion, I'm fine with removing it from the PR and we could discuss it separately.

dgeb:
As far as the earlier discussion re: PATCH / Pointer, there really is no good alternative to allowing `add` ops to the end-of-array path (e.g. `/posts/-`) if we want to support scenarios in which the client does not generate ids.

ahacking:
@dgeb. Agreed. I will always use uuids but I really would like a simple one way of doing things. Nail the plural singular thing (I prefer plural always as it matches endpoint names) I always want a collection object keyed by id as the definitive reference model given we want to use JSON pointer. I want this even for singular resources. Uniform structure always regardless of cardinality.

I am happy to have links in a nested object so that generic processing logic can wire things up.

I think there is a need for meta info in objects however and that is how I have carried errors to date in my APIs but I'll go along with problems and JSON pointers. I do want the ability to have custom attributes in my errors/problems which could be useful depending on the type of error (eg could be suggestions for alternative names or other hints for validation errors).

I am really wanting simplicity, one way for everything and minimum repetition hence not wanting id in the object but on the collection key since it has to be put there for JSON pointer to work generally. Then we have a uniform structure, just resources housed in collection objects which are housed in a root object and JSON pointers to link up the object graph.

As for polymorphic concerns I think let users define whatever named collections they like and don't worry about prescribing types. Its only when you describe a generic data collection in the spec that it became a concern. I would stay silent on it. If people want to use a brown bag let them and let them decide how to distinguish types. The id will need to be unique across types because all types will share the same key space when they are all in the same object and that is again another win for (v4) uuids and a whole lot of stuff JSON API can stay out of.

dgeb:
Thanks for everyone's feedback. I'm going to explore a simplification of this PR in a new branch, which may not be complete for another week or more. I do think that there's some fat to be trimmed from the spec and some ideas that could be more fully baked. I'm primarily concerned with  polymorphism and client ids.

Please feel free to continue to comment here or on other issues. As I mentioned above, I've got a busy week ahead and may not be as responsive as I've been.

ahacking:
@dgeb Cheer's, I look forward to seeing the updated spec

krainboltgreene:
I need an example of what an error response looks like. I wasn't able to reason with the given description.

dgeb:
Please review #237, which represents the simplification of this PR that I promised. The differences are contained in https://github.com/dgeb/json-api/commit/8f831e99219793f43d9382a0d03528e87afd18eb

dgeb:
Closing in favor of #237

PR #239, opened 2014-07-07
Update examples to use single objects instead of arrays

whatthewhat:
This replaces arrays with a single object as defined in https://github.com/json-api/json-api/pull/237

I'm not sure, but on the front page it might look better to leave the array and add a second post..

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -380,13 +380,13 @@ Here's another example:
   "links": {
     "posts.comments": "http://example.com/comments/{posts.comments}"
   },
-  "posts": [{
+  "posts": {
     "id": "1",
     "title": "Rails is Omakase",
     "links": {
       "comments": [ "1", "2", "3", "4" ]
     }
-  }]
+  }
 }
 ```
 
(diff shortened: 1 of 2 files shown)


format/index.md
```diff
@@ -380,13 +380,13 @@ Here's another example:
   "links": {
     "posts.comments": "http://example.com/comments/{posts.comments}"
   },
-  "posts": [{
```
BRMatt, line 383: I'm not sure this instance should be changed. My understanding of the spec is that a single resource such as `/posts/1` could be represented as a single object, where as a collection of resources `/posts` should be represented as a collection, even if there is only one item in the collection. That said, I could be misreading the spec - it doesn't seem to clarify whether a "singular resource" is when a single resource is returned, or when a single resource is requested.

whatthewhat, line 383: @BRMatt you are definitely right here

What got me confused is that on the front page it's not obvious what is being fetched, a resource or a collection

dgeb:
@BRMatt @whatthewhat I just created #240 which attempts to clarify the singular / plural issue with different wording. Please let me know if this change works for you.

PR #240, opened 2014-07-07
Reword singular -> individual and plural -> collection.

dgeb:
This clarifies that collections can contain any number of resources,
and that different endpoints will always return or expect an
individual resource vs. a collection of resources.


BRMatt:
:+1:

whatthewhat:
@dgeb this does read clearer :+1:

dgeb:
@BRMatt @whatthewhat thanks for confirming! 

@steveklabnik please merge if you agree.

steveklabnik:
Seems good, thanks. :)

PR #241, opened 2014-07-07
Added SQLAlchemy-JSONAPI

ColtonProvias:
I'm currently working on a JSON API serializer for SQLAlchemy (and a JSON API adapter for flask, but that I will post later).  It's still basic but I'm looking to get as close to full compliance with the spec as I can!


steveklabnik:
Oh awesome! thank you!

Issue #244, opened 2014-07-14
Mixed formats for relationship URLs?


wrenoud:
The example of relationship URLs in [Document Structure: URL Templates](http://jsonapi.org/format/#document-structure-url-templates) does not conform to the specification in [URLs: Relationship URLs](http://jsonapi.org/format/#urls-relationships).

Specifically the `/links` portion of the path is omitted, i.e.

http://example.com/posts/1/comments vs
http://example.com/posts/1/links/comments

Is this indended as an alternate url form or is this a mistake. If the later I would be very happy to submit a pull request to change it.

steveklabnik:
JSON API does not mandate the structure of your URLs, so this is not strictly wrong.

(that said, it'd be nice to have consistent examples)

dgeb:
> JSON API does not mandate the structure of your URLs, so this is not strictly wrong.
> 
> (that said, it'd be nice to have consistent examples)

Agreed. We should probably stick with examples that are consistent with the recommendations in the rest of the spec.

I'd prefer that this example follow the [filtering recommendations](http://jsonapi.org/format/#fetching-filtering) rather than contain a nested URL:

```
"posts.comments": "http://example.com/comments?posts={posts.id}"
```

@wrenoud a PR would be great - thanks for catching this.

BTW, if an API does choose to use nested URLs instead of top-level, filtered resources, I would discourage including `links` in the URL to nested resources. There is a distinction between a URL to the link itself and the related resource which can be conveyed by including `links` in a URL. This allows for distinctions between, say, creating / deleting a resource vs. creating / deleting a link to a resource.

wrenoud:
Thanks for the clarification @dgeb, that makes more sense wrt the convention for `links`. I did not glean that understanding from my scan of the format.

PR #245, opened 2014-07-14
Changed example to use filtering

wrenoud:
Changed the first URL Templates example to use filtering to avoid confusion with nested URLS. Addresses issue #244.


dgeb:
@wrenoud awesome - thanks!

PR #259, opened 2014-08-07
Simplify the language around the Location header on create

ethanresnick:
Before, there was a paragraph aimed at the case of a single resource being created and a separate one aimed at the case of a collection of resources being created. I merged these into one statement, which simply says the Location URL must locate _all_ created resources.

If, to be super explicit, you want to keep this as two separate paragraphs, then the paragraph that's talking about the one resource case should start with the clause "If a single resource is created, ". (Otherwise, its mention of a "primary resource is confusing and arguably contradictory.)

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -825,7 +825,7 @@ When one or more resources has been created, the server **MUST** return a `201
 Created` status code.
 
 The response **MUST** include a `Location` header identifying the location of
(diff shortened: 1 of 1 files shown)


dgeb:
This does seem both simpler and clearer. Thanks!

PR #261, opened 2014-08-11
Clarify the fields spec to allow extra fields, fix #260

### LINKED ISSUE #260: Should resource urls be specified when "fields" is forced?
I'm not sure what I must do when the "fields" option is used:
- let the "non-resources" fields like "href", "type" or "links"?
- remove them if the end user do not allow them explicitly

I think there is room for interpretation reading the spec,
and I find this response totally ok for a `posts?fields=body` request:

``` json
  [{
    "id": "1",
    "href": "http://example.com/comments/1",
    "body": "Mmmmmakase"
  }, {
    "id": "2",
    "href": "http://example.com/comments/2",
    "body": "I prefer unagi"
  }]
```

http://jsonapi.org/format/#fetching-sparse-fieldsets
http://jsonapi.org/format/#document-structure-resource-urls

Thx for your help!

damienalexandre:
Related to #260,
I do not know if using the word "moderately" in a specification is a good idea,
open to suggestions.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -699,6 +699,9 @@ An endpoint **SHOULD** return a default set of fields in a resource object if no
 fields have been specified for its type, or if the endpoint does not support use
 of either `fields` or `fields[TYPE]`.
 
+An endpoint **MAY** also choose to always return a limited set of 
+non-specified fields, such as `id` or `href`.
+
 Note: `fields` and `fields[TYPE]` can not be mixed. If the latter format is
 used, then it must be used for the primary resource type as well.
 


dgeb:
Hmmm... the phrasing around "moderately" does sound a bit off to me. What do you think of this:

```
An endpoint **MAY** also choose to always return a limited set of 
non-specified fields, such as `id` or `href`.
```

damienalexandre:
Thx, your version is better!

dgeb:
Great! :shipit:

Issue #250, opened 2014-07-31
Clarify top level "links" key.


ethanresnick:
If I understand correctly, there are two ways to specify a key–value pair in the top-level `links` section: 
- `"path": "url-template"`, and 
- `"path": {"type": "type", "href": "url-template"}`.

But the URL Templates section, in describing the behavior of the `links` section, only describes the first case. Specifically, it says: "The value of each key is interpreted as a URL template." But, to accommodate the second case, it should read (something like): "For each key, if its value is a string, then its value is interpreted as a url template; if its value is an object, its value's `href` property is interpreted as a url template."

dgeb:
@ethanresnick Good catch! Your understanding is correct. If you're up for it, a PR to clarify these two distinct possible representations would be appreciated.

ethanresnick:
I'm closing this, as I've already submitted the relevant PR.

PR #262, opened 2014-08-12
Clarify top-level links

ethanresnick:
Addresses #250.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -372,7 +372,7 @@ For example:
 In this example, fetching `http://example.com/comments?posts=1` will fetch the
 comments for `"Rails is Omakase"` and fetching
 `http://example.com/comments?posts=2` will fetch the comments for `"The Parley
-Letter"`. The use of a URL template obviates the need for each post to specity 
+Letter"`. The use of a URL template obviates the need for each post to specify 
 that it links to a collection of comments and to provide a hardcoded-url for that 
 collection.
 
@@ -387,18 +387,18 @@ a string, that string represents a URL template. If the value is an object, it m
 either an `"href"` key, a `"type"`key, or both; the `"href"` key holds a URL template
 while the `"type"` key specifies a resource type.
 
-For each key–value pair in the `"links"` object, the value specifies `"type"` and/or `"href"`
-information about all occurrences of the repeated relationship pointed to by the key. Specifically, 
-if the value holds a URL template (as described above), it is specifying that the `"href"` 
-for the linked resource(s) in a given occurrence of the repeated relationship is the result
-of expanding that URL template, using the resource object in which the occurrence appears as the
-data for the expansion. Similarly, if the value has a `"type"` property, it is specifying that the
-`"type"` of the linked resource(s) in all occcurences of the repeated relationship is the value of
-that property. 
+For each key–value pair in the `"links"` object, the value specifies `"type"` and/or 
+`"href"` information about all occurrences of the repeated relationship pointed to by the key. 
+Specifically, if the value holds a URL template (as described above), it is specifying that 
+the `"href"` for the linked resource(s) in a given occurrence of the repeated relationship is 
+the result of expanding that URL template, using the resource object in which the occurrence 
+appears as the data for the expansion. Similarly, if the value has a `"type"` property, it is 
+specifying that the `"type"` of the linked resource(s) in all occcurences of the repeated 
+relationship is the value of that property.
 
-NOTE: any information about a given occurrence of a relationship specified in the resource-level 
-`"links"` object always supersedes information specified by the top-level `"links"` object in case 
-of conflict.
+However, note that any information about a given occurrence of a relationship specified in the 
+resource-level `"links"` object always supersedes information specified by the top-level `"links"` 
+object in case of conflict.
 
 Here's another example:
 
--- a/format/index.md
+++ b/format/index.md
@@ -348,8 +348,9 @@ the referenced objects as a collection of resource objects.
 
 ### URL Templates <a href="#document-structure-url-templates" id="document-structure-url-templates" class="headerlink"></a>
 
-A top-level `"links"` object **MAY** be used to specify URL templates that can
-be used to formulate URLs for resources according to their type.
+A top-level `"links"` object **MAY** be used to specify URL templates that can be
+used to formulate URLs for resources. This often allows API responses to be more 
+compact.
 
 For example:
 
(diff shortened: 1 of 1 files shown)


format/index.md
```diff
@@ -348,8 +348,9 @@ the referenced objects as a collection of resource objects.
 
 ### URL Templates <a href="#document-structure-url-templates" id="document-structure-url-templates" class="headerlink"></a>
 
-A top-level `"links"` object **MAY** be used to specify URL templates that can
-be used to formulate URLs for resources according to their type.
+A top-level `"links"` object **MAY** be used to specify URL templates that can be
+used to formulate URLs for resources. This often allows API responses to be more
```
ethanresnick: Here, I removed the phrase "according to their type" because it's incorrect—the templates are according to path—and added the next sentence to state the impetus for templates upfront.

format/index.md
```diff
@@ -371,14 +372,31 @@ For example:
...
-Letter"`.
+Letter"`. The use of a URL template obviates the need for each post to specity 
+that it links to a collection of comments and to provide a hardcoded-url for that 
+collection.
+
+Each key in the top-level `"links"` object is a dot-separated path (defined from the 
+point of view of the reference document) that points at a repeated relationship. 
+Paths start with a particular resource type and can traverse related resources. For
+example `"posts.comments"` points at the `"comments"` relationship in each
+resource of type `"posts"`.
```
ethanresnick, line 383: The above paragraph is what used to be [here](https://github.com/json-api/json-api/pull/262/files#diff-26e38355b69056b0db1c2b1c612241b2L403). Just moving it, and the subsequent paragraph, up, in order to state the behavior earlier in the section.

format/index.md
```diff
@@ -371,14 +372,31 @@ For example:
...
+
+By default, each value in the `"links"` object is a URL template, given as a string.
+However, a value in the `"links"` object **MAY** also be an object with two properties: 
+`"href"`and `"type"`. In this case, the `"href"` key holds the URL template while the 
+`"type"` key specifies the type of the linked resource(s), which facilitates lookups of 
+linked resource objects by the client.
```
ethanresnick: Above paragraph is replacing the bullet point that was [here](https://github.com/json-api/json-api/pull/262/files#diff-26e38355b69056b0db1c2b1c612241b2L407) and gets at the crux of #250.

format/index.md
```diff
@@ -371,14 +372,31 @@ For example:
...
 Here's another example:
 
 ```javascript
 {
   "links": {
-    "posts.comments": "http://example.com/comments/{posts.comments}"
+    "posts.comments": {
+      "href": "http://example.com/comments/{posts.comments}",
+      "type": "comments"
+    }
```
ethanresnick, line 411: Changed this so we have an example of both allowed formats in this section.

format/index.md
```diff
@@ -398,17 +416,6 @@ the default explosion is to percent encode the array members (e.g. via
...
 comments.
 
-The top-level `"links"` object has the following behavior:
-
-* Each key is a dot-separated path that points at a repeated relationship. Paths
-  start with a particular resource type and can traverse related resources. For
-  example `"posts.comments"` points at the `"comments"` relationship in each
-  resource of type `"posts"`.
-* The value of each key is interpreted as a URL template.
-* For each resource that the path points to, act as if it specified a
```
ethanresnick, line 408: This bullet point was always pretty unclear. I tried to clean it up a bit at [line 454](https://github.com/json-api/json-api/pull/262/files#diff-26e38355b69056b0db1c2b1c612241b2R454) but it could still use to be more specific.

format/index.md
```diff
@@ -457,10 +469,6 @@ In a compound document, linked resources **MUST** be included as resource
 objects in a top level `"linked"` object, in which they are grouped together in
 arrays according to their type.
 
-The type of each relationship **MAY** be specified in a resource-level or top-
-level `"links"` object with the `"type"` key. This facilitates lookups of linked
-resource objects by the client.
```
ethanresnick, line 462: Removed this paragraph because it was now redundant: we said above that the top-level `"links"` can contain `"type"` keys, and the values at resource-level `"links"` are just resource or collection objects, which also allow `"type"` keys.

format/index.md
dgeb: How about "resource representations" instead of "API responses"?

format/index.md
dgeb: This phrasing seems a bit off: there is no concept of a "default". Each value may be either a string or an object. Similarly, there is no requirement that an object contains both `"href"` and `"type"` - it can contain either or both.

dgeb: I think of a top level link object as a [collection object](http://jsonapi.org/format/#document-structure-collection-objects) that can't contain `"ids"`.

format/index.md
dgeb, line 411: :+1:

format/index.md
```diff
@@ -444,6 +451,11 @@ In this example, the URL for the author of all three posts is
 Top-level URL templates allow you to specify relationships as IDs, but without
 requiring that clients hard-code information about how to form the URLs.
 
+For each key/path in the top-level `"links"` object, clients should do as follows. For each 
+resource that the path points to, act as if that resource specified a relationship whose
+liked resource(s) can be found at the URL formed by expanding the URL template with values 
+drawn from the resource.
+
```
dgeb, line 467: I think you're talking about expression expansion within URL templates here - is that right? Your reference to `key/path` is confusing. I'm not thrilled with the original phrasing in the bullet points, but I think this needs some work too.

format/index.md
ethanresnick: Done!

ethanresnick:
Hey @dgeb,

Sorry it took me a bit to get back to this. I've updated my PR based on your feedback.

format/index.md
ethanresnick, line 467: I went through a few iterations of this and finally came up with something I'm relatively happy with. It's now at [line 390](https://github.com/json-api/json-api/pull/262/files#diff-26e38355b69056b0db1c2b1c612241b2R390)

format/index.md
```diff
@@ -371,14 +372,43 @@ For example:
...
+Each key in the top-level `"links"` object is a dot-separated path (defined from the 
+point of view of the reference document) that points at a repeated relationship. 
+Paths start with a particular resource type and can traverse related resources. For
+example `"posts.comments"` points at the `"comments"` relationship in each
+resource of type `"posts"`.
+
+Each value in the `"links"` object may either be a string or an object. If a value is
+a string, that string represents a URL template. If the value is an object, it must have
+either an `"href"` key, a `"type"`key, or both; the `"href"` key holds a URL template
```
dgeb: needs a space to break this up: `"type"`key

dgeb:
@ethanresnick thanks for updating this. I'm now :+1: on these changes, which really explain the top-level `"links"` object more thoroughly than before. I only commented on one tiny typo in my read-through.

@steveklabnik @wycats what do you guys think?

format/index.md
ethanresnick: fixed!

ethanresnick:
@dgeb Great!

dgeb:
Related to #276 - we may want to reach a consensus on `links` in general before clarifying the language.

dgeb:
Thanks for your efforts here, @ethanresnick. Ultimately we ended up simplifying top-level links a great deal in #341.

PR #263, opened 2014-08-15
Change 204 response from MUST to MAY

feelepxyz:
Currently the spec states that you must return a 204 if an update is successful without affecting other attributes on the resource.

But if the update triggers changes to other attributes it should return a 200.

This seems overly strict and makes the implementation complex.

For simplicity we would like to always return a 200 with the full resource and not have to implement logic to figure if a there were other attributes updated by the server.

A 204 makes more sense when you replace the entire resource, but less sense when you can do partial updates.

Because it's so subtle it would be hard to rely on the current requirement for consistent behaviour across implementations.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -1074,7 +1074,7 @@ DELETE /articles/1/links/tags/1,2
 
 #### 204 No Content <a href="#crud-updating-responses-204" id="crud-updating-responses-204" class="headerlink"></a>
 
-A server **MUST** return a `204 No Content` status code if an update is
+A server **SHOULD** return a `204 No Content` status code if an update is
 successful and the client's current attributes remain up to date. This applies
 to `PUT` requests as well as `POST` and `DELETE` requests that modify links
 without affecting other attributes of a resource.


thesteady:
I agree with @harrison. I just implemented an endpoint with the 200/204 logic, and I could see it being a potential point of confusion for clients who aren't as familiar with the details of JSON API standard. What was the reasoning behind the initial decision for the "MUST"? I'm curious. :smile:

dgeb:
I do think that the `MUST` in both the 200 and 204 sections is a bit misplaced. It is more accurate to say that a 200 response `MUST` include a representation of the resource, while a 204 response `MUST` not. So that would require more rewording of both paragraphs.

I'd favor using `SHOULD` in place of the `MUST` you've changed to a `MAY`:

> A server **SHOULD** return a `204 No Content` status code if an update is successful and the client's current attributes remain up to date.

This would give you some room to maneuver in while still making a strong recommendation.

How does that sound?

@steveklabnik thoughts?

feelepxyz:
@dgeb agreed, I'm all for changing it to `SHOULD`. A strong recommendation is better.

steveklabnik:
In general, I am usually :+1: on turning **MUST**s into **SHOULD**s. @wycats originally wrote this part, so we should make sure there isn't some reason we're missing.

feelepxyz:
@steveklabnik @dgeb changed to `SHOULD`.

greysteil:
@steveklabnik / @wycats - any update on this?

steveklabnik:
I'm still :+1: but haven't heard from @wycats

PR #273, opened 2014-09-17
Clarify values for blank relationships.

dgeb:
Has-one relationships: `null`
Has-many relationships: `[ ]`

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -294,6 +294,21 @@ And here's an example of a linked author represented as a resource object:
 //...
 ```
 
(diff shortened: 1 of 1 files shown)


steveklabnik:
What about the 404 case?

dgeb:
@steveklabnik I'd like to work up some error examples (such as for 404s) and include that in a separate commit. 

I've removed the reference to issue #101 so that won't be closed by merging this commit.

steveklabnik:
Awesome, seems good. Thanks :D

PR #279, opened 2014-09-26
Clarify updating of relationships

dgeb:
This PR makes the following improvements to the consistency and correctness of updating relationships:
- A `PUT` request should be allowed to the URL of a to-one relationship
- A `POST` request sent to a to-one relationship URL that already contains a relationship should fail
- A `DELETE` request sent to a to-one relationship URL that doesn't contain a relationship should fail
- A single resource ID (or an array of IDs, as previously stated) can be POSTed to a to-many relationship URL
- 404 and 409 error responses 

Essentially, this establishes `PUT` as the safe way to replace a to-one relationship. `POST` and `DELETE` should only be used to verify expectations of the current value of a relationship.

In addition, I've pulled in the MUST -> SHOULD change from @harrison in #263 since it is compatible with these changes. There's also some whitespace cleanup in a separate commit.

One point of possible debate here is whether `409 Conflict` should be returned when POSTing a relationship that already exists. It seems like the most appropriate status code to me, but maybe @steveklabnik wants to weigh in :)

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -1074,7 +1074,7 @@ DELETE /articles/1/links/tags/1,2
 
 #### 204 No Content <a href="#crud-updating-responses-204" id="crud-updating-responses-204" class="headerlink"></a>
 
-A server **MUST** return a `204 No Content` status code if an update is
+A server **SHOULD** return a `204 No Content` status code if an update is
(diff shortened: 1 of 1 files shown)


steveklabnik:
Gave it some thought, and this all seems good to me :)

PR #281, opened 2014-09-29
Add error examples

ahx:
The [errors section](http://jsonapi.org/format/#errors) needs some examples for clarification. I did not really understand `path` and `links` until reading https://github.com/json-api/json-api/pull/234#issuecomment-45563232

This PR adds a basic example for an error response. I am omitting `id` and `code` and `detail`, because the main purpose was to make clear how `path` can be used.

I would appreciate if someone else could add another example where a `links` attribute makes sense like trying to update multiple resources at once where one fails or so.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -1405,6 +1405,40 @@ An error object **MAY** have the following members:
 
 Additional members **MAY** be specified within error objects.
 
+### Error example – Creating a resource with an empty required field
+
+This request tries to create a resource of type "articles" with a missing field:
+
+```text
+POST /articles
+Content-Type: application/vnd.api+json
+Accept: application/vnd.api+json
+
+{
+  "data": {
+    "type": "articles",
+    "title": "",
+    "body": "This is an article with a missing title"
+  }
+}
+```
+
+The error response could look like this:
+
+```text
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [{
+    "code": "field_empty",
+    "title": "The title must not be empty",
+    "links": ["/data"],
+    "paths": ["/title"]
+  }]
+}
+```
+
 [attributes]: #document-structure-resource-object-attributes
 [complex attributes]: #document-structure-resource-object-complex-attributes
 [fields]: #document-structure-resource-object-fields


ahx:
Questions: 
1. Should `path` be in JSON Pointer syntax (http://tools.ietf.org/html/rfc6901) (I assume: yes)?
2. Should `path` be `"/posts/title"` or just "`/title`" in this example and why?

format/index.md
```diff
@@ -1174,15 +1174,46 @@ An error object **MAY** have the following members:
...
+
+The response could look like this:
+
+```text
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [{
+    "title": "The title field must not be empty",
```
kevin-brown: Wouldn't it make more sense of this to be in the `detail`? I wasn't sure if

> It **SHOULD NOT** change from occurrence to occurrence of the problem, except for purposes of localization

applied on the field or resource level.

kevin-brown:
> Should path be "/posts/title" or just "/title" in this example and why?

In the JSON API [implementation for Django REST Framework](https://github.com/kevin-brown/drf-json-api/blob/3634b4110882438be122dd22ea5b42419d566982/tests/test_errors.py#L15-L36) that I'm working on, we are just using `/title`.  The [discussion for why](https://github.com/kevin-brown/drf-json-api/pull/3) came down to it matching the JSON PATCH format for the resource level (not the root level).

I do have another related question though: How should errors across multiple fields (or not field related) be represented?  Currently [we are using `/-`](https://github.com/kevin-brown/drf-json-api/blob/3634b4110882438be122dd22ea5b42419d566982/tests/test_errors.py#L84-L90) as the path.

---

I wouldn't mind seeing examples for authentication errors as well, we are [serializing them currently like](https://github.com/kevin-brown/drf-json-api/blob/3634b4110882438be122dd22ea5b42419d566982/tests/test_errors.py#L39-L61)

``` json
{
    "errors": [{
        "status": "403",
        "title": "Authentication credentials were not provided."
    }]
}
```

format/index.md
ahx: Since `path` links to a single field ("title"), i think the title should be about that specific field.

I understand `detail` as a long form of `title`. So another example could be… a Dog resource where a name field must be one of the [Disney characters](http://simple.wikipedia.org/wiki/List_of_Disney_characters) but the user tried to set it to "steve" (not in that list) instead . Now `title` could say something rather generic like "The dog must be named after a Disney character" whereas `detail` would render a very long error description like "Lorem ipsum Disney […] Goofy, Pluto, Mini Mouse, please […]". Or am i on the wrong track here?

krainboltgreene:
For what it's worth my current errors look like this:

``` json
{
  "errors": [
    { "id": "email_not_unique" },
    { "id": "password_too_small" }
  ]
}
```

I use the `id` field to look up a value in a dictionary for the appropriate i18n message for the client's determined language.

ahacking:
I have experienced quite a bit of unhappiness with the current errors definition.

I want to identify a scope to which one or more problems apply. The scope could be at a resource level or field level.

Json-api has links and a path and this just doesnt seem to fit. It doesn't allow multiple problems to be specified once you have identified the problem area.  I am not sure why we don't just have a json pointer (what json patch uses) to identify the problem scope and allow a collection of problems.

Each problem can then have a title, detail and additional information such as code, reason or anything else the server wishes to return.

IMO errors could just be:

``` json
{
  "errors": [
    { "path": "<json pointer>",
      "problems": [{
        "title": "Some text",
        "detail": "Detailed explanation",
        "code": "some_code",
        "whatever": " you need"
        }, {
          // etc
        }
      ]
    }
  ]
}
```

An alternative format is to drop the errors array and return an object using the distinct json pointers as the keys:

``` json
{
  "errors": {
    "<json pointer>": [{
      "title": "Some text",
      "detail": "Detailed explanation",
      "code": "some_code",
      "whatever": " you need"
    }, {
      // etc
    }],
    "<json pointer>": [
      // etc
    ]
  }
}
```

krainboltgreene:
This was brought up in the original errors conversation, but decided against.

ahacking:
@krainboltgreene Thanks. Any reference where it was discussed and more importantly why? I am aware a draft nottingham rfc was looked at, and I looked at it too and it looks like an il conceived spec given its intent.  I believe it had some influence on the json -API error designbut was ultimately diverged since the draft rfc had little hope of meeting json-api requirements.

Does anyone have a set of use cases or requirements for json-api? It seems there is an unspoken set of requirements and very little rationale for how things are. Given this I don't see why we cannot change anything in the interests of pragmatism.

Json API allows multiple resources to be specified in all create and update actions so errors need to provide adressability of problems balanced with unnecessary repetition.  The spec needs trimming down to a firm set of capabilities that all implementations must adhere to. Forcing a json pointer provides a definitive scope for problems, anything else is just flimsy design and ultimately the slippery slope to non interoperability.

steveklabnik:
@ahx 

Sorry about the delay here! We are certainly interested in adding examples, but this does not merge any more. Are you interested in keeping up this PR? If so, let's go over them in detail, but I'm not going to review until I hear from you :smile:

ahx:
@steveklabnik i have updated the example. Using a Rails' i18n-ish looking "code" in the example ("errors.attributes.title.blank"), because there were some questions regarding i18n in the past and I think this is how I would use the field, even without using Rails.
Is 422 the proper status code?

ahx:
I have rebased this PR. :+1: :-1: :hand: ?

ColtonProvias:
@ahx I don't think there is a need for a long error code like that if the `path` is already giving the location of the error.  It could probably be something shorter like `invalid_required`.

ahx:
@ColtonProvias I've used the long code ("errors.attributes.title.blank") to resemble a code that you can throw against [some i18n library](http://guides.rubyonrails.org/i18n.html). But this probably is just distracting. So i changed it to "field_empty".

tkellen:
Sorry for the slow/no response @ahx. If you'd be willing to rebase this one more time and ensure it lines up with 1.0 of the spec, we can get it landed.

tkellen:
Please rebase/resubmit this PR if you'd like for it to be merged.

dgeb:
Also, any examples should target the new [examples](http://jsonapi.org/examples/) page.

PR #290, opened 2014-10-14
Add Json-API package

ethanresnick:
Like I mentioned in the text, the linked package, which I wrote, is largely undocumented and doesn't have test cases (i know, i know...). Still, it's the most complete and compliant server side js implementation I know of, so I think it should be included.


steveklabnik:
I'm not comfortable with value judgements on this page, and I don't want to have to monitor the status of the library when this changes.  Can we just remove the parenthetical, please?

ethanresnick:
@steveklabnik Done!

dgeb:
@ethanresnick thanks for removing the parenthetical - looks good to me now :+1:

PR #292, opened 2014-10-16
This plurality wasn't intended

krainboltgreene:
It seemed to imply that you were querying a many to many relationship when the example clearly states a one to many relationship.


dgeb:
@krainboltgreene good catch - thanks!

PR #300, opened 2014-10-29
Properly remove the whole resource in PATCH.

loganfsmyth:
Per [RFC 6901](https://tools.ietf.org/html/rfc6901), `path: "/"` would be a command to remove a key `""` from the object, whereas `path: ""` would remove the whole object. Assuming I'm reading that spec correctly.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -1348,7 +1348,7 @@ PATCH /article/1/links/author
 Content-Type: application/json-patch+json
 
 [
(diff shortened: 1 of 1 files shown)


dgeb:
@loganfsmyth you're correct - thanks much! I see a couple other instances of `"/"` that I'll replace now as well.

loganfsmyth:
Thanks! This was submitted right as I went to bed, so not surprised I missed other instances.

PR #303, opened 2014-10-30
Change wording of `save HTTP`

immunda:
This isn't worded well, and `save` inferred a create/update operation to my mind.


dgeb:
That is more clear. Thanks @immunda !

PR #306, opened 2014-11-03
These are members, shouldn't be plural

krainboltgreene:
Our example shows a single entity and not a list, and per the
spec it should be a singular object key.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -86,7 +86,7 @@ The following post is represented as a resource object:
 
 ```javascript
 {
-  "posts": {
+  "post": {
     "id": "1",
     // ... attributes of this post
   }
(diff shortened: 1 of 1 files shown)


alexbyk:
In this api plural is used as a key and a uri path everywhere, not only in this example. Because it is simple when a key (or a path) matches a db/collection name. How do use suppose to extract a data for collection `knives`/`knife`?

krainboltgreene:
Hmm, yes, you're right. Don't know how I missed this? I think I was partially confused with request value keys.

gr0uch:
I would vote against this change, because it would force string inflection logic downstream. The client would have to recognize that `post` and `posts` are of the same type.

krainboltgreene:
I absolutely agree.

PR #309, opened 2014-11-09
Add Yaks to the list of Ruby server side examples

plexus:
Yaks has been able to do JSON API for over a year now. Would be great if it could be added to the list!

https://github.com/plexus/yaks


dgeb:
Nice - thanks! We like our yaks clean-shaven around here ;)

plexus:
shaved_yak

:+1:

Issue #314, opened 2014-11-11
Proposal: broken nested linked resources without referential links


kurko:
According to the spec, `linked` resources don't include `links`. Thus, the basic semantic format for compound documents is this:

```
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "comments": ["9"],
    }
  }],
  "linked": {
    "comments": [{
      "id": "9",
      "text": "lolwat"
    }]
  }
}
```

When using `include`, such as `GET /posts?include=comments.author`, the reference between comments and author would be broken.

```
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "comments": ["9"],
    }
  }],
  "linked": {
    "comments": [{
      "id": "9",
      "text": "lolwat"
    }],
    "authors": [{
      "id": "10",
      "author": "Wally"
    }]
  }
}
```

How would you know that `authors` belong to `comments`?
### Proposal

I propose enforcing `links` presence in every described resource.

hesalx:
Funny, I was always sure of the opposite.

http://jsonapi.org/format/#document-structure-resource-objects

> Resource objects have the same internal structure, regardless of whether they represent primary or linked resources.

http://jsonapi.org/format/#document-structure-resource-object-attributes

> There are four reserved keys in resource objects:
> - "id"
> - "type"
> - "href"
> - "links"

`linked` resources are equal in rights with their top-level siblings.
There are just lack of examples with `links` inside `linked`.

kurko:
You're right. Silly me.

Thanks a lot!

Issue #311, opened 2014-11-11
Bring consistency to top-level and resource-level links


dgeb:
I'd like to float a proposal that comes from considering points raised by @ahacking, @ethanresnick, @wvteijlingen, and others in #238, #262, #276, and other recent issues. It represents a breaking change that should be considered carefully before the spec hits 1.0.

The proposal is to make the format of the `links` object identical in both the top-level and resource-level. Each link can be represented either by an object or a string. Link objects can contain two properties: `href` and / or `type`. If a link is represented with a string, it should be assumed to be `href`.

The implications of this change are pretty drastic:
- It reduces processing rules for implementations by establishing a single rule for processing `links`.
- IDs for has-one and has-many relationships will be represented alongside other attributes in a resource. (Of course, an implementation may instead choose to exclusively provide hrefs for relationships and never provide related IDs.)
- The term `links` will never appear in URL template expressions or in relationship URLs.
- There is no need to call out a separate ["collection object"](http://jsonapi.org/format/#document-structure-collection-objects)
- It will be clear that resource-level links are redundant and used to override top-level links. I imagine that many APIs will exclusively use top-level links and never use resource-level links.
- It will be clear that expressions in URL templates can reference _any_ attribute in a resource (even complex object attributes) in order to form links to another resource (see #238).
- It will move JSON API one step closer to "standard REST APIs", which will probably increase adoption.

If there is adequate support for this change, I'll create a PR for detailed review.
### Mixed top-level and resource-level links

Before:

``` javascript
{
  "links": {
    "posts.comments": "http://example.com/comments?post={posts}"
  },
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "comments": {
        "href": "http://example.com/comments?post=1"
      }
    }
  }, {
    "id": "2",
    "title": "The Parley Letter",
    "links": {
      "comments": {
        "href": "http://example.com/comments?post=2"
      }
    }
  }]
}
```

After:

``` javascript
{
  "links": {
    "posts.comments": "http://example.com/comments?post={posts}"
  },
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "comments": "http://example.com/comments?post=1"
    }
  }, {
    "id": "2",
    "title": "The Parley Letter",
    "links": {
      "comments": "http://example.com/comments?post=2"
    }
  }]
}
```
### Top-level links, resource-level relationship IDs

Before:

``` javascript
{
  "links": {
[...]

PR #315, opened 2014-11-12
Example with links in linked resources

hesalx:
To shed some light on such questions #314. Because I have met such question at least in one more discussion somewhere.

I had added another example to not increase complexity of the first one.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -557,6 +557,70 @@ times (in this example, the author of the three posts). Along these lines, if a
 primary document is linked to another primary or linked document, it should not
 be duplicated within the `"linked"` object.
 
+Here's another example:
+```javascript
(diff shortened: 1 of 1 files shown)


kurko:
:+1:

dgeb:
@arantir - This is a great example, although it may not be completely necessary depending on the outcome of #311

Either way, I think that the inclusion of a top-level links example that is based on a linked resource (`"comments.author"` here) is a nice addition.

krainboltgreene:
That feeling when all this time you've confused `links` with `linked`.

Good example.

dgeb:
Thanks for your work here @arantir. Closing due to the simplification of top-level links in #341.

PR #321, opened 2014-11-21
Added link to JSONAPI.NET, a .NET implementation of the spec.

SphtKr:
I'm the author of JSONAPI.NET, any questions let me know. Thanks!

### DIFF
--- a/examples/index.md
+++ b/examples/index.md
@@ -71,6 +71,10 @@ has a page describing how to emit conformant JSON.
 
 * [api2go](https://github.com/univedo/api2go) is a small library to make it easier to provide a JSON API with your Golang project.
 
+### .NET <a href="#server-net" id="server-net" class="headerlink"></a>
+
+* [JSONAPI.NET](http://jsonapi.codeplex.com) is a .NET library that integrates with ASP.NET WebAPI, Json.NET, and (optionally) Entity Framework to help you quickly create JSON API compliant web services.
+
 ## Messages <a href="#messages" id="messages" class="headerlink"></a>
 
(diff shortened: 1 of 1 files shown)


steveklabnik:
Thank you! Glad to see this :)

jonnii:
@SphtKr I have something that does something similar, it's mainly focused on automatically enveloping json responses: https://github.com/jonnii/emvelope

SphtKr:
@jonnii Hi Jonnii, yes, I just found your project the other day. I found I had to do more than just "wrap" the result, I had to come up with a framework for putting ID values in for related objects, collecting and including related objects (making "Compound Documents"--does emvelope have code for that? I found that was one of the hardest things), and I have some (I think) really cool Entity Framework integration: you can mostly just subclass my EF-specific ApiController and it will already do GET, POST, PUT, and DELETE with no extra code! I see you've got some additional flexibility that I don't, like different case styles. Anyway, I'd be interested in your feedback and/or contributions.

jonnii:
@SphtKr to get full json-api compatibility is certainly not a trivial task and what I've done wouldn't work for that. Right now I can support meta-data (for paging etc...) and just wrapping/unwrapping the envelope. I think ultimately the direction you've gone is the right way, maybe the final solution will be to offer something like https://github.com/WebApiContrib/CollectionJson.Net.

btw, you should also consider moving your code to github...

SphtKr:
@jonnii :-) Yes, I'm already questioning choosing CodePlex…I thought there'd be some value, say integration with NuGet or for documentation generation from code (a la Sandcastle)…but looks like it's not so great, really.

Hmm! I haven't figured out paging yet, I'll have a look at your code.

PR #328, opened 2014-12-16
Handling of links

m-breen:
Just had the idea, thought I would share.
Here is a proposed way of handling links.
Any json object `{}` described in a resource could express a link, just by having one of the following keys : "href", "id" or "ids"

``` javascript
//...
  {
    "id": "1",
    "type":"posts",
    "title": "Rails is Omakase",
    "author": { // To-One relationship
      "id":9,
      "type":"people", // Must be specified if type of the resource is different than the parent key name
      "href":"http://example.com/people/9",
    }
    "comments" : { // To-Many relationship
      "ids" :[ "5", "12", "17", "20" ], 
      "href": "http://example.com/comments?posts=1", // use of an "opaque url"
    }
  }
//...
```

Resource "links" attribute would therefore no longer be necessary.
This would allow flexibility, as you could easily embed external resource attributes within the current resource.
This would still be compatible with top level links
Although Parsing would be a bit more complex...
Would love your feedback on this

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -247,7 +247,7 @@ collection of `comments`:
     "title": "Rails is Omakase",
     "author": { // To-One relationship
       "id":9,
-      "type":"people", // Type must be specified if different than the parent attribute
+      "type":"people", // Must be specified if type of the resource is different than the parent key name
       "href":"http://example.com/people/9",
     }
     "comments" : { // To-Many relationship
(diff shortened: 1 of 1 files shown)


wvteijlingen:
At this point we're leaning towards linking as described in https://github.com/json-api/json-api/issues/311.

PR #333, opened 2015-01-07
Relaxing the definition of what Client-Generated IDs can be.

natbusa:
Hi,

In the current version of the jsonapi format document, the server can craft all sorts of ids as long as they are strings, with alphanum chars, underscores and dashes. However the client generated id are limited to UUIDs. Why is that? 

My opinion is that the client should be allowed to define an arbitrary id, as well as granting the server the right and responsibility to either grant or deny the creation of a given resource with the client-supplied id.

By doing so,
If the user wants to POST the resource with "id" :"2", well I guess it should be possible. Weather the resource is actually created or not, well this is up to the server to determine.

My suggestion: when defining id's, provide the same semantic flexibility both for client- and server- crafted id's

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -892,7 +892,7 @@ Servers **MAY** use other HTTP error codes to represent errors.  Clients
 
 A server **MAY** accept client-generated IDs along with requests to create one
 or more resources. IDs **MUST** be specified with an `"id"` key, the value of
-which **MUST** be a properly generated and formatted *UUID*.
+which **MAY** be a properly generated and formatted *UUID*.
 
 For example:
 


wvteijlingen:
If an API uses client-supplied ids, these ids MUST be unique. The easiest way to enforce this is to require client-supplied ids to be UUIDs. Having this requirement takes away a lot of room for divergence, and makes it possible to build more generic client side tooling.

I cannot really think of a plausible use case in which a UUID poses a problem.

natbusa:
UUID do not enforce uniqueness any more than another id generation pattern. That is exactly my point in the change.

natbusa:
A case where you would like to have more flexibility is when the client is exporting the ids from another system (api or datastore) and it is necessary to keep the same order and id in the new system. Think for instance in cloning or fresh importing of a collection in the api.

My change in fact is not about saying that UUID are not suited but rather to define UUID as possible client side generation pattern and allow other patterns if required by the client application.

That said it would be always the responsibility of the server to create the resource or not, independently from the fact that it was a UUID or a different id generation pattern.

wvteijlingen:
Ah yes, I hadn't thought about such a use case :). I'm not against having some leeway here, apart from the fact that it introduces even more room for diversion in a spec that already heavily uses `MAY`s. I would like to see a more strict and opinionated spec, but others might disagree :).

Maybe using `SHOULD` would work? That leaves room for other patterns, while still recommending UUIDs.

lgebhardt:
I agree that any format for ids should be acceptable for client creation, so long as the server has the ability to reject client created ids.

I'm not sure why the spec should care if the id is a string or an integer.

natbusa:
@lgebhardt : The current definition of ids is as follow:

An ID SHOULD be represented by an "id" key and its value MUST be a string which SHOULD only contain alphanumeric characters, dashes and underscores.

According to the current definition, integer id's are not allowed ... however, this is not the point of this pull request. You could, of course, push this discussion further by opening a new issue.

dgeb:
The requirements for IDs have been simplified to:

> Each resource object MUST contain an id member, whose value MUST be a string.

Requirements for client-generated IDs have also been relaxed, and a non-normative note has been added. See http://jsonapi.org/format/#crud-creating-client-ids

Issue #336, opened 2015-01-27
ambiguities for nested includes


tkellen:
The spec states that if you request a nested relation, only the last relation should be serialized in the result. So, a request `posts?include=comments.authors`, should only include `authors` in the `linked` section of the response (even though we had to query through comments to get the authors).

So, fine, that is easy to implement.  The question I have is this: I have not seen any references to having nested relations under `links`.  Should the `links` key for a given `post` in this request appear like this?

```
{
  links: {
    "comments.authors": {
      "type": "authors",
      "ids": [...]
    }
  }
}
```

tkellen:
Upon further reflection, this seems like it is obviously what the spec intends. Perhaps an example showing this would be useful for others who might overthink this as I did? I'd be happy to add one once the next draft lands.

dgeb:
@tkellen This seems like a slippery slope that could introduce quite a bit of complexity to implementations. 

Consider the different serializations that would be required for these two requests:
- `GET /posts?include=comments.author` - Each `post` resource would have to include `comments.author` in its `links` object. _This distant relation would only appear in the `post`'s `links` when `comments` was not also requested._
- `GET /posts?include=comments,comments.author` - Each `post` resource would only have to include `comments` in `links`, and then each `comment` resource would need to include `author`. By far this is more straightforward to construct on the server and unpack on the client.

In order to avoid this complexity, which increases the more distant the relationship, I'm comfortable with only providing direct relationships in a resource's `links` object. If has-many-through relationships are required on the client, then the client can easily request each relationship. I realize this limits the use cases for requesting a distant relation without a direct relation, but I think it's worth it to keep implementations straightforward.

tkellen:
I'm surprised to hear you say that. I think the spec should support this, but the implementation of it is definitely non-trivial. Supported or not, I think some clarification around this is warranted.

dgeb:
I don't see any harm in an implementation including deep links like this, but I don't think we should make it a requirement. Either way, I agree that clarification is warranted.

dgeb:
@tkellen I may be coming around to your perspective after unsuccessfully trying to come up with scenarios where has-many-through links could be useful without including the linkage IDs in the primary resources.

This may simply be a case of an optional feature that is non-trivial to implement.

tkellen:
[...]

Issue #327, opened 2014-12-08
PATCH updates using the regular JSON API format


bintoro:
How about allowing two flavors of `PATCH` requests to accommodate implementations that do not wish to abuse the `PUT` method by employing it for partial updates?
- `PATCH` with `application/json-patch+json`: As currently specified.
- `PATCH` with `application/vnd.api+json`: Payload as in `POST` or `PUT`. Explicit intention to update the target resource with only the attributes present in the payload.

JSON API already leaves the door open for an implementation to limit `PUT` to its original purpose, which is to replace the target resource in its entirety. The proposed change would make it easier for services to establish this convention by simply announcing their interpretation of `PUT` out of two possible choices.

Right now, if a service wanted to do this, it would also need to define new rules for `PATCH` and would end up in violation of JSON API.

dgeb:
Certainly the new [Patch extension](http://jsonapi.org/extensions/patch/) can be used as an alternative to PUT. 

The use of PUT for updates is pretty firmly defined in the base spec (as of RC2) for [updating resources](http://jsonapi.org/format/#crud-updating). However, I'll link to your [other comment from #341](https://github.com/json-api/json-api/pull/341#issuecomment-74243140), which I think is a reasonable point:

> Perhaps this is a needlessly academic concern, but isn't it a reasonable requirement that a client should be able to perform a true PUT to request the resetting of some server-side properties that it doesn't have direct control over?

dgeb:
I'd like to clarify the rationale for PUT-as-update instead of PUT-as-replace because I'm aware that it is somewhat controversial.

Basically, it comes down to the rather nebulous definitions of "whole resources" that should be replaced. This is especially an issue for relationships. Should every to-one and to-many relationship also be replaced along with attributes? I think this would be surprising to many clients, who might expect only attributes, or perhaps only attributes and to-one relationships, to be replaced. How about relationships that have inverses?

We've avoided much of this murky territory in JSON API by spec'ing PUT-as-update. This also simplifies implementations that don't also need to support both PATCH and PUT to be compliant with the base spec.

With that said, I certainly appreciate points like you've made above, and I know that this won't please everyone. However, I wanted to at least be clear about the rationale.

bintoro:
Indeed, it's not immediately obvious how a true PUT should behave vis-à-vis relationships, but that's pretty much the point of specifications: to decide one way or another and then standardize.

[...]

Issue #324, opened 2014-12-01
simplify sort


alexbyk:
How about to simplify sort to `?sort=name:-1`

full example:
`?sort=name:1,price:-1` which means (order does matter) sort by:
1. by `name` asc
2. by `price` desc

Benefits:
1) In this form it should be simple to split all pairs by `,` character, then split pairs to key:value
2) Right now it's impossible to use `-field` in the DB. But the are no words about field names should not begin from the `-` character. So current implementation of sorting makes api inconsistent 

We can also make it backward compatible. But it would be simpler to drop current syntax out as inconsistent

wvteijlingen:
I don't see any benefits to this form over the current spec. The current spec uses a comma as a separator as well, so splitting is equally easy. I don't think an attribute name starting with a dash would be common. This form actually has the same problems, what happens when I have a field named `name:1`?

Perhaps some constraints on the attribute name would resolve this problem, otherwise I would leave sorting as is.

alexbyk:
if you have a field named `name:1`, you could sort by `name:1:-1`|`name:1:1` without problems, and you gave an example why not to use a short form `?sort=name` => `?sort=name:1`. But you also gave a perfect example why current impl. should be changed (except a short form from my proposal). If you have 2 fields `-name` and `name` it would be a real problem in the current impl.

About benefits. `<>` with (1,0,-1) operations looks simpler than getting  charAt(0) and checking if it equals to `-` in the most of languages

wvteijlingen:
If I have `?sort=name:-1`, and two fields `name` and `name:-1`, we cannot differentiate between an descending sort on the field `name` and a ascending sort on the field`name:-1`. It is the same problem as with `name` and `-name`. The only upside is that a field named `name:-1` would be even more uncommon.

> About benefits. <> with (1,0,-1) operations looks simpler than getting charAt(0) and checking if it equals to - in the most of languages.

I don't think we should base our decisions on whether something looks simpler in a specific language. Of course we should make parsing as easy as possible, but I don't think this would be a real improvement.

Maybe others disagree with me on this though :).

alexbyk:
@wvteijlingen i wrote that `?sort=name` without `:1|-1` is optional. I edited a comment (cleared optional confusing section). If you have 2 fields, the sorting query would be `?sort=name:-1:1,name:-1` which means sort by `name:-1` asc then by `name` desc. And the only remaining problem is `,` in fields names

wvteijlingen:
[...]

Issue #317, opened 2014-11-16
More RESTful relationships (proposal)


hesalx:
Totally mindstormed by #316
Inspired by [GitHub API](https://developer.github.com/v3/), REST and @RSSchermer.
## The problem

Impossibility of `POST` to relationship URL is non-RESTful.
If it is possible to `GET` resources from a relationship URL, it should be allowed to `POST` homogeneous resources to that relationship URL.
## The proposal

**tl;dr** The main part of proposal is to take out relationships from the payload to the URL. 
### Creating Related Resources

(There is no such paragraph in the spec now.)
#### Creating To-One Related Resources

To-one relationship MAY be created by sending a POST request to the relationship URL.

``` javascript
POST /users/1/links/photo
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json

{
  "photos": {
    "title": "John Doe",
    "src": "http://example.com/images/avatar.png"
  }
}
```

A POST request should only succeed if no relationship is currently defined.
#### Creating To-Many Related Resources

To-many relationship MAY be created by sending a POST request to the relationship URL.

``` javascript
POST /articles/1/links/comments
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json

{
  "comments": {
    "body": "I love REST!"
  }
}
```
#### Updating To-One Relationships

To-one relationships MAY be updated along with other attributes by including them in a links object within the resource object in a PUT request.  (The same as in the current spec)

To-one relationships MAY be updated by sending a PUT request to the relationship URL with identifier of the related resource included into the URL. The body of the request SHOULD be empty.

``` javascript
PUT /articles/1/links/author/1
```

A to-one relationship MAY alternatively be added by sending a POST request with identifier of the related resource included into the URL. The body of the request SHOULD be empty.

``` javascript
POST /articles/1/links/author/1
```

A POST request should only succeed if no relationship is currently defined.
#### Updating To-Many Relationships

To-many relationships MAY optionally be updated with other attributes by including them in a links object within the document in a PUT request. (The same as in the current spec)

To-one relationships MAY be updated by sending a PUT request to the relationship URL with identifiers of the related resources included into the URL. The body of the request SHOULD be empty.

``` javascript
PUT /users/1/links/articles/1,2,3
```

A PUT request sent to the URL of a relationship SHOULD completely replace every member of the relationship.

A to-many relationship MAY be added by sending a POST request to the URL of the relationship with a single resource identifier included into the URL. The body of the request SHOULD be empty.

[...]

PR #341, opened 2015-02-06
Proposal for v1.0 rc2

dgeb:
This represents a fairly thorough rewrite of JSON API by myself and @wycats. This rewrite is based upon feedback we've received and our experience building both client and server implementations of the spec. I regret that changes of this scope and magnitude are proposed in a single pull request, but this has been necessitated by the condensed time in which we had to work.

The overarching goal of this rewrite has been to bring rigor and simplicity to JSON API, transforming it from a loose set of guidelines to a specification that can be fully implemented to create compatible clients and servers. In general, SHOULDs have been minimized and replaced with MUSTs, and any MAYs that remain tend to be paired with MUSTs.

The current spec has been refactored into a base spec, rules for extending it (including official extensions for bulk and patch updates), as well as recommendations (e.g. for forming URLs) that are deemed beyond-the-scope of the spec.

Some of the major changes in the base specification include:
- Rules for extending the spec with media type parameters have been defined.
- Polymorphism is supported.
- Every resource object MUST include both `type` and `id` members.
- ~~Attribute and association keys MUST be dasherized (e.g. `"first-name"`).~~ [Note: no longer a requirement; this has been moved to the new Recommendations page]
- Primary resources MUST now appear under the top-level `"data"` key.
- Linked resources (in a compound document) MUST now appear in an array directly under the top-level `"linked"` key.
- Top-level `links` relate to the primary data, including `self`. Support for URI templates in links has been removed.
- Link objects within resources have been more fully described, and can now include `self` links,  relationship URLs (for manipulating the relationship), related resource URLs (for fetching related resource objects), linkage to other objects in a compound document, and `meta` info.
- Pagination links have been established and a `page` query param has been reserved. The spec remains agnostic about pagination strategies.
- Filtering rules have been moved to the recommendations page, but the base spec reserves the `filter` query param.
- Support for creating, updating, and deleting multiple resources in a single request has been moved to the Bulk extension.
- JSON Patch support has been moved to the Patch extension.
- URL rules have been moved to the new Recommendations page.
- Response statuses have been further defined.

Please review and leave general feedback below. I hope we can merge this soon and handle specific concerns in more targeted issues.

---

Some relevant links to modified pages:
- [Base specification](https://github.com/dgeb/json-api/blob/v1rc2/format/index.md)
- [Extensions](https://github.com/dgeb/json-api/blob/v1rc2/extensions/index.md)
  - [Bulk extension](https://github.com/dgeb/json-api/blob/v1rc2/extensions/bulk/index.md)
  - [Patch extension](https://github.com/dgeb/json-api/blob/v1rc2/extensions/patch/index.md)
- [Recommendations](https://github.com/dgeb/json-api/blob/v1rc2/recommendations/index.md)


tkellen:
cc @bobholt

aaronshaf:
Thank you for your work on this!

format/index.md
```diff
@@ -52,636 +58,370 @@ called out below.
...
-      "body": "Dependency Injection is Not a Vice"
-    }]
+  "meta": {
+    "copyright": "Copyright 2015 Example Corp.",
+    "authors": [
+      "Yehuda Katz",
+      "Steve Klabnik",
+      "Dan Gebhardt"
+    ]
+  }
```
kevin-brown: There appears to be a comma missing here.

kurko:
Sad Panda. I can see the benefits, but at the same time it means my team (backend+frontend web+mobile) either changes tons of endpoints or leaves JSONAPI entirely (at least for the time being). Nonetheless, I am +1 on this.

gr0uch:
Looks like a good revision/rewrite of the spec, clarifies a lot of ambiguities. Though I have to say that to many who expected a slow moving spec that will be upgraded extensibly at the 1.0-rc phase, massive breaking changes is a disappointment. +1 though.

With regards to versioning I don't see any elaboration on this? How can we expect the `application/vnd.api+json` media type to be consistent with our expectations?

bobholt:
I'll probably give this a full read on Monday, but in general, positive on removing ambiguities and moving toward marking the spec as stable. Thanks for the heads-up, @tkellen.

xogeny:
My interest in JSON-API comes from the fact that I think Siren, HAL and Collection+JSON are all very good first steps with respect to designing hypermedia APIs but I'm always on the lookout for new media types that help to improve on those.

Unfortunately, JSON-API takes a big step backwards (in my opinion) by making the design decision that entities MUST have an `id` but only MAY have a `self` URI.  I feel this disregards best practices with respect to hypermedia APIs.

Now, to your credit a) you don't seem to claim that this is a hypermedia format anywhere on your web site (although I think it is implicit in the presence of links) so perhaps this isn't a fair comparison (although you yourselves make comparisons to HAL) and b) I strongly suspect that this is really a compromise design to try and maximize compatibility with existing conventions for APIs (e.g., Ruby on Rails).  So, these seem to be explicit design decisions and not something you've overlooked.

You are, of course, free to make those decisions and I'm free to not like them.  I was just hoping for something that didn't reinforce what I consider to be bad practices (at least when trying to move in the HATEOAS direction).  Most people probably think I'm just being a RESTafarian, but the issue for me isn't really about RESTfulness but more about raising awareness of self-discoverable APIs where clients don't need to know/construct URLs for endpoints.

ahx:
Two questions, which i would probably have to answer to my colleagues, but i am not sure how:
1. Are underscores allowed in attribute names? (Please?)
2. Why the change from "href" to "resource" in link objects?

:hurtrealbad: :heart: 
The rest of the changes will definitly hurt (like "type is required in all requests" or "id field must be in all responses  and must be a string"), but make sense.
+1

dgeb:
@kevin-brown thanks for the correction - fixed!

@aaronshaf you're welcome - hope it meets your needs!

@kurko @daliwali As a maintainer of a couple implementations myself, I completely sympathize with your wishes to change the spec slowly and minimize breakages as we approach 1.0. As difficult as this massive set of changes might be to absorb, the goal is to stabilize the spec at 1.0 very soon and avoid such turbulence afterward. My view of the current (rc1) spec is that it's more of a "guidebook" with so much wiggle room that client and server implementations will rarely match up without customization. We've really tried to remove such ambiguities with this PR, and I'm hopeful from your +1s that you agree.

@bobholt looking forward to your feedback.

@xogeny I feel that JSON API has all the pieces in place to be a hypermedia format, but that full hypermedia support is definitely optional. Self links and richer relationship links would need to be required instead of optional, and it's a good question whether this is a bridge too far for many implementations. Perhaps we should consider a "hypermedia extension" that changes these MAYs to MUSTs?

@ahx thanks for your questions. As for our decision to go with dasherized attributes - we wanted to avoid requiring out-of-band knowledge to parse a response. One approach to this in the spec is to say that, for instance, a client MUST convert underscores to dashes (which provides some wiggle room for servers). We took a more direct approach by making the requirement on the server. Let's discuss further. Also, the change from `href` to `resource` was driven by the introduction of `self` links, and the possible confusion between the target of `self` vs. `href`.

Thanks to everyone so far for your comments and review. I know that this many changes isn't easy to process all at once and can be tough to swallow for those with current implementations. Trust me when I say it hasn't been easy for me either and I am very much looking forward to a stable 1.0.

darkbaby123:
I also don't think dasherized attribute is required. I understand that you want people to parse the response to fit the model layer or other things. But it doesn't mean we need to choose a "hard to used directly" attribute format. It feels like increasing the parse difficulty to force people to do right thing. While it doesn't have to be.

gr0uch:
hmm, also I disagree with the dasherized key names as well, seems unnecessary. I would be in favor of dropping any requirement on key naming.

barelyknown:
Are you or @wycats able to provide any insight as to the role that json:api will play with Ember Data? If the plan is for json:api to be supported as the default serializer/adapter format, the changes seem totally worth it (even if they are large and a bit of a pain to deal with for existing implementations).

jacqueline-homan:
@dgeb and @wycats : I've been knee-deep in a MAJOR F# project that I've pretty much had to do 90% of the work on all by myself so far where once the database end is completely set up, will be transformed into a web app. It is a data-heavy app (think: Intellius on steroids but for providing dossiers on NGO's instead of individuals). As of right now, I am writing data to a JSON file. How well will the JSON API play with an F# app?  Or at this stage, should I consider changing from writing to JSON to XML and then build a custom F# parser for the XML as part of the app?

wycats:
@barelyknown the plan is for Ember Data to have first class support for JSON API. Feedback from Ember Data was a big input into this round of changes.

barelyknown:
@wycats As long as that's the case, I'm all for it. Hopefully we can finalize v1 soon.

csantero:
@jacqueline-homan Check out [JSONAPI.NET](https://github.com/SphtKr/JSONAPI.NET).

mitchlloyd:
@dgeb Would it be possible to add an example to the pagination section of the document? I read through this section several times and I'm still having trouble visualizing how this is supposed to work. I think an example here would be really helpful.

 I'm very excited about these changes. Thank you for your hard work.

jerel:
Why switch to dasherized key names when that's not a valid format for variables on server or client? Curious what the reason is.

Otherwise, I'm very happy to have this stablize on this small set of requirements so we have a target to aim for.

johnnyoshika:
Dasherized attribute name would be really unfortunate, and would make it cumbersome to consume the API using JavaScript.

kurko:
> As for our decision to go with dasherized attributes - we wanted to avoid requiring out-of-band knowledge to parse a response @dgeb 

*Out-of-band knowledge*? How's dashes better? In 9 years of working, I've never seen an API using dashes, not even docs mentioning it. I really don't understand this argument. Could you elaborate?

format/index.md
```diff
@@ -52,636 +58,370 @@ called out below.
...
 
-#### Resource URLs <a href="#document-structure-resource-urls" id="document-structure-resource-urls" class="headerlink"></a>
+Any other member in a resource object represents an "attribute", which may
+contain any valid JSON value.
 
-The URL of each resource object **MAY** be specified with the `"href"` key.
-Resource URLs **SHOULD** only be specified by the server and therefore are
-typically only included in response documents.
+Attribute names **MUST** consist of only lower case alphanumeric characters
+and dashes (U+002D: HYPHEN-MINUS, "-"). Attribute names **MUST NOT** begin
```
kurko: > As for our decision to go with dasherized attributes - we wanted to avoid requiring out-of-band knowledge to parse a response @dgeb 

_Out-of-band knowledge_? How's dashes better? In 9 years of work, I've never seen an API using dashes, not even docs mentioning it. I really don't understand this argument. Could you elaborate?

dgeb:
I discussed with @wycats and we've agreed to move attribute and association naming requirements out of the base spec and into the new Recommendations page. I'll elaborate further once I've updated this PR.

sandstrom:
This looks great! (my only comment is that dasherized attribute names feels suboptimal, but it seems like they're already gone)

dgeb:
I've just amended this PR with a commit to move field naming requirements to the new Recommendations page. We're now treating all names (including those for attributes, associations, and resource types) and URLs as opaque to the spec. This keeps the spec leaner and more flexible.

The naming rules did mitigate a potential conflict with default sort order, in which names couldn’t begin with a `-`. As an alternative, the concept of a default order is removed; sort order must now be explicitly specified with a `+` or `-` prefix. For instance, `GET /posts?sort=-created,+title`.

In hindsight, I think the naming requirement did cross lines, so thanks to everyone for pushing back on it.

P.S. We can continue to discuss recommendations well after the spec hits 1.0, so let's do so in more focused issues / PRs after this is merged.

SphtKr:
:+1: Thrilled that `id` is required again!

@dgeb Questions:
- The [Sparse Fieldsets](https://github.com/dgeb/json-api/blob/v1rc2/format/index.md#sparse-fieldsets-) section doesn't have the same "A server MAY choose to support..." caveat that the Inclusion, Sorting, and Pagination sections do. I imagine this is an omission, but just in case, are servers/endpoints required to support sparse fieldsets?
- Are dot-separated paths supported for sorting as they are for inclusion? Forbidden? Agnostic?
- Slightly pedantic, but arguably "The top level of a document MUST NOT contain any additional members [besides `data`, `meta`, `links`,`linked`]." and "Error objects MUST be returned as a collection keyed by "errors" in the top level of a JSON API document" conflict. (I'd like to return errors at the same time as data but frankly I'm probably wrong).

@wycats Thrilled to hear that Ember Data is still targeting this, that's why I started [JSONAPI.NET](http://github.com/SphtKr/JSONAPI.NET) in the first place, to facilitate making Ember Data apps (@jacqueline-homan yes, come have a look we'd appreciate your feedback!)

SphtKr:
One more question:

It looks as though requesting multiple IDs via a single GET (e.g. `/posts/1,2,3`) has been removed, and I suppose the only part of the spec bearing on this now is the [Filtering Recommendations](https://github.com/dgeb/json-api/blob/v1rc2/recommendations/index.md#recommendations-for-filtering-), where you'd accomplish this via `posts?filter[id]=1,2,3`. That's okay, but it's funny to me that there's a bulk extension covering PUT, POST, and DELETE but no clear way to do for GET.

miguelsan:
One question. [In the FAQ](http://jsonapi.org/faq/#why-is-json-api-not-versioned) we read:

> Once JSON API is stable, it will always be backwards compatible using a never remove, only add strategy. #46 

What's been out of that, when we are talking about _v1.0_? Shouldn't it be _stable_? Has the strategy changed?

dgeb:
@mitchlloyd there are simple pagination examples on the [home page](https://github.com/dgeb/json-api/blob/v1rc2/index.md) and the [extensions page](https://github.com/dgeb/json-api/blob/v1rc2/extensions/index.md#profiles-). These both make use of the reserved links (`prev`, `next`, `first`, and `last`) as well as the reserved `page` query param. The example profile (on the extensions page) shows how an implementation might include additional pagination data in `meta`.

@SphtKr those are all valid points / questions that I'll try to resolve soon. I need to confer with @steveklabnik and @wycats.

@miguelsan we hope to release v1.0 very soon and then maintain the stability mentioned in the FAQ afterward. We're trying to get all the churn out now, so this is a critical time for the spec.

miguelsan:
Another one. It top-level `links` relate only to primary `data`, shouldn't they be defined inside it? Currently the spec allows only `self` and pagination inside those top-level `links`, and since `self`is already under `data`, that be a small change. Moreover when pagination has not been part of the spec till now, it would not break anything and the API would be simplified.

miguelsan:
Another one. If top-level `links` relate only to primary `data`, shouldn't they be defined inside it? Currently the spec allows only `self` and pagination inside those top-level `links`, and since `self`is already under `data`, that'd be a small change. Moreover when pagination has not been part of the spec till now, it would not break anything and the API would be simplified.

dgeb:
@miguelsan When primary data is a collection, the value of `"data"` is an array. There's no place inside that array for a member object that applies to the whole collection (a sibling of the members would be odd and difficult to parse). Thus, the top-level `"links"` is used.

If I'm misunderstanding you, please provide an example.

miguelsan:
You are right. The lack of an example of a `data` array misled me. How should it look like for the `"type"`, should it be repeated inside each item of the collection? I think that this point is not clear enough.

miguelsan:
You are right. The lack of an explicit example for a `data` array misled me. I'd be nice to have.

format/index.md
```diff
@@ -52,761 +58,507 @@ called out below.
...
+* A `resource` member, whose value is a related resource URL (as defined above).
+* Linkage to other resource objects ("object linkage") included in a compound
+  document. This allows a client to link together all of the resource objects
+  included in a compound document without having to `GET` one of the
+  relationship URLs. Linkage **MUST** be expressed as:
+  * `type` and `id` members for to-one relationships.
+  * `type` and `ids` members for homogenous to-many relationships.
+  * A `data` member whose value is an array of objects each containing `type`
+    and `id` members for heterogenous to-many relationships.
+* A `"meta"` member that contains non-standard meta-information about the
```
bobholt, line 228: Is "meta" really one of the "at least one," or is it an additional **MAY** on top of at least one of the preceding three? In other words: could I include an object that _only_ has a meta property?

miguelsan:
When we are at it, I expose my wishes in here, see if you like them.
I personally dislike having `links` and `linked` as keywords in the spec, since they are too close to each other and increase the cognitive overhead on reading.

For my taste, this structure would be nicer:

```
{
  "meta": {
    "about": "jsonapi"
  },
  "links": {
    "self": "site.com/main-resources/1",
    "last": "site.com/this/url?page=last"
  },
  "data": {
    "primary": [{
      "type": "main resource",
      "id": "1"
      "links": {
        "resource": "site.com/linked-resources/1"
        "type": "linked resource",
        "id": "1"
      }
    }, {
      "type": "main resource",
      "id": "2"
    }],
    "secondary":  {
      "type": "linked resource",
      "id": "1"
    }
  }
}
```

No `linked`, just divide data into primary and related resources. What do you think?

format/index.md
```diff
@@ -52,761 +58,507 @@ called out below.
...
-GET /people?fields=id,name,age
-```
+* Plus (U+002B PLUS SIGN, "+") to request an ascending sort order.
+* Minus (U+002D HYPHEN-MINUS, "-") to request a descending sort order.
 
-An endpoint **MAY** support requests that specify fields for any resource type
-with a `fields[TYPE]` parameter.
+> Note: By requiring a sort order prefix instead of allowing a default
+order, JSON API avoids setting requirements for the first character in field
+names.
```
bobholt: It's not clear to me why this is a good trade-off. I don't see that it's particularly bad, either, but does put the added burden on the client to explicitly state a sort order every time.

If JSON API were to set requirements for the first character in field names, it seems like they would tend to match existing programming language requirements. Requiring that field names not start with PLUS, MINUS, or other non-numerical characters doesn't seem like a huge overreach.

gr0uch:
@miguelsan that would make parsing more verbose, since the primary collection is at `data.primary` instead of `data` and the linked entities at `data.secondary` instead of `linked`. The way it is currently in this PR is more concise.

miguelsan:
Sure. It is a trade off for a better representation of the underlying concepts.

But even if the spec goes the shorter way, it should still remain semantically consistent. `linked` is also `data` in the same sense of _content_. It could at least get a less confusing naming on the top level, such as _primary/secondary_ or _main/related_ instead of _data/linked_.

lgebhardt:
@miguelsan support for compound documents is optional. Nesting data under data.primary would be more verbose for servers that do not implement compound documents. Same for requests that do not include the related data. It helps for me to read `linked` as `linked_data`.

miguelsan:
Sorry, but if you want people to read `linked_data` you cannot write `linked`. You must write `linked_data`. That's a hard fact ;-)

format/index.md
```diff
@@ -1004,125 +789,176 @@ Content-Type: application/vnd.api+json
...
-to be added and removed individually. This can be done by making fine-grained
-requests to the relationship URL.
+If the relationship is updated successfully then the server **MUST** return
+a `204 No Content` response.
+
+#### Updating To-Many Relationships <a href="#crud-updating-to-many-relationships" id="crud-updating-to-many-relationships" class="headerlink"></a>
+
+A server **MUST** respond to `PUT`, `POST`, and `DELETE` requests to a *to-many
+relationship URL* as described below.
+
```
bobholt: In the spirit of not requiring the client to stop and consider the differences between updating a to-one vs a to-many relationship, would it make sense for "Updating To-One Relationships" to support `POST` and `DELETE` with as equivalent to `PUT`when the existing relationship is `null` (erroring otherwise) and `PUT` to set the relationship to `null`, respectively?

SphtKr:
@miguelsan it might be better to have the key naming discussion over on #312 where it's already started...and where I made my own [humble suggestion](https://github.com/json-api/json-api/issues/312#issuecomment-73556579) 😜

format/index.md
```diff
@@ -52,761 +58,508 @@ called out below.
...
 
-### Creating Resources <a href="#crud-creating-resources" id="crud-creating-resources" class="headerlink"></a>
+### Creating Resources <a href="#crud-creating" id="crud-creating" class="headerlink"></a>
 
-A server that supports creating resources **MUST** support creating individual
-resources and **MAY** optionally support creating multiple resources in a single
-request.
+A resource can be created by sending a `POST` request to a URL that represents
+a collection of resources. The request **MUST** include a single resource object
+as primary data. The resource object **MUST** contain at least a `type` member.
```
bobholt, line 572: > > The request **MUST** include a single resource object-request as primary data.

This, and similar statements in updating and deleting, is violated by the bulk extension. If that violation is allowable (and it should be, at least for official extensions), then perhaps there should be a statement to that effect in /extensions/index.md

format/index.md
```diff
@@ -52,761 +58,508 @@ called out below.
...
-A server **MAY** choose to support requests to return only specific fields in
-resource object.
+* Plus (U+002B PLUS SIGN, "+") to request an ascending sort order.
+* Minus (U+002D HYPHEN-MINUS, "-") to request a descending sort order.
 
-An endpoint **MAY** support requests that specify fields for the primary
-resource type with a `fields` parameter.
+> Note: By requiring a sort order prefix instead of allowing a default
+order, JSON API avoids setting requirements for the first character in field
+names.
```
bobholt, line 501: It's not clear to me why this is a good trade-off. I don't see that it's particularly bad, either, but does put the added burden on the client to explicitly state a sort order every time.

If JSON API were to set requirements for the first character in field names, it seems like they would tend to match existing programming language requirements. Requiring that field names not start with PLUS, MINUS, or other non-numerical characters doesn't seem like a huge overreach.

format/index.md
```diff
@@ -1004,125 +790,176 @@ Content-Type: application/vnd.api+json
...
-system which may involve many editors. An alternative is to allow relationships
-to be added and removed individually. This can be done by making fine-grained
-requests to the relationship URL.
+If the relationship is updated successfully then the server **MUST** return
+a `204 No Content` response.
+
+#### Updating To-Many Relationships <a href="#crud-updating-to-many-relationships" id="crud-updating-to-many-relationships" class="headerlink"></a>
+
+A server **MUST** respond to `PUT`, `POST`, and `DELETE` requests to a *to-many
+relationship URL* as described below.
```
bobholt, line 938: In the spirit of not requiring the client to stop and consider the differences between updating a to-one vs a to-many relationship, would it make sense for "Updating To-One Relationships" to support `POST` and `DELETE` with as equivalent to `PUT`when the existing relationship is `null` (erroring otherwise) and `PUT` to set the relationship to `null`, respectively?

dgeb, line 938: @bobholt We explored that approach, but ultimately decided that it was an extra requirement on the server that didn't provide any strong benefits. We're really trying to limit code paths as much as possible at this point to simplify implementations. Plus, the `PUT` approach is already consistent for both to-one and to-many relationships.

bobholt:
I made a few line comments, but those are mostly requests for clarification or consideration. Overall, I'm :+1: with the spec as it stands. 

I was a little uneasy over the dasherized requirement, so :+1: on that having moved to recommendation status, too.

format/index.md
dgeb, line 501: That's a legitimate take on this - and I can't say that I haven't been on the fence here. We decided on this trade-off so that names would have no restrictions, and to keep sorting requirements only affecting implementations that choose to support sorting. In other words, we wanted to avoid having an optional aspect of the spec driving requirements in a very core part of the spec.

format/index.md
dgeb, line 572: Agreed! The Patch extension also dictates a different top level structure and I've been struggling with how to best formalize these conflicts.

format/index.md
SphtKr, line 501: For history there's a long discussion on this in #324.

format/index.md
```diff
@@ -52,761 +58,508 @@ called out below.
...
 ```
 
-Note: a request for `comments.author` should not automatically also include
-`comments` in the response (although comments will obviously need to be queried
-in order to fulfill the request for their authors).
-
-Multiple linked resources could be requested in a comma-separated list:
+An endpoint **MAY** support multiple sort criteria by allowing
+comma-separated (U+002C COMMA, ",") fields as the value for `sort`. Sort
+criteria should be applied in the order specified.
```
bobholt: Should this "should" be a **SHOULD**?

fcy:
+1 for the new specs. I'm starting to code a new client and will use instead of the current one on jsonaspi.org

format/index.md
```diff
@@ -52,761 +58,508 @@ called out below.
...
-    "id": "1",
-    "title": "Rails is Omakase",
-    "links": {
-      "author": "17"
-    }
-  }
-//...
-```
+> Note: This spec is agnostic about inflection rules, so `type` can be either
+plural or singular. However, the same value should be used consistently
```
fcy: Why not make simpler and have it mandatory the singular `type`?

gr0uch: Nobody can agree on naming things, so I think that no naming scheme should be mandatory in the spec.

fcy: The other members of a resource object have specific names: `links`, `id`, and `meta`, why is it important for "type" to be different?. 

All other attributes can be named freely, which is great. IMO this note just adds complexity to the spec with no gain.

dgeb: Just to be clear (and I'm probably mis-reading your comment), we're not discussing the key `type` vs `types`, but rather the value associated with the `type` member (e.g. `comment` vs `comments`, `blog-post` vs `blog-posts`). So the challenge becomes not only making rules about pluralization but also word delimiters. All together it seems consistent to leave naming decisions, including those for types, attributes, associations and URLs, as recommendations rather than requirements. If any of these can be freely named, then why not all of them?

fcy: > Just to be clear (and I'm probably mis-reading your comment), we're not discussing the key type vs types, but rather the value associated with the type member (e.g. comment vs comments, blog-post vs blog-posts).

Oh my bad, I misinterpreted the note. Makes total sense to not enforce the value of `type`.

What about adding the following - in bold - to the sentence:

> so `type`**'s value** can be either plural or singular.

dgeb: @fcy - ah, that was ambiguous. Fixed - thanks!

format/index.md
dgeb: Good catch - fixed. Thanks @bobholt!

recommendations/index.md
```diff
@@ -0,0 +1,138 @@
...
+characters and dashes (U+002D HYPHEN-MINUS, "-").
+
+It is also recommended that resource types be pluralized. Dasherized and
+pluralized resource types can be used as URL segments without translation,
+as discussed below.
+
+## Recommendations for URL Design <a href="#urls" id="urls" class="headerlink"></a>
+
+### Reference Document <a href="#urls-reference-document" id="urls-reference-document" class="headerlink"></a>
+
```
bobholt, line 23: This section nicely describes reference documents, but doesn't make an explicit recommendation. It probably should.

format/index.md
```diff
@@ -1004,125 +790,176 @@ Content-Type: application/vnd.api+json
...
 
-A to-one relationship **MAY** be removed by sending a `DELETE` request to
-the URL of the relationship. For example:
+##### 404 Not Found <a href="#crud-updating-responses-404" id="crud-updating-responses-404" class="headerlink"></a>
 
-```text
-DELETE /articles/1/links/author
-```
+A server **MUST** return `404 Not Found` when processing a request to modify
+a resource that does not exist.
```
tkellen, line 855: This should be 500 unless the client sent an if-match header, in which case it should be 412. It's definitely not 404.
http state chart

dgeb, line 855: Source for diagram - https://github.com/basho/webmachine/wiki/Diagram

cc @steveklabnik @wycats

kevin-brown, line 855: ~~I fail to see why this should be a 500 error instead of a 404, especially considering the diagram appears to point to a 404. Unless you know the resource previously existed, in which case it would be a 410.~~

Looks like I misread the diff as being for `DELETE`. [Read below for my updated thoughts.](https://github.com/json-api/json-api/pull/341/files#r24473002)

steveklabnik, line 855: I like that diagram, but would like to see some supporting evidence from the RFCs.

format/index.md
csantero, line 228: Does it make sense for a relationship link object to only have a `self` key, like the following?

```
{
    "id": "1",
    "type": "comment",
    "links": {
        "author": {
            "self": "/comments/1/links/author"
        }
    }
}
```

I understand that the relationship URL is useful for PUT/POST/DELETE, but if the server only sends that down without a `resource` key or an id/type combination, how is the client meant to know how to fetch the resource at that relationship? Should `GET /comments/1/links/author` return the same thing as `GET /comments/1/author`? If it should, then the spec should mention that (I don't think I saw anywhere that mentions what GETting a relationship url should do).

It seems to me that if the answer to that last question is no, then a relationship link object MUST include at least one of the following:
- `id`/`ids` and `type`.
- `resource`

> Is "meta" really one of the "at least one," or is it an additional MAY on top of at least one of the preceding three? In other words: could I include an object that only has a meta property?

I'd also like some clarification on this. `meta` is in the same boat as `self`.

BTW, love these changes overall, I'm just nitpicking.

format/index.md
ahacking, line 855: DELETE is an [idempotent method](http://tools.ietf.org/html/rfc7231#section-4.2.2)

[DELETE method](http://tools.ietf.org/html/rfc7231#section-4.3.5)

```
   If a DELETE method is successfully applied, the origin server SHOULD
   send a 202 (Accepted) status code if the action will likely succeed
   but has not yet been enacted, a 204 (No Content) status code if the
   action has been enacted and no further information is to be supplied,
   or a 200 (OK) status code if the action has been enacted and the
   response message includes a representation describing the status.
```

One can argue that because DELETE is idempotent then a 204 response for an already deleted resource is most appropriate if it is known that the resource did exist at some point. If it is not known then an error response seems most appropriate.

HTTP response codes are designed to allow the server to be as specific as it can with errors partitioned into classes.

The error classes that seem most appropriate in order of specificity are:

[Response 410](http://tools.ietf.org/html/rfc7231#section-6.5.9)

[Response 404](http://tools.ietf.org/html/rfc7231#section-6.5.4)

[Response 5xx](http://tools.ietf.org/html/rfc7231#section-6.6):

```
   The 5xx (Server Error) class of status code indicates that the server
   is aware that it has erred or is incapable of performing the
   requested method.
```

Given the above definition, 500 should be seen as the error of last resort as it reflects **the server** has _erred_ or cannot perform the operation (implying it ordinarily would if it was able to). It certainly does not help the client or intermediate proxy/caches to know the state of the resource.

Maybe a better approach is some guidance around the priority of appropriate responses anchored to the RFC.  It would need to cover the the use cases of deleting a resource that once existed, vs deleting a resource that never existed, vs deleting a resource where previous existence is not known.

tkellen, line 855: @ahacking If you look at the diff, you'll see this section of the spec is not speaking about DELETE.

Here is the content:

> 404 Not Found
> A server MUST return 404 Not Found when processing a request to modify a resource that does not exist.
> A server MUST return 404 Not Found when processing a request that references a related resource that does not exist.

kevin-brown, line 855: The diagram looks like it is special casing `PUT` for PUT-as-create situations, which JSON API doesn't appear to be advocating. In that situation, if the server was not able to create the resource, it would be an internal server error (500) and the diagram makes sense.

The JSON API spec sounds like it is not dealing with PUT-as-create, so updating a nonexistent resource still sounds like a client error instead of a server error. The server _did not fail_ to process the request, _it wasn't able to_ because the resource doesn't exist.

From the [HTTP/1.1 spec on PUT](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.6)

> If the Request-URI does not point to an existing resource, **and that URI is capable of being defined as a new resource** by the requesting user agent, the origin server can create the resource with that URI.

_(emphasis mine)_

The specification does not require PUT-as-create, it actually makes it pretty clear that it should only happen if the server chooses.

> If the resource could not be created or modified with the Request-URI, an appropriate error response SHOULD be given that reflects the nature of the problem.

This leaves the response code open to interpretation, which explains why the idea of returning a 500 or a 404/410 has come up.

ahacking, line 855: @tkellen my bad, misread the comments as applying to DELETE.  At least there is some supporting documentation from the RFCs.

I stand by what I said for DELETE and response code specificity so if JSON API is counter to what I wrote then that would be another issue to deal with.

bintoro:
This is starting to look really good.

I was a little disappointed by how the spec now mandates the use of `PUT` for partial updates. Perhaps this is a needlessly academic concern, but isn't it a reasonable requirement that a client should be able to perform a true `PUT` to request the resetting of some server-side properties that it doesn't have direct control over?

In #327 I proposed that a JSON API server could declare whether it supports true `PUT` along with `PATCH` or if all updates are partial. (In the latter case `PUT` and `PATCH` requests with `application/vnd.api+json` would be considered synonymous).

The _Recommendations_ document is going to be very useful, and I'd like to suggest that it be expanded to cover the manipulation of nested resources as well. What it means to modify `/posts/1/tags` vs. `/posts/1/links/tags` is quite clear once you understand the concepts, but this is almost guaranteed to be a point of confusion.

Would there be interest in publishing, as some sort of appendix, a table that puts together all the various HTTP method / URL structure / relationship permutations and gives a summary of the intended outcome of each operation? Kind of like what I sketched [here](https://github.com/json-api/json-api/issues/317#issuecomment-65967368), although that's now outdated since #311 didn't quite materialize.

tkellen:
Are nested relations readable via `/resource/:id/links/:relation`? Updatable?

bobholt:
@tkellen: It's a recommended URL structure: https://github.com/dgeb/json-api/blob/v1rc2/recommendations/index.md#relationship-urls-and-related-resource-urls-

tkellen:
@bobholt Yup, I understand that!  I'm asking if the spec can be more explicit about whether it supports/recommends nested relations, like this: `/resource/1/links/nested.relation`

format/index.md
dgeb, line 855: I agree with @kevin-brown's [comment above](https://github.com/json-api/json-api/pull/341#discussion_r24473002) and the importance of the clause "and that URI is capable of being defined as a new resource by the requesting user agent".

The [PUT spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.6) goes on to say:

> The recipient of the entity MUST NOT ignore any Content-\* (e.g. Content-Range) headers that it does not understand or implement and MUST return a 501 (Not Implemented) response in such cases.

A client could reasonably expect a 501 if the server does not support the JSON API `Content-Type`. However, if the server does support the JSON API content type, then there is an understanding between client and server about this specification. Namely, in this scenario, the understanding is that PUT-as-create is not supported by the server.

So, having passed the `Content-*` negotiation, I believe that this spec is within its rights to recommend status codes that will be most clear and useful to clients, as long as they don't violate HTTP. And I believe that a 404 / 410 response is much more useful to clients than further overloading 501, which I think will only lead to confusion.

dgeb:
I talked with @wycats last night about many of the questions raised so far and some remaining items to finish.
- Should sparse fieldset support be a requirement?
  - Favoring YES. We don't think this is much of an onus on implementators, and provides a big win for clients.
- Are dot-separated paths supported for sorting?
  - YES. PR needs to be updated to reflect this.
- Need to refine statement about possible elements in the top-level of a document.
  - Ok, PR needs to be updated to call out `errors` as a possible top-level member.
- We've lost the ability to fetch multiple individual resources (e.g. `/posts/1,2,3`).
  - We considered adding `GET` to allow this in the bulk extension, but requiring any payload kills cacheability. So, we're favoring leaving this special case out of the spec. Resources can still be filtered by `type` and `id`.
- Can a link object contain only `meta`?
  - YES. This member alone could be extended to provide a lot of useful information.
- Appropriateness of `data` / `linked`?
  - Continue discussion in #312.
- TODO: Add section related to fetching relationships (e.g. `GET /posts/1/links/comments`)
- TODO: Clarify that 204 is an appropriate response to DELETE to indicate that a client's representation is correct.

I'll update this PR soon to reflect the above. If we still haven't addressed your questions / concerns, please let us know and we'll do our best. As a reminder - even after this gets merged, there will be some time (hopefully brief) to discuss everything in more focused issues before finalizing the spec at 1.0.

dgeb:
> I'm asking if the spec can be more explicit about whether it supports/recommends nested relations, like this: `/resource/1/links/nested.relation`

@tkellen Would it cause conflict anywhere to assume that would be allowed implicitly because the name of the relation is opaque to the spec (and therefore could contain a `.`)?

tkellen:
This is similar to the concern I brought up in #336.

> Say you have an organization with projects and the projects have utilizations (days worked by employees). A request to utilizations?include=project.organization is handy, whereas a request to utilizations?include=project.organization.projects drills into a relation that has nothing to do with the primary resource.

Similarly, posting to `/utilizations/links/project.organization.projects` to alter that "relationship" is kind of nonsensical.

bintoro:
@tkellen Sorry if I misunderstand, but are you suggesting that the spec should weigh in on what kinds of deep references should be supported? I firmly believe this sort of thing must be left to applications to decide. I've written a server-side implementation that allows arbitrarily deep URLs, and the only thing I've found inherently nonsensical is if a resource appears twice, i.e., the path becomes circular. And even that is not really a case that can't be handled so much as an indication that the client is buggy.

Would `/utilizations/links/project.organization.projects` be synonymous to `/utilizations/project/organization/links/projects` or is there some subtle difference I'm not seeing?

Because if there's no difference, then we already have the recommendation you're looking for. The slash-separated path logically follows from the current recommendations, although it might be worthwhile to spell this scenario out explicitly as well.

**EDIT:** Observe how the `links` segment only appears at the end, right before the actual relationship we're requesting. A URL such as `/utilizations/links/project/organization` would be invalid because `/utilizations/links/project` is already a link — it doesn't contain a member called `organization`. This is why I'm not fond of the dotted notation that would place `links` after the root.

mattecapu:
Could someone elaborate on the example reported [here](https://github.com/dgeb/json-api/blob/v1rc2/format/index.md#updating-to-one-relationships-)?
Shouldn't we use the `DELETE` method?

bobholt:
@mattecapu You can see the response to a similar question I asked above: https://github.com/json-api/json-api/pull/341#diff-for-comment-24350137

mattecapu:
@bobholt thanks, I didn't saw that.
However, my question still holds... Why To-One relationships support only PUT, while to-many support also DELETE and POST?

dgeb:
@mattecapu It's a subtle distinction. PUT serves to replace the current value; it is equivalent for both to-many and to-one relationships. POST / DELETE act upon _individual members_ of a to-many relationship, rather upon the whole relationship, so there is no direct equivalent for to-one relationships.

mattecapu:
@dgeb if, as the spec says, I refer to HTTP semantics, `DELETE` doesn't make any distinction between a single resource of a resource part of a collection.

Citing from [RFC7231](http://tools.ietf.org/html/draft-ietf-httpbis-p2-semantics-22#section-4.3.5)

> The DELETE method requests that the origin server remove the
>    association between the target resource and its current
>    functionality.  In effect, this method is similar to the rm command
>    in UNIX: **it expresses a deletion operation on the URI mapping of the
>    origin server, rather than an expectation that the previously
>    associated information be deleted.**

(emphasis mine)

I think it reflects perfectly what we intend to do with relationships URLs: make them act on the relationships instead of the resources behind.
`PUT`ting a `null` value to delete a relationships makes sense when we [update a relationships along with the resource attributes](https://github.com/dgeb/json-api/blob/v1rc2/format/index.md#updating-a-resources-to-one-relationships-). Otherwise I can't see any reason why `DELETE` wouldn't be preferrable.

dgeb:
@mattecapu The bottom line is that we're trying to avoid _requiring_ support for multiple paths to solve the same problem. In other words we're trying to avoid saying that an implementation MUST support updating a to-one relationship with PUT _and_ POST / DELETE. Clearly you think that DELETE is preferable. However, DELETE needs to be paired with POST for creation / updating, and I think the argument for supporting POST for updating a (possibly pre-existing) to-one relationship is much weaker.

I should also add the obvious that implementations can choose to support other paths that the spec doesn't address.

mattecapu:
@dgeb oh now it's clear
I agree we need to state that implementations MAY support other methods

nwjsmith:
@dgeb great work on this. I see all of the changes as out-and-out wins over RC1. Is there anything preventing this PR from getting merged? It seems like any issues raised in this thread could be addressed in subsequent smaller PRs/issues.

tkellen:
:+1: to merging.

steveklabnik:
I'm :+1:  to merging as well, mostly waiting until discussion dies down.

fcy:
+1 to merging

ttasanen:
:+1:

gr0uch:
:+1: :beer:

bintoro:
@dgeb, could you perhaps elaborate a bit on this:

> DELETE needs to be paired with POST for creation / updating

I'm not arguing against the use of PUT for deleting associations, just curious about whether this is simply a part of the JSON API philosophy or if it's based on something else. The HTTP spec doesn't seem to mandate that DELETE be paired with POST.

From RFC 7231 section 4.3.5. DELETE:

> For example, a resource that was previously created using a PUT request, or identified via the Location header field after a 201 (Created) response to a POST request, might allow a corresponding DELETE request to undo those actions.

I agree with others that this PR is ripe to be merged as is. Any remaining concerns can be discussed in separate issues.

dgeb:
> I'm not arguing against the use of PUT for deleting associations, just curious about whether this is simply a part of the JSON API philosophy or if it's based on something else. The HTTP spec doesn't seem to mandate that DELETE be paired with POST.

@bintoro I'm definitely arguing for consistency across JSON API, which supports POST but not PUT for creation of resources. My phrase "DELETE needs to be paired with POST" overstated this point.

dgeb:
Thanks everyone for taking the time to review. I don't want to sneak changes in to this PR that you may or may not have a chance to review, so I'm going to stop making adjustments.

I made one minor tweak above to indicate that `type` is not required in link objects for empty relationships (this was a change I had made and forgot to commit). I imagine this is non-controversial, but let me know otherwise in a separate issue after the merge.

@steveklabnik @wycats let's merge unless you feel there are any major blockers. There are a few issues I'd still like to raise, and I'm sure others feel the same, but it's probably best to wait until after the merge.

mgenev:
@wycats is there a timeline for ember data matching this spec?

wycats:
@mgenev Stay tuned! You will be happy with the answer!

barelyknown:
@wycats That is such good news. Awesome.

mtparet:
> Every resource object MUST include both type and id members

Usually the name of the resource is the type so why do we need to specify again although there is no need ? (it does not mean we can't specify it when needed)

nwjsmith:
@mtparet AFAIK it's to make supporting polymorphic relationships trivial to support, and it greatly simplifies the lookup of resources in client caches/`linked` resource lists. See: http://jsonapi.org/format/#document-structure-resource-identification

ahacking:
The type is needed for heterogenous collections generally, and polymorphism is a particular case of that.

Always having a type makes processing uniform for all cases and avoids the need to partition results into typed collections and the associated problems that come with that, primarily the inability to order results. Having a type as a field of a resourve also means where supported, the type can be changed.

ethanresnick:
I know I'm way late on this—I sort of tuned out of JSON-API after @ahacking got me thinking more about annotated, backwards-compatible syntaxes that could be enabled with JSON-LD—but I wanted to offer some feedback.

I think these are very good changes overall and I appreciate all the work that @dgeb and others have done to get us here. In particular, I think going all in on the `(type, id)` requirement makes a lot of sense, given the central role that plays in Ember Data. I also like the overall simplifications and the moving of a bunch of stuff out of the base spec.

But I am really sad to see that this PR still doesn't enable linking in embedded objects (i.e. #238). We got so close to solving that with #311, so this feels like a step backward. The current resolution to #238 (create new, normalized resources) seems to impose SQL-driven logic on JSON-API API's that may be backed by non-SQL datastores. And it just makes a lot of APIs more complicated: by far, the most annoying aspect of the current API I've built on this spec is the extra, normalized resources I've had to create. 

If I think there might be a solution to the embedded linking problem that's relatively simple—sort of an adaptation of the ideas in #311 to the changes in v1.0rc2—should I open a new issue for it? Or is this issue considered settled?

bintoro:
> But I am really sad to see that this PR still doesn't enable linking in embedded objects

Pardon my naivete, but why can't we support a `links` structure within complex attributes (a.k.a. embedded objects) as well?

``` json
{
  "data": {
    "type": "people",
    "name": "Victor Mayonnaise",
    "address": {
      "street": "12 Mean Street",
      "links": {
        "city": {
          "type": "cities",
          "id": "123"
        },
        "state": {
          "type": "states",
          "id": "12"
        },
        "country": {
          "type": "countries",
          "id": "1"
        }
      }
    }
  }
}
```

I've read through the relevant threads and never saw this option even floated, so there must be something I'm missing.

ahacking:
@bintoro those discussions occurred well before the current changes. You should be able to search the issues (which are prossibly closed now).

@ethanresnick yes I would like to see specific linking out of rich/complex/embedded fields. In my use cases those rich/complex attributes are not standalone resources and are not addressable at an endpoint, they exist purely as structured fields within a resource. Thanks for opening #369 !

@bintoro I think the links bag applies at a resource object level, not at an arbitrary field level, but I may be wrong.

bintoro:
@ahacking Looking at #311 again, it appears that what I suggested above was indeed the standard practice sometime in the days of yore, pre-RC1 or so, but was later disallowed. Then #311 came along and proposed to move all links to the top level. Great. However, when that was backtracked on, `links` was rehabilitated on the main resource object only, not on complex values. But I'm unable to figure out why.

There's nothing that suggests `links` must be inherent to the root of a resource object. I see `links` simply as a means of packaging the representation of relationships, no matter where they appear.

Unless someone can point out why `links` within complex attributes is a problem, I believe they should be explicitly allowed. Having to normalize nested objects for API purposes is far from a trivial transformation.

I also considered maintaining just one `links` object per resource and storing everything there:

``` json
{
  "data": {
    "type": "people",
    "name": "Victor Mayonnaise",
    "address": {
      "street": "12 Mean Street",
    },
    "links": {
      "address.city": {
        "type": "cities",
        "id": "123"
      },
      "address.state": {
        "type": "states",
        "id": "12"
      },
      "address.country": {
        "type": "countries",
        "id": "1"
      }
    }
  }
}
```

But this is no good because then `address` is not self-contained. Ideally, you should be able to `GET /people/1/address` and receive a full representation, not just the street. (EDIT: Illustration only! Not saying servers must expose such partial resources.)

ethanresnick:
@bintoro After thinking about this some more, I really like your proposal [to nest `"links"` keys within complex attributes](https://github.com/json-api/json-api/pull/341#issuecomment-75886668). I originally wanted to have only one `"links"` key, more inline with the proposal you made directly above, but I agree that it's awkward to have the complex attribute split up, like `"address"` is in your example. (Still, if there are strong objections to parsers having to recursively search for `"links"` keys, splitting up the complex attribute wouldn't be the end of the world.)

I also considered a couple other solutions that tried to get the best of the above (i.e. a single `"links"` key and not splitting up the complex attribute), but both of them caused more trouble than they were worth.

I'm going to open up a separate issue for this, since I haven't heard otherwise from @dgeb and it doesn't make sense to keep discussing it in this PR.

dgeb:
> I'm going to open up a separate issue for this, since I haven't heard otherwise from @dgeb and it doesn't make sense to keep discussing it in this PR.

I agree this deserves its own issue for discussion. I begin to lose track of issues once they're closed. Thanks @ethanresnick.

PR #353, opened 2015-02-19
Introduce `supported-ext` media type parameter.

dgeb:
The `supported-ext` media type parameter is used by servers to indicate support for one or more media type extensions. The actual media type extension used in a particular document is still specified with the `ext` media type parameter.

This allows servers to clearly indicate support for multiple media type extensions AND specify the particular extension used to format a document.

### DIFF
--- a/extensions/bulk/index.md
+++ b/extensions/bulk/index.md
@@ -10,7 +10,7 @@ extension](/extensions/#official-extensions) of the JSON API specification.
 It provides support for performing multiple operations in a request,
 including adding and removing multiple resources.
 
-Servers **SHOULD** indicate support for the Bulk extension by including the
+Servers **MUST** indicate support for the Bulk extension by including the
 the media type query parameter `supported-ext=bulk` in the `Content-Type`
 header of every response. For example: `Content-Type:
 application/vnd.api+json; supported-ext=bulk`.
--- a/extensions/bulk/index.md
+++ b/extensions/bulk/index.md
@@ -10,9 +10,15 @@ extension](/extensions/#official-extensions) of the JSON API specification.
 It provides support for performing multiple operations in a request,
 including adding and removing multiple resources.
 
-Servers **SHOULD** indicate support for the JSON API media type's Bulk
-extension by including the header `Content-Type: application/vnd.api+json;
-ext=bulk` in every response.
+Servers **SHOULD** indicate support for the Bulk extension by including the
+the media type query parameter `supported-ext=bulk` in the `Content-Type`
+header of every response. For example: `Content-Type:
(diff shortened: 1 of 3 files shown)


wecc:
This looks great! :+1:

wycats:
@dgeb What is the point of making `supported-ext` a **SHOULD** instead of a **MUST**.

dgeb:
I was keeping it consistent with rc2, which currently reads:

```
Servers that support one or more extensions to JSON API SHOULD return 
those extensions in every response in the ext media type parameter of the 
Content-Type header. The value of the ext parameter MUST be a 
comma-separated (U+002C COMMA, ",") list of extension names.
```

I don't remember why we went with SHOULD there tbh - MUST seems more useful. Shall I change it in this PR?

wycats:
Yeah, let's do **MUST**.

dgeb:
@wycats Works for me. Updated.

bintoro:
This is a very good idea.

I'm just wondering if it would be better to omit this stuff from the extension docs and link to the appropriate section of the main spec.

Reasons:
-   Repeating this information gives the impression there's some important extra detail, when in actuality there's not.
-   Stuff may have to be reworded anyway, since the main spec talks about a comma-separated list of extensions, while the extension docs strictly speaking require a `supported-ext=<extension>`, no list. 

The same goes for the subsequent paragraph (not edited in this PR):

> Clients **MAY** request the JSON API media type's _NNN_ extension by specifying the header `Accept: application/vnd.api+json; ext=<nnn>`. Servers that do not support the _NNN_ extension **MUST** return a `415 Unsupported Media Type` status code.

This is also covered in the main spec. More to the point, if a server doesn't implement an extension, it's under no obligation to follow a clause laid out by that extension, so the last MUST doesn't really apply to anyone. So unless this info is provided purely for client implementors' convenience, it's kinda superfluous.

extensions/bulk/index.md
```diff
@@ -10,9 +10,15 @@ extension](/extensions/#official-extensions) of the JSON API specification.
 It provides support for performing multiple operations in a request,
 including adding and removing multiple resources.
 
-Servers **SHOULD** indicate support for the JSON API media type's Bulk
-extension by including the header `Content-Type: application/vnd.api+json;
-ext=bulk` in every response.
+Servers **MUST** indicate support for the Bulk extension by including the
```
steveklabnik, line 13: I would prefer wording like "If a server supports the Bulk extension, it MUST..."

steveklabnik:
@bintoro I agree with you, actually. @dgeb , let's do this.

dgeb:
Closed via #402, which follows the suggestions above from @bintoro and @steveklabnik.

PR #356, opened 2015-02-20
describe response code for bad or missing accept header

bobholt:
Handy to have referenced in the spec.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -398,7 +398,8 @@ A resource, or collection of resources, can be fetched by sending a `GET`
 request to an endpoint.
 
 JSON API requests **MUST** include an `Accept` header specifying the JSON
-API media type.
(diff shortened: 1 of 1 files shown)


dgeb:
Thanks for this PR. I might tweak it just a bit to reflect that a single endpoint may provide alternative content types that are not JSON API (which of course would be beyond the scope of this spec). 

So instead of: 

> Servers **MUST** return a `406 Not Acceptable` status code if this header is missing or incorrect.

Perhaps:

> Servers **MUST** return a `406 Not Acceptable` status code if this header is missing or specifies an unsupported media type.

bobholt:
Sounds good to me. I went ahead and updated the PR, but feel free to tweak away.

dgeb:
Thanks @bobholt - looks good to me.

Issue #312, opened 2014-11-11
Rename "data" to "primary"


ethanresnick:
Right now, top-level resources can be keyed under `"data"` when they identify their type within the resource. I think we should rename this key to `"primary"` as that more clearly expresses the relationship between those resources and any resources that may be in the `"linked"` key

steveklabnik:
More generic words are better, because they're less likey to be reserved for some purpose.

lgebhardt:
In this case only reserved keys are allowed at the top level, so a change to `primary` should be doable. However I'm not in favor of the term `primary`. I'd rather see something that is more specific like `primary_resources`. Then there are `linked_resources`. However I think it's fine to shorten those to `resources` and `linked`.

dgeb:
The thinking with `data` was that it could sound natural whether used with a single object or an array.

dgeb:
My above comment is just explaining the rationale behind the choice of the generic `data`. I agree that it doesn't exactly scream "primary resources". Let's leave this open to mull over and decide firmly before 1.0.

bintoro:
Another vote for `primary`! While I tend to use actual resource names in my own implementation, `primary`/`related` makes a ton of sense.

SphtKr:
Giving some thought to this...and some thesaurus searching. What about: `principal` and `related`? Agree that `data` is maybe a little over generic. `principal` pairs very well with `related`, but also stands well on its own and via-a-vis `meta` and `links` in a non-compound document.

dgeb:
Just a reminder that we're trying to avoid multi-word reserved keys so that the spec can remain agnostic about word delimiters (see #341).

miguelsan:
@SphtKr thanks for pointing me out this issue, I was already starting to feel uncomfortable discussing this at #341.

@everybody
We currently have two kinds of resources on the top level: a main content chunk plus the rest of related content wrapped together. In the specification the main content is called _the document's "primary data"_ and the related content is called _a list of resource objects that are linked to the primary data_. We probably agree in that we need not to use those terms in the keys, but they should be a guide for the final decision. The keys should be one-worded, not mistakable for other ones and meaningful by themselves.

My point is that the requirement of distinctness bans `linked` as a candidate, since it is too close to the everywhere present `links` keyword. `resource` (or `resources`) would also fall out, since both `primary_resources` and `linked_resources` could be meant under it, which does not give enough significance to this keyword candidate. The same could be applied to `data`, since both resources include data in the same sense.

[...]

PR #357, opened 2015-02-20
Clarify responses for fetching resources and relationships.

### LINKED ISSUE #348: Need to specify fetching of relationships
I believe relationship URLs MUST support GETs. We need to clarify responses for to-one and to-many relationships.

For example, a to-one relationship response:

```
GET /articles/1/links/author

{
  "links": {
    "self": "/articles/1/links/author",
    "resource": "/articles/1/author"
  },
  "data": { 
    "type": "people", "id": "12" 
  }
}
```

An empty to-one relationship response:

```
GET /articles/1/links/author

{
  "links": {
    "self": "/articles/1/links/author",
    "resource": "/articles/1/author"
  },
  "data": null
}
```

A homogenous to-many relationship:

```
GET /articles/1/links/tags

{
  "links": {
    "self": "/articles/1/links/tags",
    "resource": "/articles/1/tags"
  },
  "data": { 
    "type": "tags", 
    "ids": ["2", "3"] 
  }
}
```

A heterogenous to-many relationship:

```
GET /articles/1/links/tags

{
  "links": {
    "self": "/articles/1/links/tags",
    "resource": "/articles/1/tags"
  },
  "data": [{ 
    "type": "system-tags", 
    "id": "2" 
  }, {
    "type": "user-tags", 
    "id": "5" 
  }]
}
```

An empty to-many relationship:

```
GET /articles/1/links/tags

{
  "links": {
    "self": "/articles/1/links/tags",
    "resource": "/articles/1/tags"
  },
  "data": []
}
```

dgeb:
Introduces a new top-level section to the document, "Fetching Data",
which includes subsections for both "Fetching Resources" and 
"Fetching Relationships". Possible response codes and example documents
are now specified.

[Fixes #348]

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -415,10 +415,10 @@ The top-level links object **MAY** contain the following members:
 * `"self"` - a link for fetching the data in the response document.
 * Pagination links for the primary data, as described below.
 
-## Fetching Resources <a href="#fetching" id="fetching" class="headerlink"></a>
+## Fetching Data <a href="#fetching" id="fetching" class="headerlink"></a>
 
-A resource, or collection of resources, can be fetched by sending a `GET`
-request to an endpoint.
+Data, including resources and relationships, can be fetched by sending a
+`GET` request to an endpoint.
 
 JSON API requests **MUST** include an `Accept` header specifying the JSON
 API media type. Servers **MUST** return a `406 Not Acceptable` status code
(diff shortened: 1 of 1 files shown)


nwjsmith:
I like that this susses out the concept of a **missing** relationship (returns a `404`) vs an **empty** relationship (which returns `[]`/`null` in the data member). :heart:

bintoro:
I wholeheartedly support bringing nested resource URLs to the base spec. :+1:

However, I'm not comfortable with mangling the representation of relationships when they're accessed directly.

If `GET /articles/1` returns

``` json
{
  "data": {
    "type": "articles",
    "id": "1",
    "links": {
      "comments": {
        "self": "http://example.com/articles/1/links/comments",
        "resource": "http://example.com/articles/1/comments",
        "type": "comments",
        "ids": ["5", "12"]
      }
    }
  }
}
```

then logic dictates that `GET /articles/1/links` should return

``` json
{
  "data": {
    "comments": {
      "self": "http://example.com/articles/1/links/comments",
      "resource": "http://example.com/articles/1/comments",
      "type": "comments",
      "ids": ["5", "12"]
    }
  }
}
```

and `GET /articles/1/links/comments` should return

``` json
{
  "data": {
    "self": "http://example.com/articles/1/links/comments",
    "resource": "http://example.com/articles/1/comments",
    "type": "comments",
    "ids": ["5", "12"]
  }
}
```

instead of

``` json
{
  "links": {
    "self": "http://example.com/articles/1/links/comments",
    "resource": "http://example.com/articles/1/comments"
  },
  "data": {
    "type": "comments",
    "ids": ["5", "12"]
  }
}
```

To me it would make the most sense to treat relationships as (sub)resources in their own right. Why do we have a subtly different syntax for this one case?

Treating relationships as a particular kind of resource would also simplify the language and obviate the need to refer to "resources and relationships" collectively as "data".

I already raised this point in [#312](https://github.com/json-api/json-api/issues/312#issuecomment-75550923) and was going to open a dedicated issue for clarifying the meaning of _resource_, but I suppose this PR is relevant enough.

miguelsan:
Actually, relationships are already been treated as resources, since they got their own URL. Bintoro is right in what he suggests here.

dgeb:
> I wholeheartedly support bringing nested resource URLs to the base spec.

They're actually in [recommendations](http://jsonapi.org/recommendations/#urls), not the base spec, but I'm glad you like the recommendations :)

> Why do we have a subtly different syntax for this one case?

I initially had the same thought as you to completely reproduce the structure of the link object as it appears embedded in a resource's links.

To do so for heterogenous relationships, you'd end up with this mess (note the nesting of `data` within `data`):

```
{
  "data": {
    "self": "http://example.com/articles/1/links/comments",
    "resource": "http://example.com/articles/1/comments",
    "data": [{
      "type": "user-comments",
      "id": "5"
    }, {
      "type": "system-comments",
      "id": "5"
    }]
  }
}
```

I concluded that it is much more appropriate to consider the _linkage data_ itself to be the primary data. And the top-level links then relate to the primary data.

This allows for the much more sane representation of the above linkage:

```
{
  "links": {
    "self": "http://example.com/articles/1/links/comments",
    "resource": "http://example.com/articles/1/comments"
  }
  "data": [{
    "type": "user-comments",
    "id": "5"
  }, {
    "type": "system-comments",
    "id": "5"
  }]
}
```

Note that linkage data can easily be paginated with this structure.

bintoro:
> They're actually in recommendations, not the base spec

I was referring to [this segment](https://github.com/json-api/json-api/pull/357/files#diff-26e38355b69056b0db1c2b1c612241b2R424) introduced in the PR.

> To do so for heterogenous relationships, you'd end up with this mess

Aha, I didn't realize polymorphic links ended up with that kind of format. But it doesn't change my view because 
-   it's not _that_ messy
-   you still have to deal with the same "mess" when fetching the full resource where the inner `data` is embedded within `links.comments`.

I would rather tweak the representation of heterogeneous relationships than special-case the top-level structure when requesting `/foo/links/bar`. If the ugliness is in the two `data` members, then that's an easy fix.

Sure, it looks pretty when you completely replace the redundant container with a list of objects, but it's not entirely logical and leads to two different ways of parsing the same thing:
-   inside an embedded `links.foo` object: look for a member called `data`
-   when requesting `/links/foo`: check the type of the value itself (object vs. array).

bintoro:
I've kinda backtracked on the idea of calling everything a resource because then eventually another term would be needed for the specific things that resources are today.

The idea had its roots in the fact that everything is represented by a _resource object_. But looking at this PR more closely, everything is now collectively called _data_. Well, was the term **data object** considered?

Each of the following would then be a specialized kind of data object:
- resource object
- links object
- link object

This would be easier to grasp than "everything is data and is packaged in resource objects, but not all that is data is a resource".

Any kind of data object could be delivered as primary data. (And I still maintain that the representation should be independent of the HTTP endpoint. Let's have a clear definition of _link object_ and then refer to that instead of basically defining the same thing in different terms under _Fetching Relationships > Responses_.)

**EDIT:** It turns out there are already three occurrences of "data object" in random places in the spec.

ethanresnick:
@dgeb Because we have a bit of hypermedia mixed in, it seems like there could be a case in which the server doesn't know the linkage data, and so can't respond with (or set) that. For example, what if my API exposes a resource that links to a JSON-API resource on another server. The only data in that link object to get/set, then, is the `"resource"` url.

dgeb:
Reviewed with @wycats

format/index.md
```diff
@@ -430,6 +430,279 @@ via a web browser.
...
+A server **MUST** respond to a successful request to fetch a relationship
+with a `200 OK` response.
+
+The primary data in the response document **MUST** be one of the following:
+
+* `null` for empty to-one relationships.
+* an object containing `type` and `id` members for non-empty to-one
+  relationships.
+* an empty array (`[]`) for empty to-many relationships
+* an object containing `type` and `ids` members for non-empty homogenous
```
ethanresnick, line 596: @dgeb This key was renamed to `id`!

ethanresnick:
@dgeb But wait: what happens in the case I described, in which the server doesn't know the linked entities type/id, but instead only has a `resource` url that responds with the linked entities (e.g. because those entities are controlled by another site)? Should the server's response include `data: null` even if there may in fact be entities at the `resource` url? Or is this PR really saying that linked resources have to be such that the server can provide their type(s) and id(s), if they exist? If that's what the PR's implying, then I think it should state that explicitly.

PR #359, opened 2015-02-20
change URL for jsonapi.net

csantero:


dgeb:
@csantero - thanks for the update!

PR #360, opened 2015-02-20
explicitly state type member is required in update

bobholt:
- Note explaining type is always required was only under 'Creating Resources'
  - Move note to general CRUD section introduction
  - Remove blockquote formatting, so it is more visible
- type member was explicitly required in 'Creating Resources,' not 'Updating Resources'
  - Explicitly called it out in 'Updating Resources'
  - Did _not_ call it out in any relationship updating sections

Feel free to tweak this PR. Mainly a convenient way of pointing out the issue locations.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -565,19 +565,19 @@ include media type extensions relevant to the request.
 A request **MUST** completely succeed or fail (in a single "transaction"). No
 partial updates are allowed.
 
+The `type` member is required in every resource object throughout requests and
+responses in JSON API. There are some cases, such as when `POST`ing to an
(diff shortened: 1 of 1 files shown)


dgeb:
I believe a good portion of this needs to remain a non-normative note, rather than official language of the spec. But thanks for pointing out that it might get overlooked in its current location. I'll review with @wycats, who wrote this particular note.

dgeb:
Merged via c09a00c17253d332173eb439d657fef97f14f8df and 2bc4a9c0687a7b0d4cd9fc0de76e68c7eacb9960 which returns this to being a non-normative note.

Issue #358, opened 2015-02-20
provide accordance for viewing api in a browser


tkellen:
Currently the spec states 

> JSON API requests MUST include an Accept header specifying the JSON API media type.

This makes it difficult to browse an api in a browser.  Any thoughts on augmenting the spec to support this use case?

PR #364, opened 2015-02-22
provide accordance for viewing api in a browser

tkellen:
I'm open to any derivation of this clause that allows a user to view json-api documents in a standard browser without modification. This addresses issue #358.


dgeb:
Seems good as a non-normative note - thanks @tkellen!

PR #366, opened 2015-02-22
Clarify media types for JSON Patch messages

bintoro:
Only the example requests made it clear that from now on `application/vnd.api+json` is to be used instead of the JSON Patch IANA media type. This is now explicitly stated. Also, there was one occurrence of `application/json` left in the _Responses_ section.


dgeb:
Thanks for your careful review @bintoro!

PR #367, opened 2015-02-23
add SHOULD to filtering section

tkellen:


dgeb:
Thanks @tkellen. I suppose a similar recommendation should be added above in the nearly identical pagination section.

> The page query parameter is reserved for servers to use to specify pagination.

tkellen:
Sure thing @dgeb.  You're absolutely right, PR incoming.

PR #371, opened 2015-02-25
Prefer heterogeneous/homogeneous to the variants without the e

csantero:
I know I am being excessively pedantic but according to http://grammarist.com/usage/heterogeneous-heterogenous/, the words we want should have the extra "e".


dgeb:
Thanks for this, @csantero! I never realized that there's a distinction, but it's a good one to know.

Issue #373, opened 2015-02-26
Why is (id,type) tuple a unique identifier?


mehaase:
I'm interested in json:api for a RESTful API that I'm starting to work on. I sat down to start digging into the details. I was surprised almost at the outset by this requirement:

> A resource object MUST contain at least the following top-level members:
> - "id"
> - "type"

Later...

> A resource object's type and id pair MUST refer to a single, unique resource.

And...

> A resource object MAY include a URL in its links object, keyed by "self", that identifies the resource represented by the resource object.

Given that a resource's URL already uniquely identifies that resource, what is the design rationale for requiring a unique `(id, type)` tuple instead? Why not make the `self` URL mandatory? Then you would have a unique identifier, and this would allow RESTful APIs for resources that don't necessarily have a distinct ID.

I'll offer an example. I have an authentication API with the URL `/api/authentication`. A client can `POST` a username and password to this resource. If the username and password are valid, the API returns a resource representation that includes an API token for the client to use with future requests. I can dream up a "type" for this resource but I really don't have any good ID for it, because it's a singleton resource.

_Feedback? Am I doing it wrong?_

json:api is the first standard I'm looking at, but at a glance I don't think JSON-LD or HAL require this same (id, type) identifier. The json:api FAQ doesn't explain this design decision and I can't find any other issues related to this design, so apologies if I'm asking an obvious question. Thanks.

bintoro:
This issue seems to pop up repeatedly (recently in #354), and I'm about to submit a PR that clears it up in the spec.

My interpretation is that there's no way you can equate HTTP resources and JSON API resources. The choice of the word "resource" in JSON API is kinda unfortunate, but it may be too late to change that.

In HTTP, a _resource_ refers to the entirety of the thing(s) affected by a request, hence they can be uniquely identified by a URI. JSON API resources, in turn, are the underlying things that HTTP resources provide an interface to. To put it another way, HTTP resources can be represented by JSON API resources.

From Fielding's dissertation:

> A resource is a conceptual mapping to a set of entities, not the entity that corresponds to the mapping at any particular point in time.

The way I see it, JSON API resources are those entities. They are just objects with attributes and relationships.

A URI like `http://example.com/people/1` might be a one-to-one mapping between an HTTP resource and a JSON API resource. But the very same JSON API resource may be exposed via any number of other HTTP resources as well; for instance, at `http://example.com/articles/1/author` or as part of a collection at `http://example.com/people`.

[...]

Issue #354, opened 2015-02-20
Related resources: URLs of `resource` links


tilboerner:
I'd like some clarification on the purpose of `resource` members in links to related resources.

The [relevant section](https://github.com/json-api/json-api/blob/dfaf60ee962a1807aa9563fafa75cef703a854ce/format/index.md#resource-relationships-) of the current spec states:

> If a relationship is provided as a link object, it MUST contain at least one of the following:
> - A self member, whose value is a URL for the relationship itself (a "relationship URL"). This URL allows the client to directly manipulate the relationship. For example, it would allow a client to remove an author from an article without deleting the people resource itself.
> - A resource member, whose value is a related resource URL (as defined above).
> - ...

This seems rather clear to me, but the associated example contains a surprising value for a `resource` URL:

``` json
// ...
{
  "type": "articles",
  "id": "1",
  "title": "Rails is Omakase",
  "links": {
    "self": "http://example.com/articles/1",
    "author": {
      "self": "http://example.com/articles/1/links/author"
      "resource": "http://example.com/articles/1/author",
      "type": "people",
      "id": "9"
    },
    "comments": {
      "resource": "http://example.com/articles/1/comments"
    }
  }
}
// ...
```

`"http://example.com/articles/1/author"` – Why is there an article-specific URL for this linked resource? If I'm understanding the spec correctly, I would expect the resource URL in this particular case to be `http://example.com/people/9`. (This is the URL given in the `linked` member of the example compound document below, which also contains the strange author URL.)

Since the relationship itself is already identified by the `self` member, I don't see what's to be gained by this extra indirection. It seems to merely introduce the ambiguity of making the same resource available at different URLs.

I could see this kind of URL being used for a composition-like relationship, where a resource can't exist independently of its container; however, that's obviously not the case where many articles can have the same author. So, am I missing some important point here, or is this just an unfortunate example?

dgeb:
Relationships are mutable. Related resource URLs (identified by `resource`) are immutable for the life of a resource. Related resource URLs should always return the _current_ resource(s) in a relationship. This concept was missing in the pre-RC2 spec, but I think it's very useful.

Essentially, related resource URLs are valuable beyond the life of the request. Consider that you may not want to immediately fetch this related resource, or you want to know if it's been updated at a later time.

[...]

PR #376, opened 2015-02-26
Add definition of resource & FAQ entry

bintoro:
Currently no definition of _resource_ exists in the spec, yet the very first sentence establishes that it is a key concept.

This PR adds a definition for resource in the intro section along with a FAQ entry detailing the difference from an HTTP resource and explaining why this leads to the ID+type requirement. It also felt appropriate to bring some mention of HTTP into the very beginning.

Addresses: #354, #373

### DIFF
--- a/faq/index.md
+++ b/faq/index.md
@@ -79,3 +79,30 @@ significant. It's necessary to separate primary and related resources by more
 than type because it's possible that a primary resource may have related
 resources of the same type (e.g. the "parents" of a "person"). Nesting related
 resources in `linked` prevents this possible conflict.
+
+### Since HTTP resources are already identified by URIs, why are `ID` and `type` required attributes? <a href="#resources-ids" id="resources-ids" class="headerlink"></a>
+
+*Resource* is a core abstraction in both JSON API and HTTP [[RFC
+7231](https://tools.ietf.org/html/rfc7231)]. The concepts are not equivalent,
+however, and each specification has its own definition for the term.
+
+In HTTP, a resource is simply the service available at an HTTP endpoint. In order
+to interact with it, *representations* are needed. JSON API provides structure
+for those representations.
+
+An HTTP resource can be thought of as an interface (or a mapping) to a set of
+underlying JSON API resources. An HTTP request always targets a single HTTP
+resource but may affect any number of JSON API resources either in full or
+partially.
+
+Importantly, the same JSON API resource may be exposed via several different HTTP
+resources (URIs). An application is free to implement one-to-one correspondence
+between JSON API resources and HTTP resources by always using unique canonical
+URIs. But this is not required, and therefore an ID scheme is necessary.
+
+While JSON API resources often represent persistent objects that have a
+meaningful ID, this need not be the case. For example, a server might provide an
+authentication service (an HTTP resource) that returns instances of
+authentication events (JSON API resources) that do not need an ID. In such a
+situation, it is suggested that a UUID be generated for each resource object
+originating at the server.
--- a/format/index.md
+++ b/format/index.md
@@ -7,8 +7,17 @@ title: "Format"
 
 ## Introduction <a href="#introduction" id="introduction" class="headerlink"></a>
 
-JSON API is a specification for how a client should request that resources be
-fetched or modified, and how a server should respond to those requests.
+JSON API is a specification for how an HTTP client should request that
+resources be fetched or modified, and how an HTTP server should respond to
+those requests.
+
+A *resource* is an entity that has *attributes* and may be associated with other
+resources via *relationships*. JSON API does not impose constraints on what a
+resource may represent. For example, a resource could describe an object in a
+database, an event that has occurred, or the state of a service.
+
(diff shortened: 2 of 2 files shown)


dgeb:
Thanks for this @bintoro. I think this will clear up many misunderstandings. 

:+1:

kjwierenga:
I'm a bit confused by the overloading of the word 'resource'.

So we have the HTTP definition of what a 'resource' is. We have the Fielding definition of what a 'resource' is. Are the HTTP definition and the Fielding definitions the same? I assume they are.

Then we have the JSON API definition of what a resource is and somewhere I read:

> In terms of JSON API, resources are just objects.

As far as I can see the JSON response documents are representations of resources (both terms in Fielding definition). In my opinion we should not use the word 'resource' in the JSON API spec to mean a different thing.

For example:

> Primary data MUST appear under a top-level key named "data". Primary data MUST be either a single resource object, an array of resource objects, or a value representing a resource relationship.

Could be changed to (remove 'resource' before 'object').

> Primary data MUST appear under a top-level key named "data". Primary data MUST be either a single object, an array of objects, or a value representing a resource relationship.

In fact everywhere 'Resource object' could be replaced by just 'object'.

Come to think of it 'object' isn't really a good name, because it's just 'data'; there is no behavior... (The objects in JavaScript Object Notation are not really objects actually).

In this Pull Request I read:

> A [JSON API] _resource_ is an entity that has _attributes_ and may be associated with other resources via _relationships_. JSON API does not impose constraints on what a resource may represent. For example, a resource could describe an object in a database, an event that has occurred, or the state of a service.

Why overload the word 'resource'

One way to make the difference would be to use the word '_entity_'. The second sentence to mention 'resource' in the spec would then read like this

> The document's "primary data" is a representation of the entity, collection of entities, or entity relationship primarily targeted by a request.

Sections would be renamed like this:
- 'Resource Objects' becomes 'Entities'
- 'Resource Attributes' becomes 'Entity Attributes'
- 'Resource Identification' becomes 'Entity Identification'
- 'Resource Types' becomes 'Entity Types'
- 'Resource Relationships' becomes 'Entity Relationships'
  etc.

When you use a different word, then every occurence of the word 'resource' in the current spec raises an important question: 'Is this an entity or is this a [HTTP/Fielding] resource'.

Am I completely off track here? Or would using a different word clarify the spec?

gr0uch:
I'm curious what the initial motivation for using the word "resource" instead of "entity" was.

kjwierenga:
The definition of resource within the JSON API spec is different from the definition in HTTP. This leads to confusion/ambiguity in the spec. By using a different word, be it 'object' or 'entity' we can remove this ambiguity. I chose 'entity' because I think it fits the purpose. A JSON fragment like the following isn't an object (it has no behavior. In fact it's just data.

```
{
    "id":"1",
    "type":"people",
    "name":"Steve"
}
```

Maybe there is a better word?

bintoro:
@kjwierenga, your interpretation is correct. The word "object" is currently used to refer to JSON objects, but "entity" would be workable.

I don't want to get too deep into the debate whether the terminology should be changed, as it doesn't bother me that much, but I would certainly support such a change.

I don't find the word "resource" particularly important. It could be "entity" or whatever. The success of a standard hinges mostly on its concepts, not the choice of an individual word, and nobody's proposing to change the _concept_ of what is today called "resource".

kjwierenga:
Upon scanning the spec for the word 'resource' it actually isn't a big problem because most if not all uses are in fact the JSON API definition of 'resource'. The HTTP definition of 'resource' is avoided by using 'endpoint' instead or by the phrase 'the URL that represents the resource'.

Sometimes [JSON API] 'resource' and 'resource object' are used in the same sentence. What is the difference between [JSON API] 'resource' and 'resource object' if any?

I guess some problems of definition come up in the 'Responses' section. I think it uses 'resource' in the HTTP definition.

I don't think it's a big deal although using a different name would remove all ambiguity.

dgeb:
> What is the difference between [JSON API] 'resource' and 'resource object' if any?

A _resource object_ is the JSON representation of a _resource_ (which is a more abstract term).

---

I'm not crazy about the alternative "entity". It doesn't seem to me that an `entity` member in a link object would be as clear as `resource`. But then again, it's hard for me to look at this with fresh eyes.

We should at least clarify terminology as @bintoro is attempting with this PR.

/cc @steveklabnik @wycats

bintoro:
@kjwierenga The spec never uses "resource" in the HTTP meaning, or if it does, it's a bug. A _resource object_ is the JSON representation of a resource.

tkellen:
@bintoro would you be willing to rebase and ping when ready?

bf4:
@bintoro @tkellen This seems to have become forgotten :( also ref https://github.com/json-api/json-api/issues/966

kjwierenga:
This can be closed?

jelhan (approved):
@dgeb I think we want to revisit this one. I think this clarification if the term _resource_ used in the spec would help a lot to prevent confusion.

Issue #369, opened 2015-02-24
Allow JSON-LD


ethanresnick:
Currently, the spec says "The top level of a document **MUST NOT** contain any additional members." I'd suggest an exception to allow `@context` and any other members that authors would need to support JSON-LD. 

[More radical](https://github.com/json-api/json-api/issues/159#issuecomment-70868301) proposals to bake JSON-LD ideas into the spec may not have gained support, but there's no reason the spec should _explicitly prevent_ authors from adding JSON-LD annotations should they desire. Hopefully that's relatively uncontroversial!

bintoro:
I've wondered about this as well. Not so much w.r.t. JSON-LD, but more generally, a custom extension might want to place some stuff on the top level, and I think it should be allowed. Or, at least, let's reserve an extra free-for-all member.

gr0uch:
+1 to allowing custom properties on the top level.

dgeb:
> The top level of a document MUST NOT contain any additional members.

The reasoning behind this is to future proof the spec. After 1.0, we can't break backward compatibility and will make only additive changes. If the spec allowed any other members at the top level, then JSON API would be prohibited from making additive changes to the top level.

With that said, it's probably very safe to include a member such as `@context` at the top level, but the onus is on you to maintain compatibility with JSON API if for some crazy reason the same member was reserved in the spec at the top level.

> a custom extension might want to place some stuff on the top level, and I think it should be allowed.

Extensions can make changes to the base spec as needed - they are not purely additive to the base spec. Changes will of course only affect clients and servers that negotiate for the extensions.

bintoro:
> Extensions can make changes to the base spec as needed - they are not purely additive to the base spec.

Gotcha. It's been kinda unclear what extensions may and may not do. Up until now, `Content-Type` negotiation has not been mandatory, implying that extensions mustn't break compatibility with the base spec. This is partly resolved with PR #353. If we add to this a clause stating that servers must only provide extended/incompatible services to clients who explicitly indicate support for them, then it becomes safe for extensions to do whatever they want. (This, in turn, implies that clients MUST specify an extension in `Content-Type` or `Accept` if they want to use its features.)

ethanresnick:
@dgeb Re...

[...]

PR #378, opened 2015-02-28
Allow extra properties

ethanresnick:
A PR for #369. 

@dgeb For now, the text in this PR just describes the rule; it doesn't include any of the reasoning behind the rule. That seems consistent with the rest of the spec, but I'm happy to add a sentence about the reasoning if you want.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -95,7 +95,7 @@ If any of these members appears in the top-level of a response, their values
 
 The top level of a document **MUST NOT** contain any additional members whose
 names start with alphanumeric characters. Additional members whose names start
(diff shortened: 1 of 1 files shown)


ethanresnick:
By the way, right now "link objects" don't seem to reserve any keys. It seems like, for extensibility, we might want to apply the same policy to them as to the top-level of the document. I can add that to the PR if desired.

dgeb:
Merged via 4e5383f19c832ea4c7e98e636cef6d08e743498d

ethanresnick:
Awesome! Thanks @dgeb

PR #385, opened 2015-03-02
clarify "A JSON object **MUST** be at the root of every JSON API response."

bobholt:
The [statement here](http://jsonapi.org/format/#document-structure-top-level) conflicts with the inability to send a body with a `204 No Content` response. I interpret a `204` as still falling under the category of a "JSON API response," though it is one without a document.

This is a stab at rewording, but feel free to tweak or clarify in your own way.

Also from [RFC 7231](http://tools.ietf.org/html/rfc7231#section-6.3.5): "A 204 response is terminated by the first empty line after the header fields because it cannot contain a message body."


steveklabnik:
I agree that this is nice to call out. Thank you!

PR #392, opened 2015-03-04
unify id and ids

tkellen:
This distinction is unnecessary. Checking if `id` is a string, an
array or null is trivial for clients.

### DIFF
--- a/extensions/bulk/index.md
+++ b/extensions/bulk/index.md
@@ -89,7 +89,7 @@ Multiple resources can be deleted by sending a `DELETE` request to a URL that
 represents a collection of resources to which they all belong.
 
 The body of the request **MUST** contain a `data` member whose value is an
-object that contains `type` and `ids`, or an array of objects that each
+object that contains `type` and `id`, or an array of objects that each
 contain a `type` and `id`.
 
 For example:
@@ -100,6 +100,6 @@ Content-Type: application/vnd.api+json; ext=bulk
 Accept: application/vnd.api+json; ext=bulk
 
 {
-  "data": { "type": "articles", "ids": ["1", "2"] }
+  "data": { "type": "articles", "id": ["1", "2"] }
 }
 ```
--- a/extensions/patch/index.md
+++ b/extensions/patch/index.md
@@ -138,7 +138,7 @@ A server **MUST** respond to Patch operations that target a *to-many
 relationship URL* as described below.
 
 For all operations, the `"value"` **MUST** contain an object that contains
-`type` and `ids` members, or an array of objects that each contain `type`
+`type` and `id` members, or an array of objects that each contain `type`
 and `id` members.
 
 If a client requests a `"replace"` operation to a *to-many relationship URL*, the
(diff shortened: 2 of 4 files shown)


steveklabnik:
I am 👍 on this. Seems like a very minor, odd case.

bintoro:
Back in the day people hoped that the keys could be unified, but for unknown reasons the idea didn't make its way into the spec.

Just a couple of days ago, @ethanresnick brought this up again in #312. It's not just about `id` and `ids` but the `data` member too (for heterogeneous relationships). I'm also strongly in favor of unifying all three. Not only is it way more elegant, it's simpler from a parsing point of view, too.

I've investigated the old threads, and it appears there was disagreement on singular vs. plural for the key, and that may even be why it all fell apart. Based on this observation, I suggested in #312 the following:

> [...] let's simply define a _reference_ as one of:
> - an ID,
> - an array of IDs, or
> - an array of (ID, type) tuples.
> 
> Clearly, a **reference** could be labeled `ref`. No semantic controversy there.

I don't insist on any specific key, but I would like there to be just one. If it's an array, you know it's to-many, and if the array elements are objects, you know it's heterogeneous (the same can also be inferred from the absence of a shared `type` member). Separate keys are totally redundant.

ethanresnick:
Yeah, this discussion is happening in #312 too. 

I think we should use a new word rather than just chopping off the `s`, since `id` almost always denotes a singular id outside of this spec. I'm not a huge fan of `ref` either, because it seems more cryptic than the other keys in the spec—the only other key that's an abbreviation is `id`, and that's an abbreviation that's so familiar to most programmers as to not even sound like an abbreviation anymore.

What do you guys think of `"target"`, which, as I proposed in #312, could be used to replace the `"data"` member too. So the three cases of link objects would look like this.

```
"links": {
      "employer": {
        "type": "companies",
        "target": "4"
      },
      "skills": {
        "type": "skills",
        "target": ["1", "2", "3"]
      },
      "favorites": {
        "target": [
          {"type": "posts", "id": "4"}, 
          {"type": "people", "id": "2"}
        ]
      }
}
```

bintoro:
The reason I like "reference" is that it's perhaps the only word whose actual meaning fits the use case extremely well. It should also be acceptable to those who are concerned about grammatical number.

If the spec wants to avoid abbreviations, then I can appreciate that. But it didn't even occur to me that `ref` would be cryptic to anyone using JSON API.

`target` is definitely a possibility. It's just that, in typical usage, the target of a relationship is the whole entity being targeted, generally not just the ID. Does this matter? I could live with it. And `target` works very well in the heterogeneous case.

One more option that occurs to me is `key`. It's succinct without being an abbreviation, makes sense semantically, is instantly understandable, and might be justifiable to the singular/plural crowd as well:

```
"links": {
      "employer": {
        "type": "companies",
        "key": "4"
      },
      "skills": {
        "type": "skills",
        "key": ["1", "2", "3"]
      },
      "favorites": {
        "key": [
          {"type": "posts", "id": "4"}, 
          {"type": "people", "id": "2"}
        ]
      }
}
```

**EDIT:** I guess the thing that bugs me about `target` is that it's not conceptually parallel to `id`. The link object itself represents the target, while an ID is just one way of referring to it. The member whose meaning is closest to "target" would probably be the `resource` URL.

**EDIT 2:** Let's also consider what an empty to-many would look like:

`"target": []` — Good ("no targets")
`"ref": []` — Good ("no references")
`"id": []` — Not that good
`"key": []` — Seems the worst of all

bintoro:
@steveklabnik, does the merging of this PR indicate that you would be opposed to fusing the third overlapping member (`data`) into the `id/ids` combo, or that it should be discussed in a new issue/PR?

Issue #249, opened 2014-07-31
Ambiguous whether link points to a collection or single resource


ethanresnick:
In cases where referenced resources are fetched from an external url (i.e. we're not dealing with a compound document), and the uri template in the `links` keys is expanded using the primary resources' ids—rather than the id(s) of the referenced resources—it's ambiguous whether the document returned at that url will contain a single resource or a collection of resources.

For example, in the following response for `GET /mytype`, it's unclear whether each `mytype.prop` would ultimately hold a single resource or multiple:

```
{
  links: {
    "mytype.prop": {
      "href": "http://.../{mytype.id}",
      "type": "type2"
    }
  }
  mytype: [{
    "id": "3",
    "attrName": "val"
  }, 
  //...
  ]
} 
```

The same thing occurs if you use a resource-level `links` key instead of a top level one. Here's a response for `GET /mytype/3`

```
{
  mytype: {
    "id": "3",
    "attrName": "val",
    "links": {
      "prop": {
        "href": "http://.../3",
        "type": "type2"
      }
    }
  }
} 
```

My question is: is this a problem? (Seems like it might complicate client-side code a bit). And, if it is, what to do about it?

hesalx:
Late answer, but I think the question is reasonable.

**tl;dr** What would your client do is you make such request alone? Where to know if it's a collection or a single resource?

From the REST perspective either collection or single resource is just "a resource". A good client should not relies on this difference. (A related resource may not be a JSON API resource at all. A client should determine it according to the media type of response.)

In some sense REST tells you to know what the resource actually is at the moment of the response, not at the moment of the request. Otherwise it means that one request depends of another (earlier one). REST stands for the opposite behaviour.

JSON API provides the unified format for collections and single resources. It's either the `"data"` object or the object named after the `"type"`. And the type of this object is either array or object. Which determines the resource as a collection or a single resource clearly.

Actually, there is the proposal for schemas #310, which, if accepted, would help to define additional properties of related resources in such special cases, despite it would deviates from REST a bit, if it reasonable urgent for you.

ethanresnick:
Yeah, what I'm talking about here would be addressed well by schemas, and that's the real issue. Because, for some clients, it's useful to know in advance what kind of response you're going to get back (e.g. how Ember Data will mock a collection of models vs a single model before making the request).

steveklabnik:
Sorry for taking so long to respond here :(

> My question is: is this a problem?

I don't think that it is. as @arantir mentions:

[...]

PR #398, opened 2015-03-05
Prescribe primary data format: object for singular, array for collection

bintoro:
Currently, it's not possible for a generic client to look at a response and reliably figure out if the server is exposing a singular resource that can be manipulated via PUT or a collection that can be appended to via POST (short of attempting the operation, of course).

This PR adds a paragraph specifying that the representation (single object vs. array) must reflect the nature of the endpoint.

Specifically, the new clauses preclude
- representing a logically singular resource as an object wrapped in an array
- representing a collection with a length of 1 as an individual object.

A related issue is #249. Some applications might want to provide this information (to-one vs. to-many) directly in a link object. While this PR alleviates the issue somewhat, a client still needs to perform a dummy request for the resource(s) to get the info.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -83,6 +83,11 @@ objects, or a value representing a resource relationship.
 }
 ```
 
+A logical collection of resources (e.g., the target of a to-many relationship)
+**MUST** be represented as an array, even if it only contains one item.
+A logically singular resource (e.g., the target of a to-one relationship)
+**MUST** be represented as a single resource object.
+
 Error objects **MUST** appear under a top-level key named `"errors"`.
 
 A document's top level **MAY** also have the following members:


dgeb:
Concepts such as "logically singular" or "logically a collection" are a meta / schema concern. Given that a relationship MAY be expressed simply with a URL, I don't see how this could be rigorously specified.

bintoro:
The intent here was that a server shouldn't pick the representation based on item count (_n_ = 1 vs. _n_ > 1). While it's indeed not possible to rigorously specify _when_ an endpoint should serve a collection, it's perfectly possible to specify _how_ to serve it.

Besides, JSON API is already in the schema business because it has the concepts of to-one and to-many. :smirk:

dgeb:
Might be ok - sorry for the hasty close @bintoro - will review.

dgeb:
@wycats and I agree - it's fine after all - thanks @bintoro :+1:

bintoro:
😊

Issue #386, opened 2015-03-02
Guidance for links to empty relationships


nwjsmith:
The spec currently gives guidance on how to structure relationships when the exist, but gives no guidance for when they do not.

It is obvious (but not stated in the spec) that when the value of a relationship is a related resource URL, it should be `null` when missing:

**Present**

``` json
// ...
{
  "type": "articles",
  "id": "1",
  "title": "Rails is Omakase",
  "links": {
    "comments": {
      "resource": "http://example.com/articles/1/comments"
    }
  }
}
// ...
```

**Empty**

``` json
// ...
{
  "type": "articles",
  "id": "1",
  "title": "Rails is Omakase",
  "links": {
    "comments": {
      "resource": null
    }
  }
}
// ...
```

Things are less clear when a relationship is provided as a link object because of the `self` link. Even when the related resource is empty, the `self` link is a valid link for modifying the relationship.

**Present**

``` json
// ...
{
  "type": "articles",
  "id": "1",
  "title": "Rails is Omakase",
  "links": {
    "self": "http://example.com/articles/1",
    "author": {
      "self": "http://example.com/articles/1/links/author",
      "resource": "http://example.com/articles/1/author",
      "type": "people",
      "id": "9"
    }
  }
}
// ...
```

What does this look like if the `author` relationship is empty? There are two strategies: omitting keys, or `null` values.

**Missing, omitting keys**

``` json
// ...
{
  "type": "articles",
  "id": "1",
  "title": "Rails is Omakase",
  "links": {
    "self": "http://example.com/articles/1",
    "author": {
      "self": "http://example.com/articles/1/links/author"
    }
  }
}
// ...
```

**Missing, `null` values**

``` json
// ...
{
  "type": "articles",
  "id": "1",
  "title": "Rails is Omakase",
  "links": {
    "self": "http://example.com/articles/1",
    "author": {
      "self": "http://example.com/articles/1/links/author",
      "resource": "http://example.com/articles/1/author",
      "type": null,
      "id": null
    }
  }
}
// ...
```

I propose that the `null` value approach be taken by the spec. @dgeb work on PR https://github.com/json-api/json-api/pull/357 might help to inform where we go on this?

bintoro:
To signify an empty to-one relationship, you're supposed to set `id: null`. If it's a to-many relationship, you say `ids: []`. If you normally use `data`, I guess you could have an empty array in there, too.

[...]

PR #399, opened 2015-03-05
Remove language that associates ID-based linking with compound documents

bintoro:
RC2 introduced the concept of _object linkage_. Link objects of the (ID, type) variety are now explicitly tied to compound documents only.

This ignores another obvious use case for ID-based linking: the construction of canonical URLs.

Given a link object

```
"author": {
  "type": "people",
  "id": "9"
}
```

an application-specific client could determine that the resource is available at `http://example.com/people/9`.

It's pretty clear that applications are going to do this, so it would be wise to at least recognize the possibility, hence this PR.

With link templates, such URL construction would be possible for generic clients as well. Link templates were removed with the advent of heterogeneous collections, but that was throwing the baby out with the bathwater. For _canonical URLs_ they are still perfectly possible, keyed by type instead of association name. This, however, is a discussion to be had in a separate issue, and it can happen post-1.0.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -217,10 +217,8 @@ one of the following:
   relationship. For example, it would allow a client to remove an `author` from
   an `article` without deleting the `people` resource itself.
 * A `resource` member, whose value is a related resource URL (as defined above).
-* Linkage to other resource objects ("object linkage") included in a compound
-  document. This allows a client to link together all of the resource objects
-  included in a compound document without having to `GET` one of the
-  relationship URLs. Linkage **MUST** be expressed as:
+* Linkage to other resources based on their identifying `id` and `type` members
+  ("resource linkage"). Linkage **MUST** be expressed as:
   * `type` and `id` members for to-one relationships. `type` is not required
     if the value of `id` is `null`.
   * `type` and `id` members for homogeneous to-many relationships. `type` is
@@ -234,7 +232,9 @@ A link object that represents a to-many relationship **MAY** also contain
 pagination links, as described below.
 
 If a link object refers to resource objects included in the same compound
-document, it **MUST** include object linkage to those resource objects.
+document, it **MUST** include resource linkage to those resource objects.
+This allows a client to link together all of the included resource objects
+without having to `GET` one of the relationship URLs.
 
 For example, the following article is associated with an `author` and `comments`:
 


dgeb:
I think these belong as non-normative notes.

bintoro:
@dgeb, can you clarify? The PR doesn't _add_ any mention of URL construction; it merely removes the part that suggests IDs are to be used in compound documents only.

Both mentions of the importance of IDs to compound document linkage remain:

In _Resource Identification_:

> This identification is used for linkage in compound documents and batch operations that modify multiple resources at a time.

In _Resource Relationships_:

> If a link object refers to resource objects included in the same compound document, it **MUST** include resource linkage to those resource objects.

The only thing removed here was "included in a compound document" in the definition of _object linkage_ (now _resource linkage_ since the relationship target might not be currently represented as an object).

(Btw, I really don't think JSON API needs to be ashamed of the fact that it defines a relational object model. That's pretty much its _raison d'être_. The _Resource Identification_ quote above completely ignores the primary purpose of IDs: the ability to specify relationships among the resources. For that, you _need_ IDs for the actual entities. It's not just a representational thing.)

format/index.md
```diff
@@ -234,7 +232,9 @@ A link object that represents a to-many relationship **MAY** also contain
 pagination links, as described below.
 
 If a link object refers to resource objects included in the same compound
-document, it **MUST** include object linkage to those resource objects.
+document, it **MUST** include resource linkage to those resource objects.
+This allows a client to link together all of the included resource objects
+without having to `GET` one of the relationship URLs.
```
dgeb, line 237: I consider this last sentence non-normative and think it would be better as a note, perhaps with even further explanation.

dgeb, line 237: I realize it wasn't a note before, and that it was moved, but this brought my attention to it. You don't have to change it as part of this PR.

format/index.md
```diff
@@ -217,10 +217,8 @@ one of the following:
   relationship. For example, it would allow a client to remove an `author` from
   an `article` without deleting the `people` resource itself.
 * A `resource` member, whose value is a related resource URL (as defined above).
-* Linkage to other resource objects ("object linkage") included in a compound
-  document. This allows a client to link together all of the resource objects
-  included in a compound document without having to `GET` one of the
-  relationship URLs. Linkage **MUST** be expressed as:
+* Linkage to other resources based on their identifying `id` and `type` members
+  ("resource linkage"). Linkage **MUST** be expressed as:
```
dgeb, line 221: I am fine with the more explicit "resource linkage".

dgeb:
I'll mark this to review before 1.0 since it has some nuance to examine carefully.

bintoro:
One more observation about this: Even when related resources are fetched using the `resource` URL, it's a nice feature that a server can signal an empty relationship right away in the link object by adding `id: null` or `id: []` in the payload, saving the client a futile request.

This might be worthy of (at least) a Note because it's not obvious to those who weren't around back when IDs had a more central role in linking.

**EDIT:** This is the same thing as the second half of the Note that was part of #425 but wasn't included in #433. Original motivation was issue #386, which made it clear that there's some confusion around this.

dgeb:
> One more observation about this: Even when related resources are fetched using the resource URL, it's a nice feature that a server can signal an empty relationship right away in the link object by adding id: null or id: [] in the payload, saving the client a futile request.

@bintoro Sorry that you've had to shuffle this note around several PRs / issues. I suppose it deserves its own PR for consideration. Since it's non-normative, there's no rush before 1.0.

PR #400, opened 2015-03-05
Correct "collection of error objects" to read "array of..."

bintoro:
As I understand it, a _collection_ is a more abstract thing, while an _array_ is a representational thing. Let's stick to this usage.


dgeb:
Agreed. Thanks @bintoro.

PR #401, opened 2015-03-05
Clarify that `fields` only affects attributes and relationships

ethanresnick:
Previously, we [discussed](https://github.com/json-api/json-api/issues/260) whether a request with the `fields` parameter could still return the basic members of each resource object, like `type` and `id`, even if those weren't included in the `fields` param. Back then, the answer was yes, but changes in RC2 made this illegal. This PR makes the old behavior legal again, and actually mandates it.

The reasoning is that allowing the user to exclude `type` and `id`:
1. conflicts with the current definition of a resource object, which says that "A resource object MUST contain at least the following top-level members: id, type`.
2. accordingly, it creates an edge case for clients, which might otherwise reasonably rely on `type` and `id` always being present;
3. and it can complicate server-side implementations, which have to keep track of these properties anyway (e.g. to handle `include`s) throughout the process of building up the resource objects to respond with.

This PR resolves the issue by more clearly noting the separation between reserved top-level members (`"type"`, `"id"`, `"links"`, and `"meta"`) and attributes (the other top level members) and relationships (the contents of the `"links"` object, excluding `"self"`). Then, it specifies that the `fields[TYPE]` parameter only applies to attributes and relationships.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -114,7 +114,7 @@ Here's how an article (i.e. a resource of type "articles") might appear in a doc
 // ...
 ```
 
-#### Resource Attributes <a href="#document-structure-resource-object-attributes" id="document-structure-resource-object-attributes" class="headerlink"></a>
+#### Resource Objects <a href="#document-structure-resource-objects" id="document-structure-resource-objects" class="headerlink"></a>
 
 A resource object **MUST** contain at least the following top-level members:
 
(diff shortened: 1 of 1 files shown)


bintoro:
Good catch. The actual issue must be addressed, but in my mind `id` and `type` are definitely "attributes". Could we maybe call them "required attributes" and just make it clear that they are always included regardless of the value of `fields`?

Here's how I see the various terms:
- _attributes_ — `id`, `type`, and type-specific resource attributes
- _relationships_
- _fields_ — attributes + relationships
- _members_ — everything, i.e. fields + `links` + `meta`

Also note that you now have a subheading "Resource Objects" following an identically named section heading.

ethanresnick:
@bintoro I just fixed the extra "Resource Objects" headings; that was a mistake. Re your point about terminology: I agree that the current terminology is not ideal, but can we open another issue to discuss that? That way, this PR and discussion can stay small and actually get merged soon.

bintoro:
@ethanresnick You know what, you're completely right here. It's better if `id` and `type` remain "members" and not attributes.

Consider: "A resource's attributes and relationships can be updated by sending a `PUT` request...". But `id` and `type` cannot be updateable.

dgeb:
Merged via b7fe0e06c8a0e8e3fb6929470323f3cb70785c8b and 45e53d556e6c61ee43075857037183befedb057b

Thanks for the discussion and clarification @ethanresnick and @bintoro :+1:

PR #405, opened 2015-03-06
Retire "linked resource" in favor of "related resource"

bintoro:
This is the best outcome. "Link" remains as a representational concept only (_link object_).


dgeb:
Consistency FTW

PR #411, opened 2015-03-06
Fix error object definition

ethanresnick:
Error objects are _not_ "specialized resource objects". That would imply that they have the same base semantics as resource objects (just with some additions) when, in fact, `"links"` means something totally different in error objects, `"id"` and `"type"` don't appear to be required.

This PR just lets them stand as their own type of object, rather than as subtypes of standard resource objects.

@dgeb


dgeb:
Thank you @ethanresnick - good catch!

PR #412, opened 2015-03-06
Reserve keys in link objects using the same strategy as top-level keys

ethanresnick:
See title. This should preserve more extensibility down the line, as the current spec doesn't reserve any link object keys.


dgeb:
_reviewing_

ethanresnick:
Cool. And now, I'm going to bed :)

dgeb:
I will apply a similar statement to top-level links. Thanks for this @ethanresnick.

Issue #383, opened 2015-03-01
Linking from complex attributes


ethanresnick:
## The use case

Consider the following cases:
1. A resource representing a product has a `"price"` attribute, which is an object of the form: `{"amount": ..., "currency": ...}`, where the `"currency"` property refers to a particular `currencies` resource.
2. A resource representing a project has a `"team-members"` attribute, which is an array of objects of the form: `{"person": ...., "role-played": ...}`, where the `"person"` and `"role-played"` properties refer to particular `people` and `roles` resources.

In both of the cases above, the attribute mentioned is an object or an array of objects that includes a link to another resource. But, crucially, the values for these attributes are only likely to occur in, or are only relevant in the context of, one resource. That is the (amount, currency) pair that represents a product's price doesn't make sense outside the context of that product, and the (person, role) pair that describes the role of a person on a particular project likely does not need to be considered outside of that project.

Accordingly, there's little semantic or functional justification for making separate, addressable resources from these  "complex attributes"'s values. Meanwhile, turning them into separate resources can have a performance cost for backends implemented with document databases, which allow the complex attributes' values to be stored directly in the main resource's document. Therefore, the JSON-API spec shouldn't mandate that these complex attributes be turned into separate resources. But, if it's not going to mandate that, it has to somehow allow these complex attributes' links to other resources to be expressed in the response.
## Prior Discussions

This issue has been discussed many times before, including in #238, #276, and #311. But many of the proposed solutions no longer apply post rc2. Post rc2, though, @bintoro made [two](https://github.com/json-api/json-api/pull/341#issuecomment-75886668) [proposals](https://github.com/json-api/json-api/pull/341#issuecomment-75944727) that offer good starting points for discussion.
## My _initial_ preferred approach

As I [said previously](https://github.com/json-api/json-api/pull/341#issuecomment-76412541), I think @bintoro's first proposal makes the a lot of sense. It's syntactically consistent with the current approach, and it seems easier to implement server-side than splitting up the complex attribute would be. Its downside, of course, is that it makes a parser's search for `"links"` somewhat more complicated.

An in between option, that keeps it reasonably simple to find links but still keeps a complex attribute's value together, could be the following:

```
{
  “data”: {
    “id”: “1”,
    “type”: “people”,
    “name”: “Victor Mayonnaise”,
    “links”: [“employer”, “address.city”],
    “employer”: {
      “relationship”: “http://example.com/people/1/links/employer”,
      “resource”:  “http://example.com/companies/2”
[...]

Issue #313, opened 2014-11-11
Prefix reserved keys in resources


lgebhardt:
There is a strong possibility of a name conflict within the key space of a resource with the four reserved keys, [`id`, `type`, `href`, `links`]. I'd like to float the idea of prefixing these keys with an underscore (or other character). Thus we would have the reserved keys of [`_id`, `_type`, `_href`, `_links`].

The primary benefit is that it opens up the possibility of using a key with one of the currently reserved key names. I think it is quite likely for some existing datasets to be using the `href` and `type` keys and this will save developers the need to rename these existing keys. A secondary (and minor) benefit is that it's easy to pick out the special keys visually.

The only downside I see is the adding of an extra character to each key, but with compression that is mostly a non issue. It's also possible there could still be conflicts, but its probability is probably a couple of orders of magnitude less.

An example:

``` json
{
  "posts": {
    "id": "1",
    "title": "Rails is Omakase",
    "post_href": "/rails_is_omakase"
  }
}
```

would become:

``` json
{
  "posts": {
    "_id": "1",
    "title": "Rails is Omakase",
    "href": "/rails_is_omakase"
  }
}
```

or 

``` json
{
  "posts": {
    "_id": "1",
    "_type": "posts",
    "_href": "http://example.com/posts/1",
    "title": "Rails is Omakase",
    "href": "/rails_is_omakase",
    "_links": {
      "author": "9",
      "comments": [ "4", "5" ]
     }
  }
}
```

Thoughts?

steveklabnik:
I am very, very strongly against adding an underscore prefix. It's quite
ugly.

lgebhardt:
Fair enough. I find it a bit ugly too, but I worry the namespace collisions may prevent some existing apps from converting to JSON API compliant tools. Is there a more appealing way to mitigate the namespace collisions?

steveklabnik:
FWIW, Hal is also considering removing their underscore prefixes.

I don't think that this will actually affect a significant number of
people, as they are already probably using id, so that's not a conflict. If
they're not using hypermedia already, then links and href aren't going to
be used.

The only one I can see is 'type', and we would probably be in line with
whatever that is already.

dgeb:
I also oppose the prefixes. I haven't personally heard of any conversion pains, perhaps because the keys are already pretty natural (as @steveklabnik points out). And the bottom line is that I'd rather not compromise the usability / aesthetics of the spec.

lgebhardt:
[...]

PR #417, opened 2015-03-06
Explain complex attributes and linking from within

### LINKED ISSUE #289: q: nested attrs without links?
Hello. I have a collection in mongoDB with sizes as array and prices as subdocuments of subdocuments

```
{
  products => {
    id: 'myId'
    sizes: ['m','l'],
    prices: {
      min: {uah: 100, usd: 8},
      max: {uah: 200, usd: 16}
    }

  }
}
```

How should I represent it according to the specification? Is json above ok? Could a 'sizes' field be an array? Could a 'prices' field be a nested object without links?

bintoro:
Submitting for last-minute review because this cannot happen post-1.0 in the form proposed here due to the reserved-key issue.

Resolves: #289, #383

The representation of outbound links follows my suggestion [in this comment](https://github.com/json-api/json-api/pull/341#issuecomment-75886668) that was later vetted against alternative approaches in #383.

(cc @ethanresnick)

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -144,8 +144,10 @@ In addition, a resource object **MAY** contain any of these top-level members:
 Any other top-level member in a resource object represents an "attribute".
 An attribute may contain any valid JSON value.
 
-If the value of an attribute is a JSON object or array, the member is called
-a [complex attribute](#document-structure-complex-attrs).
+If the value of an attribute is a JSON object or array, the member is called a
+*complex attribute*. The value is allowed to be any valid JSON structure.
+However, a JSON object that constitutes or is contained in a complex attribute
+must reserve the `id`, `type`, `links`, and `meta` members for future use.
 
 Although has-one foreign keys (e.g. `author_id`) are often stored internally
 alongside other information to be represented in a resource object, these keys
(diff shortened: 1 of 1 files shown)


ethanresnick:
+1. I don't know if I'm for #313 but, if we don't do #313 or this, we'll likely have to break backwards-compatibility (at least in a limited sense) to add linking from complex attributes down the line.

dgeb:
Adding for 1.0 final review.

tkellen:
I'm :+1: to finding a way to explicitly allow nested links for future consideration, but I think it is premature to have examples of this usage in the spec. What about simply reserving the `links` key at any level of nesting and leaving it at that?

tkellen:
Actually, reserving any reserved keywords at any level of nesting.

dgeb:
> Actually, reserving any reserved keywords at any level of nesting.

Yes, this is sufficient I believe.

bintoro:
"Any reserved keywords" is not well-defined. Would this mean `id`, `type`, `links`, and `meta`?

dgeb:
Yes - any keywords reserved in a resource object.

bintoro:
@dgeb Gotcha. What about the example? Remove altogether, remove linking, or leave as is?

Example usage was requested in #289.

dgeb:
@bintoro I definitely don't want to include an example here. We are simply reserving the keys, without making a statement about their usage.

bintoro:
Updated!

tkellen:
:+1:

dgeb:
@bintoro please rebase

bintoro:
Rebased.

By the way, if there's a short answer to why linking from complex attributes is so contentious, I'd love to hear it!

tkellen:
For my part, it's lack of experience with implementations doing this in the wild. It could well be that the resulting impact is minor, but I feel it bears more consideration before we explicitly support this. I'd love to continue the discussion post 1.0, though!

I am out of town for most of March but when I return I am going to create a test suite using a reference database that any implementor can use to validate compliance with the spec. Once that is in place I think it will be much easier for us to reason about additions like this--we can temporarily add them to the test suite and reference database to see the real world implications.

dgeb:
@bintoro thanks for rebasing!

> By the way, if there's a short answer to why linking from complex attributes is so contentious, I'd love to hear it!

The possibility of nested, n-level deep links that must be maintained and accessed independent of the containing object (ie. resource) adds a large burden of complexity to implementations. I would prefer that this be implemented as an extension first (and perhaps exclusively) as a proof of concept.

>  I am going to create a test suite using a reference database that any implementor can use to validate compliance with the spec.

Very excited for this @tkellen!

seanrucker:
Has linking from complex attributes made any headway since this discussion?

PR #418, opened 2015-03-06
Clarify definition of 'resource object'

bintoro:
With the latest changes, _resource object_ no longer seems to be used in the umbrella meaning (resources + relationships), so it's finally possible to give it a more focused definition without introducing other changes.

It also felt appropriate to explicitly waive the ID requirement for POST here, since it was only implied in the _Creating Resources_ section.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -123,13 +123,15 @@ with non-alphanumeric characters are allowed.
 
 ### Resource Objects <a href="#document-structure-resource-objects" id="document-structure-resource-objects" class="headerlink"></a>
 
-"Resource objects" appear in a JSON API document to represent primary data
-and included resources.
+"Resource objects" appear in a JSON API document to represent resources.
 
 A resource object **MUST** contain at least the following top-level members:
 
 * `"id"`
 * `"type"`
+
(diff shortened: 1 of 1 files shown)


dgeb:
I'm fine with this definition and I think it's good to explicitly waive the ID requirement here  :+1:

bintoro:
I just noticed the part on updating resources says:

> The `PATCH` request MUST include a single resource object as primary data. The resource object MUST contain a `type` member.

So, does the ID requirement not apply when updating, or is this an oversight in the PATCH section?

**EDIT:** If the spec wants to accommodate the rare case of changing the ID of an existing resource, then it would make sense for ID to remain optional for PATCH.

I'll update the PR as soon as someone lets me know what the intent here is.

dgeb:
@bintoro Very good catch - appreciate the careful review. I'd prefer: "The resource object **MUST** contain `id` and `type` members."

Thanks in advance for updating.

> If the spec wants to accommodate the rare case of changing the ID of an existing resource, then it would make sense for ID to remain optional for PATCH.

I don't believe we do want to accommodate that. If someone can make a compelling case to allow that, I'm (we're) all ears.

PR #421, opened 2015-03-06
Unify rules about reserved object members

### LINKED ISSUE #415: Attributes of relationships
IMHO, the wording on how relationships can be specified via linkage objects is not quite clear. Are the listed attributes (`type`, `id`, `data`) the only ones allowed in specifying the relationships, or custom attributes of relationships can be listed as well?

An example of a relationship with attributes: suppose that we have a tree of directories and files in a Unix filesystem, and that we need to describe two kinds of relationships between them: direct descendance (file or subdirectory belonging to a directory) and symlink. Does the spec allow to do it by including a custom attribute directly with the linkage object? I believe that `meta` is unsuitable for this because it describes relationship as a whole and not its specific members.

Specifically, is the example below compliant?

``` http
{
  "type": "directories",
  "id": "12",
  "name": "usr",
  "links": {
    "descendants": {
      "data": [{
        "type": "directories",
        "id": "34",
        "kind": "direct"
      },{
        "type": "directories",
        "id": "56",
        "kind": "symlink"
      },{
        "type": "files",
        "id": "78",
        "kind": "symlink"
      }]
    }
  }
}
```

To allow this explicitly, I propose to change the wording in the specification of linkage as follows (proposed addition is italicized):
- Linkage to other resource objects ("object linkage") included in a compound
  document. This allows a client to link together all of the resource objects
  included in a compound document without having to `GET` one of the
  relationship URLs. Linkage **MUST** be expressed as:
  - `type` and `id` members for homogeneous to-many relationships. `type` is
    not required if the value of `id` is an empty array (`[]`).
  - A `data` member whose value is an array of objects each containing `type`
    and `id` members for heterogeneous to-many relationships.
  
  _In all three cases, other members can be  OPTIONALLY included  alongside `type` and `id`
  to describe custom attributes of a relationship._
- A `"meta"` member that contains non-standard meta-information about the
  relationship.

Also, I think that the "complete example document with multiple included relationships" should be extended to include the illustration of `data` linkage.

Any chance to still consider this before the spec hits 1.0? 
I would be glad to promptly provide a PR for both proposed changes.

bintoro:
Simplifies the spec ~~and resolves #415~~.

This is possible because the restrictions all follow the "alphanumeric" rule.

I may occasionally write some weird quasi-English, so if the wording can be improved, please edit away.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -75,10 +75,7 @@ certain aspects are only applicable to one or the other. These differences are
 called out below.
 
 Unless otherwise noted, objects defined by this specification **MUST NOT**
-contain any additional members whose keys start with an alphanumeric character.
-
-Any object **MAY** contain additional members whose keys start with a
-non-alphanumeric character.
+contain any additional members.
 
 ### Top Level <a href="#document-structure-top-level" id="document-structure-top-level" class="headerlink"></a>
 
@@ -143,9 +140,8 @@ In addition, a resource object **MAY** contain any of these top-level members:
 * `"meta"`: non-standard meta-information about a resource that can not be
   represented as an attribute or relationship.
 
-A resource object **MAY** contain additional top-level members whose keys begin
-with an alphanumeric character. These members represent "attributes" and may
-contain any valid JSON value.
+A resource object **MAY** contain additional top-level members. These members
+represent "attributes" and may contain any valid JSON value.
 
 If the value of an attribute is a JSON object or array, the member is called a
 *complex attribute*. The value is allowed to be any valid JSON structure.
--- a/format/index.md
+++ b/format/index.md
@@ -118,10 +118,6 @@ A document's top level **MAY** also have the following members:
 If any of these members appears in the top-level of a response, their values
 **MUST** comply with this specification.
 
-The top level of a document **MUST NOT** contain any additional members whose
-names start with alphanumeric characters. Additional members whose names start
-with non-alphanumeric characters are allowed.
-
 ### Resource Objects <a href="#document-structure-resource-objects" id="document-structure-resource-objects" class="headerlink"></a>
 
 "Resource objects" appear in a JSON API document to represent resources.
@@ -283,11 +279,6 @@ without having to `GET` one of the relationship URLs.
 > Note: If present, a *related resource URL* must be a valid URL, even if the
 relationship isn't currently associated with any target resources.
 
-Besides the members described above (`self`, `resource`, `type`, `id`, `meta`,
-and the keys for pagination), a link object **MUST NOT** contain any additional
-members whose names start with alphanumeric characters. Additional members whose
-names start with non-alphanumeric characters are allowed.
-
 For example, the following article is associated with an `author` and `comments`:
 
 ```javascript
@@ -448,9 +439,22 @@ The top-level links object **MAY** contain the following members:
   data represents a resource relationship.
 * Pagination links for the primary data (as described below).
 
-The top-level links object **MUST NOT** contain any additional members whose
-names start with alphanumeric characters. Additional members whose names
-start with non-alphanumeric characters are allowed.
+### Additional Object Members <a href="#document-structure-additional-members" id="document-structure-additional-members" class="headerlink"></a>
+
+This specification describes various JSON objects with a specific purpose and
+format. Each description enumerates a set of *reserved members* that carry a
+special meaning for the object. Additional members are allowed subject to the
+following conditions.
+
+Any additional members are allowed in these objects:
+* resource objects (as *attributes*)
+* `meta` objects
+* error objects.
+
+Other objects **MUST NOT** contain any additional members whose keys start with
+an alphanumeric character. Additional members whose keys start with a
+non-alphanumeric character are allowed. This also applies to the object that
+constitutes the document's top level.
 
 ## Fetching Data <a href="#fetching" id="fetching" class="headerlink"></a>
 
@@ -1427,4 +1431,3 @@ An error object **MAY** have the following members:
   to the resource path(s) expressed in the error object's `"links"` member
   [e.g. `["/first-name", "/last-name"]` to reference a couple attributes].
 
-Additional members **MAY** be specified within error objects.
--- a/format/index.md
+++ b/format/index.md
@@ -143,8 +143,9 @@ In addition, a resource object **MAY** contain any of these top-level members:
 * `"meta"`: non-standard meta-information about a resource that can not be
   represented as an attribute or relationship.
 
-Any other top-level member in a resource object represents an "attribute".
-An attribute may contain any valid JSON value.
+A resource object **MAY** contain additional top-level members whose keys begin
+with an alphanumeric character. These members represent "attributes" and may
+contain any valid JSON value.
 
 If the value of an attribute is a JSON object or array, the member is called a
 *complex attribute*. The value is allowed to be any valid JSON structure.
@@ -435,6 +436,8 @@ For example:
   }
 }
 ```
+
+Any members **MAY** be specified within `meta` objects.
 
 ### Top-level Links <a href="#document-structure-top-level-links" id="document-structure-top-level-links" class="headerlink"></a>
 
--- a/format/index.md
+++ b/format/index.md
@@ -74,6 +74,12 @@ Although the same media type is used for both request and response documents,
 certain aspects are only applicable to one or the other. These differences are
 called out below.
 
+Unless otherwise noted, objects defined by this specification **MUST NOT**
+contain any additional members whose keys start with an alphanumeric character.
+
+Any object **MAY** contain additional members whose keys start with a
+non-alphanumeric character.
+
 ### Top Level <a href="#document-structure-top-level" id="document-structure-top-level" class="headerlink"></a>
 
 A JSON object **MUST** be at the root of every JSON API response containing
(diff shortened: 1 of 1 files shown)


dgeb:
I think I'm :+1: on this move to DRY up the spec. After repeating similar verbiage in a few places, I was considering consolidation too.

Let's review the verbiage before 1.0 ...

bintoro:
IMO the intro about "reserved members" is somewhat superfluous because it's plain and obvious what the "additional" members are in addition to. It's there because the removed parts had some context that is absent from the dedicated section.

A condensed version could be as simple as this:

> Additional members in JSON objects described by this specification are allowed as follows:
> 
> [...]

I think this makes it perfectly clear even without any convoluted explanations.

dgeb:
@bintoro agreed - no need to provide a name for something that's only referenced in one place. Please go with the simpler form.

bintoro:
Updated. Short and sweet!

tkellen:
I think this needs another pass.  If the goal is simply to say "keys beginning with non-alphanumeric characters are allowed in any object", I think we can do that even more concisely. I find it a bit hard to extract that meaning with the current wording.

dgeb:
I think it's fine to leave the first bullet point in the respective sections. And perhaps a central note such as this would be sufficient:

"The top-level object, resource objects, and link objects **MAY** contain additional members whose keys start with a non-alphanumeric character."

dgeb:
Or perhaps:

"Any objects described in this specification, including the top-level object, **MAY** contain additional members whose keys start with a non-alphanumeric character."

bintoro:
@dgeb Precisely. A key motivation here was to include the various unnamed objects too. But it gets too messy if the "**MUST NOT** contain any additional members" clause is added for those objects too. Ignore them or retain the global rules?

dgeb:
@bintoro I see. I'm sure we can work this out cleanly. Perhaps we provide a name for these optional custom keys, and then we can mention them wherever that clause appears. So, what to call them...

bintoro:
I think the global "Other objects **MUST NOT** contain any additional members" statement is quite useful because then you don't have to specify it separately for things like the `{id, type}` objects that are used with relationship URLs. That's the main benefit of the dedicated section. Perhaps the wording can still be simplified a bit, but some amount of legalese is necessary to be able to have it all in one place.

bintoro:
The second bullet can be cleaned up a bit by using a formal MAY clause and dropping the redundant part about the top-level object:

> Other objects **MUST NOT** contain any additional members whose keys start with an alphanumeric character. They **MAY** contain additional members whose keys start with a non-alphanumeric character.

If the enumeration of the "arbitrary members allowed" objects in the first bullet is somehow confusing, I guess it could be turned into a sublist.

Other than that, I'm running short on ideas on how to make the section any clearer than this. When I looked at it with fresh eyes earlier today, the thing seemed fine...

**EDIT:** Alternatively, we could enumerate nothing and just say:

>  Unless otherwise noted, objects defined by this specification **MUST NOT** contain any additional members whose keys start with an alphanumeric character. They **MAY** contain additional members whose keys start with a non-alphanumeric character.

This could go directly in the _Document Structure_ main section.

dgeb:
> Unless otherwise noted, objects defined by this specification MUST NOT contain any additional members whose keys start with an alphanumeric character. They MAY contain additional members whose keys start with a non-alphanumeric character.

Well done, @bintoro. I really like the idea of moving this to the top of the _Document Structure_ main section. I would prefer that broad conditions like this be made early.

I would re-phrase the start of the second sentence from "They MAY ..." to "Any object MAY ...". Maybe that should even be its own paragraph, since it is isolated from the "Unless otherwise noted" clause. So two separate paragraphs:

> Unless otherwise noted, objects defined by this specification **MUST NOT** contain any additional members whose keys start with an alphanumeric character.
> 
> Any object **MAY** contain additional members whose keys start with a non-alphanumeric character.

What do you think?

bintoro:
Yeah that's perfect. I amended the commit accordingly.

bintoro:
One thing to think about is if resource objects and `meta` objects could use a formal MAY clause to allow arbitrary keys. Currently it's only stated indirectly. The intent is quite clear, though.

tkellen:
Huge :+1: to this revision--thanks for being willing to go another round on this!

dgeb:
> One thing to think about is if resource objects and meta objects could use a formal MAY clause to allow arbitrary keys. Currently it's only stated indirectly. The intent is quite clear, though.

Upon review, I am in favor of a formal MAY clause for resource attributes and `meta` object members. Might as well be mind-numbingly clear about these things.

@bintoro do you want to add a commit to this PR? We could also handle it separately.

bintoro:
> In addition, a resource object **MAY** contain any of these top-level members:
> - `links`
> - `meta`
> 
> Additional members **MAY** be specified [...]

That doesn't feel quite right, so I was thinking of talking about "reserved members" explicitly:

> In addition, a resource object **MAY** contain any of these reserved top-level members:

Yea/nay?

dgeb:
How about we change:

> Any other top-level member in a resource object represents an "attribute". An attribute may contain any valid JSON value.

to:

> A resource object **MAY** contain additional top-level members, each having a key that begins with an alphanumeric character, that represent "attributes". An attribute **MAY** contain any valid JSON value.

Note that I corrected the final **MAY** too.

tkellen:
> A resource object **MAY** contain additional top-level members whose keys begin with alphanumeric characters. These members represent "attributes" which **MAY** contain any valid JSON value.

dgeb:
@tkellen yes, that reads more clearly.

I actually question the final **MAY** (for both of ours, and the current spec).

bintoro:
The informal "may" was intentional, I believe. What else are you going to put in an attribute besides a valid JSON value? If we're going to use a formal keyword it needs to be a MUST.

dgeb:
@bintoro that's what I was thinking. It's weird to use the informal "may" in a spec.

bintoro:
I always thought "may" was fine, but it could just as well be a MUST.

Can't members with non-alphanumeric initials be attributes?

dgeb:
Let's keep the informal "may".

> Can't members with non-alphanumeric initials be attributes?

No, those are to be ignored by the spec.

tkellen:
Presumably not. For example, if you were using json-api and json-ld together (ick), I wouldn't expect sparse fieldsets to interact with a `@content` member.

bintoro:
> > Can't members with non-alphanumeric initials be attributes?
> 
> No, those are to be ignored by the spec.

Well, in that case the new phrasing is better because this was never pointed out anywhere, I think.

Have a look at the new commit. Used @tkellen's version but changed "which" to "and" and "MAY" to "may".

dgeb:
Solid :+1: from me - thanks for seeing this through @bintoro!

tkellen:
I'm good with this pending one final concern: if the spec dictates ignoring members that begin with something other than alphanumeric characters, how will this affect other languages (e.g. unicode surrogate pairs)?

dgeb:
@tkellen this is a completely legitimate concern and may be a showstopper for the whole non-alphanumeric exception.

bintoro:
The non-alphanumeric characters people are going to want to use probably fall within ISO-8859-1 or so. It would be fairly easy to explicitly define the available custom prefixes, and then, anything else is valid for attributes etc.

bintoro:
Any object defined by this specification **MAY** contain additional members whose keys start with an _extension prefix_, defined as one of `! # $ % & * + - . / : ; < > ( ) [ ] { } = ? @ \ ^ _ | ~`.

Unless otherwise noted, objects defined by this specification **MUST NOT** contain any additional members whose keys do not start with an extension prefix.

---

Even a subset of that would be plenty of choice.

dgeb:
@bintoro I think you are on a reasonable track here. We'll need to define a specific character or set of characters as reserved, and provide a name for that set. I think "extension prefix" is confusing since JSON API already has a concept of "extensions".

We would also need to call out each character in unicode, like (U+0040 COMMERCIAL AT, "@").

Since we are discussing narrowing the set to such an extreme, we could simplify implementations by limiting it to just the `@` prefix. With the set you've defined, I'm imagining some messy regexes in every client.

Very interested in others thoughts here, since this exception has been largely a concession to interoperability with other specifications.

bintoro:
Yeah, that entire set of characters is excessive IMO.

Were it up to me, I'd pick at least two from `! # $ @ ^`. Having just one may cause regret down the road.

It's sleepy time for me now, but I've got a feeling this won't be resolved today :laughing:

dgeb:
@bintoro thanks for all your help today - have a great night!

steveklabnik:
This clause is becoming increasingly problematic... :/

dgeb:
I think @steveklabnik is right. I'm now leaning toward leaving this exception to a custom extension. Extensions can provide breaking changes to the base specification. If a client and server are able to agree on custom prefixed keys, then they should be able to negotiate that formally through an extension.

tkellen:
:+1:

dgeb:
@bintoro you've been through a lot with this PR, I understand. Are you interested in removing the custom exceptions for non-alphanumeric keys? If not, I can work this up in a separate PR.

If anyone would like to make a case for keeping any exceptions, rather than doing so through custom extensions that codify special keys, now would be the time.

bintoro:
@dgeb Sure thing. But I've got another viewpoint on this extension thing. The way I see it, extensions are for functionality that cannot be provided compatibly. This case doesn't really fall into that category.

~~Consider a server that provides extra information in objects (e.g., JSON-LD), but a client isn't required to understand that information in order to use the server. Having to declare an extension would leave generic clients scratching their heads.~~ What am I talking about... This can be handled by extension negotiation.

Or when an application that uses custom members wants to use another extension like Bulk, it would have to define a proprietary Bulk extension that incorporates support for custom members (since there can only be one extension per request).

Reserving a couple of magic prefixes in the base spec would seem to be good for interoperability.

dgeb:
@bintoro Thanks for updating this PR one more time.

I really see no problem with extensions being lightweight layers over other extensions. They certainly can be additive only and reference another extension. I think this is the best way to reduce surprises and ensure that clients and servers completely agree entirely on a format.

steveklabnik:
Thank you for your hard work, @bintoro. So much effort so so few lines, such is specification work. I'm happy with the way this turned out.

ethanresnick:
@steveklabnik @bintoro @dgeb I totally missed this thread, so I'm sorry for jumping in late. I'm fine with the spirit of this change (i.e. handling these cases through extensions), but I wanted to comment on the language...

If we say that additional members **MUST NOT** be present, doesn't that mean that a conforming implementation is required to respond with an error if additional members are present? If so, then this accomplishes the opposite of what we intend (i.e. allowing us to add properties later that won't break existing clients). So I think we actually mean to say that "additional members **MUST** be ignored by clients and servers, as future versions of JSON-API may give these members special meanings".

bintoro:
@ethanresnick That's a very legitimate point. While the spec doesn't require that an error should be raised when extra keys are encountered, it's perfectly conceivable that a strict implementation might choose to do exactly that. Maybe raise a new issue/PR for this?

ethanresnick:
@bintoro Yup, I'll submit a pr tomorrow!

PR #423, opened 2015-03-07
Remove obsolete note about missing fields when updating

bintoro:
The removed note was intended to justify PUT-as-update and is no longer relevant.


dgeb:
Agreed. This note is no longer relevant for PATCH.

Issue #370, opened 2015-02-25
Make 'self' a direct attribute of a resource object


bintoro:
As it stands, the workings of `self` are inconsistent.

Inside a _link object_, `self` indicates the URL of the thing represented by the **object itself**. The `self` member appears together with the object's `id` and `type`.

Inside a _links object_, `self` points to the URL of the thing represented by the **parent object**. Given the previous observation, this is illogical.

This may not seem like a huge deal, but there are a couple of another points here too:
- In JSON API, the `links` member is intended as a container for _relationships_. It seems like a needless complication to stuff an extra key there just for the fun of it, when it has nothing to do with relationships. This relates to my ongoing effort to put some sense in the link/links/linked/related/included conceptual mess (#380).
- Clients may need to look up canonical URLs on the `self` members of related resource objects. Assuming you want to exclude the actual resource attributes, how do you request the inclusion of `self` when it's neither a relationship nor an attribute?

I propose to move the offending `self` member one level up to the context it actually belongs to:

```
{
  "data": [{
    "type": "articles",
    "id": "1",
    "self": "http://example.com/articles/1",           <---
    "title": "JSON API paints my bikeshed!",
    "links": {
      "author": {
        "self": "http://example.com/articles/1/links/author",
        "resource": "http://example.com/articles/1/author",
        "type": "people",
        "id": "9"
      },
      "comments": {
        "self": "http://example.com/articles/1/links/comments",
        "resource": "http://example.com/articles/1/comments",
        "type": "comments",
        "ids": ["5", "12"]
      }
    }
  }]
}
```

Now `self`, `id`, and `type` are aligned both at the resource level and the link level.

There's no reason why `self` needs to be inside `links`. I presume this idea came from HAL, but the ways in which `links` is used in HAL and JSON API are not really analogous.

dgeb:
@wycats and I aren't seeing this as an inconsistency. Will comment more later.

bintoro:
@dgeb, understood.

Canonical URL troubles would be better addressed directly in the link object anyway (#379). Closing because this is just an annoyance, not a huge defect.

But if #379 doesn't happen (the two-key solution), and `self` can only be delivered within`links`, then JSON API isn't going to be a very elegant format for applications that use canonical URLs for linking.

PR #424, opened 2015-03-07
Define "fields" & add details about 'fields' param

bintoro:
[Not needed for 1.0]

"Field" still appears a few times but is not defined anywhere. It used to appear more, but #401 expanded them to "attributes and relationships".

This PR formally defines _fields_ as _attributes_ + _relationships_.

The _Sparse Fieldsets_ section is edited a bit:
- The format of the `fields` param is spelled out (it's a list).
- It is noted that servers MAY return `self` regardless of `fields`.
- Occurrences of "attributes and/or relationships" are changed back to "fields" (partly reverses #401).

The term _field_ seems like a useful one, so defining it rather than avoiding it felt appropriate. I'll amend if others don't share this view.

Note: `self` is an odd one out, as it's not a field but not properly a _member_ either (#370), so I loosely refer to everything as "properties".

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -149,6 +149,9 @@ If the value of an attribute is a JSON object or array, the member is called a
 However, a JSON object that constitutes or is contained in a complex attribute
 must reserve the `id`, `type`, `links`, and `meta` members for future use.
 
+A resource object's attributes and relationships are collectively called its
+"fields". <a href="#document-structure-resource-fields" id="document-structure-resource-fields"></a>
+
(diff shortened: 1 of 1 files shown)


tkellen:
:+1: @dgeb and I waffled about this a bit on Friday and I think this ties things up nicely.

dgeb:
I'm also :+1: on this.

@bintoro could you please rebase?

bintoro:
Rebased.

PR #425, opened 2015-03-07
Introduce 'canonical' and clarify 'resource'

### LINKED ISSUE #379: Clarify the meaning of "related resource URL"
Is it the intent of the spec that a _related resource URL_ (as defined in [Resource Relationships ](http://jsonapi.org/format/#document-structure-resource-relationships)) SHOULD always reflect an up-to-date state of the relationship? In other words, are we saying that the URL SHOULD NOT be the canonical URL of a to-one related resource?

Currently the definition is followed by:

> A related resource URL SHOULD remain constant even when the resource(s) it represents mutate.

I find it obvious that when _resources_ mutate, their URLs should remain unchanged.

I suspect this is trying to say: "A related resource URL SHOULD remain constant even when the _relationship_ (the set of referenced resources) mutates". That would be the equivalent of the rule stated in the first paragraph, i.e., no canonical URLs please.

It's important to set this straight because when a client performs an operation on a related resource, it needs either the related URL or the canonical URL depending on intent. They're not (generally) interchangeable, so there must be no room for confusion.

(cc @tilboerner)

bintoro:
[Please review before 1.0]

Resolves: #379

Currently, the `resource` member in link objects may be tied either to the relationship or to the identity of the target resource(s). This ambiguity sets the stage for all kinds of elusive bugs and race conditions. It would be prudent to resolve this conclusively before 1.0.

This PR addresses the issue in the following way:
- The role of `resource` is clarified by limiting it to "relationshippy" URLs only.
- A new member named `"canonical"` is allowed in link objects to hold a canonical URL.

An alternative naming scheme for `resource` and `canonical` would be `href-related` / `href-canonical`, which is more descriptive but involves changing the established key `resource`.

Additionally, a Note is added that points out the following, which otherwise may not be obvious:
- a related resource URL cannot be replaced with `null` to indicate "empty" (#386)
- an empty relationship can be signaled by including `id: null` or `id: []` (#386, #399).

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -222,12 +222,16 @@ The name of the relationship declared in the key **SHALL NOT** be `"self"`.
 
 The value of a relationship **MUST** be one of the following:
 
-* A string, which represents a URL for the related resource(s) (a "related
-  resource URL"). When fetched, it returns the related resource object(s) as the
-  response's primary data. For example, an `article`'s `comments` could specify a
-  URL that returns a list of comment resource objects when retrieved through a
-  `GET` request. A related resource URL **SHOULD** remain constant even when the
-  relationship it represents mutates.
+* A URL for the related resource(s) (a "related resource URL"). When fetched, it
+  returns the related resource object(s) as the response's primary data. For
+  example, an `article`'s `comments` relationship could specify a URL that
+  returns a list of comment resource objects when retrieved through a `GET`
+  request.
+
+  A related resource URL **MUST** remain constant even when the relationship (the
+  set of identities of the referenced resources) mutates. That is, the response
+  from a related resource URL always reflects the current state of the
+  relationship.
 
 * An object (a "link object").
 
@@ -239,6 +243,11 @@ one of the following:
   relationship. For example, it would allow a client to remove an `author` from
   an `article` without deleting the `people` resource itself.
 * A `resource` member, whose value is a related resource URL (as defined above).
+* A `canonical` member, whose value is a canonical URL for the related
+  resource(s). A canonical URL **MUST** be permanently tied to the identity of
+  the target resources. That is, a previously retrieved canonical URL
+  continues to point to the same resources even if the original relationship
+  mutates.
 * Linkage to other resource objects ("object linkage") included in a compound
   document. This allows a client to link together all of the resource objects
   included in a compound document without having to `GET` one of the
@@ -257,6 +266,11 @@ pagination links, as described below.
 
 If a link object refers to resource objects included in the same compound
 document, it **MUST** include object linkage to those resource objects.
+
+> Note: If present, a *related resource URL* must be a valid URL, even if the
+relationship isn't currently associated with any target resources. If a
+relationship is empty, the server can indicate this in the link object by
+including an `id: null` or `id: []` member depending on relationship cardinality.
 
 Besides the members described above (`self`, `resource`, `type`, `id`, `meta`,
 and the keys for pagination), a link object **MUST NOT** contain any additional
--- a/format/index.md
+++ b/format/index.md
@@ -290,37 +290,7 @@ For example, the following article is associated with an `author` and `comments`
     "author": {
       "self": "http://example.com/articles/1/links/author",
       "resource": "http://example.com/articles/1/author",
-      "type": "people",
-      "id": "9"
-    },
-    "comments": {
-      "resource": "http://example.com/articles/1/comments"
-    }
-  }
-}
-// ...
-```
-
-The `author` relationship includes a URL for the relationship itself (which
-allows the client to change the related author without deleting the `people`
-object), a URL to fetch the resource objects, and linkage information for
-the current compound document.
-
-The `comments` relationship is simpler: it just provides a URL to fetch the
-comments. The following resource object, which provides the `comments`
-relationship as a string value rather than an object, is equivalent:
-
-```javascript
-// ...
-{
-  "type": "articles",
-  "id": "1",
-  "title": "Rails is Omakase",
-  "links": {
-    "self": "http://example.com/articles/1",
-    "author": {
-      "self": "http://example.com/articles/1/links/author",
-      "resource": "http://example.com/articles/1/author",
+      "canonical": "http://example.com/people/9",
       "type": "people",
       "id": "9"
     },
@@ -329,6 +299,14 @@ relationship as a string value rather than an object, is equivalent:
 }
 // ...
 ```
+
+The `author` relationship includes a URL for the relationship itself (which
+allows the client to change the related author without deleting the `people`
+object), URLs to fetch the resource objects, and linkage information for
+the current compound document.
+
+The `comments` relationship is simpler: it just provides a related resource URL
+to fetch the comments, so the URL can be provided as a string value.
 
 ### Compound Documents <a href="#document-structure-compound-documents" id="document-structure-compound-documents" class="headerlink"></a>
 
@@ -352,6 +330,7 @@ A complete example document with multiple included relationships:
       "author": {
         "self": "http://example.com/articles/1/links/author",
         "resource": "http://example.com/articles/1/author",
+        "canonical": "http://example.com/people/9",
         "type": "people",
         "id": "9"
       },


bintoro:
While adding `canonical` to the example payloads, I realized that it's totally unnecessary to have a dedicated example just to illustrate the changing of

```
"comments": {
  "resource": "http://example.com/articles/1/comments"
}
```

to 

```
"comments": "http://example.com/articles/1/comments"
```

so I removed it and put the short form in the remaining example. Let me know if you'd like to have this reverted.

dgeb:
> Currently, the resource member in link objects may be tied either to the relationship or to the identity of the target resource(s).

I disagree here. The spec is pretty clear that "A related resource URL SHOULD remain constant even when the relationship it represents mutates."

I do see that the addition of a field such as `canonical` further reinforces the purpose of `resource`. However, I don't see the addition of `canonical` as a blocker for 1.0, since it's entirely additive.

dgeb:
I will mark this for 1.0 because we should be absolutely certain of the name `resource`, to ensure that `canonical` (or whatever is decided) is a nice complement.

bintoro:
@dgeb That SHOULD is a problem. Change it to MUST and the biggest issue goes away. Then `canonical` can become an additive change down the road.

But as @ethanresnick speculated in #379:

>  I have a feeling that that SHOULD was put in the spec carefully, given that nearly everything else has been converted to a MUST. (Perhaps it's a SHOULD to save servers the requirement of having to support non-canonical urls at all.)
> 
> So, if that SHOULD isn't going to change, then I'm on board with your two-key solution, which would certainly allow us to specify a MUST for each key individually.

dgeb:
@bintoro fair enough - we SHOULD examine that wording carefully before 1.0 :+1:

steveklabnik:
I brefily mentioned to @dgeb that I would rather re-use the existing `canonical` link relation rather than introducing canonical into this spec. It would mean just allowing any link relation to be the key.

bintoro:
It occurred to me that the spec actually defines **resource URL** as the canonical URL for singular resources.

Therefore, placing the **related resource URL** in a member named `resource` is perhaps not an ideal arrangement, regardless of what happens with `canonical`.

ethanresnick:
Re the naming question....

In REST terms, we have two resources here: 
1. `/articles/1/comments`, which represents the contents of the comments link
2. `/comments/12,13,14`, which represents comments 12, 13, and 14

Now, the entities picked out by the first resource might happen, at some points in time, to be the same as the entities picked out by the second resource. But `"canonical"` really doesn't apply here; we have two different conceptual resources, not two IRIs for the same resource, one of which we want to designate as the canonical IRI.

Moreover, for someone who doesn't quite understand the resource picture described above, but only has a mental model in which the `"canonical"` url is the permanent one, I can imagine them thinking that `"canonical"` means the exact opposite of the meaning being proposed. (I.e. they'd expect that the unchanging url in the `"resource"` key should be the one in the `"canonical"` key, and vice-versa.)

So, if we want a second key, I'd probably call it `"current"`, to say that "this url represents the value of the link".

But, honestly, for 1.0, I think it'd be fine to just change the SHOULD to a MUST for the `resource` key. Supporting `"resource"`-style urls isn't that onerous on servers. Then, we can add a second key later.

bintoro:
@ethanresnick I'm confused by your comment.

The related resource URL (the `resource` member) is defined like this: "[...] the response from a related resource URL always reflects the current state of the relationship".

Isn't that pretty much the definition of "current" and opposite of "canonical"? :smile: 

Or is the concern that people don't know what "canonical" means? I think the _Resource URL_ section (where `self` is defined) is the right place to highlight this and clearly point out that individual resources have identities.

This PR defines the proposed `canonical` member like this:

> A canonical URL **MUST** be permanently tied to the identity of the target resources. That is, a previously retrieved canonical URL continues to point to the same resources even if the original relationship mutates.

**EDIT:** If you're saying that `resource` should be named something like `current` and `canonical` should be named `resource` then I agree.

bintoro:
Given @ethanresnick's observation that the term _canonical_ is not a perfect fit conceptually (since the various URLs do not expose permanently duplicate content), I present two naming plans that do not involve "canonical":

One:
- `"static"` for the permalink
- `"href"`, `"current"`, `"related"`, or _maybe_ `"resource"` for the related resource URL

Two:
- `"resource"` for the permalink
- `"href"`, `"current"`, or `"related"` for the related resource URL

ethanresnick:
@bintoro I think I understand the confusion. I was proposing `"current"` for what you're calling the permalink because I read it as "This is the url that (always) returns the specific entities that are linked _currently_ (i.e. at the time that the containing payload was downloaded)". Your reading, which also makes sense, has it mean "This is the url that always returns the current entities that the relationship is about." Because of this double reading, which also applies to `"static"`, I think we should avoid those terms.

**EDIT** Would using `"present"` or `"now"` instead of `"current"` solve the issue? So we'd have `"present"`/`"now"` for the permalink and `"resource"` for the related resource url?

bintoro:
> Because of this double reading, which also applies to `"static"`, I think we should avoid those terms.

Or, you know, we could just assume that implementors read the spec :smirk:

Realistically, there's no set of keys that would be completely self-explanatory.

I do see your point about `"current"` and how someone might get the wrong idea. But for `"static"`, it's a stretch... Even with conscious effort, I have a hard time associating the word "static" with a mutable set of resources.

I'm also strongly opposed to using a word like "current" or "now" in the permalink meaning because, although I concede both readings are possible, the _related resource_ interpretation feels way way _way_ more natural to me, and I'm sure I'm not the only one. So let's just forget about "current".

In the initial comment I suggested `"href-related"`, and I still think it's quite reasonable, given that the term is **related resource URL**. I would be happy with `static` / `related`. The term for what is now _canonical_ would become **static resource URL**.

dgeb:
I would welcome a separate PR that changes the current `resource` to `related` and changes the associated SHOULD to a MUST. This URL can continue to be called the _related resource URL_.

This was a less clear path when `linked` was at the top-level (instead of `included`), but there is no longer the possibility for confusion between `related` and `linked`.

I believe that this change is worth making, and not just bike-shedding, because it removes the word "resource" from the structure of a JSON API document. This frees us to change this term in the future, if necessary, to disambiguate JSON API resources from HTTP resources.

I am less clear on the need for a "static resource URL". The spec no longer supports the concept of links to multiple individual resources, and the former comma-delimited-ID approach was not appropriate for long lists. A different "recommendation" would be needed for forming these URLs. Let's discuss that in a new issue (not PR) to be addressed post-1.0.

@bintoro please let me know if you agree with the above proposal and whether you'd be willing to narrow this PR in scope. Thanks.

dgeb:
@steveklabnik my concern about allowing any link relation is that it means that the keys we reserve for 1.0 (`self`, `related`, `first`, `last`, `prev`, `next`) are the only ones we'll ever be able to reserve and narrowly define.

Maybe the answer is to reserve a prefix for custom keys? We're already reserving keys that begin with non-alphanumeric characters for custom usage. Do we want another reserved prefix in `links`, such as `href-`?

Or perhaps I misunderstood your suggestion?

dgeb:
> please let me know if you agree with the above proposal and whether you'd be willing to narrow this PR in scope.

@bintoro of course, feel  free to open a new PR instead. And if you don't have time, I'll be glad to make this change.

bintoro:
@dgeb Sounds perfect. How do you feel about the slightly amended wording in the definition of related resource URL, as per this PR? And the Note about representing empty relationships?

format/index.md
```diff
@@ -222,12 +222,16 @@ The name of the relationship declared in the key **SHALL NOT** be `"self"`.
...
+* A URL for the related resource(s) (a "related resource URL"). When fetched, it
+  returns the related resource object(s) as the response's primary data. For
+  example, an `article`'s `comments` relationship could specify a URL that
+  returns a list of comment resource objects when retrieved through a `GET`
+  request.
+
+  A related resource URL **MUST** remain constant even when the relationship (the
+  set of identities of the referenced resources) mutates. That is, the response
+  from a related resource URL always reflects the current state of the
+  relationship.
```
dgeb, line 235: @bintoro :+1: to this clarification

format/index.md
```diff
@@ -239,6 +243,11 @@ one of the following:
   relationship. For example, it would allow a client to remove an `author` from
   an `article` without deleting the `people` resource itself.
 * A `resource` member, whose value is a related resource URL (as defined above).
```
dgeb, line 245: change `resource` to `related`

format/index.md
```diff
@@ -258,6 +267,11 @@ pagination links, as described below.
 If a link object refers to resource objects included in the same compound
 document, it **MUST** include object linkage to those resource objects.
 
+> Note: If present, a *related resource URL* must be a valid URL, even if the
+relationship isn't currently associated with any target resources. If a
+relationship is empty, the server can indicate this in the link object by
+including an `id: null` or `id: []` member depending on relationship cardinality.
+
```
dgeb, line 274: Let's keep the first sentence, but not the second. The second is covered in the fetching relationships section (and btw, `"data": null` and `"data": []` are valid responses for empty relationships)

dgeb:
@bintoro I've commented on the sections I'd like to keep for this PR. Let's deal with other issues separately.

ethanresnick:
@dgeb's proposal sounds good to me too!

bintoro:
> The second is covered in the fetching relationships section

@dgeb Just to clarify, the intent of the note was to point out that an empty relationship can be indicated in the link object, too, saving the client a request to the relationship URL when there's nothing there.

The _Fetching Relationships_ section doesn't seem to address this at all; it only deals with responses from a relationship URL.

dgeb:
> Just to clarify, the intent of the note was to point out that an empty relationship can be indicated in the link object, too, saving the client a request to the relationship URL when there's nothing there.

Thanks for the clarification. I didn't get that on first reading (my fault, not yours). Still not sure we need that note, but I'm more open to it now.

dgeb:
Closing in favor of #433

steveklabnik:
> that the keys we reserve for 1.0 (self, related, first, last, prev, next) are the only ones we'll ever be able to reserve and narrowly define.

of those, only `related` is not already a standard relation, and we're using them in compliance with the standard relations.

PR #427, opened 2015-03-07
Change "data object" to "resource object"

bintoro:
I wasn't able to make the two-item list make grammatical sense, so I reworded the whole thing. Hope it's fine.


dgeb:
Yes, this is consistent. Thanks @bintoro!

PR #428, opened 2015-03-07
Non-normative note about responses without message-body.

dgeb:
The statement about 204s should be both non-normative and more 
inclusive. Other responses, such as 404, may not contain a message-body 
either.

The relevant normative line is:

> A JSON object **MUST** be at the root of every JSON API response 
> containing data.


steveklabnik:
Sounds good

Issue #365, opened 2015-02-22
Do not hurry for 1.0


AlexParamonov:
Hello.

Recenty merged pull request https://github.com/json-api/json-api/pull/341 contains breaking changes to the previous JSON API resource structure. It actually not the resource anymore, but data. Everything is a data now.

And I completely disagree with that. I believe there are other people who thinks differently that PR author.
Previous version was online for quite a long time, and it was adopted and loved for it old structure. **Don't rush** for irreversible changes and lock the 1.0. Listen to other people before changing the standard.

Now, lets get closer to what I am disagree with exactly.
It is the `data` root namespace and required `type` key.
Don't let the bad code practices fool you. It is absolutely not needed and adds more problems that it could possibly solve (if any):
### Naming.

"data" keyword means nothing. Server returns a set or unknown things. Query the /videos enpoint and get some unknown data objects that could be identified only by the type key.

Linked objects. ungrouped and comes in a mess:

```
"linked": [{
    "type": "people",
    "id": "9",
    "first-name": "Dan",
    "last-name": "Gebhardt",
    "twitter": "dgeb",
    "links": {
      "self": "http://example.com/people/9"
    }
  }, {
    "type": "comments",
    "id": "5",
    "body": "First!",
    "links": {
      "self": "http://example.com/comments/5"
    }
  }, {
    "type": "comments",
    "id": "12",
    "body": "I like XML better",
    "links": {
      "self": "http://example.com/comments/12"
    }
  }]
```

Clients are forced to go item by item to build required collection (comments or people).
### Pseudo help for sti.

Don't leak your DB structure to the API!

If response should contain objects of the different type, don't return unnamed data mess.
return the "covers" or "thumbnails" instead:

```
{
  "thumbnails": [
    {
      "id": "laonet",
      "title": "LAO NET TV",
      "cover_image": {
        "sd": "http://example.com/sd.jpg",
        "hd": "http://example.com/hd.jpg"
      },
      "href": "http://example.com/channels/laonet",
      "type": "channel"
    },
    {
      "id": "dexter",
      "title": "Dexter",
       "cover_image": {
        "sd": "http://example.com/sd.jpg",
        "hd": "http://example.com/hd.jpg"
      },
      "href": "http://example.com/shows/dexter",
      "type": "show"
    },
}
```
### ~~Resource~~ Data creation.

POST /photos

```
{
  "data": {
    "type": "photos",
    "title": "Ember Hamster",
[...]

PR #431, opened 2015-03-08
Attributes of relationships #415

### LINKED ISSUE #415: Attributes of relationships

hhware:
Resolves: #415

### DIFF
--- a/extensions/jsonpatch/index.md
+++ b/extensions/jsonpatch/index.md
@@ -27,7 +27,7 @@ operations allowed in this top level array.
 ### Request URLs and Patch Paths <a href="#patch-urls" id="patch-urls" class="headerlink"></a>
 
 The request URL and each Patch operation's `"path"` are complementary and
-**MUST** combine to target a particular resource, collection, attribute, or
+**MUST** combine to target a particular resource, collection, resource attribute, or
 relationship.
 
 If a server supports the Patch extension, it **MUST** allow Patch operations at
@@ -65,9 +65,9 @@ Accept: application/vnd.api+json; ext=jsonpatch
 ]
 ```
 
-### Updating Attributes <a href="#patch-updating-attributes" id="patch-updating-attributes" class="headerlink"></a>
+### Updating Resource Attributes <a href="#patch-updating-attributes" id="patch-updating-attributes" class="headerlink"></a>
 
-To update an attribute, perform a `"replace"` operation with the attribute's
+To update a resource attribute, perform a `"replace"` operation with the attribute's
 name specified by the `"path"`.
 
 For instance, the following request should update just the `src` property of the
@@ -227,8 +227,8 @@ Accept: application/vnd.api+json; ext=jsonpatch
 #### 204 No Content <a href="#patch-responses-204" id="patch-responses-204" class="headerlink"></a>
 
 A server **MUST** return a `204 No Content` status code in response to a
-successful Patch operation in which the client's current attributes remain up to
-date.
+successful Patch operation in which the client's current resource attributes 
+remain up to date.
 
 #### 200 OK <a href="#patch-responses-200" id="patch-responses-200" class="headerlink"></a>
 
--- a/extensions/jsonpatch/index.md
+++ b/extensions/jsonpatch/index.md
@@ -162,11 +162,12 @@ To add an element to a to-many relationship, request an `"add"` operation that
 targets the relationship's URL. Because the operation is targeting the end of a
 collection, the `"path"` must end with `"/-"`.
 
-In the following example, the comment with ID `123` is added to the list of
-comments for the article with ID `1`:
+In the following example, the new reprint with ID `123` is added to the list of
++publications of the article with ID `1` ():
+
 
 ```text
-PATCH /articles/1/links/comments
+PATCH /articles/1/links/publications
 Content-Type: application/vnd.api+json; ext=jsonpatch
 Accept: application/vnd.api+json; ext=jsonpatch
 
@@ -175,7 +176,7 @@ Accept: application/vnd.api+json; ext=jsonpatch
     "op": "add", 
     "path": "/-", 
     "value": [
-      { "type": "comments", "id": "123" }
+      { "type": "magazines", "id": "123", "publication_type": "reprint" }
     ]
   }
 ]
--- a/format/index.md
+++ b/format/index.md
@@ -308,7 +308,17 @@ Resource linkage **MUST** be represented as one of the following:
 * an array of linkage objects for non-empty to-many relationships.
 
 A "linkage object" is an object that identifies an individual related resource.
-It **MUST** contain `type` and `id` members.
+It **MUST** contain at least the following members:
+
+* `"id"`
+* `"type"`
+
+A linkage object **MAY** contain additional members. If defined, 
+they represent "[relationship attributes]" and may contain any valid JSON value. 
+[Relationship attributes] **MUST** abide by the same constraints as 
+[resource attributes].
+Complex [relationship attributes] **MUST** abide by the same constraints as 
+[complex resource attributes].
 
 > Note: Resource linkage in a compound document allows a client to link
 together all of the included resource objects without having to `GET` any
@@ -377,6 +387,23 @@ A complete example document with multiple included relationships:
           { "type": "comments", "id": "5" },
           { "type": "comments", "id": "12" }
         ]
+      },
+      "publications": {
+        "self": "http://example.com/articles/1/links/publications",
+        "related": "http://example.com/articles/1/publications",
+        "linkage": [{
+          "type": "websites",
+          "id": "6",
+          "publication_type": "original"
+        }, {
+          "type": "magazines",
+          "id": "9",
+          "publication_type": "reprint"
+        }, {
+          "type": "websites",
+          "id": "29",
+          "publication_type": "excerpts"
+        }]
       }
     }
   }],
@@ -402,6 +429,27 @@ A complete example document with multiple included relationships:
     "body": "I like XML better",
     "links": {
       "self": "http://example.com/comments/12"
+    }
+  }, {
+    "type": "websites",
+    "id": "6",
+    "hostname": "the-art-of-bike-shedding.com",
+    "links": {
+      "self": "http://example.com/websites/6"
+    }
+  }, {
+    "type": "magazines",
+    "id": "9",
+    "title": "Bike Shed Reviews",
+    "links": {
+      "self": "http://example.com/magazines/9"
+    }
+  }, {
+    "type": "websites",
+    "id": "29",
+    "hostname": "silver-bullet-news.com",
+    "links": {
+      "self": "http://example.com/websites/29"
     }
   }]
 }
@@ -929,6 +977,12 @@ Accept: application/vnd.api+json
     "links": {
       "photographer": {
         "linkage": { "type": "people", "id": "9" }
+      },
+      "staff": {
+        "linkage": [
+          { "type": "people", "id": "11", "function": "gaffer" },
+          { "type": "people", "id": "12", "function": "make_up_artist" }
+        ]
       }
     }
   }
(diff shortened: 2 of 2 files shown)


dgeb:
@hhware Thanks for working up this PR in time for consideration.

I personally don't agree with this blurring of lines between resources and relationships. In my opinion, this adds a lot of complexity for little benefit. If relationships are allowed to have attributes, then the spec will need to provide a means to edit those attributes (PATCH I suppose). Furthermore, it will create dissonance for API designers who need to decide between creating resources vs. "complex relationships".

dgeb:
Moving to 1.0 milestone.

@wycats @steveklabnik are either of you in favor?

hhware:
@dgeb, thanks for reviewing it!

> then the spec will need to provide a means to edit those attributes (PATCH I suppose)

It is addressed by this PR

> I personally don't agree with this blurring of lines between resources and relationships. In my opinion, this adds a lot of complexity for little benefit... Furthermore, it will create dissonance for API designers who need to decide between creating resources vs. "complex relationships".

IMHO, this "blurring" can be viewed as an intrinsic feature of some real-life systems, which have objects which are "in between" resources and links, but much closer to links. As for the "dissonance", it could indeed be a dissonance to some, but longed-for flexibility to others. There is always a balance between rigidity and flexibility, and I think it would in the spirit of JSON API to be more on a rigid side of things, but not entirely rigid. In my view, allowing simple attributes of relationships is 90% of rigidity and 10% of flexibility.

Let me elaborate more with two examples.

First one extends the filesystem example from #415. Imagine a CMS-like system (this is the case which I personally care about), consisting of a tree of containers, which can contain either some objects or other containers. A container with all its content can be referred to from another container (to allow people share content without copying it). The difference from the filesystem example is that there is more than one types of such references (to allow different approach to sharing, either with read-only rights, with some admin rights, etc.) Surely, each of this types of sharing links can be represented as an object of its own, necessitating several types of such objects and complexity associated with that, but it would be very convenient for the API designer to _hide_ this complexity and simplify life for himself and for the API consumers by _abstracting_ this as a relationship with just one attribute (type of the link). I wish JSON API allowed to do that.

Second one is some modification of audit-like example from #415. Suppose some endpoints of an API represent some committees, and other endpoints of that API represent people, who are members of these committees. People move from one committee to another over time, which can be reflected by `joined_at` / `left_at` timestamp attributes of the relationships. It would be easy to represent the history of each committee over time this way, showing when each member joined and when she left. I think it is a perfect abstraction _if_ the API is only concerned with such superficial representation of the situation. If the API needs to deal with the particulars of the processes of joining / leaving of members, the API designer could opt for resources for them, which would allow him to represent details of running for membership, voting for it, etc. Is it a dissonance? I would say it is a reasonable freedom of expression, and I wish JSON API allowed it.

Last but not least, `meta` of relationships can be very useful for reporting computed data about them, e.g., some statistics, which can only be returned by GET, but is never stored by POST / PUT / PATCH.

tkellen:
@hhware Would you be willing to mock up a set of example requests/payloads showing how you'd accomplish your desired interaction within the constraints of json-api now, and a companion set within the constraints you are proposing?

I think it would be very helpful to see them side by side.

tkellen:
This change is additive and can be considered post 1.0. I'm going to pull it from the milestone.

hhware:
@tkellen Definitely, I will.

tkellen:
@hhware Are you still willing to provide examples for this? If so, is there any chance you could do so this week? We are planning on doing a group review in the very near term and I would love to have them as a reference so we can consider this carefully.

hhware:
> Are you still willing to provide examples for this?

@tkellen, yes, I have been working on this PR and examples for the last few weeks, time permitting. I guess it's finally time to let it go :)

@dgeb, @tkellen, I very much appreciate your time and interest in reviewing my earlier posts on this topic. I reworked and significantly simplified this proposal to account for the discussion.

@lsylvester, I would appreciate your opinion on whether this satisfies your expectations wrt attributes of relationships.

#### Structure of this PR

This PR is split into several commits for convenience of reviewers:
1. the _substance_ of the proposal, 
2. _terminology_ updates and formatting changes,
3. _optional_ improvements (sparse fieldsets, sorting, `meta`).

The 4th commit is the example, it can be viewed in readable form [here](https://rawgit.com/hhware/json-api/i415-p431-example/i415-p431.html).

#### Key points

I think the PR now articulates the goals and limitations of the proposed changes much better than before. In particular, I would like to emphasize the following points:
1. The line between resources and relationships is much less blurred in this version compared to the previous one.
2. It was targeted at to-many relationships before, now it includes to-one relationships because it is more consistent. It does not seem to conflict with anything: e.g., if an API backed by RDBMS wants to provide to-one relationship with attributes, it can store them in a join table with appropriate UNIQUE indices.
3. The main difference between resources and "complex relationships" is that while resources can have their own types and can be referred to by `type`/`id`, relationships do not have their own `type`/`id` and cannot be referred to by them. I believe this is the reason why there should be no "dissonance": relationships can never compete with resources because they are only auxiliary entities, exactly the same as in the current spec.
4. Another difference between resources and relationships is that the latter cannot have `links`.
5. Answering the [comment](https://github.com/json-api/json-api/pull/431#issuecomment-77752602) of @dgeb: there is simply _no need_ to provide methods for updating attributes of relationships: if such attribute has to be changed, the corresponding relationship can be deleted and created from scratch, just like it is done in the current spec.
6. The only similarity between resources and relationships is that both allow arbitrary attribute members (and `meta`, in the optional part).
7. The spec is updated so that the word "attribute" is never used alone, it is always either "resource attribute" or "relationship attribute" (primarily in the terminology part).
8. Not sure whether there is any practical use in allowing complex attributes of relationships, but I think it is more consistent to keep limitations on attributes of resources and attributes of relationships identical.
9. Suggestions are made on incorporating relationship attributes into sparse fieldsets and sort requests. Suggestion on sorting  could have its merits even without relationship attributes. I will open a new issue on it if it does not go through as part of this PR.
10. For the purposes of pushing this PR forward, I tried to update as little as necessary. In particular, examples became less convincing after my changes, they should probably be replaced altogether.
11. In this PR, most of `GET` examples do not illustrate attributes of relationships, but the "complete example document with multiple included relationships" does. To the contrary, a good portion of `POST`/`PATCH` examples do illustrate them. The primary reason for this discrepancy is that there is no "complete example document" for `POST`/`PATCH`. 
12. I expect that one common use case of this would be ordered relationships (touched in #473). In RDBMS setting, a to-many relationship can be stored with an additional column, e.g., `order_number INT`. Relationships in the spec have semantics of sets, so one way of addressing this question could be support of _totally ordered sets_, but then a mechanism to distinguish sets from totally ordered sets would be required. This will likely bring a lot of confusion since in most cases relationships are not ordered. Support for relationship attributes removes the issue: order can be explicitly communicated and sorting by it can be supported.

#### Milestoning

I am well aware that the official position is to postpone additive changes until post 1.0. While some additive changes are mere optimizations with available workarounds (e.g., #477), I think it would be appropriate to consider this one for 1.0 because it modifies one of the core concepts of the spec. If relationship attributes were _proposed post 1.0_ and then accepted, then obviosuly implementors of libraries would have to deal with them as with additive change. As @steveklabnik [put](https://github.com/json-api/json-api/issues/365#issuecomment-77287668): "I can appreciate frustration with changes, we are making them _now_, so that we can not make them in the future" -- I think it absolutely applies to the particular change proposed here.

Another reason to consider this now is that this change is additive for the JSON API spec and its generic implementations, but it will _not be additive_ for specific JSON-API-compliant APIs which could make use of this feature. They will rarely be able to afford a luxury of switching from relationships-as-entities to relationships-with-attributes because it is a backward-incompatible change for an API. Implementors of such APIs make choices on API design and then stick to them. Even if this feature becomes availabe in JSON API 1.X, APIs implemented before that would not be able just to "add" it, the way the spec and libraries can. Obviously, some people are waiting for 1.0 to start using JSON API or to consider it "for real" because of stability expectations --  it is not unreasonable to expect that after 2 years of preparations for 1.0, the core concepts of the spec will remain stable for a while.

Commit "substance" is what I believe has to be considered for 1.0. Commit "terminology" would be useful in 1.0 too, but the main reason I provide it here is to avoid rejection of the proposal because of misunderstandings of terminology (the commits are non-overlapping though, so some terminology changes have leaked into other commits as well). Commit "optional" is not blocking for 1.0, but provided here to complete the picture.

#### Examples

Two alternatives to relationship attributes have been suggested in discussions in #415 and here:
1. make relationships heterogeneous, with value of `type` incorporating the attribute(s),
2. disguise relationships as resources.

After having a closer look, I realized that I do not even understand how the first option can be implemented.  But even if it could be, I think it is not a workable solution for the following reasons:
1. it pollutes type system with multitude of technical entities, creating complexity out of nowhere,
2. it is impractical for relationships with multiple attributes -- number of types will grow in geometrical progression,
3. it is impractical for relationships attributes which can be represented as ENUMs with large number of members for the same reason,
4. it is inapplicable to relationship attributes which cannot be represented as ENUMs, e.g., which have numeric values.

So I skip the first option and compare the second one to the proposed solution [here](https://rawgit.com/hhware/json-api/i415-p431-example/i415-p431.html). If these examples are unclear or need improvement, please let me know, so I could update them.

For completeness sake, other alternatives also exists: define an extension or simply handle these attributes within `meta`. However, these alternatives produce less portable APIs, which defeats the purpose of following JSON API spec.

lsylvester:
@hhware For my purposes, I wanted to embed information about the linked resource without having to include the linked resource, so the information wasn't particular to the association. 

The reason for this was to be able to invalidated locally cached resources.

I think for my use case, the best approach would be to allow for `meta` in the linkage objects, which is included in this PR, but I do not require the relation attributes which are the main feature of this PR.

hhware:
Made a minor update, left a note in the message of the corresponding commit (to preserve PR structure).

steveklabnik:
>  In my opinion, this adds a lot of complexity for little benefit. 

Ultimately, this is my thought as well. I've read through these tickets (sorry for the delay, work is getting tough) but I don't understand why relationships need attributes. As soon as they do, they become resources in their own right, and don't need special casing.

dgeb:
We sincerely appreciate your effort here, @hhware. However, after consideration and discussion, @tkellen, @steveklabnik, and myself share the belief that relationships with attributes should become resources in their own right. This has been a tough call as we struggle to balance complexity and utility.

An edge case we should consider is the allowance of `meta` on linkage objects, as you and @lsylvester have mentioned.

Thanks again for all your work, @hhware.

ethanresnick:
/reffing #794, as the extension system should make something like this possible

PR #434, opened 2015-03-08
convert non-normative note into constraint

tkellen:
This non-normative note included RFC2119 keywords, making its
status as a note ambiguous. Per our previous discussion, json-api
should not disallow implementors from including hasOne fkeys as
attributes, but it should require that any relations which are
provided must be represented under the links object.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -136,18 +136,18 @@ the client and represents a new resource to be created on the server.
 
 In addition, a resource object **MAY** contain any of these top-level members:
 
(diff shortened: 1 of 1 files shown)


dgeb:
:+1:

bintoro:
@tkellen There's one occurrence of `*SHOULD NOT*` instead of `**SHOULD NOT**`.

tkellen:
Thanks for catching that @bintoro, will push a fix after we land #417.

PR #435, opened 2015-03-09
Welcome, @dgeb! :heart:

steveklabnik:
Dan has had commit for a long time, but we never announced it. Let's fix that.


dgeb:
Hey, thanks @steveklabnik <3

PR #437, opened 2015-03-09
Merge relationship bullet points.

dgeb:
This should not change the meaning, but hopefully clarifies the
commonalities of `type` and `id` members for both to-one and
homogeneous to-many relationships.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -261,10 +261,12 @@ one of the following:
 * A `related` member, whose value is a related resource URL (as defined above).
 * Linkage to other resources based on their identifying `id` and `type` members
   ("resource linkage"). Linkage **MUST** be expressed as:
-  * `type` and `id` members for to-one relationships. `type` is not required
-    if the value of `id` is `null`.
-  * `type` and `id` members for homogeneous to-many relationships. `type` is
-    not required if the value of `id` is an empty array (`[]`).
+  * `type` and `id` members for to-one and homogeneous to-many relationships.
+    For a to-one relationship, the value of `id` **MUST** be either a string ID
+    or `null` to indicate an empty relationship. For a to-many relationship,
+    the value of `id` **MUST** be either an array of string IDs or an empty
+    array (`[]`) to indicate an empty relationship. If a relationship is empty,
+    `type` **MAY** be omitted.
   * A `data` member whose value is an array of objects each containing `type`
     and `id` members for heterogeneous to-many relationships.
 * A `"meta"` member that contains non-standard meta-information about the


hhware:
@dgeb I hope it would be appropriate to ask this question here: the current wording is such that the `data` option is only available for heterogeneous relationships, although homogeneous relationships are obviously a subset of heterogeneous ones. Would it make sense to explicitly state that the `data` option is available for both homogeneous and heterogeneous relationships?

bintoro:
The _Fetching Relationships_ section needs this too. One option to consider is having a small separate section explaining how to represent nonempty relationships, and then referring to it where needed. It would help keep the object definitions short.

hhware:
The same probably applies to creating and updating...

Also, there are some examples here: 
https://github.com/hhware/json-api/commit/00d8de1b70f6d264993f1c610f0707c5e35a31b3#diff-26e38355b69056b0db1c2b1c612241b2R1383
https://github.com/hhware/json-api/commit/00d8de1b70f6d264993f1c610f0707c5e35a31b3#diff-26e38355b69056b0db1c2b1c612241b2R1326
If relationship attribute is removed from them, would such examples be useful in the spec?

format/index.md
```diff
@@ -261,10 +261,12 @@ one of the following:
...
   ("resource linkage"). Linkage **MUST** be expressed as:
-  * `type` and `id` members for to-one relationships. `type` is not required
-    if the value of `id` is `null`.
-  * `type` and `id` members for homogeneous to-many relationships. `type` is
-    not required if the value of `id` is an empty array (`[]`).
+  * `type` and `id` members for to-one and homogeneous to-many relationships.
+    For a to-one relationship, the value of `id` **MUST** be either a string ID
+    or `null` to indicate an empty relationship. For a to-many relationship,
+    the value of `id` **MUST** be either an array of string IDs or an empty
+    array (`[]`) to indicate an empty relationship. `type` is not required if
```
tkellen: If a relationship is empty, the `type` member **MAY** be omitted?

tkellen:
:+1:

format/index.md
dgeb: Yeah, that's good.

dgeb:
> Would it make sense to explicitly state that the data option is available for both homogeneous and heterogeneous relationships?

I think this might muddy the waters. If servers could return either form, then clients would be missing a clear indicator for which form to send (and which type of relationship it is).

dgeb:
I agree we may need to flesh out a few "empty relationship" cases in the fetching and updating relationships sections.

hhware:
> I think this might muddy the waters. If servers could return either form, then clients would be missing a clear indicator for which form to send (and which type of relationship it is).

Can these two forms (`type`/`id` and `data`) be viewed as two possible representations for to-many relationships, letting the API creator to decide on a preferred form? Perhaps I am missing the point, but what is the benefit of using homo-/heterogeneity as a criterium to decide between the two forms?

Is the client expected to deduce the type of the relationship from the form being used in the response? In the absence of schema, I would think that the contract is specified elsewhere, most likely in the API docs. What is wrong with providing this "clear indicator" in the API docs then, e.g., to specify that one endpoint is sending/accepting one form and another endpoint is sending/accepting another form, or that all endpoints are accepting only one form, or that any endpoint is accepting either form? Why does JSON API have to be rigid here?

Another consideration has occurred to me. Let's assume for a moment that the proposal for relationship attributes (#415) is accepted in the future. The current spec is saying that the link object _MUST_ be of `type` / `id` form for homogeneous relationships and of `data` form for heterogeneous relationships. The only form which allows attributes is the `data` form. However, if relationship attributes were allowed, they would make equal sense for both homogeneous and heterogeneous types, it would be weird to allow them only for the latter. However, in my understanding, this MUST cannot be overridden once 1.0 is out. Does it mean then that #415 will become impossible (or crippled and therefore impractical) and should be decided upon before 1.0? If so, I can try to work out the mockups requested by @tkellen quicker (it would be helpful to know the deadline...)

I suggest to lift this restriction and make a change to the linkage section along these lines:
- Linkage to other resources based on their identifying `id` and `type` members
  ("resource linkage"). Linkage **MUST** be expressed as:
  - two forms are allowed for to-many relationships:
    - `type` and `id` members. `type` is not required if the value of `id` is an empty array (`[]`).
    - `data` member whose value is an array of objects each containing `type` and `id` members.
      `type` is not required if the value of `id` is `null`.

Yet another alternative, less fancy but more robust/consistent (also eliminates the ambuguity of `id` being string or array):
- Linkage to other resources based on their identifying `id` and `type` members
  ("resource linkage"). Linkage **MUST** be expressed as:
  - `type` and `id` members for to-one relationships. 
  - `data` member whose value is an array of objects each containing `type` and `id` members, 
    for to-many relationships. 
  
  `type` is not required if the value of `id` is `null`.

bintoro:
Allow me to suggest:

Homogeneity can be signaled by whether `type` appears at the link object's top level.

It would be fine to use `data` for homogeneous relationships, as long as the individual items don't include a `type` member.

These would be equivalent:

```
"things": {
  "type": "things",
  "id": ["1", "2", ...]
}
```

```
"things": {
  "type": "things",
  "data": [{"id": "1"}, {"id": "2"}, ...]
}
```

dgeb:
So the alternative that I considered when working on RC2 was to treat all resource linkage as (potentially) heterogeneous and nest it all under a `linkage` key (not `data`). 

The example on the home page would be reworked as follows:

```
 ...
  "data": [{
    "type": "posts",
    "id": "1",
    "links": {
      "author": {
        "linkage": {"type": "people", "id": "9"}
      },
      "comments": {
        "linkage": [
           {"type": "comments", "id": "5"},
           {"type": "comments", "id": "12"}
        ]
      }
    }
  }]
  ...
```

This simplifies the rules for fetching and processing linkage data by eliminating the homogeneous / heterogeneous distinction. It also nests linkage data consistently under a single key for both to-one and to-many relationships.

However, it is _very verbose_, so much so that I worried it would hurt adoption of JSON API.

tkellen:
As someone who never uses polymorphic relations, my initial reaction to this wasn't positive... but taken from the standpoint that we are putting homogeneous and heterogeneous collections on a level playing field, this makes a ton of sense.

I think @dgeb's suggestion is the ideal case both from an implementation and consumption standpoint. The loss of succinctness feels minor compared to the internal consistency it affords.

bintoro:
@dgeb Interesting.

Verbose? Yes. More elegant than `data`? Yes!

I just have slight reservations about repeating the shared `type` for homogeneous collections. Not because of verbosity, but because there's probably at least some value in being able to look at a single attribute and get certainty about what's in the collection.

That said, I definitely prefer this idea over the current situation. It would also get rid of the annoying incongruity between link objects and responses from relationship URLs. (I suppose relationship URLs would simply return the contents of `linkage` as primary data.)

dgeb:
> I just have slight reservations about repeating the shared type for homogeneous collections. Not because of verbosity, but because there's probably at least some value in being able to look at a single attribute and get certainty about what's in the collection.

Yeah, this is a concept lacking for primary resource collections as well, so this really just makes relations consistent. 

> I suppose relationship URLs would simply return the contents of linkage as primary data

Yes.

> I definitely prefer this idea over the current situation.

I will work up a PR so that everyone can see the before / after, since this approach has generated  interest and it seems a shame to go to 1.0 before fully giving it a chance.

dgeb:
Closing in favor of #447

PR #439, opened 2015-03-09
Condense relationship example

bintoro:
- Removes an example payload that seems to just draw out the _Resource Relationships_ section without adding value (previously part of #425). Specifically, this case hardly needs to be illustrated:
  
  ```
  "comments": {
  }
  ```
- Replaces the slightly puzzling statement
  
  > allows the client to change the related author without deleting the `people` object
  
  with a more general one ("[...] the related author directly").
- Spells out _related resource URL_.
- Removes a mention about "current compound document", since resource linkage may or may not refer to an included resource.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -297,42 +297,19 @@ For example, the following article is associated with an `author` and `comments`
       "type": "people",
       "id": "9"
     },
-    "comments": {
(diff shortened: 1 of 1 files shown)


dgeb:
:+1:

tkellen:
:+1:

Issue #271, opened 2014-09-04
No guidance for use with json-schema


HelloGrayson:
I'm guessing that jsonapi should have a position on whether json-schema should or could be used alongside of. 

Thoughts?

Why - I'm working on a multi-language SOA where there is a benefit in knowing what makes up a given "type".

steveklabnik:
JSON-Schema can never totally describe JSON API, but that doesn't mean it's not helpful at times.

HelloGrayson:
@steveklabnik I'm not talking about using json-schema to describe json-api. I'm talking about guidelines for those who want to use json-schema WITH json-api. Make sense?

xiaohanyu:
+1 for this

eneuhauser:
**Updated:** Moved this to a [pull request](https://github.com/json-api/json-api/pull/440)

I am currently documenting my API that is trying to conform to JSON API using RAML and JSON-Schema. Unfortunately, I am new to JSON API, RAML, and JSON-Schema. I've been following JSON API for some time and just now had an opportunity to start exploring it. I am even newer to RAML and JSON-Schema, but find it a very powerful way to express an API.

As I built out my API, I found it useful to express JSON API in JSON Schema notation. I'm sure I may have misrepresented some things, but I believe [this schema](https://github.com/eneuhauser/json-api/blob/schema/schema) expresses JSON API very concisely and would be a good intro to someone new to JSON API.

Additionally, I feel seeing it in JSON Schema helps solidify some concepts. One question I had while writing the schema is if an `error` response could have a `links` block?

kjwierenga:
Although json-schema can never describe JSON API exhaustively there is merit in having a json-schema that covers as much of the spec as possible. I'm currently using an updated version of @eneuhauser schema to validate JSON API responses using rspec

In `spec/rails_helper.rb`

```
  config.include JSON::SchemaMatchers
  config.json_schemas[:json_api] = 'spec/schema/json-api-schema.json'
```

In my spec:

```
      expect(response.body).to match_json_schema(:json_api)
```

The logical place to host this schema would be at `http://jsonapi.org/schema`. Is there an intention to host this schema within the json-api project?

tkellen:
Closing in favor of the PR. I'm working on house-cleaning so we can focus on stuff like this!

PR #440, opened 2015-03-10
Adding a detailed json-schema

eneuhauser:
I saw that you just removed the json-schema, but I believe this schema is a good representation of JSON API and concisely explains the format.

### DIFF
--- a/faq/index.md
+++ b/faq/index.md
@@ -66,7 +66,14 @@ alternative ways of requesting certain operations (e.g., resource creation via
 
 ## Is there a JSON Schema describing JSON API? <a href="#is-there-a-json-schema-describing-json-api" id="is-there-a-json-schema-describing-json-api" class="headerlink"></a>
 
-Not currently, no. JSON Schema cannot fully represent the semantics of JSON API, and so any such schema would be partial anyway.
+Yes, you can find the JSON Schema definition at
+[http://jsonapi.org/schema](http://jsonapi.org/schema). This schema is as
+restrictive as possible, but has flexibility to be extended within your
+documentation. Validation will not yield false negatives, but could yield false
+positives for the sake of flexibility.
+
+You can find more information about the JSON Schema format at
+[http://json-schema.org](http://json-schema.org).
 
 ## Why are resource collections returned as arrays instead of sets keyed by ID? <a href="#resource-collections-returned-as-arrays" id="resource-collections-returned-as-arrays" class="headerlink"></a>
 
--- a/schema
+++ b/schema
@@ -1,0 +1,357 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "title": "JSON API Schema",
+  "description": "This is a schema for responses in the JSON API format. For more, see http://jsonapi.org",
+  "oneOf": [
+    {
+      "$ref": "#/definitions/success"
+    },
+    {
+      "$ref": "#/definitions/failure"
+    },
+    {
+      "$ref": "#/definitions/info"
+    }
+  ],
+  
+  "definitions": {
+    "success": {
+      "type": "object",
+      "required": [
+        "data"
+      ],
+      "properties": {
+        "data": {
+          "$ref": "#/definitions/data"
+        },
+        "included": {
+          "description": "To reduce the number of HTTP requests, servers **MAY** allow responses that include related resources along with the requested primary resources. Such responses are called \"compound documents\".",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/resource"
+          },
+          "uniqueItems": true
+        },
+        "meta": {
+          "$ref": "#/definitions/meta"
+        },
+        "links": {
+          "description": "Link members related to the primary data.",
+          "allOf": [
+            {
+              "$ref": "#/definitions/links"
+            },
+            {
+              "$ref": "#/definitions/pagination"
+            }
+          ]
+        },
+        "jsonapi": {
+          "$ref": "#/definitions/jsonapi"
+        }
+      },
+      "additionalProperties": false
+    },
+    "failure": {
+      "type": "object",
+      "required": [
+        "errors"
+      ],
+      "properties": {
+        "errors": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/error"
+          },
+          "uniqueItems": true
+        },
+        "meta": {
+          "$ref": "#/definitions/meta"
+        },
+        "jsonapi": {
+          "$ref": "#/definitions/jsonapi"
+        }
+      },
+      "additionalProperties": false
+    },
+    "info": {
+      "type": "object",
+      "required": [
+        "meta"
+      ],
+      "properties": {
+        "meta": {
+          "$ref": "#/definitions/meta"
+        },
+        "links": {
+          "$ref": "#/definitions/links"
+        },
+        "jsonapi": {
+          "$ref": "#/definitions/jsonapi"
+        }
+      },
+      "additionalProperties": false
+    },
+    
+    "meta": {
+      "description": "Non-standard meta-information that can not be represented as an attribute or relationship.",
+      "type": "object",
+      "additionalProperties": true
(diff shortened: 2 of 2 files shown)


tkellen:
Thanks for this @eneuhauser! I'm going to hold off on merging until we have marked 1.0.

amaltson:
With 1.0 released, will this be considered for merging? This would help when documenting APIs with RAML.

tkellen:
Yes! Can you squash this to a single commit please?

schema
```diff
@@ -0,0 +1,243 @@
...
+          "description": "A `self` member, whose value is a URL for the relationship itself (a \"relationship URL\"). This URL allows the client to directly manipulate the relationship. For example, it would allow a client to remove an `author` from an `article` without deleting the `people` resource itself.",
+          "type": "string",
+          "format": "uri"
+        },
+        "related": {
+          "description": "A URL for the related resource(s) (a \"related resource URL\"). When fetched, it returns the related resource object(s) as the response's primary data. For example, an `article`'s `comments` relationship could specify a URL that returns a list of comment resource objects when retrieved through a `GET` request.",
+          "type": "string",
+          "format": "uri"
+        },
+        "linkage": {
```
ethanresnick: This should be `data`

schema
```diff
@@ -0,0 +1,243 @@
...
```
ethanresnick: ditto

schema
```diff
@@ -0,0 +1,243 @@
...
+      "description": "\"Resource objects\" appear in a JSON API document to represent primary data and linked resources.",
+      "type": "object",
+      "required": [ "type", "id" ],
+      "properties": {
+        "type": { "type": "string" },
+        "id": { "type": "string" },
+        "links": { "$ref": "#/definitions/links" },
+        "meta": { "$ref": "#/definitions/meta" }
+      },
+      "additionalProperties": true
```
ethanresnick: There should be attributes and relationships definitions here, I think, and then additionalProperties should be false

schema
```diff
@@ -0,0 +1,243 @@
...
+          "uniqueItems": true
+        },
+        "paths": {
+          "description": "An array of JSON Pointers to the relevant attribute(s) within the associated resource(s) in the request document. Each path **MUST** be relative to the resource path(s) expressed in the error object's \"links\" member [e.g. [\"/first-name\", \"/last-name\"] to reference a couple attributes].",
+          "type": "array",
+          "items": { "type": "string" },
+          "uniqueItems": true
+        }
+      },
+      "additionalProperties": true
```
ethanresnick: additionalProperties should be false here too, though meta is now allowed

slaskis:
I've been using this schema and updated my version of it now to pass with 1.0 (including your comments @ethanresnick).

Since you did most of the work feel free to update this PR with my gist: https://gist.github.com/slaskis/f4f32beeea0080320e00

eneuhauser:
@tkellen I've updated the schema to match the new 1.0 format and squashed into a single commit. I validated the schema complies with the example on the homepage using [JSON Schema Lint](http://jsonschemalint.com/draft4/#).

@slaskis, thanks for your help. I used your changes to start out, but I did make some modifications based on the current format of the spec. Please let me know if you see any issues.

slaskis:
@eneuhauser happy to help! 

I was a bit lazy with the descriptions but I see you took care of that :+1:

schema
```diff
@@ -0,0 +1,357 @@
...
+          "$ref": "#/definitions/resource"
+        },
+        {
+          "description": "An array of resource objects, an array of resource identifier objects, or an empty array ([]), for requests that target resource collections.",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/resource"
+          },
+          "uniqueItems": true
+        }
```
slaskis, line 114: Perhaps this could be moved into something like "#/definitions/collection"? Not that it's re-used right now but just for clarity. If that's even the correct term, but "an array of resource objects" doesn't really fit in a nice and tidy string.

schema
```diff
@@ -0,0 +1,357 @@
...
+					}
+				}
+			]
+		},
+
+    "attributes": {
+			"description": "Members of the attributes object (\"attributes\") represent information about the resource object in which it's defined.",
+      "type": "object",
+			"patternProperties": {
+				"^\\w[\\w-_]*$": {
```
slaskis: Is it too strict to filter out the [reserved member names](http://jsonapi.org/format/#document-resource-object-attributes) here? That is adding `relationships` and `links` to the regex.

eneuhauser: I had originally tried this, but felt it was unnecessary; however, I never turn down a regex challenge, so I came up with the following: `^(?!relationships$)(?!links$)\\w[\\w-_]*$`.

slaskis: Nice, regex golf time! What about merging those into `^(?!relationships|links$)\\w[\\w-_]*$`? ;)

schema
eneuhauser, line 114: I don't know about this one. You wouldn't create a new object to represent an array of resources, so I don't think it needs its own definition.

slaskis, line 114: Very good points, ignore my ramblings please :)

faq/index.md
```diff
@@ -66,7 +66,14 @@ alternative ways of requesting certain operations (e.g., resource creation via
 
 ## Is there a JSON Schema describing JSON API? <a href="#is-there-a-json-schema-describing-json-api" id="is-there-a-json-schema-describing-json-api" class="headerlink"></a>
 
-Not currently, no. JSON Schema cannot fully represent the semantics of JSON API, and so any such schema would be partial anyway.
+Yes, you can find the JSON Schema definition at
+[http://jsonapi.org/schema](http://jsonapi.org/schema). Please note that this
+schema is not a perfect document. Just because a JSON document may validate
+against this schema, that does not necessarily mean it is a valid JSON API
+document. The schema is provided for a base level sanity check.
```
tedepstein: Following this PR, and the preceding issue #271, I agree that a schema is helpful.  If it doesn't validate all of the semantics of JSON API, then I would say it's like most schemas in that respect.  It can yield some "false positives," where a given JSON input passes schema validation, but breaks some JSON API rules. 

Still, I assume that whatever this schema does validate is in fact a requirement of the JSON API spec.  So it should not yield any "false negatives."  If a given input fails validation against this schema, then it's definitely not valid JSON API.

That's an important guarantee, and one worth stating in the FAQ, I think.  Rather than saying it's "not a perfect document," say something like "Please note that this schema only enforces a subset of the JSON API specification."

eneuhauser: This wording was taken from the previous JSON Schema. The previous schema was a lot more simplistic, so your suggestions are more appropriate for this version of the schema.

I reused the wording because I didn't want to get into the politics of how it should read, but I can take a crack at updating the copy match what this schema does and doesn't do.

tkellen: Would you mind doing that @eneuhauser and pinging back when ready? I'd love to get this merged but it would be nice for it to be a bit more clear.

eneuhauser:
@tkellen, I've updated the FAQ with the spirit of the schema definition. I also cleaned up some whitespace issues with the schema if you notice differences there. Please review and let me know if it is sufficient. Thanks!

tkellen:
thanks @eneuhauser!

cc @dgeb @ethanresnick @steveklabnik

dgeb:
I haven't reviewed the schema definition in detail, but it looks like a great start.

I suggest we merge and allow for refinement (if needed).

Thanks so much for your work here @eneuhauser!

bf4:
Just want to confirm, unless I'm not seeing it, that this schema was manually generated and is not as authoritative as the `/format` page? e.g. it includes `info` which isn't there.

ethanresnick:
> Just want to confirm, unless I'm not seeing it, that this schema was manually generated and is not as authoritative as the /format page

Correct. The current schema is non-normative and has a few known issues (see #867)

bf4:
ref multiple schemas PR for discussion https://github.com/json-api/json-api/pull/1127

PR #442, opened 2015-03-10
Updates Markdown code blocks with correct types

jfinkels:
Code blocks for HTTP requests should be of the following form.

``````
```http
GET /articles/1
```
``````

Code blocks for JSON data should be of the following form.

``````
```json
{"key": "value"}
```
``````

### DIFF
--- a/extensions/index.md
+++ b/extensions/index.md
@@ -88,7 +88,7 @@ GET http://api.example.com/posts
 
 That document will de-reference to explain your link relations:
 
-```text
+```http
 GET http://api.example.com/profile HTTP/1.1
 ```
 
--- a/format/index.md
+++ b/format/index.md
@@ -480,19 +480,19 @@ A server **MUST** support fetching resource data for every URL provided as:
 
 For example, the following request fetches a collection of articles:
 
-```text
+```http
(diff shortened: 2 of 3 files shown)


bintoro:
I've been wanting to get rid of `javascript` too.

But some of the payloads contain comments and/or ellipses, and these will be highlighted as invalid syntax if marked `json`. That's the reason for `javascript`. I'm sure this could be addressed through CSS, though.

There are quite a few blocks that contain both HTTP and JSON, and this PR turns them all into `http`. However, the highlighter seems to be smart enough to recognize an HTTP response header and a JSON payload inside a `text` block and format them accordingly. Makes no sense to turn those into `http`.

The magic doesn't apply to request examples, but I think I still prefer `text` — then, at least the code is uniformly non-formatted.

tkellen:
Thanks for taking the time to submit this PR @jfinkels. Would you mind re-basing this and updating to only use `http` in examples where the only content of the block is http?

jfinkels:
Okay, now the pull request only changes HTTP request blocks.

tkellen:
Thanks @jfinkels!

PR #443, opened 2015-03-10
correct invalid refs to `resource` links (should be `related`)

tkellen:


dgeb:
Good catch @tkellen :+1:

PR #444, opened 2015-03-10
add MUST to include clause

tkellen:


tkellen:
Assuming you agree this makes sense @dgeb, I'll update our clauses for `fields` and `sort` as well.

dgeb:
>  I'll update our clauses for fields and sort as well.

Sounds good - thanks :)

Issue #347, opened 2015-02-17
Are embedded objects allowed for fields?


counterleft:
Hey @dgeb and friends,

Given #341, and

> Any other member in a resource object represents an "attribute", which may contain any valid JSON value.

what is the spec's stand on embedded json objects as attributes? For example, we have the models, "product", "product-option" and "price". A product has options and options have prices. A couch is a product. The black leather and brown fabric are the options. Each option is priced in USD and CAD. A price will never stand on its own; it always belongs to an option. An option always belongs to a product; an option isn't shared with other products.

Should the options and prices be modeled with `links` and `linked` or can they be embedded into the representation of the primary resource (`data`) since their lifecycle is tied to that resource?

The two different approaches shown here: https://gist.github.com/winterchord/58391f01dd0bc87944ac

csantero:
We support embedded json object attributes in [jsonapi.net](https://github.com/SphtKr/JSONAPI.NET). I feel like that approach makes more sense than coming up with surrogate ids for these objects and then extracting them into `links`.

dgeb:
I believe that objects are fine as attributes as long as they are wholly dependent upon the parent resource. This means that requests can never address the object directly, but can only modify it through the parent resource. Furthermore, wholly dependent objects should not be shared across resources.

If you feel tempted to add `type` and `id` fields to your object attributes, then they are probably resources themselves.

Last but not least: if you are in doubt, just create separate resources.

PR #446, opened 2015-03-10
clarify fields and attributes

tkellen:
this clearly denotes what constitutes a field/attribute and then
re-uses that reference when describing how sparse fieldsets
should operate.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -137,22 +137,6 @@ In addition, a resource object **MAY** contain any of these top-level members:
 * `"meta"`: non-standard meta-information about a resource that can not be
   represented as an attribute or relationship.
 
-A resource object **MAY** contain additional top-level members. These members
-represent "attributes" and may contain any valid JSON value.
-
-If the value of an attribute is a JSON object or array, the member is called a
-*complex attribute*. The value is allowed to be any valid JSON structure.
-However, a JSON object that constitutes or is contained in a complex attribute
-must reserve the `id`, `type`, `links`, and `meta` members for future use.
-
-A resource object's attributes and relationships are collectively called its
-"fields". <a href="#document-structure-resource-fields" id="document-structure-resource-fields"></a>
-
-Although has-one foreign keys (e.g. `author-id`) are often stored internally
-alongside other information to be represented in a resource object, these keys
-**SHOULD NOT** appear as attributes. If relations are provided, they **MUST**
-be represented under the "links object", as described below.
-
 Here's how an article (i.e. a resource of type "articles") might appear in a document:
 
 ```javascript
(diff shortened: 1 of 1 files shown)


tkellen:
This also provides an example of how the document could be inter-linked. Assuming ya'll like this, I'm going to put a manifest of terms and links at the bottom of the doc and link the hell out of everything.

bintoro:
I like the new subsections for _Attributes_ etc., but I think the special mention about `id`, `type`, and `self` in _Sparse Fieldsets_ would have been useful. It was a deliberate addition after all. A non-normative note would be enough, though.

bintoro:
Oh yeah and :+1: for the reference-style Markdown links. I was going to do the same thing. The spec could use more intra-doc linking, particularly where a term is used before it has been properly introduced.

bintoro:
@tkellen Sorry to bug you about this, but may I suggest simply:

> Note: Only _fields_ are affected. `type`, `id`, and `self` are included as normal.

tkellen:
No bugging, that's _much_ better.

tkellen:
Updated!

bintoro:
Yay!

One more thing: in _Complex Attributes_, the definition includes "[complex attribute]", but the actual link is `[complex attributes]` in plural, so the link doesn't format.

Other than that, this looks good to me.

format/index.md
```diff
@@ -173,6 +157,28 @@ Here's how an article (i.e. a resource of type "articles") might appear in a doc
 // ...
 ```
 
+#### Attributes <a href="#document-structure-resource-object-attributes" id="document-structure-resource-object-attributes"></a>
+
+A resource object **MAY** contain additional top-level members. These members
+represent "[attributes]" and may contain any valid JSON value.
+
+Although has-one foreign keys (e.g. `author_id`) are often stored internally
```
dgeb, line 165: Earlier, I changed this to match the dasherized recommendation - so `author-id`. Perhaps you're trying to represent how these fields might be stored internally in a database table?

tkellen, line 165: Precisely!

dgeb, line 165: Ok, then ... makes sense on careful reading :+1:

format/index.md
```diff
@@ -753,25 +758,21 @@ GET /articles/1?include=author,comments,comments.author
...
 
-A client **MAY** request that an endpoint return only specific
-[fields](#document-structure-resource-fields) in the response on a per-type basis
-by including a `fields[TYPE]` parameter. The value of the parameter is a
-comma-separated (U+002C COMMA, ",") list that refers to the names(s) of the
-fields to be returned.
+A client **MAY** request that an endpoint return only specific [fields] in the
+response on a per-type basis by including a `fields[TYPE]` parameter.
+
+> Note: Only [fields] are affected; `type`, `id`, and `self` are included as normal.
```
dgeb: since `self` links aren't required, then perhaps we should say "optionally `self` links" instead of just "`self`"?

dgeb:
Great job at consolidation and clarification here, @tkellen!

bintoro:
Hmm... I gotta say, I kinda liked the original definition of a complex attribute that specifically said it can be _anything_. This has not been entirely clear to users (e.g., #289, #347).

In case the rewording has to do with accommodating the Markdown link, it's possible to use reference-style links even if the text doesn't match the link title:

`If the value of an attribute is a JSON object or array, the member is called a [complex attribute][complex attributes]. The value is allowed to be any valid JSON structure.`

As for the note, I figured "as normal" gives it away, but might as well make it foolproof:

> `type`, `id`, and (optionally) `self` are included as normal

**EDIT:** I just don't want anyone to get an idea that they should drop `self` because `fields` is present, since there's no way for a client to request is specifically.

dgeb:
> `type`, `id`, and (optionally) `self` are included as normal

Yes, that works for me. It is an important note.

tkellen:
I've updated that note.  @bintoro can you suggest a rewriting of that entire section if you feel it is unclear? I think it is fine, and I'm not sure how to incorporate your suggestion in a sensical way.

bintoro:
@tkellen I thought it was just fine earlier when I noticed the link wasn't being formatted :smile:

```
If the value of an attribute is a JSON object or array, the member is called a
["complex attribute"][complex attributes]. The value is allowed to be any valid
JSON structure. However, an object that constitutes or is contained in a complex
attribute must reserve the `id`, `type`, `links`, and `meta` members for future
use.
```

The statement "is allowed to be any valid JSON structure" was a deliberate addition. The word "structure" was meant to imply that you can have nested objects too. If it doesn't accomplish this, I can reword.

But, if you think it's unnecessary to mention that at all, then the current revision is good to go as is!

tkellen:
Updated!  Can you take another look?

bintoro:
Thanks @tkellen! That should make it clear.

dgeb:
:+1:

PR #448, opened 2015-03-10
In examples move Ruby JsonApiClient under Clients

ms-ati:
It appears that the Ruby library `JsonApiClient` was erroneously listed under Server, when it should been under Client. Or am I mistaked?


steveklabnik:
Seems good, thanks!

PR #449, opened 2015-03-11
Provide conditions to allow for additive changes.

dgeb:
Implementations MUST ignore unspecified members encountered in objects.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -75,7 +75,11 @@ certain aspects are only applicable to one or the other. These differences are
 called out below.
 
 Unless otherwise noted, objects defined by this specification **MUST NOT**
-contain any additional members.
+contain any additional members. Client and server implementations **MUST**
+ignore object members not recognized by this specification.
+
+> Note: These conditions for object members allow this specification to
+evolve through additive changes.
 
 ### Top Level <a href="#document-structure-top-level" id="document-structure-top-level" class="headerlink"></a>
 


bintoro:
"unrecognized"?

dgeb:
@bintoro "unrecognized" is better - thanks. Updated.

tkellen:
I don't think it is meant to, but this wording could be read to say that you can't return an error when a payload includes some attribute that is invalid for the resource being created/updated (e.g. specifying a field that doesn't exist in the underlying database table). I personally think it is fine to be liberal in what you accept and all that, but this seems overly prescriptive.

dgeb:
So the spec still says:

> A resource object MAY contain additional top-level members. These members represent "attributes" and may contain any valid JSON value.

Therefore, all additional top-level members in a resource object are considered attributes, and it would be appropriate for a server to return an error if it doesn't recognize one of them.

I believe the clause in this PR just applies to all members of objects that the spec doesn't specifically address. Maybe it could be worded more clearly.

dgeb:
Maybe this is why I went with "unspecified" instead of "unrecognized", come to think of it.

bintoro:
"[...] MUST ignore object members not recognized by this specification"

EDIT: Or something like that... There probably isn't one word for this. "Unspecified" != "not defined by a specification".

dgeb:
> "Unspecified" != "not defined by a specification".

@bintoro True.

> "[...] MUST ignore object members not recognized by this specification"

This is an improvement. Updated again. And thanks again.

@tkellen does this now work for you?

tkellen:
I guess i'm still confused.  You said:

> I believe the clause in this PR just applies to all members of objects that the spec doesn't specifically address. 

If the spec doesn't address a member, and it is in the top level, it's an attribute. So, how can clients or servers ignore members not recognized by the spec? The thing that makes something a member **is** it being recognized by the spec.

bintoro:
@tkellen My thinking is that "recognized by the spec" == "allowed by the spec".

A clause allowing arbitrary members in an object (e.g., attributes in resource objects) makes them "recognized", even if the extra keys are not explicitly enumerated.

dgeb:
@tkellen The resource object has a "catch all" rule for unreserved members - they're all attributes. Therefore, the spec can never evolve to reserve more keys in a resource object. 

However, links objects and the top-level object don't have such a "catch all" rule. If we introduce a top-level object member such as `templates` some time in the future, we don't want current implementations to throw errors when parsing that unrecognized member. We want them to skip it.

steveklabnik:
This is usually the default position. Do we need to spell this out in normative fashion?

tkellen:
Ah, please ignore me. I thought we were talking about reserving new members in resource objects, not the top level. :+1:

dgeb:
@tkellen ok, cool - glad we're on the same page now :)

@steveklabnik I am concerned that implementations may interpret the following as permission to reject documents when coming across unspecified members:

> Unless otherwise noted, objects defined by this specification **MUST NOT** contain any additional members.

Since this is normative, I added a normative clause to balance it out. I'm open to suggestions to better express both conditions.

steveklabnik:
Ah, that's fair. Hmm.

dgeb:
@steveklabnik is this good enough for now or would you prefer it be all non-normative?

steveklabnik:
Let's do it!

dgeb:
Great! The wording can be finessed after 1.0 if necessary.

PR #453, opened 2015-03-13
Remove relationship url option

lsylvester:
The option for a url for a relationship is equivalent to a "link object" with a "related" member.

There is no need to have both options as it just increased the complexity of implementation.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -261,27 +261,21 @@ can be specified by including a member in a resource's links object.
 
 The name of the relationship declared in the key **SHALL NOT** be `"self"`.
 
-The value of a relationship **MUST** be one of the following:
-
-* A URL for the related resource(s) (a "related resource URL"). When fetched, it
-  returns the related resource object(s) as the response's primary data. For
-  example, an `article`'s `comments` relationship could specify a URL that
-  returns a list of comment resource objects when retrieved through a `GET`
-  request. A related resource URL **MUST** remain constant even when the
-  relationship (the set of referenced resources) mutates. That is, the response
-  from a related resource URL always reflects the current state of the
-  relationship.
-
-* An object (a "link object").
-
-If a relationship is provided as a link object, it **MUST** contain at least
-one of the following:
+The value of a relationship **MUST** be a "link object", and **MUST** contain at
+least one of the following:
 
 * A `"self"` member, whose value is a URL for the relationship itself (a
   "relationship URL"). This URL allows the client to directly manipulate the
   relationship. For example, it would allow a client to remove an `author` from
   an `article` without deleting the `people` resource itself.
-* A `"related"` member, whose value is a related resource URL (as defined above).
+* A `"related"` member, whose value is a URL for the related resource(s)
+  (a "related resource URL"). When fetched, it returns the related resource
+  object(s) as the response's primary data. For example, an `article`'s
+  `comments` relationship could specify a URL that returns a list of comment
+  resource objects when retrieved through a `GET` request. A related resource
+  URL **MUST** remain constant even when the relationship (the set of
+  referenced resources) mutates. That is, the response from a related resource
+  URL always reflects the current state of the relationship.
 * A `"linkage"` member, whose value represents "resource linkage".
 * A `"meta"` member that contains non-standard meta-information about the
   relationship.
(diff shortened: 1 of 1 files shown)


hhware:
I support this. Another argument in favor: there are several ways to specify relationships, why is one of them effectively preferred in a sense that the spec provides a shortcut for it? Removing the shortcut will also make top-level `links` and `links` of relationships more aligned.

Suggestion in #478 (actually inspired by this PR) can be viewed as a further improvement to it in terms of "readability at the first glance".

@lsylvester, BTW, the paragraph containing "The URL can therefore be specified directly as the attribute value" and the example which it describes should also be updated as part of this PR.

lsylvester:
@hhware Thanks. I have updated to remove the comments from the example as it added no further value of the author example.

ethanresnick:
+1! I was also going to suggest this. It's a further move toward consistency over concision. Not only are the responses containing links now more uniform but, pending #482, there'll be more uniformity between link input and output formats as well! Finally, I agree with @hhware that it seems odd to privilege the hypermedia format here when it sits on equal footing with linkage everywhere else in the spec.

I'm curious about what @dgeb thinks of this, though, since he did a pretty thorough revision of the spec and this exception survived.

bintoro:
I support this change on the grounds that providing a related resource URL as a direct attribute value causes a mismatch between the JSON structure and the recommended URL structure:
- `links/someresource` is a relationship URL, while
- `links.someresource` contains the related resource URL.

I think it's a good rule that if there's a URL path that corresponds to a member in the JSON representation, the two should have similar semantics.

ethanresnick:
@dgeb @tkellen What's your take on this?

bintoro:
Something to think about is if the definition of _related resource URL_ is just too long to sit comfortably in the middle of the bulleted list.

Here's an alternative version where both remaining definitions (relationship URL and related resource URL) are moved to their own paragraphs, just like _linkage_:

https://gist.github.com/bintoro/344c19a7ac29345b00b9

ethanresnick:
Another consideration I just realized: if we remove the url-only format, then the format for the resource-level `self` link becomes:

```
"self": {
  "related": "..." 
}
```

Maybe that's not awful, but it is worth noting.

bintoro:
@ethanresnick `self` is a separate thing, defined in the section [Resource URLs](http://jsonapi.org/format/#document-structure-resource-urls).

ethanresnick:
@bintoro I guess...but do we really want implementations to have to have one rule for parsing the `"self"` value in a link object (and the values of the top-level `links`) and a different rule for parsing every other value in a link object? So it seems to me that keeping this format actually ends up being simpler.

bintoro:
Hmm. My first impression is that this is totally a non-issue, but let's see what others have to say. Think of it this way: a `self` member always contains a URL. That's pretty consistent, no? :smile:

If we really want consistency among links object members, then let's do what I've been saying for a long time and place `self` directly on the resource object.

dgeb:
I don't consider it much of a burden to parse strings vs. objects. It adds a little bit of complexity but also offers benefits. The string option for link values provides a clean and simple approach to include fetch-only related resource URLs. It allows deeply nested related resource URLs, such as `"comments.author"`, to be included alongside direct relationships, such as `"comments"`.

bintoro:
> I don't consider it much of a burden to parse strings vs. objects.

Agreed.

For me this issue is mostly about the incongruity between `links/foo` and `links.foo`. It can also be argued that the most logical (albeit useless) scalar-value replacement for a link object would be its `self`.

So, there are in fact _two_ reasons why one might expect the value to be a _relationship URL_ instead of the related resource URL.

> It allows deeply nested related resource URLs, such as `"comments.author"`, to be included alongside direct relationships

Hmm, I don't see why deep associations should need a string value any more than direct relations...

dgeb:
That last sentence doesn't stand well on its own. My emphasis on "clean and simple" from the previous sentence applies to the last one:

> The string option for link values provides a clean and simple approach to include fetch-only related resource URLs. It allows deeply nested related resource URLs, such as "comments.author", to be included alongside direct relationships, such as "comments".

Anyway, your prior argument is fairly compelling:

> It can also be argued that the most logical (albeit useless) scalar-value replacement for a link object would be its self.

I will milestone this for consideration, but I should say that we are very reluctant to make even minor breaking changes at this point.

tkellen:
From an implementors standpoint this is six of this half a dozen of the other to me.

dgeb:
I'm going to close this having spoken with several implementors who consider it rather trivial to implement the alternative string representation.

Thanks for the good discussion.

Issue #452, opened 2015-03-12
Combining extensions


bintoro:
Guys, I'm growing increasingly concerned about the extension system's ability to do what it's supposed to.

The prevailing view is that the base spec will be kept very lean, and extensions will provide all bells and whistles. It has even been pointed out that moving all optional functionality to extensions would give us feature detection for free in the form of extension negotiation.

This all sounds great until you consider that extensions can only be used one at a time.

Let's say I'm writing a public API and that there are four well-known extensions I'd like to use:
- sorting
- pagination
- sideposting
- PUT resource creation.

None of these are mutually exclusive or overlapping in functionality. In fact, I want to use the first two all the time and sometimes combine the last two as needed. Wat do?

Well, I have to define a custom extension _A_ that covers the first two and another extension called _B_ for the other half. This now means that a generic client is unable to use the full capabilities of the server even if it supports all four extensions individually.

At this point, why would anyone implement extension negotiation at all? It seems to add zero value on either the client or server. I could just as well itemize the supported functionality in the API docs, and client implementors could save the trouble of teaching their clients what _A_ and _B_ mean since no one else uses the same definitions anyway.

**TL;DR:**
- Things are directly interoperable only when you use one well-known extension at a time, but that may not be the most common scenario if everything's going to be an extension.
- There seems to be no incentive to formally declare a custom extension as opposed to just documenting the API as "JSON API + extensions this & that" (and skipping the `ext` choreography).

Maybe I'm thinking about this all wrong, so can someone please alleviate my concerns?

hhware:
> Things are directly interoperable only when you use one well-known extension at a time, but that may not be the most common scenario if everything's going to be an extension.

Indeed, this seems to be quite an artificial limitation, if applied to arbitrary extensions, which might not overlap at all. Following that logic, things are the most directly interoperable if there are no extensions allowed! :)

Should not the spec simply tell the following:
- The value of the `ext` parameter **MUST** be a comma-separated (U+002C COMMA, ",") list of extension names.
- Servers that do not support a requested combination of extensions **MUST** return a `406 Not Acceptable` status code.

bintoro:
@hhware, if only it were that simple. Extensions can contradict one another or have interactions that can be resolved in many equally plausible ways. It's pretty much impossible to define universal rules for dealing with those situations.

[...]

PR #454, opened 2015-03-13
Combining extensions #452

hhware:
Here are the changed discussed in #452.

I wonder whether it is worth making recommendation about the order of elements in `ext` and `supported-ext` or just leave it completely open. IMHO, the order could only make any practical difference if that particular header is used in `Vary`, should it just be a concern of the implementation?

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -45,18 +45,20 @@ server supports both the "bulk" and "jsonpatch" extensions.
 
 If an extension is used to form a particular request or response document,
 then it **MUST** be specified by including its name in the `ext` media type
-parameter with the `Content-Type` header. The `ext` media type parameter
-**MUST NOT** include more than one extension name.
+parameter with the `Content-Type` header. 
+The value of the `ext` media type parameter **MUST** be formatted 
+in the same way as the value of `supported-ext` media type parameter 
+and **MUST** be limited to a subset of extensions listed in `supported-ext`.
 
 For example: a response that includes the header `Content-Type:
-application/vnd.api+json; ext=jsonpatch; supported-ext=bulk,jsonpatch`
-indicates that the document is formatted according to the "jsonpatch"
-extension.
+application/vnd.api+json; ext=ext1,ext2; supported-ext=ext1,ext2,ext3`
+indicates that the document is formatted according to the 
+extensions "ext1" and "ext2".
 
 Clients **MAY** request a particular media type extension by including its
-name in the `ext` media type parameter with the `Accept` header. Servers
-that do not support a requested extension **MUST** return a `406 Not
-Acceptable` status code.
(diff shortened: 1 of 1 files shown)


bintoro:
Thanks for filing the PR, @hhware. I just think we'll either need to use fictitious extensions for the example headers or leave them alone for now. See, how are you going to combine `bulk` and `jsonpatch` in the same request? They don't seem additive to me.

Also, it might make things more clear if the example value for `ext` was a strict subset of `supported-ext`.

format/index.md
```diff
@@ -45,18 +45,21 @@ server supports both the "bulk" and "jsonpatch" extensions.
...
+indicates that the document is formatted according to the "bulk" and "jsonpatch"
+extensions.
 
 Clients **MAY** request a particular media type extension by including its
-name in the `ext` media type parameter with the `Accept` header. Servers
-that do not support a requested extension **MUST** return a `406 Not
-Acceptable` status code.
+name in the `ext` media type parameter with the `Accept` header. 
+The value of the `ext` media type parameter 
+**MUST** be a comma-separated (U+002C COMMA, ",") list of extension names.
```
dgeb: This sentence is repeated from above and unnecessary here.

format/index.md
```diff
@@ -45,18 +45,21 @@ server supports both the "bulk" and "jsonpatch" extensions.
...
 
 For example: a response that includes the header `Content-Type:
-application/vnd.api+json; ext=jsonpatch; supported-ext=bulk,jsonpatch`
-indicates that the document is formatted according to the "jsonpatch"
-extension.
+application/vnd.api+json; ext=bulk,jsonpatch; supported-ext=bulk,jsonpatch`
+indicates that the document is formatted according to the "bulk" and "jsonpatch"
+extensions.
```
dgeb: Since bulk and jsonpatch can't be combined, let's leave the example as it was OR use completely fictitious names as @bintoro suggested. For instance, `ext=a,b; supported-ext=a,b,c,d`.

hhware:
@bintoro, @dgeb, thank you for reviewing it! Sorry for not making it right the first time.
The updated version includes corrections for all 3 suggestions of yours:
- on subset,
- on fictition names,
- removed duplication.

dgeb:
Thanks @hhware!

bintoro:
Well that was easy! Thanks guys!

I thought multi-extension requests would be a hard sell :smile:

hhware:
@bintoro, somehow I was sure that multiple extensions are allowed until you brought this up, so thank you!

dgeb:
@bintoro I had come to accept it as an inevitable addition by the time you opened the issue, so thanks to you and @hhware for moving it along :)

PR #455, opened 2015-03-13
Corrections for patch extension

hhware:
I suggest the following minor changes:
- The reminder about `ext` seems unnecessary here since it repeats the base spec. In any case, it should be either reworded to account for changes in #454, or similar reminder should be added to bulk extension page, where it is currently missing. I think it is best to remove it from this page.
- Formatting typo corrected.


dgeb:
Thanks again @hhware :+1:

PR #456, opened 2015-03-13
More consistent resource linkage in the official extensions

hhware:
I think these corrections should be applied to bring the spirit of #447 to the official extensions.


dgeb:
@hhware thanks! I've been meaning to revisit the extensions to bring them inline with those changes.

Issue #419, opened 2015-03-06
PATCH-related clarifications


bobholt:
For consideration pre-1.0:

Commit 01ce3454602e394b078303b8853d725b7981c472 changes all references of `PUT` to `PATCH`. :+1: for RFC 5789 compliance, but it opens a couple of questions:
1. There is no longer explicit support for full-replacement updates via the `PUT` method. Devs familiar with `PUT` but not `PATCH` may find it strange that it isn't mentioned at all.
   - Suggestion: specify that servers **MAY** support the `PUT` method, but it **MUST** only be used for full-replacement updates.
2. RFC 5789 says ["If the Request-URI does not point to an existing resource, the server MAY create a new resource, depending on the patch document type (whether it can logically modify a null resource) and permissions, etc."](http://tools.ietf.org/html/rfc5789#section-2). The json-api spec does not currently offer guidance on this.
   - Suggestion: explicitly say whether or not this is allowable under json-api. If there is no strong guidance deemed necessary, then a best-practice suggestion could be added to Recommendations.

bintoro:
The spec now says nothing about PUT. This means that servers are free to support resource creation and replacement via PUT. No need to even declare an extension, since it poses no conflict (yet).

bobholt:
@bintoro Sure, that makes sense to those of us who have been immersed in the spec for a while, but I'm thinking of Random Serverdev, who sees this for the first time. 

Documentation is more useful when it errs on the side of being explicit rather than making newcomers have to figure something out based on a lack of evidence. I picture someone coming here, not seeing a reference to `PUT` and going like this:

 _scroll_ _scroll_... _scroll_... <ctrl-F> PUT 

"Huh?"

I'm for saving lots of people that time.

bintoro:
@bobholt, fair point. I suppose we could say something like "JSON API does not currently define functionality for PUT, but may do so in the future".

Stating that "servers MAY support the PUT method, but it MUST only be used for full-replacement updates" is not rigorous enough. Implementations would get the details differently, and many would inevitably end up non-compliant once we have an official PUT spec.

With this in mind, I should qualify my previous comment. It's of course prudent to declare a custom extension to avoid a possible incompatibility down the line if/when PUT appears in the base spec.

dgeb:
If the spec says that:

> servers MAY support the PUT method, but it MUST only be used for full-replacement updates.

[...]

Issue #397, opened 2015-03-05
Allow DELETE requests to remove a to-one relationship


jfinkels:
In the section _[Updating To-One Relationships](http://jsonapi.org/format/#crud-updating-to-one-relationships)_, a to-one relationship is removed by using a `PUT` request with the request body `{"data": null}`. This makes sense, but might it also make sense to allow a `DELETE`request to the relationship URL? The _[Updating To-Many Relationships](http://jsonapi.org/format/#crud-updating-to-many-relationships)_ section allows `DELETE` requests to remove members of a to-many relationship.

bintoro:
This is because the spec reserves DELETE for removing items from logical collections. For collections, you have POST/DELETE symmetry, while logically singular entities are always manipulated via PUT.

bintoro:
@jfinkels, is this issue ripe for closing?

Btw, it's perfectly fine if a server wants to additionally support DELETE in the to-one case. It's allowed for sure, just not required.

jfinkels:
That's fine, but I think this should at least be documented...

bintoro:
@jfinkels Do you think [this recent FAQ entry](http://jsonapi.org/faq/#wheres-put) adequately addresses this issue, or did you have a more explicit explanation in mind?

jfinkels:
That's fine, thanks!

PR #457, opened 2015-03-13
Add FAQ entry about PUT

bintoro:
I spent way too long coming up with that first sentence. If it sounds stupid, please suggest an improvement and make me facepalm.

Addresses: #397, #419

### DIFF
--- a/faq/index.md
+++ b/faq/index.md
@@ -52,6 +52,14 @@ and capabilities and use the errors response to let users know. This error featu
 is still pending to be included in the standard since is still in
 [discussion](https://github.com/json-api/json-api/issues/7).
 
+### Where's PUT? Can I use method *X* to do *Y*? <a href="#wheres-put" id="wheres-put" class="headerlink"></a>
+
+JSON API does not currently specify the use of the `PUT` method for any purpose.
+
+Servers may complement the base specification by providing extra capabilities and
+alternative ways of requesting certain operations (e.g., resource creation via
+`PUT` in addition to `POST`).
+
 ### Is there a JSON Schema describing JSON API? <a href="#is-there-a-json-schema-describing-json-api" id="is-there-a-json-schema-describing-json-api" class="headerlink"></a>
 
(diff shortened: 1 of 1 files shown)


dgeb:
@bintoro ha! That is a tricky one to express, and I think you've just about got it.

I'm :+1: to adding this and we can of course refine FAQ later.

faq/index.md
```diff
@@ -52,6 +52,14 @@ and capabilities and use the errors response to let users know. This error featu
 is still pending to be included in the standard since is still in
 [discussion](https://github.com/json-api/json-api/issues/7).
 
+### Where's PUT? Can I use method *X* to do *Y*? <a href="#wheres-put" id="wheres-put" class="headerlink"></a>
+
+JSON API does not currently specify the use of the `PUT` method for any purpose.
+
+Servers may complement the base specification by providing extra capabilities and
+alternative ways of requesting certain operations (e.g., resource creation via
+`PUT` in addition to `POST`).
```
mtvx, line 61: I was expecting answer to the question of why `PATCH` was chosen to represent things that `PUT` is normally used for (to overwrite data in resources)?

So I suggest to open it up a bit more.

bintoro, line 61: The reason is that the HTTP spec reserves PUT for complete resource replacements, while PATCH is intended for partial updates. If the payload only specifies a strict subset of fields, a true PUT request would cause the omitted fields to be reset. PATCH causes their values to remain intact.

This PR was not originally intended to answer that question, hence #488. But if people think it would make sense to address the PUT/PATCH issue directly in this FAQ entry instead of adding a new one, then I'm certainly ready to add that explanation here.

mtvx, line 61: Ok, thanks for the explanation. I was in the assumption that PUT can be used for partial updates, but not the first nor the last time I was wrong :)

dgeb:
Merging now. This is definitely a FAQ, which we can of course refine as needed.

PR #458, opened 2015-03-13
Bulk responses

hhware:
Extension "jsonpatch" describes responses in detail, but similar section is missing from the description of the extension "bulk". The suggested change is copied almost word-to-word from the former to the latter, which is not elegant, but is there a better place for this?

### DIFF
--- a/extensions/bulk/index.md
+++ b/extensions/bulk/index.md
@@ -98,3 +98,33 @@ Accept: application/vnd.api+json; ext=bulk
   "data": { "type": "articles", "id": ["1", "2"] }
 }
 ```
+
+## Responses <a href="#responses" id="responses" class="headerlink"></a>
+
+For successful Bulk requests, a server **MUST** return 
+the corresponding success HTTP status code of the performed operation.
+
+When a server encounters one or more problems while processing a Bulk
+operation, it **SHOULD** specify the most appropriate HTTP error code in the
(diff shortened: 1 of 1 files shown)


extensions/bulk/index.md
```diff
@@ -98,3 +98,33 @@ Accept: application/vnd.api+json; ext=bulk
...
+Internal Server Error` might be appropriate for multiple 5xx errors.
+
+A server **MAY** return error objects that correspond to each operation. The
+body of the response **MUST** contain an array of JSON objects, which
+**MUST** be in sequential order and correspond to the operations in the
+request document. Each response object **SHOULD** contain only error objects
+keyed by `"errors"`, since no operations can be completed successfully when
+any errors occur. Error codes for each specific operation **SHOULD** be
+returned in the `"status"` member of each error object.
+
```
dgeb, line 130: This last paragraph will need to be revisited because the Bulk extension doesn't contain the concept of multiple "operations" like the JSON Patch extension does. Instead, a single operation is performed on one or more resources sent as primary data.

hhware, line 130: I believe it does: "It provides support for performing multiple operations in a request". I do not mind revising it, but which term to use then?

hhware:

dgeb:
We are not versioning the official extensions along with the base spec.

hhware:
Resolved by #655 in a more generic way.

PR #461, opened 2015-03-14
Clarifications on request extensions #452

hhware:
I think that wording on relation of `ext` and `supported-ext` is clear for responses, but a bit unclear for requests. This is my attempt to fix it.


dgeb:
Makes sense to include request and response document examples. :+1:

PR #462, opened 2015-03-14
Remove reference to paginating non-primary resources

ethanresnick:
Right now, the spec says: "To paginate an included collection returned in a compound document, supply pagination links in the corresponding link object." But: what corresponding link object is this talking about? The only link objects in the `"included"` section are for part of individual resources; they're not associated with any collections. 

Even if the "included" section were to be restructured such that there were clear collections, presumably one for each path asked to be included, I don't think we'd want to paginate those collections; it should be enough to just paginate the primary resources, and only include resources related to the primary resources that are part of the current "page" (which is what the spec would do if this PR were merged). Doing otherwise could get messy, with query parameters like `page[posts.author][size]`. Ick.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -817,17 +817,11 @@ alphabetical order.
 
 ### Pagination <a href="#fetching-pagination" id="fetching-pagination" class="headerlink"></a>
 
-A server **MAY** choose to limit the number of resources returned in a response
-to a subset ("page") of the whole set available.
+A server **MAY** choose to limit the number of primary resources returned 
+in a response to a subset ("page") of the whole set available.
 
 A server **MAY** provide links to traverse a paginated data set ("pagination
-links").
-
-Pagination links **MUST** appear in the link object that corresponds to a
-collection. To paginate the primary data, supply pagination links in the
-top-level `links` object. To paginate an included collection returned in
-a compound document, supply pagination links in the corresponding link
-object.
+links") links in the top-level `links` object.
 
 The following keys **MUST** be used for pagination links:
 


hhware:
> I don't think we'd want to paginate those collections; it should be enough to just paginate the primary resources

@ethanresnick, I thought it is a feature, and quite a useful one! 

For instance, suppose the API is built around a hierarchy of some items. Suppose the API is providing a method to search those items by some attribute and a custom parameter for that search method to allow clients to request all sub-hierarchy of items rooted at the matching item -- they will be returned in `included`. If the tree is deep, the typical number of objects in `data` will be much less than the typical number of objects in `included`. What is worse, total number of objects in `data` and `included` will be unpredictable for any given request because the match can occur at any depth of the tree. Hence, if the goal of pagination is to return consistent number of objects in response, this goal becomes unreachable. 

"Doing otherwise" would provide a possibility for the API designer to handle this problem.

dgeb:
> But: what corresponding link object is this talking about? The only link objects in the "included" section are for part of individual resources; they're not associated with any collections.

The corresponding link object is not in the `included` section, but rather in one of the primary resource objects. To illustrate, I've added pagination links to the home page example which allow you to fetch subsequent comments without re-fetching the included ones:

``` json
{
  "links": {
    "self": "http://example.com/posts",
    "next": "http://example.com/posts?page[offset]=2",
    "last": "http://example.com/posts?page[offset]=10"
  },
  "data": [{
    "type": "posts",
    "id": "1",
    "links": {
      "author": {
        "linkage": { "type": "people", "id": "9" }
      },
      "comments": {
        "next": "http://example.com/posts/1/comments?page[offset]=3",
        "last": "http://example.com/posts/1/comments?page[offset]=10",
        "linkage": [
        ]
      }
    }
  }],
  "included": [{
    "type": "people",
    "id": "9",
    "first-name": "Dan",
    "last-name": "Gebhardt",
    "twitter": "dgeb",
    "links": {
    }
  }, {
    "type": "comments",
    "id": "5",
    "body": "First!",
    "links": {
      "self": "http://example.com/comments/5"
    }
  }, {
    "type": "comments",
    "id": "12",
    "body": "I like XML better",
    "links": {
      "self": "http://example.com/comments/12"
    }
  }]
}
```

ethanresnick:
Ahh, ok. I totally misunderstood this. I'll close this PR.

PR #463, opened 2015-03-14
Required extensions #452

hhware:
I think one more situation is not covered: when a particular extension is vital for the API, so that it is not possible to form requests / responses without it. The server should have a way to communicate this. Here is my attempt to fix this problem.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -60,6 +60,23 @@ name in the `ext` media type parameter with the `Accept` header. Servers
 that do not support a requested extension or combination of extensions
 **MUST** return a `406 Not Acceptable` status code.
 
+Servers **MAY** require a particular extensions to be used in every request
+and response. In this case, servers **MUST** return those extensions 
+in every response in the `required-ext` media type parameter of the 
+`Content-Type` header. The value of the `required-ext` parameter 
+**MUST** be a comma-separated (U+002C COMMA, ",") list of extension names
+and must be limited to a subset of the extensions listed in `supported-ext`.
+
+For example: a response that includes the header `Content-Type:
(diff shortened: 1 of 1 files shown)


dgeb:
I'm not comfortable with the concept of required extensions, which seem to be a roundabout way of versioning the json-api spec (and every extension in use by an API) via extensions. Generic clients simply couldn't work with a spec that has required extensions.

ethanresnick:
I think we could always add something like this post 1.0 but, pre 1.0, it's worth noting that this can already be done with standard HTTP: if a server wants to require an extension, it can just 415 any requests that don't have that extension in the `ext` parameter.

hhware:
> Generic clients simply couldn't work with a spec that has required extensions.

@dgeb, is it a requirement that generic clients must be able to work with every API based on JSON API spec? Suppose one creates an API with a custom pagination approach via top-level `meta`, and with custom relationships via link-object-level `meta`, none of which is readable by a generic client. Should not there be a way to communicate to the client that custom extensions must be turned on, besides specifying this in the API docs?

jelhan:
I agree with @dgeb that required extensions goes against the intent of the JSON:API specification. The specification tries giving control to the client. E.g. which related resources should be included, which fields of a resource should be returned, which profile or extension to use. Of course it's up to each server to decide if a feature is supported. But the decision if a supported feature should be used is given to the client. This is also the case for content negotiation for profiles and extensions, which has been added in v1.1.

In addition, I struggle to see an use case for required extensions or profiles. Especially if the extension would be required for all requests regarding of endpoint and HTTP method. Hard to imagine a profile or extension, which would be meaningful required for all operations on every endpoint.

I'm going to close this PR. If there are any new arguments, please feel free to comment anyways. Reopening is always an option.

PR #464, opened 2015-03-15
"Must" in non-normative notes

hhware:
Does it actually make sense there? IMHO, either "must" should be downgraded to a softer term or non-normative note should be upgraded to a constraint.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -346,7 +346,7 @@ meta-information about linkage.
 together all of the included resource objects without having to `GET` any
 relationship URLs.
 
-> Note: If present, a *related resource URL* must be a valid URL, even if the
+If present, a *related resource URL* **MUST** be a valid URL, even if the
 relationship isn't currently associated with any target resources.
 
 For example, the following article is associated with an `author` and `comments`:
@@ -747,7 +747,7 @@ a relationship URL that does not exist.
 does not exist. For example, when `/articles/1` does not exist, request to
 `/articles/1/links/tags` returns `404 Not Found`.
 
-> Note: If a relationship URL exists but the relationship is empty, then
+If a relationship URL exists but the relationship is empty, then
 `200 OK` **MUST** be returned, as described above.
 
 ##### Other Responses <a href="#fetching-relationships-responses-other" id="fetching-relationships-responses-other" class="headerlink"></a>


bintoro:
I see no reason to avoid words like "must". The specially formatted keywords exist for a purpose.

As for why some things are written as non-normative segments: I believe this is based on having just one normative clause per rule. Any explanations that repeat or clarify a previously stated rule are expressed in a non-normative way.

I agree that formal MUSTs in non-normative notes are not ideal. My preferred approach would be just to turn each of them into an informal "must".

tkellen:
:+1: to these edits as is. I agree that 2119 keywords do not belong in non-normative notes (I have updated a few instances of this as well).

hhware:
@bintoro I personally do not quite understand what is the meaning of plain-text "must", whether in the spec or in notes. How should I read it? Is it a constraint or not? If not, why use "must"?

Perhaps it would make sense to convert all verbs covered by RFC2119 to boldfaced all-caps keywords? The RFC specifies how these words should be read: any document can just refers to it and be done with this formality. If non-RFC versions of these words are used, what are they supposed to mean, where is the difference between RFC and non-RFC versions defined? Why introduce this confusion in the first place?

If for some reason the spec wants to stay intentionally ambiguous about something, some synonym could be used ("ought to", "can"?), although I personally fail to see a reason why this **MUST** be necessary...

FYI, a PR on a similar issue: #474

bintoro:
@hhware The plaintext "must" carries the normal meaning of the word; it's a requirement. But the lack of a formal keyword is an indication to the reader that the statement doesn't introduce a _new_ normative requirement. That is, the sentence can be skipped without loss of critical information.

Isn't it good that we can liberally add explanatory statements and notes without bloating the text with redundant normative clauses? IMO it's silly to have a number of normative rules for the exact same thing. It has nothing to do with "intentionally ambiguous".

hhware:
> Isn't it good that we can liberally add explanatory statements

@bintoro, thank you for the explanation, got it now. 

(I still think that all editions suggested here are appropriate.)

hhware:

hhware:
A humble reminder: this issue is still awaiting fulfillment of @dgeb's [promise](https://github.com/json-api/json-api/issues/507#issuecomment-86633286).

hhware:
Rebased this PR. Two items are left out of initial three.

tkellen:
:+1: from me @dgeb?

dgeb:
Thanks @hhware!

hhware:
I have a suspicion that this and 3 preceding merges were lost because of forced push to gh-pages...

tkellen:
Checking on this briefly.

tkellen:
You're correct @hhware. I'm on my wife's computer today (mine is being repaired) and her git config is set to push matching, not push simple. My force pushes without specifying the top-level-attributes branch force pushed over gh-pages. Fixing this now.

PR #465, opened 2015-03-15
Binding sections

hhware:
It appears that sections before Extending also contain clauses which must be considered binding. I think it would be convenient to make Extending a borderline between binding and non-binding sections.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -17,6 +17,8 @@ without compromising readability, flexibility, or discoverability.
 JSON API requires use of the JSON API media type
 ([`application/vnd.api+json`](http://www.iana.org/assignments/media-types/application/vnd.api+json))
 for exchanging data.
+JSON API documents are defined in JavaScript Object Notation (JSON)
+[[RFC4627](http://tools.ietf.org/html/rfc4627)].
 
 ## Conventions <a href="#conventions" id="conventions" class="headerlink"></a>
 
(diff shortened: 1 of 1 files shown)


hhware:

hhware:
A humble reminder: this issue is still awaiting fulfillment of @dgeb's [promise](https://github.com/json-api/json-api/issues/507#issuecomment-86633286).

hhware:
A humble reminder: 1.0 seems to be around the corner, can this PR be considered? I believe that this is a non-controversial change. It makes structure of the document a bit more robust.

tkellen:
Seems fine to me. @dgeb?

dgeb:
Marking the Extending section as binding, regardless of extensions, seems like a good idea.

The explicit ordering of binding sections seems a bit odd, considering it only adds the introduction, which is pretty much exposition. And then moving JSON usage out of document structure takes a bit away from that section.

I'm not sure exactly how to handle this, but this seems pretty close. Marked for resolution before 1.0.

dgeb:
The Extending section contains:

> An extension MAY make changes to and deviate from the requirements of the base specification apart from this section, which remains binding.

I believe this is sufficient.

hhware:
@dgeb, section Extending is binding, so section Conventions must be binding too, otherwise what is the point to make Extending binding if RFC2119 words used in it are not strictly defined? The clause "JSON API documents are defined in JSON" must be binding too -- wouldn't it be funny to allow XML in the standard called JSON API? So maybe move  "JSON API documents are defined in JSON" to Conventions (it will not be totally out of place there) and declare Extending and Conventions binding? I can redo the PR if there is an agreement. Alternatively, I still think that what is suggested by this PR is fine because section Introduction does not contain anything which should not be binding.

dgeb:
The JSON requirement simply doesn't belong in Conventions. Nor does it belong in the intro. It belongs with the document structure definition.

There are plenty of absurdist extensions that could be created and would simply be ignored. Extensions must state where they deviate from the base spec. They must be negotiated by client and server, so there will be no surprises.

PR #467, opened 2015-03-16
Apparent error in Updating To-Many Relationships

hhware:
Perhaps I am missing something, but in which case it can be an object and not an array?


dgeb:
This is indeed an error, but it is also corrected by #480, which is bigger in scope. Let's decide on that first.

hhware:
Resolved by #480

Issue #105, opened 2013-05-27
What does creating a document with links look like?


gabrielgrant:
In "Updating" > "Creating a Document" the example document doesn't have any links.

When creating a document with links in the URL style, I'm assuming the top-level "links" template is still required, yes?

Example 1

```
POST /photos
Content-Type: application/json
Accept: application/json

{
  "links": {
    "photos.comments": "http://example.com/comments/{posts.comments}"
  },
  "photos": [{
    "title": "Ember Hamster",
    "src": "http://example.com/images/productivity.png",
    "links": {
      "comments": [ "1", "5", "12", "17" ]
    }
  }]
}
```

Or is it valid to just include the ID(s) and assume the server knows what type of object you're referencing?

Example 2

```
POST /photos
Content-Type: application/json
Accept: application/json

{
  "photos": [{
    "title": "Ember Hamster",
    "src": "http://example.com/images/productivity.png",
    "links": {
      "comments": [ "1", "5", "12", "17" ]
    }
  }]
}
```

I'm pretty sure the client should have to be explicit about what it is linking to, despite there not being any way to explicitly communicate whether a server supports ID-based JSON, URL-based JSON or both.

As with #101 it's possible that this is implicit, but I think at least having one of these examples in an appendix makes sense.

steveklabnik:
If you want the links to be something specific, then you'd include them, but if you don't, then you don't.

That said, an example with including them would make sense.

drgomesp:
I'm also not sure about how to implement the object creation from the links point of view.

What would be the best option?

```
POST /categories
Content-Type: application/json
Accept: application/json

{
  "category": [{
    "name": "My category",
    "links": {
      "products": [ "1", "5", "12", "17" ]
    }
  }]
}
```

Versus

```
POST /categories
Content-Type: application/json
Accept: application/json

{
  "category": [{
    "name": "My category",
    "products": [ 
        "1", "5", "12", "17" 
    ]
  }]
}
```

?

dgeb:
It's possible to create a resource together with relationships. For example:

```
POST /categories
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json

{
  "data": {
    "type": "category",
    "name": "My category",
    "links": {
      "products": { "type": "products", "ids": ["1", "5", "12", "17"] }
    }
  }
}
```

PR #468, opened 2015-03-16
An example of creating relationships together with resource

### LINKED ISSUE #375: Examples with relations in "Creating Resources" section
Creating resources with relations needs to be done just like in updating, with a links object under data informing the relations (instead of foreign keys).

The [Creating Resources](http://jsonapi.org/format/#crud-creating) section doesn't mention this or show any examples including links, making this an easy oversight for an implementer.

hhware:
This is just an example to illustrate https://github.com/json-api/json-api/issues/105#issuecomment-74985964. 

It should close #375.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -908,6 +908,33 @@ Accept: application/vnd.api+json
 }
 ```
 
+Resources can be created together with relationships:
+
+```text
+POST /photos
+Content-Type: application/vnd.api+json
+Accept: application/vnd.api+json
+
+{
+  "data": {
+    "type": "photos",
+    "title": "Ember Hamster",
+    "src": "http://example.com/images/productivity.png",
+    "links": {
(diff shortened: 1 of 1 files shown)


bintoro:
More important than an example would be even a basic mention of how relationships are represented when creating resources. The format is currently detailed only under _Updating Resources_.

This is being addressed in #480 and #482. Once the format is properly described, I'd say a dedicated example becomes unnecessary, but it would be a good candidate for the "additional usage examples" section I'm trying to establish in #493.

**EDIT:** PR #480 now incorporates a relationship in the base example for creating resources.

tkellen:
Thanks for the PR @hware! I'm going to close this in favor of #480, which is a little more comprehensive. If you feel anything is missing still, could you please comment on that issue?

PR #469, opened 2015-03-16
Additional pagination members

### LINKED ISSUE #384: Clarifying total for pagination links.
The documentation lists that `"first"`, `"last"`, `"prev"`, and `"next"` must be used for pagination.

For systems that use an offset/limit scheme, maybe a `"total"` or `"count"` key MAY be included in the `"meta"` key for the resource being paginated.  This way it can prevent requests with an offset outside of the range capable of being returned.

hhware:
Perhaps I am missing something, but if the server needs to define additional members of `links`, related to pagination, i.e. something beyond `prev`, `next`, `first`, `last`, how can it be done? For example, the server wishes to return the total number of pages, or something else?

Perhaps it would be even better to group pagination-related members under a separate member of `links`, e.g., under `links.pagination`, to avoid polluting the namespace of `links` with arbitrary custom keys?

@ColtonProvias, this should hopefully resolve #384. I would appreciate your opinion on it.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -287,7 +287,7 @@ one of the following:
   relationship.
 
 A link object that represents a to-many relationship **MAY** also contain
-pagination links, as described below.
+pagination members, as described below.
 
 If a link object refers to resource objects included in the same compound
 document, it **MUST** include resource linkage to those resource objects.
@@ -445,7 +445,7 @@ The top-level links object **MAY** contain the following members:
 * `"self"` - a link for fetching the data in the response document.
 * `"related"` - a related resource URL (as defined above) when the primary
   data represents a resource relationship.
-* Pagination links for the primary data (as described below).
+* Pagination members for the primary data (as described below).
 
 ## Fetching Data <a href="#fetching" id="fetching" class="headerlink"></a>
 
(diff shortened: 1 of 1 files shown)


tkellen:
I'm :+1: to this adjustment. @dgeb?

dgeb:
A link object can also contain a `meta` member "that contains non-standard meta-information about the relationship."

In this way, meta-information can be included about relationships in the same way that it can be included about primary resources.

ethanresnick:
I'm also pro keeping the pagination keywords at the top-level of the links object and just putting extra stuff in `"meta"`. Because the user can't add custom pagination members anyway without using `"meta"` (to preserve future extensibility for us), and keeping the keys at the top level makes them much more easily interpretable as standard [link relations](http://www.iana.org/assignments/link-relations/link-relations.xhtml)—the keywords for which we're not-so-accidentally reusing.

ahacking:
In my apis I never use pagination, I find offset and limit to be far more versatile and more fundamental than an artificial pages.

I support just using meta as @dgeb has suggested and not try to pick a flavor of partial collections.

dgeb:
I'm in favor of closing this, but willing to hear rebuttals.

hhware:
> that contains non-standard meta-information about the relationship

@dgeb, perhaps I'm missing something, but I thought top-level `links` does not allow `meta`. That said, I agree that `meta` would be a better place for this. 

Should I update this PR to add `meta` to top-level `links` instead of its current content?

dgeb:
> I thought top-level links does not allow meta

That's correct - I was referencing the top-level `meta`.

> Should I update this PR to add meta to top-level links instead of its current content?

I don't think this is necessary.

steveklabnik:
I am also :-1: , for the same reasons as @dgeb

hhware:
@dgeb, I see now, thanks! Closing it, although I still find it somewhat weird that 
- custom pagination information must be placed under another top-level key (`meta`) than the reserved 4 pagination links, and
- if an API chooses to define other pagination links besides the reserved 4, those links will also have to go under top-level key `meta`.

PR #470, opened 2015-03-16
Inclusion of meta in fields

hhware:
Since `meta` may contain arbitrary data, the server may need a way to include/exclude `meta` as a whole or its members from the response. For example, consider an API built around a content tree, with some statistics about subtrees being returned for each node in `meta`. It can be resource-intensive to generate such statistics, and it would be helpful to have a way to turn it off.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -760,11 +760,22 @@ GET /articles/1?include=author,comments,comments.author
 A client **MAY** request that an endpoint return only specific [fields] in the
 response on a per-type basis by including a `fields[TYPE]` parameter.
 
-> Note: Only [fields] are affected; `type`, `id`, and (optionally) `self` are
-included as normal.
+An implementation of this specification **MAY** also define custom rules 
+for inclusion of `meta` member or specific descendant members of `meta` 
+member in `fields[TYPE]` parameter.
+
+> Note: Only [fields] and `meta` are affected; 
+`type`, `id`, and (optionally) `self` are included as normal.
+
+> Note: The rules regulating inclusion of `meta` should define whether `meta` 
+as a whole can be included, whether its members or their descendants can be 
+selectively included and which specific members can be included, whether 
+`meta` or its descendant members are included by default if not listed in 
+`fields[TYPE]`, etc.
 
 The value of the `fields` parameter **MUST** be a comma-separated (U+002C
-COMMA, ",") list that refers to the name(s) of the fields to be returned.
+COMMA, ",") list that refers to the name(s) of the fields to be returned
+(including [fields], `meta` and its descendant members).
 
 If a client requests a restricted set of [fields], an endpoint **MUST NOT**
 include additional [fields] in the response.


hhware:

ethanresnick:
I'm +1 on this. There should be a way to exclude `meta` information, I think, and, from the perspective of the API's user, using the `fields` parameter seems like a natural choice. Granted, allowing this would complicate our definition of a field a bit, and also complicate server-side implementations a tad, but I don't think either of those outweigh the benefit. 

If we do merge this, though, I do think the PR should offer a bit more detail. For example, would a `fields[posts]=title` parameter mean that `posts` resources would not have `meta` information because `meta` wasn't explicitly included in the `fields` list? Or would it mean that, depending on the implementation, `posts` resources might have `meta`?

(I'd prefer the first option, which treats meta just like any other field.)

hhware:
> Granted, allowing this would complicate our definition of a field

@ethanresnick, the way it is suggested, definition of a field remains intact, `meta` or its members are added as options to `fields[TYPE]`. I think it is better that way, would not you agree?

Added a clarification on it.

> If we do merge this, though, I do think the PR should offer a bit more detail. 

I think that this should be defined by implementations because
1. This spec is never concerned with anything about `meta` or its contents, and this case should not be an exception. 
2. An implementation may want to allow selective inclusion of some members of `meta`, while requiring that other members of `meta` are always included. According to (1), there is no way this spec can possibly get into these details.

Added clarifications.

ethanresnick:
> the way it is suggested, definition of a field remains intact, meta or its members are added as options to fields[TYPE]. I think it is better that way, would not you agree?

To me, it's really weird and potentially confusing to have a parameter called "fields" and to define something called a "field", but then to allow that "fields" parameter to take a value (namely `meta`) that isn't defined as a "field". That's why I was suggesting that, if we merge this PR, we'd probably want to expand our definition of a "field" so that "meta" counts as one. Otherwise, including it in the `fields` parameter is just too bizarre from a naming point of view.

> I think that this should be defined by implementations

If we follow my logic above and define `meta` to count as a standard field, then it seems weird that we wouldn't make the same rules apply to `meta` that apply to every other field (i.e., it **MUST** be excluded entirely if not explicitly mentioned in the `fields` parameter).

> An implementation may want to allow selective inclusion of some members of `meta`

If this is really what we want to support, then I don't think we should use the `fields[type]` parameter at all for it, since that parameter completely shows or removes a field in all other cases; it doesn't allow for _partial_ inclusion of a field. My suggestion would be that API's that want partial `meta` inclusion should use a custom extension to define the behavior they need; trying to bolt it on to a standard spec parameter (i.e. `fields`) that otherwise acts very differently doesn't make much sense to me.

bintoro:
> it doesn't allow for _partial_ inclusion of a field

Right. But I actually think it would be completely sensible to request individual field members using a dot-separated syntax as in `fields=foo,bar.x,bar.y,meta.i,meta.j`, where `bar` is a complex attribute.

Requiring support for this in the base spec would be too much, however.

ethanresnick:
> I actually think it would be completely sensible to request individual field members using a dot-separated syntax

I totally agree! And if there were an effort to support partial inclusion of fields (including `meta`) as an extension, I'd support it.

I think my basic point still stands though: if `meta`'s inclusion is going to be affected by `fields[type]`, our definition of a field should encompass `meta`, whether or not an extension makes partial inclusion possible.

bintoro:
>  if `meta`'s inclusion is going to be affected by `fields[type]`, our definition of a field should encompass `meta`

Yes.

Including `meta` in the definition of _fields_ while continuing to exclude `id` and `type` is difficult to rationalize, however. If the `fields` param really needs to control the inclusion of `meta`, then I'd prefer to regard `id` and `type` as fields as well and state that they remain unaffected by `fields[type]`.

IMO it's less strange to exclude some fields from the effects of `fields` than to say that the parameter affects something that is not a field. (EDIT: Plus, we already have a note reminding that `id` and `type` are special cases, so this would barely change a thing.)

hhware:
After pondering over this a bit more, I now think the best way would be indeed to manage inclusion of `meta` or its descendants via a separate query string parameter, defined either in an extension or just by specific APIs. The primary reason is that `meta` is a potentially complex custom thing with undefined scope, so its handling may need to be custom as well. I now think that this PR should just be closed. Some good and consistent approach appears, a new issue/PR could be open. Objections?

bintoro:
@hhware That would be my preference as well. There's something we could do right now, though: mention `meta` in the note along with `self`:

>  Note: Only [fields] are affected; `type`, `id`, and (optionally) `self` and `meta` are included as normal.

I just don't like the "(optionally)" part because making it optional to include something "as normal" implies that an implementation may in fact use `fields` to govern `self` and `meta` as well.

ethanresnick:
@hhware Telling implementors to just use a custom extension for `meta` exclusion was, actually, my initial gut reaction to this PR. But, after thinking about it, I did think there was something to be gained from adopting a version of this proposal, which is why I gave my original +1. And I still do think that. So, before we close this, let me take one more stab at arguing for a version of this proposal:

The point of sparse fieldsets is to make the request faster by saving the server from having to generate,<sup>*</sup> the network from having to transmit, and the client from having to deserialize, data that the client doesn’t need.

All of these performance concerns apply just as much to `meta` as they do to what we’re currently calling the resource’s fields. That is, the `meta` data might take a long time to generate and/or there might be a lot of it to send. So, when the whole `meta` object isn’t needed, there’s no reason it shouldn’t be excludable, like resource fields. 

Now, you guys are right that the `fields` parameter is an imprecise mechanism for the client to specify what it doesn’t need, as it doesn’t support the possibility of excluding part of a complex attribute. But that limitation doesn’t make the ability to exclude a complex attribute with `fields` useless, as there may be cases where the entire attribute’s value isn’t needed. So, likewise, this limitation wouldn’t make it useless to be able to exclude `meta` altogether through the `fields` parameter.

And the solution to this limitation (namely, an extension that allows for dot-separated paths in `fields`), would also apply just as much to `meta` as to existing fields.

So, I guess my bottom line point is this: from the perspective of sparse fieldsets’ performance goals, the data in `meta` is completely analogous to the data in any other attribute, and so we should treat them the same by allowing `meta` to be controlled by fields just like anything else.

If you guys agree with this, then we can talk about how best to tweak the definition of a field so that it encompasses meta, while simultaneously keeping in place the requirement that `fields` doesn’t affect `id` and `type`. But that really shouldn’t be too hard, and I do think it would add value.

<em>*</em> i.e. query for, serialize, otherwise check/transform, etc.

bintoro:
@ethanresnick, I'm not opposed to what you're suggesting, per se, but it seems to me that the only sensible option then would be to call everything a field and add a clause stating that `id` and `type` are unaffected by `fields`. What's your position on this?

ethanresnick:
@bintoro Cool. 

On the definition front: we're only using the definition of a field in the sparse fieldsets section, so whether we carve out the exception for `type` and `id` in the definition of a field or in the definition of the `fields` parameter seems to make little functional difference to me, and I could probably support either approach. Out of curiosity, why do you prefer the latter one? (I.e., what is it about defining a field as "All members of a resource object except for those required for resource identification, namely: `type` and `id`." that seems weird to you?)

bintoro:
> what is it about defining a field as "All members of a resource object except for those requires for resource identification, namely: type and id." that seems weird to you?

I think `id` and `type` are much closer than `meta` to what is generally understood by "field". In a backend datastore, for example, you would likely consider `id` as just another attribute among the others. So yeah, it seems very counterintuitive to call a metadata object a field yet exclude the resource ID from the definition.

I also believe it's simpler to special-case `id` and `type` in the _Sparse Fieldsets_ section. In fact, we're already doing this, as there's a note reminding implementors that they're not fields.

ethanresnick:
> I think id and type are much closer than meta to what is generally understood by "field"

That totally makes sense! Then let's do it as you say: call all attributes, relationships, identifiers, and meta information fields, and just specify that the `fields` parameter doesn't affect `type` and `id`.

hhware:
> So, I guess my bottom line point is this: from the perspective of sparse fieldsets’ performance goals, the data in meta is completely analogous to the data in any other attribute, and so we should treat them the same by allowing meta to be controlled by fields just like anything else.

@ethanresnick, sorry for not expressing myself clearly. I did not actually suggest to resolve this by leaving to extensions to decide how to extend the meaning of `fields[TYPE]` (they can do it anyway). 
Since scope of `meta` is undefined, the application may need to handle it in ways totally incompatible with semantics of `fields[TYPE]`. Instead, my thinking was that if sparse `meta` is desired, extensions or API implementations may define special query string parameters for that. It would allow them to handle sparse `meta` in a variety of different ways, such as
- Allow to include/exclude `meta` as a whole via `include_meta=(true|false)`.
- Allow sparse inclusion of `meta` or specific descendants of `meta` by directly specifying them via `sparse_meta=meta.f1,meta.f5.f102` or `sparse_meta=meta` with some default convention, possibly different for different endpoints.
- Allow sparse inclusion based on values of other query string parameters. E.g., parameter `mode=(admin|viewer)` could (a) command inclusion of specific of members of `meta` and (b) modify response in some other ways. This use case is completely incompatible with `fields[TYPE]`.

This is in line with suggestion of @bintoro. Perhaps it could be phrased like that:

> Note: Only [fields] are affected; `type`, `id`, `self` (if returned by the server) and `meta` are not affected by `fields[TYPE]`. Sparse inclusion of `meta` or its members can be accomplished, for example, by providing query string parameters for this purpose.

ethanresnick:
@hhware Right. What I was saying was just that the `include_meta=(true|false)` use case could _and should_ be built into the spec by allowing the `fields[type]` parameter to apply to `meta` too. Applications that need more advanced `meta` handling would still define an extension.

tkellen:
I don't agree that meta is a field, and I think marking it a such would be a mistake. I understand the desire to control the visibility of meta, but using sparse fieldsets seems like the wrong approach. Most especially because it cannot control the top level `meta` key.

dgeb:
I also disagree with the consideration of `meta` as a field. The nature of `meta` is very application-specific and should not be considered uniformly. `meta` objects may appear in multiple locations and have very different purposes. For instance, an API may include copyright information in `meta` that simply MUST be included at all times to satisfy a legal arrangement. We don't want to force such an API to be non-compliant by saying that they ever MUST exclude that information from a response.

I also understand the desire to request the omission of `meta` at different levels, but would prefer to discuss this in an issue rather than a PR. I'm closing this PR now since I don't want to pursue treating `meta` as a field.

ethanresnick:
> For instance, an API may include copyright information in meta that simply MUST be included at all times to satisfy a legal arrangement.

@dgeb Good point! I hadn't thought about use cases like that. I'm fine with closing this then, too.

@bintoro Could you put together the PR you proposed [here](https://github.com/json-api/json-api/pull/470#issuecomment-87970410), just to note that `meta` is NOT effected?

hhware:
I have updated this PR with the wording suggested in https://github.com/json-api/json-api/pull/470#issuecomment-88224787, which, I think, was in line with suggestion of @bintoro, and also hopefully made "optionally" more clear.

PR #471, opened 2015-03-16
Namespace of attributes and relationships

hhware:
It seems implied from rules of forming references (e.g., `articles.author` and `articles.title`), but it is not  stated anywhere explicitly.


hhware:

ethanresnick:
I'm +1 on this! 

I might change the wording a bit, but this does seem like a good thing to state.

tkellen:
:+1:

dgeb:
Good to be explicit about this - thanks @hhware :shipit:

PR #472, opened 2015-03-16
Clarifying note for 404 on relationships

hhware:
Maybe I am missing something -- what are other cases when the URL could be non-existing? IMHO, it would be helpful to list them.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -701,6 +701,10 @@ Content-Type: application/vnd.api+json
 A server **MUST** return `404 Not Found` when processing a request to fetch
 a relationship URL that does not exist.
 
+> Note: This can happen when the parent resource of the relationship 
+does not exist. For example, when `/articles/1` does not exist, request to
+`/articles/1/links/tags` returns `404 Not Found`.
+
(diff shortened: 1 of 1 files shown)


fotinakis:
I agree with this—just read the spec for the first time yesterday and the 404 handling was the part I got confused by. I think it would be helpful to have much more info on the philosophies behind why related resources 404 but normal resources 200 with empty data, and what situations it helps alleviate.

hhware:

tkellen:
I'm :+1: to merging this. @dgeb?

format/index.md
```diff
@@ -700,6 +700,10 @@ Content-Type: application/vnd.api+json
 A server **MUST** return `404 Not Found` when processing a request to fetch
 a relationship URL that does not exist.
 
+> Note: This can happen when the parent resource of the relationship 
+does not exists. For example, when `/articles/1` does not exist, request to
```
dgeb: s/exists/exist/

hhware:
@dgeb, sorry for the typo! Updated.

Issue #460, opened 2015-03-14
Creating More Extensions From Base Spec Contents


ethanresnick:
It seems like, broadly, there are two types of **MAY**s in the base spec. 

The first describe elements that the server can optionally provide without the client explicitly requesting them, but that the client _must_ know how to handle if they're present. Pagination links are a good example of this: if the client doesn't know how to follow them, it won't know how to get the whole collection. Meta information, `"related"` urls, and which style of linkage to use are other examples too. Let's call these "server opt-in" features.

The second type are features that the client can optionally ask the server to provide, but that the server may or may not support. Here, I'm talking mainly about sorting and the inclusion of related resources. Let's call these "client opt-in" features. With these features, the client _must have some way to detect whether it's request to use a feature was honored_. For example, there's currently no way for a client that appends a `sort` parameter to a request to know whether the server actually sorted the resources or just ignored the parameter. That's not good. With included resources, the story's a little better, because the client can check the response payload for the `"included"` key, but might still be nice for clients to know in advance whether their `include` parameter would be honored.

My proposal is that, because the client must be able to tell whether a "client opt-in" request was honored, all of these features should be defined as extensions, so that support for them can be detected with the `supported-ext` parameter. And, now that we've allowed for multiple extensions to be used in tandem, this shouldn't cause any problems.

I think this distinction between "server opt-in" and "client opt-in" optional features, and only making extensions from the latter type, addresses @dgeb's comment in #432 that:

> If every MAY in the base spec was eliminated, then a base implementation couldn't include self links, pagination links, related links, meta data, etc. and would require clients to negotiate for inclusion of any these. This seems unnecessarily strict and limiting by default.

Here's another advantage to this distinction: it captures the fact that "server opt-in" features are actually less optional than "client opt-in" ones. Yes, a server may or may not provide `"related"` links but, if it does, a 1.0 client MUST know how to use these links to fetch the related resources. So, for clients, these optional features aren't optional. Meanwhile, the "client opt-in" features truly are optional for both parties: the client needn't request them and the server needn't support them. So it makes sense to treat these two classes of things differently, which is what making the latter into extensions does.

Going forward, all additions to JSON API are going to be either: 1) ignorable by older clients or 2) "client opt-in" features. None of them will be "server opt-in" features, in the sense that I've defined the term above, because requiring a client to know how to interpret payload fields that we add post 1.0 wouldn't be backwards compatible.

Thoughts on this?

[...]

PR #473, opened 2015-03-16
Ordering of primary data

hhware:
IMHO, not enough attention is given in the spec to the way of specifying of order of results in the response, especially given that the specification of sorting in the request is quite detailed. I believe it is just assumed that the way to specify order of results is to order elements in the top-level data, but AFAIK it is not mentioned explicitly anywhere. The only reference to ordering I was able to find is in the Recommendations: "because order is significant" . I suggest to mention this explicitly in the base spec.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -825,6 +825,13 @@ The above example should return the newest articles first. Any articles
 created on the same date will then be sorted by their title in ascending
 alphabetical order.
 
+If sorting is supported by the server and requested by the client via 
+query parameter `sort`, the server **MUST** return elements of the 
+top-level `data` array of the response ordered according to the criteria 
+specified. 
+The server **MAY** apply default sorting rules to top-level `data` 
+if request parameter `sort` is not specified. 
+
 ### Pagination <a href="#fetching-pagination" id="fetching-pagination" class="headerlink"></a>
 
 A server **MAY** choose to limit the number of resources returned in a response


hhware:

ethanresnick:
Note that merging this would make supporting sorting non-optional, which is a change from what we currently have. (Right now, the spec says "A server MAY choose to support requests to sort resource collections".) 

Making sort support mandatory could be a good thing anyway because, right now, a generic client has no way to know whether their sort request was honored. The other solution is to leave sorting optional, but require servers to state whether they support it by making it an extension. I've discussed that problem/possibility more in #460.

hhware:
@ethanresnick, thanks for pointing this out! The intention of this PR was not to make sorting non-optional, but to explicitly define ordering in the response, so I corrected it.

ethanresnick:
Cool!

I think the PR should also specify how the sorts would be applied to the to-many relationships, as that's a little vague.

hhware:
@ethanresnick, thanks again for pointing it out! I actually did not suggest that sorting could possibly apply to relationships, that sentence was only about ordering. Corrected again.

ethanresnick:
@hhware, After thinking about this more, I don't think sorts should be able to order the to-many relationship linkage at all (at least not for 1.0), but rather only the resources themselves in a collection. This is what the spec currently specifies, and it's probably not worth changing. So ignore my comment about the vagueness. I'd now suggest removing the text and note about to-many linkage objects all together. After that, I'm +1 on this PR!

hhware:
@ethanresnick, I do not quite understand your point. This PR is not about sorting directive in the request, it is about ordering of objects in the response. If the to-many relationship is stored in the database in ordered form (i.e., it is stored in a separate table with a column `order_number INT`), the server may return them in that order. The suggested **MAY** is only saying that the order of linkage objects in the array may indeed be a representation of an ordered relationship, similar to order of elements in primary `data` in the absence of `sort` may reflect default ordering used by the server.

ethanresnick:
Ahh, ok, I understand now! 

As far as the ordered relationships go: saying even that the relationship entries could represent an ordered list seems inconsistent with the other places in the spec where we're currently saying that set semantics apply (e.g. when POSTing to or DELETEing from a to-many relationship).

I'm tentatively in favor of treating relationships as lists rather than sets, which would allow for repeated entries and for order to matter, but I think that discussion would have to happen in a separate issue or PR, since it's a big change. Meanwhile, while the current (if implicit) set-based definition is implicit, I still think the note about ordering doesn't belong.

hhware:
@ethanresnick, I agree with you. I will work on addressing ordering of relationships elsewhere. Updated.

ethanresnick:
Cool. This looks good to me now!

tkellen:
:+1: to merge this PR as is.

dgeb:
Thanks @hhware :shipit: 

FWIW, I just moved the other paragraph that covers support for sorting closer to this one. See 12d5708a4ed3c22c9a6039c916adcbad17da070c

hhware:
> how the sorts would be applied to the to-many relationships

@ethanresnick, a way to sort relationship linkage object arrays is proposed in https://github.com/json-api/json-api/pull/431#issuecomment-90378980

PR #474, opened 2015-03-16
Reserved members of complex attributes

hhware:
I suggest to turn it to a RFC verb. IMHO, it should be **MUST** instead, but it could be argued that **SHOULD** could be more appropriate.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -191,7 +191,7 @@ be represented under the "links object".
 
 "[Complex attributes]" are [attributes] whose value is an object or array with
 any level of nesting. An object that constitutes or is contained in a complex
-attribute must reserve the `id`, `type`, `links`, and `meta` members for future
+attribute **MUST** reserve the `id`, `type`, `links`, and `meta` members for future
 use.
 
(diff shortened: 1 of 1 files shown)


bintoro:
Oh yeah, this one. I intentionally used a non-normative word for consistency with other objects. There's no object in the spec for which reserved members are actually formally reserved. I agree they should be, but then this is far from the only case to be considered. And I think the word here needs to be a MUST.

hhware:
> There's no object in the spec for which reserved members are actually formally reserved

@bintoro, sorry, do not quite understand what you mean by it. Is not this a formal reservation:
"A resource object **MUST** contain at least the following top-level members"

> And I think the word here needs to be a MUST.

I also think it should be **MUST**

bintoro:
@hhware I've changed my mind and would be happy to see this PR merged if you could just replace the SHOULD with a MUST. I concluded that the situation's not exactly comparable to the other objects.

hhware:
@bintoro, thanks for reviewing! Updated.

tkellen:
:+1:

PR #479, opened 2015-03-16
Clarify link object requirements

ethanresnick:
Under the current spec, this is technically valid:

```
{
  "links": {
    "author": {
      "meta": {}
    }
  }
}
```

But clearly that's not enough information for a client to figure out the link.

This PR clarifies that it's actually at least one of the `related`or `linkage` keys that's required, with `self` and `meta` then being optional extras.

We could also say that at least one of: `"related"`, `"linkage"`, or `"self"` is required, with only `"meta"` being completely optional. If we do that, then this would be valid:

```
{
  "links": {
    "author": {
      "self": "http://example.com/people/1/links/author"
    }
  }
}
```

Technically, that provides enough information for a client to figure out the linked resources, but I don't think we should allow this. Having implementors put either a `related` or `linkage` key doesn't seem like a big burden, and it saves the client a request (since in the `self`-only case it has to request that url just to get the linkage).

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -277,12 +277,15 @@ The value of a relationship **MUST** be one of the following:
 If a relationship is provided as a link object, it **MUST** contain at least
 one of the following:
 
+* A `"related"` member, whose value is a related resource URL (as defined above).
+* A `"linkage"` member, whose value represents "resource linkage".
+
+The link object **MAY** also contain:
 * A `"self"` member, whose value is a URL for the relationship itself (a
   "relationship URL"). This URL allows the client to directly manipulate the
   relationship. For example, it would allow a client to remove an `author` from
   an `article` without deleting the `people` resource itself.
-* A `"related"` member, whose value is a related resource URL (as defined above).
-* A `"linkage"` member, whose value represents "resource linkage".
+
 * A `"meta"` member that contains non-standard meta-information about the
   relationship.
 


hhware:
IMHO, these restrictions are unnecessary. I think if several options are provided, they should all be treated equally (same logic as in https://github.com/json-api/json-api/pull/453#issuecomment-81445679). In particular, `meta` is provided to accommodate obscure cases when none of other 3 approaches works, why enforce something traditional along with it? Same applies to `self`: it's not unimaginable that some implementations may prefer it to be the only member of relationship, why disallow that?

Instead, one could require that "`meta` **MUST** be non-empty if it is the only member of a relationship."

ethanresnick:
@hhware Re the `"meta"` only case: the more I think about it, the more I agree with you. I thought that promising generic clients that they'll be able to find any standardly-linked resources was the most important thing, and allowing `"meta"`-only links obviously precludes this. But, as I've thought about it more, I can imagine cases in which a server might want to implement a non-standard linking strategy (e.g. to allow for agent-driven negotiation) while still getting the benefits of responding with a compliant json-api doc.

Re the `"self"` only case: it's always been a question of balance. Allowing `"self"` only obviously makes things harder on the client, but there could be cases where it helps the server a non-trivial amount. So I guess I'm with you on that too.

I'm going to close this.

PR #480, opened 2015-03-16
Clarify relationship formatting

ethanresnick:
**Update** I've edited this description to capture the current state of the PR. It does three things:
1. When relationship information is being sent to the server as part of a PATCH or POST done on a _full resource_ (as opposed to patching/posting to the relationship directly), the format for sending this information is the same as its representation in GET requests. (I.e. there's a `links` key with a proper link object.)
2. This link object containing the new relationship information MUST have a linkage key. This clarifies that relationships can only be set by providing "linkage", not by providing `related` urls.
3. It simplifies some of the definitions for PATCHing/DELETEing relationships directly, by introducing the concept of a linkage object. It doesn't go so far as to use link objects to represent single relationships retrieved at/sent to relationship URLs. That discussion is happening here and in #482. If that's desired, I can update the PR further.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -295,11 +295,10 @@ document, it **MUST** include resource linkage to those resource objects.
 Resource linkage **MUST** be represented as one of the following:
 
 * `null` for empty to-one relationships.
-* an object containing `"type"` and `"id"` members for non-empty to-one
-  relationships.
+* a "linkage object", which is an object containing `"type"` and `"id"` members,
+  for non-empty to-one relationships.
 * an empty array (`[]`) for empty to-many relationships.
-* an array of objects each containing `"type"` and `"id"` members for non-empty
-  to-many relationships.
+* an array of linkage objects for non-empty to-many relationships.
 
 > Note: Resource linkage in a compound document allows a client to link
 together all of the included resource objects without having to `GET` any
@@ -1210,7 +1209,7 @@ described below.
 The `PATCH` request **MUST** include a top-level member named `data` containing
 one of:
 
-* an object with `type` and `id` members corresponding to the related resource.
+* a linkage object (defined above) corresponding to the new related resource.
 * `null`, to remove the relationship.
 
 For example, the following request updates the author of an article:
@@ -1246,8 +1245,7 @@ A server **MUST** respond to `PATCH`, `POST`, and `DELETE` requests to a
 *to-many relationship URL* as described below.
 
 For all request types, the body **MUST** contain a `data` member whose value
-is an object that contains `type` and `id` members, or an array of objects
-that each contain `type` and `id` members.
+is a linkage object or an array of linkage objects.
 
 If a client makes a `PATCH` request to a *to-many relationship URL*, the
 server **MUST** either completely replace every member of the relationship,
--- a/format/index.md
+++ b/format/index.md
@@ -912,9 +912,9 @@ Accept: application/vnd.api+json
 }
 ```
 
-If the resource object being sent contains any relationships in its `"links"`
-key, those relationships **MUST** each have a `"linkage"` key that includes the
-linkage the new resource is to have.
+If a relationship is provided in the `links` section of the resource object, its
+value **MUST** be a link object with a `linkage` member. The value of this key
+represents the linkage the new resource is to have.
 
 #### Client-Generated IDs <a href="#crud-creating-client-ids" id="crud-creating-client-ids" class="headerlink"></a>
 
--- a/format/index.md
+++ b/format/index.md
@@ -903,10 +903,19 @@ Accept: application/vnd.api+json
   "data": {
     "type": "photos",
     "title": "Ember Hamster",
-    "src": "http://example.com/images/productivity.png"
+    "src": "http://example.com/images/productivity.png",
+    "links": {
+      "photographer": {
+        "linkage": { "type": "people", "id": "9" }
+      }
+    }
   }
 }
 ```
+
+If the resource object being sent contains any relationships in its `"links"`
+key, those relationships **MUST** each have a `"linkage"` key that includes the
+linkage the new resource is to have.
 
 #### Client-Generated IDs <a href="#crud-creating-client-ids" id="crud-creating-client-ids" class="headerlink"></a>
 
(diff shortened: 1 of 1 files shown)


ethanresnick:
Note: To match the behavior with `PATCH` exactly, it might make more sense to require the representation to simply be:

```
"links": {
  "photographer": { "type": "people", "id": "9" }
}
```

I.e., with no linkage key. But then we'd have to update the text to say that what's being `POST`ed isn't a strictly-speaking resource object.

ethanresnick:
Ok, this PR now makes sure that the full-resource representations mirror one another, whether the resource is sent in a PATCH or POST or retrieved with a GET. The representations of link objects at relationship urls are also consistent whether sending or retrieving. But this PR doesn't go so far as to use link objects to represent single relationships retrieved at relationship URLs. If that's desired, I can update the PR further.

bintoro:
Nice job @ethanresnick!

> But this PR doesn't go so far as to use link objects to represent single relationships retrieved at relationship URLs.

Whether we want relationship URLs to return the extra layer provided by _link objects_ is one discussion, but in any case I would like to reuse the concept of _linkage_ in that section, too. Makes no sense to talk about "arrays of objects with `type` and `id` members" when the format has already been defined elsewhere.

Of course, linkage objects might gain further members that are not applicable to create/update operations, but I don't see a problem with just excluding them as they're defined.

ethanresnick:
Hey guys,

I've updated my original post to describe exactly what this PR now does.

I think there's only one thing left to discuss here... Right now, we only allow the user to set relationship values by providing linkage, rather than, e.g. a `related` url. While this PR makes that design decision more explicit, I'm wondering if we want to preserve the flexibility for JSON API to become truly hypermedia-based/distributed in the future, which would [require the ability](https://github.com/json-api/json-api/pull/357#issuecomment-76952278) to set a resource's value to an arbitrary url.

If we do want to preserve ourselves that option, I think it's as easy as replacing language like "the value **MUST** be a link object which **MUST** contain a `linkage` key" with language like "the value **MUST** be a link object which **SHOULD** contain a `linkage` key. Servers **MAY** reject attempts to set the value of a relationship when `linkage` is not provided."

format/index.md
```diff
@@ -903,11 +902,20 @@ Accept: application/vnd.api+json
...
+      }
+    }
   }
 }
 ```
 
+If the resource object being sent contains any relationships in its `"links"`
+key, those relationships **MUST** each have a `"linkage"` key that includes the
+linkage the new resource is to have.
+
```
dgeb, line 920: This phrasing seems awkward. How about keeping it similar to the note below for PATCH:

> If a relationship is provided in the `links` section of a resource object, its value **MUST** be a link object with a `linkage` member.

format/index.md
```diff
@@ -1244,8 +1245,7 @@ A server **MUST** respond to `PATCH`, `POST`, and `DELETE` requests to a
 *to-many relationship URL* as described below.
 
 For all request types, the body **MUST** contain a `data` member whose value
-is an object that contains `type` and `id` members, or an array of objects
-that each contain `type` and `id` members.
+is a linkage object or an array of linkage objects.
```
dgeb: This correction and the original are both incorrect - it should read:

> is an empty array or an array of linkage objects.

format/index.md
ethanresnick, line 920: Updated!

format/index.md
ethanresnick: Fixed! This is why I should think while updating, rather than blindly find-and-replace-ing :)

hhware:
> I'm in favor of the term "linkage object", which seems consistent with "link object" and "resource object". 

I am totally fine with linkage objects and have no special feelings for relationship tuples, but I have a small reservation. If one compares linkage object with link object, then the latter is what is keyed by `links`, whereas one cannot say that linkage object is what is keyed by `linkage` -- it is wrong in case of to-many relationship. This can be confusing for readers. If this minor thing does not bother anyone else, please ignore it.

Also, I would suggest to extract definition of linkage object into a separate paragraph, not introduce it "on the run":

> a "linkage object", which is an object containing...

ethanresnick:
@hhware 

> Also, I would suggest to extract definition of linkage object into a separate paragraph, not introduce it "on the run":

I completely agree with this in principle. I only introduced the definition in line for consistency with how the spec currently introduces many of its smaller definitions (e.g. "link object" and "relationship urls" are both defined in parentheticals). 

That said, I think the spec would really benefit from having a glossary of terms at the bottom, with every usage of a defined term in the body text linked to its entry in the glossary. If others are onboard with having such a glossary, we could define "linkage object" there.

> If one compares linkage object with link object, then the latter is what is keyed by links, whereas one cannot say that linkage object is what is keyed by linkage... If this minor thing does not bother anyone else, please ignore it.

I'll let others chime in about this, but it doesn't particularly bother me.

bintoro:
> I would suggest to extract definition of linkage object into a separate paragraph

Agreed. Defining the thing as we're explaining to-one relationships doesn't seem right.

I would also like to add a tiny definition for it, similar to what I'm trying to do to _linkage_ here: https://github.com/json-api/json-api/pull/498/files#diff-26e38355b69056b0db1c2b1c612241b2R301

Something like

> A "linkage object" identifies an individual related resource. It **MUST** be represented by an object containing `"type"` and `"id"` members.

ethanresnick:
Ok! I've updated the PR to include the "linkage object" definition in its own paragraph, along the lines @bintoro suggested.

bintoro:
> I've updated the PR to include the "linkage object" definition in its own paragraph

:+1:

> I think the spec would really benefit from having a glossary of terms at the bottom

Sure, but I'm not crazy about the idea of actually removing definitions from their main context and stashing them only in some appendix.

As I've said elsewhere, I'm planning a big PR that would improve the formatting of primary definitions and add a ton of intra-document links. I just haven't been able to submit it because of the current backlog of conflicting PRs.

Let's re-evaluate the need for an index/glossary after the intra-doc links are in place. As long as stuff is defined in the body text, I view a full glossary as overkill. At most an alphabetical index linking to the primary definitions might be useful.

ethanresnick:
@bintoro sounds good :)

dgeb:
Thanks @ethanresnick et al - merging now. I believe that this removes the last remaining asymmetry in request / response documents. As a bonus, it tightens up the terminology used in the spec with the definition of "linkage object".

ethanresnick:
Great! Thanks @dgeb!

PR #494, opened 2015-03-19
Added Spine library (Swift) to examples

wvteijlingen:
I've been working on this Swift library for some time now, and while it is still in flux, I thought it would be nice to have it in the list of examples.


steveklabnik:
Thanks!

PR #498, opened 2015-03-21
Explain linkage & relationship URL

bintoro:
Adds a basic definition for _resource linkage_ and a better explanation for why a relationship URL is useful. ~~Also moves the _related resource URL_ note next to the definition, since it had ended up in a funny place.~~ (No longer applicable.)


tkellen:
This is great @bintoro! I'm :+1: to this; @dgeb?

bintoro:
Actually, I think it's better to settle #453 first and incorporate these changes later.

bintoro:
This one has now been rebased in case there's still interest in merging it.

tkellen:
Thanks for the effort here @bintoro, I think we're good now though!

Issue #500, opened 2015-03-23
Remove top-level "self" link


ethanresnick:
The `self` link in the top-level `links` object doesn't seem to add anything. Meanwhile, it actually introduces a bit of confusion, namely: whether the link should point to the HTTP resource being requested or the JSON API resource (i.e. the HTTP entity).

Consider a request to an `/articles?filter[is-latest]=true` endpoint that always returns a single resource. If the top-level `self` key's value should be the URI of the HTTP resource (i.e. `/articles?filter[isLatest]=true`), then it's redundant, as the client already knows that URI. On the other hand, if the top-level `self` link is supposed to point to the entity—which seems like a bad idea anyway—then it's also redundant because that link can be given within the primary resource object.

So where's the value here?

ethanresnick:
Edit: from the examples [here](http://jsonapi.org/format/#fetching-resources-responses-200), it does seem like this key's value is supposed to be the HTTP resource, and not the entity. So that's good (though perhaps worth clarifying).

And, as far as the value goes...I guess it could help clients that don't realize they've been redirected or have thrown away the original url they used for the request (but for some reason have hung onto the payload). That's probably reason enough to allow this as an option. So I'm going to contradict myself and close this.

bintoro:
>  it does seem like this key's value is supposed to be the HTTP resource

Yes. I think it's pretty clear actually: "a link for fetching the data in the response document".

However, what's not at all clear is whether top-level `self` is supposed to be a static URL like resource-level `self` or if a request to, e.g., a related resource URL can return that same URL as `self`.

In other words, are repeated requests to a top-level `self` URL allowed to return completely different resources over time?

ethanresnick:
@bintoro Whether "repeated requests to a top-level self URL [are] allowed to return completely different [json api] resources over time" is exactly what I was asking. But, seeing as they seem to point to the HTTP resource, the answer should be yes. Check out the example where the top-level `self` url is `"http://example.com/articles/1/author"`

bintoro:
@ethanresnick Aha, I see. Well, in that case the definition "a link for fetching the data in the response document" is wrong because the response is not at all guaranteed to return the same data.

The definition should be something like "the URL that returned the response document". And if that's the definition, then it's a legitimate question what the whole thing is needed for.

PR #501, opened 2015-03-23
Clarify top-level self's description

ethanresnick:
As I discovered in the process of opening #500, reading the spec, and then convincing myself that #500 wasn't necessary, the current definition of the top-level `self` link is a little vague. This PR aims to fix that.


tkellen:
:+1:

PR #503, opened 2015-03-23
Restore "Examples" page to hold actual usage examples

bintoro:


dgeb:
We'll need a wider layout such as in #531 in order to accommodate the extra heading.

dgeb:
I think we could merge this now with the new layout. I'd like to add examples like in #667 outside of the base spec.

tkellen:
:+1:

joananeves:
:+1:

PR #485, opened 2015-03-16
Reorganize the Extending section

bintoro:
The _Extending_ section is growing a bit unwieldy and is difficult to understand, particularly the relationship between the two request headers.

This PR shuffles parts of it around to achieve a story like this:
1. What it's about
2. Advertising extension support
3. Requesting an extension
4. Providing the extension
5. Formulating request/response documents
6. Examples

Changes in addition to reordering + examples:
- Removes `Content-type` from the "Servers **MUST NOT** provide extended functionality..." clause. That is, when a feature is such that it must be requested in a header, `Accept/ext` is always required, and `Content-type/ext` is only used to describe the body when applicable.
- Describes the `ext`/`supported-ext` media type parameter format only once.
- Removes unnecessary mention that `ext` must be a subset of `supported-ext`.
- Condenses the "Since extensions can contradict one another..." note a lot.
- Adds explanatory note about requesting/providing additive functionality.
- Minor wording tweaks.

PR #506, opened 2015-03-25
Require quoting of media type parameters in the Accept header

ethanresnick:
The HTTP spec requires that the value of parameters in the `Accept` header be quoted for the header to parse correctly. This PR merely makes that requirement explicit, and updates the example headers to show the requirement in action. The relevant part of the HTTP spec is [section 2.2](http://tools.ietf.org/html/rfc2616#section-2.2), which says that “These special characters MUST be in a quoted string to be used within a parameter value”.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -40,7 +40,7 @@ parameter **MUST** be a comma-separated (U+002C COMMA, ",") list of
 extension names.
 
 For example: a response that includes the header `Content-Type:
-application/vnd.api+json; supported-ext=bulk,jsonpatch` indicates that the
+application/vnd.api+json; supported-ext="bulk,jsonpatch"` indicates that the
 server supports both the "bulk" and "jsonpatch" extensions.
 
 If an extension is used to form a particular request or response document,
@@ -52,10 +52,10 @@ extensions supported by the server, which are listed in `supported-ext`
 of every response.
 
 For example: a response that includes the header `Content-Type:
-application/vnd.api+json; ext=ext1,ext2; supported-ext=ext1,ext2,ext3`
+application/vnd.api+json; ext="ext1,ext2"; supported-ext="ext1,ext2,ext3"`
 indicates that the response document is formatted according to the
 extensions "ext1" and "ext2". Another example: a request that includes
-the header `Content-Type: application/vnd.api+json; ext=ext1,ext2`
+the header `Content-Type: application/vnd.api+json; ext="ext1,ext2"`
 indicates that the request document is formatted according to the
 extensions "ext1" and "ext2".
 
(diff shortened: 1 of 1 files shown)


bintoro:
> The HTTP spec requires that the value of parameters in the Accept header be quoted for the header to parse correctly.

Doesn't it just boil down to whether the value contains a comma? Both `Accept` and `Content-Type` seem to allow either `token` or `quoted-string` as the param value.

So, while it's true that `ext="ext1,ext2"` must indeed be quoted, I'm confused by the note that makes a distinction between `Accept` and `Content-Type`.

ethanresnick:
@bintoro D'oh! Of course you're right. The distinction that I was making in my head was that a parser can probably resolve the ambiguity in the `Content-Type` case because it's value isn't a comma-separated list. But, from HTTP's point of view, the value is still invalid if there's an unquoted comma. I've updated the PR accordingly.

dgeb:
@ethanresnick could you please rebase?

ethanresnick:
@dgeb What do you want me to rebase exactly? I squashed my two earlier commits, but I'm not sure if you're asking for something else as well.

steveklabnik:
@ethanresnick github says "we cannot merge this PR", which means that you need to rebase off of the current `gh-pages`.

bintoro:
I'm wondering if the separate note is necessary. The Extending section is getting ridiculously long as it is...

If the rationale really must be mentioned, can't we just append it to the requirement? Like so:

> When the value of the `ext` or `supported-ext` media type parameter contains a comma (U+002C COMMA, ","), the value **MUST** be surrounded with quotation marks (U+0022 QUOTATION MARK, """) in order to conform to the HTTP specification.

Also, to avoid specifying the code point for a comma again:

> When the value of the `ext` or `supported-ext` media type parameter contains more than one item, it **MUST** be [...]

ethanresnick:
@steveklabnik Gotcha. Done!

@bintoro I agree in theory, but I feel like we can worry about making the spec more concise post 1.0. Let's just lock down the content now, so we can tell implementors that things are set, and then we'll worry about readability.

dgeb:
@ethanresnick I'm sorry, but I agree with @bintoro's more concise rewording here. You make a good point about the correctness of the spec, but if it's not concise and readable then we have trouble determining correctness. Would you mind updating one more time?

bintoro:
If #485 is going somewhere, the edits I suggested can be incorporated there while I rebase it.

ethanresnick:
@dgeb No worries! I've updated the PR to use more-concise language based on @bintoro's suggestions.

And if this change can be accomplished as part of #485, that's fine too!

dgeb:
Thanks @ethanresnick!

@bintoro I'm not opposed to reorganizing this verbose section, as you've done in #485. If you're up for rebasing with these new additions, we can re-evaluate it.

Issue #278, opened 2014-09-26
Add support for heterogeneous linked data collection


ahacking:
This issue follows some brief discussion in #276, and proposes that the `linked` key ought to accept a heterogeneous collection, just as the `data` key currently does.

The motivation for this proposal is to:
- support identical processing rules for both `data` and `linked` collections
- allow ordering on attributes which may span resource types, eg `updatedAt`
- reduce server side collation overhead, specifically having the server sort resources into typed collections is not friendly to large result sets or streaming.

The above proposal does not intend to address streaming, if there is a desire to solve the streaming/fire-house type problem then additional consideration should also be given to the rule regarding only including a resource once.  That particular rule introduces an unavoidable requirement that a server must perform additional processing or book-keeping to prevent multiple instances of a resource being serialised.  This means servers must solve the [Hamiltonian path problem](http://en.wikipedia.org/wiki/Hamiltonian_path_problem) in order to emit a valid JSON API response, see also [Hamiltonian Path](http://en.wikipedia.org/wiki/Hamiltonian_path).  This is an NP-Complete problem and not easy to solve given an arbitrary graph of interconnected resources.  The partitioning of resources into`data` and `linked` collections also introduces the need to defer or double process results to first emit the match set, and then emit the linked set.  If one NEVER emits linked data, this entire problem reduces to a simpler book keeping exercise, but still requires keeping history as time goes by. If the duplicate resource rule is relaxed, no book keeping history is required and streaming results of any size/length over and arbitrary time period is possible and if partitioning linked data from the match set is removed, linked data can also be serialized.  It is hard to reason that repeating a resource could actually cause trouble in practice, also given a long enough running process this could actually be used to refresh resources and be beneficial. Simple optimisation techniques _could_ be used (eg a small history window) to avoid repetition in small clusters of interconnected resources, but by no means provide a hard guarantee.

dgeb:
I feel that this is consistent with the `data` key for the primary resource. If that is allowed in the spec, then I believe this should be.

I suppose the larger question, as we talk about simplification, is whether both should be allowed in the spec for 1.0. Support for heterogeneous collections is more implicit than explicit currently. If the spec explicitly allows heterogeneous collections, then that leaks into complexities with polymorphic relationships, which [I've explored unsuccessfully](https://github.com/json-api/json-api/issues/276#issuecomment-56960142).

[...]

PR #513, opened 2015-03-28
Relax restriction on repeated, included resources

ethanresnick:
Back in #278, everyone agreed that this should be a **SHOULD NOT** rather than a **MUST NOT**. 

@ahacking pointed out that:

> If the duplicate resource rule is relaxed, no book keeping history is required and streaming results of any size/length over and arbitrary time period is possible...It is hard to reason that repeating a resource could actually cause trouble in practice, also given a long enough running process this could actually be used to refresh resources and be beneficial.

My guess is that this got rather mindlessly converted into a **MUST** during the _remove all the **SHOULD**s crusade_ that was part of rc3's simplification effort. If I'm wrong on that, though, and there was a more nuanced reason for making this a **MUST**, let me know!

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -399,7 +399,7 @@ A complete example document with multiple included relationships:
 }
 ```
 
-A compound document **MUST NOT** include more than one resource object for
+A compound document **SHOULD NOT** include more than one resource object for
 each `type` and `id` pair.
 
 > Note: In a single document, you can think of the `type` and `id` as a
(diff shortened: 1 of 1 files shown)


tkellen:
From an implementation standpoint this bookkeeping is trivial and does not prevent streaming.

Could you or @ahacking please provide a concrete example of the need for this?

ethanresnick:
@tkellen I've been thinking more about this and would actually be happy to withdraw this PR, unless @ahacking wants to defend it

tkellen:
Roger that @ethanresnick, I'm going to give this a close. @ahacking, if you can bring a concrete use-case I would reconsider opening this.

PR #514, opened 2015-03-28
Handling of unsupported sorting requests

hhware:
IMHO, it would be appropriate to require servers which do not support sorting requested via `sort` (either no sorting, or only particular fields do not work) to respond with an error rather than allow them to ignore the `sort` query string parameter.


tkellen:
:+1:

tkellen:
Thanks @hhware!

PR #515, opened 2015-03-28
Clarify relationship formatting: JSON Patch

hhware:
Three things are changed here:
- Replacements with "linkage objects" which I believe were forgotten in #480.
- "Should" is replaced with "**MUST**" for to-one relationships to make it consistent with all similar cases.
- An empty array option is added for to-many relationships -- I guess it is an omission, because it is allowed in the base spec.
- [added in the second attempt] Same for to-one relationship

### DIFF
--- a/extensions/jsonpatch/index.md
+++ b/extensions/jsonpatch/index.md
@@ -96,8 +96,8 @@ target a particular relationship's URL.
 #### Updating To-One Relationships <a href="#patch-updating-to-one-relationships" id="patch-updating-to-one-relationships" class="headerlink"></a>
 
 To update a to-one relationship, perform a `"replace"` operation with a URL and
-`"path"` that targets the relationship. The `"value"` should be a link object
-that contains `"type"` and `"id"` members.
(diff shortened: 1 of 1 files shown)


extensions/jsonpatch/index.md
```diff
@@ -96,8 +96,7 @@ target a particular relationship's URL.
 #### Updating To-One Relationships <a href="#patch-updating-to-one-relationships" id="patch-updating-to-one-relationships" class="headerlink"></a>
 
 To update a to-one relationship, perform a `"replace"` operation with a URL and
-`"path"` that targets the relationship. The `"value"` should be a link object
-that contains `"type"` and `"id"` members.
+`"path"` that targets the relationship. The `"value"` **MUST** be a linkage object.
```
dgeb: The last sentence should read:

> The `"value"` **MUST** be a linkage object or `null`, to remove the relationship.

hhware:
@dgeb, sorry to miss that! Updated.

dgeb:
@hhware thanks for correcting.

:+1: LGTM.

tkellen:
:+1:

PR #526, opened 2015-03-29
Include request accepted in the possible response.

milap-neupane:
Inspired from the git-hub API for the statistics there are many API's that respond with the stats immediately if the cache is available else the server response with a 202 ACCEPTED response for the request being accepted and processes the request in background. Once there is a cache available the same request can return the response with the required resource.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -1016,6 +1016,10 @@ Content-Type: application/vnd.api+json
 }
 ```
 
+##### 202 Accepted <a href="#creating-resources-responses-202" id="creating-resources-responses-202" class="headerlink"></a>
+
+A server **MUST** return `202 Accepted` when the request needs background processing.
+
 ##### 204 No Content <a href="#crud-creating-responses-204" id="crud-creating-responses-204" class="headerlink"></a>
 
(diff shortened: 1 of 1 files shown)


ethanresnick:
I'm absolutely +1 on this. Seems like a worthy case to support in 1.0. The PR should be updated though to cover creating and updating resources as well, rather than just fetching them.

milap-neupane:
I was just wondering if it is actually needed for creating and updating. So initially i just added them to the fetching part. I too think it would be better to update it. I'll update the pull request soon.

josisusan:
:+1:

tkellen:
Thanks for this @milap-neupane, you're absolutely right that this should be supported. I agree with your PR but I don't want to merge it because it addresses a symptom of a larger problem rather than the problem itself.

I think it has become clear that we've gone a little too far with our MUSTs in the specification around response codes. I'm going to close this--we are going to relax these requirements before 1.0 (#534), vs special casing every possible status code.

milap-neupane:
Thank you @tkellen for the response. Looking forward for the status codes to be implemented in future versions.

PR #529, opened 2015-03-30
Allow 200 OK when client's attributes are up to date after PATCH

ethanresnick:
Right now, on resource creation, we allow a `200 OK` response with a representation of the newly-created resource instead of requiring a `204`, even when the client's attributes remain up to date. We say:

> the server MUST return _either_ a 201 Created status code and response document (as described above) or a 204 No Content status code with no response document. (emphasis added)

This PR applies that same rule to `PATCH` updates. Allowing a `200 OK` response even when the client's content remains up-to-date—instead of mandating a `204`—does little harm, but it can make it much easier to write a conforming implementation, as a library that implements JSON API won't have to check whether other fields (e.g. an `updated-at` field) were updated behind it's back, which would require diffing the resource and/or coordinating with the ORM/database schema. Instead, it can air on the safe side by always sending a 200 with the new resource respresentation, unless it explicitly knows that this is unecessary.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -1087,8 +1087,9 @@ If a relationship is provided in the `links` section of a resource object in a
 `PATCH` request, its value **MUST** be a link object with a `linkage` member.
 The relationship's value will be replaced with the value specified in this member.
 
-For instance, the following `PATCH` request will update the `title` attribute
-and `author` relationship of an article:
+A resource's attributes and relationships **MAY** be updated together in a single
+request. For instance, the following `PATCH` request will update the `title`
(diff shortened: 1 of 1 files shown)


tkellen:
Thanks for this @ethanresnick, you're absolutely right that this should be supported. I agree with your PR but I don't want to merge it because it addresses a symptom of a larger problem rather than the problem itself.

I think it has become clear that we've gone a little too far with our MUSTs in the specification around response codes.  I'm going to close this--we are going to relax these requirements before 1.0 (#534), vs special casing every possible status code.

PR #530, opened 2015-03-30
Clarify null vs. 404 responses; possible shapes of primary data

ethanresnick:
Filters allow for the creation of endpoints that conceptually represent a  single json-api resource or `null`, in the case that no resource currently exists which satisfies the filter's conditions. In the PR, I give as an example: `/posts?filter[is-latest]=true`, which will always return one post, except when the `posts` collection is empty in which case it doesn't pick out any posts. Another example would be `/events?filter[up-next]=true`, which would return the next upcoming event, or null if all the events in the `events` collection are in the past (or if the `events` collection is empty). 

Responses from these filtered endpoints should be allowed to be `null` too, as that's more appropriate than responding with an empty array (since the endpoints don't conceptually represent a collection) and it's more consistent than responding with 404. (If we were to 404 these requests, we'd probably be compelled to 404 the empty to-one relationship case, which I don't think we want to do.) So this PR merely updates the spec to note this edge case as a circumstance that allows `null` primary data.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -140,9 +140,14 @@ of resources, or resource relationship primarily targeted by a request.
 
 A document **MUST** contain either primary data or an array of [error objects](#errors).
 
-Primary data **MUST** appear under a top-level key named `"data"`. Primary
-data **MUST** be either a single resource object, an array of resource
-objects, or a value representing a resource relationship.
+Primary data **MUST** appear under a top-level key named `"data"`.
+
+Primary data **MUST** be either:
+
+* a single resource object or `null`, for requests that target single resources
+* an array of resource objects or an empty array ([]), for requests that target
+resource collections
+* resource linkage, for requests that target a resource's relationship
 
 ```javascript
 {
@@ -156,8 +161,6 @@ objects, or a value representing a resource relationship.
 
 A logical collection of resources (e.g., the target of a to-many relationship)
 **MUST** be represented as an array, even if it only contains one item.
-A logically singular resource (e.g., the target of a to-one relationship)
-**MUST** be represented as a single resource object.
 
 Error objects **MUST** appear under a top-level key named `"errors"`.
 
@@ -577,8 +580,13 @@ A server **MUST** respond to a successful request to fetch an individual
 resource with a *resource object* or `null` provided as the response
 document's primary data.
 
-> Note: `null` is only an appropriate response for fetching a to-one
-related resource URL to indicate the absence of a resource in the relationship.
+`null` is only an appropriate response when the requested URL is one that
+might correspond to a single resource, but doesn't currently.
+
+> Note: Consider, for example, a request to fetch a to-one related resource URL.
+This request wold respond with `null` when the relationship is empty (such that
+the URL is corresponding to no resources) but with the single related resource's
+resource object otherwise.
 
 For example, a `GET` request to an individual article could return:
 
@@ -620,12 +628,9 @@ Content-Type: application/vnd.api+json
 
 ##### 404 Not Found <a href="#fetching-resources-responses-404" id="fetching-resources-responses-404" class="headerlink"></a>
 
-A server **MUST** return `404 Not Found` when processing a request to fetch
-a resource that does not exist.
-
-> Note: When requesting a related resource that is not present, a server
-**MUST** respond with `200 OK` and `null` or an empty array (`[]`) as the
-response document's primary data, as described above.
+A server **MUST** respond with `404 Not Found` when processing a request to
+fetch a single resource that does not exist, except when the request warrants a
+`200 OK` response with `null` as the primary data (as described above).
 
 ##### Other Responses <a href="#fetching-resources-responses-other" id="fetching-resources-responses-other" class="headerlink"></a>
 


bintoro:
If we go down this path, then I'd prefer eliminating the note altogether and leaving it completely up to the server whether an endpoint exposes a conceptually singular resource or a collection.

dgeb:
I'm fine removing the note altogether as well. These kinds of edge cases will just keep cropping up and probably don't merit mention.

tkellen:
Chiming in to agree with @dgeb and @bintoro.

ethanresnick:
Ok, I'll rework this PR tomorrow to remove the note. Do we want to remove the following note as well?

> Note: When requesting a related resource that is not present, a server MUST respond with 200 OK and null or an empty array ([]) as the response document's primary data, as described above.

I'll also update the PR to correct the following line: 

> Primary data MUST be either a single resource object, an array of resource objects, or a value representing a resource relationship.

which should include an explicit mention of `null`, since this PR shows an example where `null` can be the appropriate response even though it's not representing a relationship.

In general, would people support expanding the above line to outline the different "primary data" cases in a bit more detail? I know that, as an implementor, that would've been valuable, as it was a bit of a pain to go through the spec and piece together what all the primary data values could be.

tkellen:
@ethanresnick Yeah, that note needs to go in light of these changes.

I think the current version of this PR is good otherwise.

tkellen:
@ethanresnick Can you update this PR as mentioned so we can land it?

ethanresnick:
@tkellen yep, i'll rework it tonight!

ethanresnick:
@tkellen @dgeb Ok guys, I've updated the PR! 

I've removed the notes that were special casing related resource urls and filters. Instead I've tried to describe, in more generic terms and in the spec's body, when `null` is an appropriate response. I've also updated the definition of what can go in the primary data, to make it truly exhaustive.

Finally, I tried to add a bit more rigor to the "when to 404 vs. when to `null`" question. The current 404 rules use the ambiguous phrase "a resource that does not exist"; that makes sense in the HTTP meaning of a resource, but not in the JSON API one, since a matching JSON API resource doesn't exist in the `null` case either.

ethanresnick:
@tkellen @dgeb Can we get this merged?

tkellen:
Sure can! I'm just jumping on a flight to Boston---can you squash this to a single commit and ping here when you're ready? I'll merge when I land!

format/index.md
```diff
@@ -113,9 +113,14 @@ of resources, or resource relationship primarily targeted by a request.
...
-data **MUST** be either a single resource object, an array of resource
-objects, or a value representing a resource relationship.
+Primary data **MUST** appear under a top-level key named `"data"`.
+
+Primary data **MUST** be either:
+
+* a single resource object or `null`, for requests that target single resources
+* an array of resource objects or an empty array ([]), for requests that target
+resource collections
+* resource linkage, for requests that target a resource's relationship
```
bintoro, line 150: Actually, this is already covered here:

> A logical collection of resources (e.g., the target of a to-many relationship) **MUST** be represented as an array, even if it only contains one item. A logically singular resource (e.g., the target of a to-one relationship) **MUST** be represented as a single resource object.

format/index.md
```diff
@@ -550,8 +555,14 @@ A server **MUST** respond to a successful request to fetch an individual
...
-> Note: `null` is only an appropriate response for fetching a to-one
-related resource URL to indicate the absence of a resource in the relationship.
+`null` is only an appropriate response when the requested URL is one that might
+correspond to a single resource, but doesn't currently.
+
+> Note: To clarify, consider a to-one related resource URL. This URL can
+correspond to a single resource (the related one), but it can also correspond to
+no resources when the relationship is empty. Therefore, a request to fetch a
+to-one related resource URL would respond with `null` when there is no resource
+in the relationship and with a single resource object otherwise.
```
bintoro: I believe the essential content of the note could easily be condensed to a brief mention following the main clause. Consider:

> `null` is only an appropriate response when the requested URL is one that might correspond to a single resource, but doesn't currently (such as the related resource URL of an empty to-one relationship).

format/index.md
ethanresnick, line 150: @bintoro what this change is meant to address is the fact that the current definition of primary data, as containing "a single resource object, an array of resource objects, or a value representing a resource relationship" isn't exhaustive--it excludes the null case, when null is being returned instead of a single resource (i.e. not as linkage). It also makes more explicit the possibility that the array can be empty.

It does mean that the sentences you quoted are somewhat less necessary, but I'd left them for now for clarity

ethanresnick, line 150: I can delete the second sentence you quoted though, to remove some redundancy

bintoro, line 150: Sorry, I totally missed the `null` case. I had actually sketched the exact same list as you did here (sans the nulls) before settling on the separate paragraph I quoted. You're right that the second sentence seems kinda redundant now, but maybe it could remain for completeness. I don't really have an opinion either way.

format/index.md
ethanresnick: I'm working to shorten the note now, which is definitely an improvement, but I do think a standalone note with a bit more text would be helpful for people who aren't super familiar with the spec

ethanresnick:
@tkellen awesome! I'm out now, but will try to push my responses to @bintoro's comments and then squash everything when I'm back at my computer later today! I'll ping you then

ethanresnick:
@tkellen Ok, all squashed and ready to go!

tkellen:
:+1:

Issue #492, opened 2015-03-18
Constraint on types in Bulk POST


ethanresnick:
Currently, the bulk extension just says that every resource I POST must contain a `"type"` member, but it doesn't constrain those types at all, or say how the server should respond to bizarre types.

For example, this should pretty obviously not be allowed:

```
POST /photos
{
  "data": [{"type": "people", "name": "New Person!"}]
}
```

That implies a basic restriction like "The `"type"` value provided for each resource must be among types that constitute the collection represented by the URI that you're `POST`ing to."

But, beyond that basic restriction, I'm wondering if there should be more guidance. E.g., perhaps the server MAY reject the request if all the `"type"`s don't match? Etc.

dgeb:
I think a 422 (Unprocessable entity) would be appropriate. I'm not sure we need to provide guidance for every single error condition, but this might be a good one to include.

ethanresnick:
Yup. And, if we do provide guidance here, presumably it should match the guidance we give for the non-bulk equivalent of this error. For that, we currently require a 409:

> A server MUST return 409 Conflict when processing a POST request in which the resource's `type` does not match the server's endpoint.

tkellen:
This seems good @ethanresnick, would you mind opening a PR?

ethanresnick:
I opened [a PR](https://github.com/json-api/json-api/pull/545) that starts work on this!

PR #545, opened 2015-04-08
POSTing resources with invalid type

ethanresnick:
Start of a PR for #492.

We need to figure out: 

If we're going to explicitly mention the `409` response in the bulk extension, which is what #492 was originally about, we should probably mention the other status codes/responses too. But right now the bulk extension doesn't define any responses at all. So, how do you guys want to handle that?

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -1034,7 +1034,7 @@ A server **MUST** return `409 Conflict` when processing a `POST` request to
 create a resource with a client-generated ID that already exists.
 
 A server **MUST** return `409 Conflict` when processing a `POST` request in
-which the resource's `type` does is not among the type(s) that constitute the
+which the resource's `type` is not among the type(s) that constitute the
 collection represented by the endpoint.
 
 ##### Other Responses <a href="#crud-creating-responses-other" id="crud-creating-responses-other" class="headerlink"></a>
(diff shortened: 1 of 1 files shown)


format/index.md
```diff
@@ -1034,7 +1034,8 @@ A server **MUST** return `409 Conflict` when processing a `POST` request to
 create a resource with a client-generated ID that already exists.
 
 A server **MUST** return `409 Conflict` when processing a `POST` request in
-which the resource's `type` does not match the server's endpoint.
+which the resource's `type` does is not among the type(s) that constitute the
```
dgeb: s/does is not/is not/

dgeb:
Looks good except for that minor typo, @ethanresnick.

As for bulk responses: I think we need to clarify that responses should match those for a single request. In addition, we'll need a section similar to that in the [JSON Patch extension](http://jsonapi.org/extensions/jsonpatch/#patch-responses-other) that describes encountering multiple problems from a single request:

> When a server encounters multiple problems from a single request, the most generally applicable HTTP error code should be specified in the response. For instance, 400 Bad Request might be appropriate for multiple 4xx errors or 500 Internal Server Error might be appropriate for multiple 5xx errors.

ethanresnick:
@dgeb Typo fixed!

I won't have time to work up the bulk responses stuff for a while though. (I think it's a bit less straightforward than it appears.) If that's ok, then we can leave this open and I'll get to it ASAP. Otherwise, feel free to close/merge this as is and open a new issue for the rest, which can be addressed faster.

dgeb:
No need to delay for bulk responses, all of which need to be filled out. Merging now.

PR #547, opened 2015-04-12
Use "linkage object" term in bulk delete explanation

ethanresnick:


ethanresnick:
(Closed this because, on second though, a linkage object is actually slightly different—enough so that it's probably better not to use the term here.)

Issue #476, opened 2015-03-16
Possible conflict between parameters "fields" and "include"


hhware:
In the section Sparse Fieldsets:

> A client **MAY** request that an endpoint return only specific [fields] in the response on a per-type basis by including a `fields[TYPE]` parameter.

If we use an example from the spec, `GET /articles?include=author&fields[articles]=title,body`, and assume that both `title` and `body` are attributes, does it mean that `author` should not be returned in `links` of articles? 

If yes, it seems to contradict the following:
-  in the section Inclusion of Related Resources: "Each relationship name **MUST** be identical to the key in the links section of its parent resource object",
- in the section Resource Relationships: "If a link object refers to resource objects included in the same compound document, it **MUST** include resource linkage to those resource objects".

Alternatives which I see:
1. Provide a requirement to coordinate values of `include` and `fields` in case the latter includes relationship names, and return 400 if they are not properly coordinated.
2. Change the first quoted phrase to something like: "Client **MAY** request that an endpoint return only specific [attributes] in the response on a per-type basis by including a `fields[TYPE]` parameter."
3. Change the first quoted phrase to something like: "If a compound document is returned, client **MAY** request that an endpoint return only specific [attributes] in the response on a per-type basis by including a `fields[TYPE]` parameter. If a non-compound document is returned, `fields[TYPE]` may additionally include relationship names." Also, add a "Note: In case of compound documents, inclusion of relationship in `links` is effectively managed via parameter `include`."
4. State that inclusion of a relationship name into `include` also automatically adds it to `fields[TYPE]`.

IMHO, the alternative (3) is best because it is most versatile. It will especially make sense for cases when API supports only compound documents -- no ambiguity is possible in this case.

Realistically, servers would probably return either compound or non-compound documents for all endpoints, but I guess regulating it is beyond this scope of this spec.

FYI, a PR on related topic: #471

bintoro:
Good point.

But wouldn't the simplest solution be just to state that `include=author` implies `fields[...]=author`? Clearly that's the intent here.

hhware:
@bintoro, right, the 4th option exists (added it to the initial comment). To me, it is not an attractive option at all: I would prefer the client to do all these string manipulation and provide explicit and noncontradictory instructions to the server. I think such approach is much less error-prone...

PS Sorry, somehow did not add the option (4) to the initial comment, fixed now

ethanresnick:
[...]

PR #548, opened 2015-04-12
Possible conflict between parameters "fields" and "include" #476. Option 1: bonding

hhware:


hhware:
See https://github.com/json-api/json-api/issues/476#issuecomment-92096528 for details.

format/index.md
```diff
@@ -355,6 +360,10 @@ responses are called "compound documents".
 In a compound document, all included resources **MUST** be represented as an
 array of resource objects in a top-level `"included"` member.
 
+In a compound document, any resource which references other resources included 
+into the same document (either in top-level `"data"` or `"inlcluded"` members)
```
slaskis: I think you misspelled "included" here...

hhware: @slaskis, thanks! Corrected

tkellen:
I'm :+1: to merging this. It's clear and concise and it solves the problem at hand.

format/index.md
```diff
@@ -785,6 +794,15 @@ include additional [fields] in the response.
 GET /articles?include=author&fields[articles]=title,body&fields[people]=name
 ```
 
+If both query parameters `include` and `fields[TYPE]` are provided by the client, 
+values of `TYPE` **MUST** either refer to the primary data or be listed in 
+`include`, otherwise the server **MUST** return `400 Bad Request`.
```
dgeb: I'd like this paragraph removed. I think this is an unnecessary restriction. Consider heterogeneous collections: the client may not know the members' types in the response.

format/index.md
```diff
@@ -300,6 +300,11 @@ pagination links, as described below.
 If a link object refers to resource objects included in the same compound
 document, it **MUST** include resource linkage to those resource objects.
 
+> Note: a resource may actually be referring to another resource returned 
+in the same compound document, but this reference may be omitted 
+from the link object of the parent resource because of sparse fieldset 
+restrictions.
+
```
dgeb, line 334: I think the exception on **MUST** must be normative.

Instead of:

> If a link object refers to resource objects included in the same compound document, it **MUST** include resource linkage to those resource objects.

Let's consider:

> If a link object refers to resource objects included in the same compound document, it **MUST** include resource linkage to those resource objects unless those fields are excluded via sparse fieldset restrictions.

The non-normative note could then be removed.

format/index.md
```diff
@@ -785,6 +794,15 @@ include additional [fields] in the response.
...
 ```
 
+If both query parameters `include` and `fields[TYPE]` are provided by the client, 
+values of `TYPE` **MUST** either refer to the primary data or be listed in 
+`include`, otherwise the server **MUST** return `400 Bad Request`.
+
+If both query parameters `include` and `fields[TYPE]` are provided by the client,
+and a given related object is requested in `include` but reference to it is not 
+requested in the corresponding `fields[TYPE]`, the server **MUST** return 
+`400 Bad Request`.
```
dgeb: Again, I don't think this is appropriate for heterogeneous collections and should probably be removed.

dgeb:
@hhware @tkellen I regret not fully thinking through the implications of [our suggestion](https://github.com/json-api/json-api/issues/476#issuecomment-88513509). I don't think we can realistically control this so tightly given that primary data may be a heterogeneous collection. I've now come around to thinking that this should all just be condensed to a single additional clause as I marked in the commit.

format/index.md
ethanresnick, line 334: @dgeb If I'm understanding correctly, the approach you're now aiming for is the one I advocated [here](https://github.com/json-api/json-api/issues/476#issuecomment-86700323). If so, then the clause that reads "Each relationship name MUST be identical..." must also be amended, as I mentioned in that comment.

**EDIT**: And actually, I don't see why the "If a link object refers to resource objects..." clause mentioned above would need to be changed at all, since the link object _simply wouldn't appear_ if the field has been excluded.

format/index.md
bintoro: But isn't the first paragraph the entire point of this PR? (I'm referring to 797 to 799 here...)

The purpose of the second paragraph, however, is lost on me. It seems to suggest that `fields` must be present for each `include`. Why?

And even the first paragraph needs to be amended to reflect the fact that `include` contains relationship names, not types, so it's not right to say that "values of `TYPE` must [...] be listed in `include`".

format/index.md
bintoro, line 334: > the link object _simply wouldn't appear_ if the field has been excluded

My thoughts exactly. An excluded field is supposed to be dropped from `links` altogether, so the note is redundant.

format/index.md
```diff
@@ -355,6 +360,10 @@ responses are called "compound documents".
 In a compound document, all included resources **MUST** be represented as an
 array of resource objects in a top-level `"included"` member.
 
+In a compound document, any resource which references other resources included 
+into the same document (either in top-level `"data"` or `"included"` members)
+**MUST** include references to those resources in its own `"links"` member.
+
```
bintoro, line 393: Again, I don't understand the purpose of this paragraph. How can a resource reference other resources apart from listing them under `links`?

format/index.md
dgeb, line 334: > And actually, I don't see why the "If a link object refers to resource objects..." clause mentioned above would need to be changed at all, since the link object simply wouldn't appear if the field has been excluded.

I believe you're right @ethanresnick and @bintoro. I no longer see this as necessary at all.

tkellen:
Oof. The heterogenous use-case completely escaped me here ::facepalm::.

format/index.md
hhware, line 334: I agree with all of you, @dgeb, @ethanresnick and @bintoro, but at the same time I'd like to defend my suggestion. The reason I made it a note was that it did not actually add anything normative to the spec, but just clarified the (indeed obvious) point that, if sparse fieldset requires to drop some relationship, it will not be listed in `links`. There are many notes in the spec which are of purely explanatory nature, to help the reader of the spec to put pieces together, and this one was meant to be of that kind. If you think it is insultingly unnecessary, I do not mind removing it.

format/index.md
hhware, line 393: The spec says: 

> If a link object refers to resource objects included in the same compound document

But there is no requirement for the resource to include that "link object", which "refers to resource objects included in the same compound document". This is the position which this PR reflects: the spec does not wish to control `fields` and `include` separately.

Another way to say what I meant: if object A _actually_ (in the entity-relationship space of the system behind the particular API) references object B, which happens to be returned in the same response in which A in returned, it is mandatory for the the object A to include B in its `links`. This sentence is the central point of this PR: objects are either in both `links` and `included`, or in neither of them. The other option, namely controlling `fields` and `include` independently, is discussed in #549.

I have added that "actually", although I realize that this wording is far from ideal.

format/index.md
hhware: This PR is only discussing the situation when the client provides both `fields[TYPE]` and `include`. Sparse fieldsets operate "on a per-type basis", not on per-relationship-name basis (the latter would be impossible: if we allow to specify different sparse filedsets for the same type, referred to via different relationships, it may result in a situation when we have two different fildsets requirements for a resource with the same `(type,id)`). If the client does "not know the members' types in the response", how could it require specific `fields[TYPE]` for the unknown `TYPE`? 

I believe there is no intrinsic problem here. The server knows which types constitute a particular heterogeneous relationships, so it is capable of performing the verification. The client apparently knows that too, since it specified `fields[TYPE]` in the first place. IMHO, this clause is not more and not less mandatory than other clauses which require the server to return specific errors if the input it receives makes no sense. Please correct me if I am wrong.

I have corrected the wording to account for the case of heterogeneous collections, which I indeed missed.

format/index.md
hhware: The logic from the above comment applies here too. This is the enforcement of the second clause of the PR. Reworded it to make it more clear, although it looks bulky now.

>  It seems to suggest that `fields` must be present for each `include`. Why?

Not sure I understand. It says "If both query parameters `include` and `fields[TYPE]` are provided by the client". It does not suggest that they both "must be present", it only considers what should the server do when client decided to provide both of them.

hhware:
I updated the PR, although it seems to me from the comments it received that this option is not the one everyone prefers, and that it could make sense to look at the option described in #549 instead...

format/index.md
```diff
@@ -382,6 +387,10 @@ responses are called "compound documents".
 In a compound document, all included resources **MUST** be represented as an
 array of resource objects in a top-level `"included"` member.
 
+In a compound document, any resource which actually references other resources included 
+into the same document (either in top-level `"data"` or `"included"` members)
+**MUST** include references to those resources in its own `"links"` member.
+
```
bintoro, line 393: I believe this is already stated under [Resource Relationships](http://jsonapi.org/format/#document-structure-resource-relationships):

> If a link object refers to resource objects included in the same compound document, it **MUST** include resource linkage to those resource objects.

format/index.md
```diff
@@ -812,6 +821,23 @@ include additional [fields] in the response.
...
+- the client provides parameter `fields[B]`, requiring sparse fieldsets for the type `B`,
+- resource of type `A` is actually referenced by type `B`, but 
+reference to it is not requested in `fields[B]`, 
+
+the server **MUST** return `400 Bad Request`.
+
+> Note: In two clauses preceding this note, "refer" is assumed to correctly track 
+heterogeneous relationships, where the type of the resource being referred to 
+in the relationship can be different from the name of the relationship itself.
+
```
bintoro, line 840: **1st clause**

So, this part deals with `fields[TYPE]` when `TYPE` is not applicable to the response.

Why is this a problem to begin with? Why does the server need to complain about a redundant query param? And since a heterogeneous collection may or may not contain items of a particular type, the validity of the request would have to be contingent on what types _might_ theoretically be present in the response. This is just too burdensome, especially considering that this restriction doesn't seem to have any tangible utility.

**2nd clause**

I know what this PR is about, yet after a few readings I still don't understand. Perhaps I've been hit by early-onset dementia, but I would like us to find some ways to simplify this whole ordeal.

**The Note**

Relationship names and heterogeneity are two independent properties, so again I'm not following.

hhware:
Closing in favor of #579

PR #551, opened 2015-04-13
Changed relative paths to URLs in examples

jaspervz:
Most examples contain URLs, some contain relative paths. According to the specification, all must be URLs. This has been fixed.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -173,8 +173,8 @@ Here's how an article (i.e. a resource of type "articles") might appear in a doc
   "title": "Rails is Omakase",
   "links": {
     "author": {
-      "self": "/articles/1/links/author",
-      "related": "/articles/1/author",
+      "self": "http://example.com/articles/1/links/author",
+      "related": "http://example.com/articles/1/author",
       "linkage": { "type": "people", "id": "9" }
     }
   }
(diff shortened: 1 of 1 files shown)


nwjsmith:
Where in the spec does it say URLs must be absolute?

tkellen:
Thanks for taking the time to put this together @jaspervz. JSON-API makes no assertions about URI structure so I'm going to give this a close. If you have any questions about this, please feel free to continue the discussion here.

jaspervz:
The spec doesn't say that URLs can be relative. Now the server can choose per URL if it wants to issue an absolute or relative URL and the client needs to take this into account.

tkellen:
@jaspervz that's correct. Are you suggesting it should be otherwise?

jaspervz:
I think this can be confusing for a client when it receives a JSON where e.g. the self URL is absolute and the related relative etc. I think it's a good idea to explicitly mention that URLs can be absolute or relative and that the server SHOULD always issue absolute or relative URLs.

tkellen:
I disagree. If the server can provide one or the other, a generic client must be able to handle both. Bearing this in mind, I see no benefit in restricting what servers can do.

PR #555, opened 2015-04-15
More complete update history

greysteil:
Add dates of previous release candidates and the correct date for Media type registration.


dgeb:
Thanks @greysteil! You beat me to it. I just gave a talk that included these dates and have been meaning to send a PR just like this.

PR #531, opened 2015-03-30
New visual design including document outline navigation

wvteijlingen:
Pull request for https://github.com/json-api/json-api/issues/527. There are probably still some things to tweak.

PR #570, opened 2015-04-22
Adds table of contents to Format document.

jfinkels:
There seems to be no simple way to automatically generate a table of contents when rendering Markdown as HTML via GitHub Pages, so I generated this manually.


steveklabnik:
Thanks! This will be nice to have until https://github.com/json-api/json-api/pull/531 lands

PR #579, opened 2015-04-27
Prohibit conflicting 'fields' and 'include' params

bintoro:
I don't understand most of what goes on in PR #548, so I decided to offer this alternative that addresses the core issue in two sentences.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -812,6 +812,11 @@ include additional [fields] in the response.
 GET /articles?include=author&fields[articles]=title,body&fields[people]=name
 ```
 
+A client **MUST NOT** use the `fields` parameter to exclude a relationship that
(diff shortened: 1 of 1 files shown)


hhware:
@bintoro, I agree that #548 stretched the original scope of the issue and became too clumsy -- my apologies to tired readers! Your alternative provides exactly what is required.

dgeb:
I appreciate the simplification, but still have concerns about how appropriate this is for heterogeneous collections (as expressed in an [earlier comment](https://github.com/json-api/json-api/pull/548#issuecomment-94789662)). There are going to be cases where one type in a collection has a specific relationship and another type in the same collection does not. In those cases, this seems unnecessarily strict.

dgeb:
Closing in favor of #666

PR #549, opened 2015-04-12
Possible conflict between parameters "fields" and "include" #476. Option 2: independence

hhware:

Issue #497, opened 2015-03-21
Explain how "in-between" resources for the "include" parameter should be handled


gmta:
The format documentation states the following:

> In order to request resources related to other resources, a dot-separated path for each relationship name can be specified:
> 
> GET /articles/1?include=comments.author
> 
> Note: A request for comments.author should not automatically also include comments in the response.

Now if the relation `comments.author` references to resources of type `people`, how do I know that the resources in the `included` key reference the `comments.author` relationship? In other words, how would the response look like if I were to execute the following request:

```
GET /articles/1?include=comments.author,author
```

Given that there exists a link called `author` that references a resource of type `people`. How would I differentiate between the two relations when I do not have any information about the `comment` resources that live in between the `article` and `people` resources?

I could not find anything explaining how I should handle such a request and I think that [this paragraph](http://jsonapi.org/format/#fetching-includes) could include an example.

bintoro:
I believe the `articles` resources would look like this:

``` json
{
  "type": "articles",
  "id": "1",
  "title": "Rails is Omakase",
  "links": {
    "author": {
      "self": "/articles/1/links/author",
      "related": "/articles/1/author",
      "linkage": { "type": "people", "id": "9" }
    },
    "comments.author": {
      "self": "/articles/1/comments/links/author",
      "related": "/articles/1/comments/author",
      "linkage": [{
        "type": "people",
        "id": "11"
      }, {
        "type": "people",
        "id": "12"
      }]
    }
  }
}
```

We do need an example of this in the spec, for sure. I'm trying to set up a separate section for advanced examples. This one would fit right in.

gmta:
Right, but wouldn't that require more changes to the format? For example, if I look at the definition for `links` I get:

> The value of the "links" key is a JSON object (a "links object") that represents related resources, keyed by the name of each association.

The word "association" appears one other time in the entire format description - in an unrelated note.

I'm new to JSON API and although I like the format, I find it hard to pinpoint definitions such as these.

bintoro:
@GMTA, you're right. Definitions and terminology have not been strong points of JSON API, but lately there's been huge progress towards fixing all that. Let's keep the issue open until this case has been addressed, too.

As for pinpointing definitions, I'm planning a PR that overhauls the formatting of all primary definitions, but I can't submit it right now with so many PRs pending.

bintoro:
[...]

PR #580, opened 2015-04-28
Resolve formal contradiction between `fields` and `include`

ethanresnick:
In #476, @hhware pointed out that there's a contradiction between:
1. the requirement that a relationship name in the `include` parameter "MUST be identical to the key in the links section of its parent resource object" and 
2. the possibility that the key for the relationship might not exist in the parent resource object, because it's been excluded by `fields`.

We devised two ways to fix this. 
1. Require that linkage pointing to the included related resources must not be excluded by `fields`. This was expressed originally in #548 and then in the simplified #579.
2. Remove the formal contradiction, but explicitly allow the linkage pointing to the `included` resources to be removed by fields. This approach was represented in #549 and I'm offering a simplified PR for it here.

Whether to go with this PR or #579 depends on the semantics we want for `include`, which we should finish debating first in #497. I just wanted to open this PR to have a concise representation of option 2, as #549 snuck in considerations (like [restrictions on `TYPE`](as a counterpart to #579)) that are beyond the scope of this issue.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -887,9 +887,10 @@ section of the compound document.
 
 The value of the `include` parameter **MUST** be a comma-separated (U+002C
 COMMA, ",") list of relationship paths. A relationship path is a dot-separated
-(U+002E FULL-STOP, ".") list of relationship names. Each relationship name
-**MUST** be identical to the key in the `"relationships"` member of its parent
-resource object.
+(U+002E FULL-STOP, ".") list of relationship names.
+
+If a server is unable to identify a relationship path or does not support
+inclusion of resources from a path, it **MUST** respond with 400 Bad Request.
 
 > Note: For example, a relationship path could be `comments.author`, where
 `comments` is a relationship listed under a `articles` resource object, and


hhware:
IMHO, this option is better than #579 (I explained my reasoning in detail [here](https://github.com/json-api/json-api/issues/497#issuecomment-87162174)).

ethanresnick:
**EDIT**: Ok, I've rebased this to account for our renaming of `"links"` to `"relationships"` in resource objects. 

Also, the concept of a relationship name is actually defined now in the Relationships section, where we say "The name of the relationship is its key in the relationship object." Therefore, I think one could argue that we don't need this language at all about "Each relationship name **MUST**...". However, I don't think it hurts anything, and it arguably adds clarity, so I've left it in for now.

But before I actually add try to write the exception to the "full linkage" requirement, @dgeb, can you comment on https://github.com/json-api/json-api/pull/641#issuecomment-104376564? Because, if we restricted `include` only to endpoints that return resources or collections of resources, I think specifying the "full linkage" requirement, and the `fields` exception to it, would be much easier.

Note also that the current wording of the "full linkage" requirement could still leave us with problems like that identified in #624.

ethanresnick:
Thanks for responding in #641, @dgeb. As I mentioned there, I have been trying to write up a PR for this recently, but all my attempts at specifying the correct behavior have come out very, very messy. Maybe you can take a stab at it?

Here are some observations from my explorations, though, in case they help:
1. Redefining a "relationship path" as "a list of _potential_ relationship names" seemed to simplify things. This idea of potential relationship names comes from the fact that (as I discussed [here](https://github.com/json-api/json-api/pull/674#issuecomment-105496201)) an entry in what the relationship path might _sometimes_ point to a valid relationship and sometimes not, depending on whether or not the data has a schema and the collection on which the relationship is being looked up is homogeneous. This is distinct from the issue, which our prior rewordings were addressing, that a relationship that actually does _exist_ might not be _visible_ in the payload because of `fields`. (Also, considering the above, we might want to rename "relationship path" to "include path" or similar.)
2. When the user provides an `include` parameter, the items in the `included` collection must be a function of the `include` paths; the server shouldn't be able to provide arbitrary extra items. And, yet, when no `include` parameter is provided, the server can provide (as a default) any arbitrary items it wishes to include, so long as they're given linkage to the primary data—even if no include path could pick out that particular set of resources for inclusion. For example, the server might include only the first 5 comments in the response, even though `?include=comments` would bring in all of them. (Or, if the `comments` relationship is paginated, and `include` is supposed to respect that pagination, then `include` still would require bringing in the entire first page, which might be larger. Also: we need to specify that include works with pagination too, if its supposed to. Ugh.)
3. Because the server can include whatever it wants when no `include` parameter is present, the full linkage requirement is adequate for that case. But, when an `include` path is present, it seems intuitively to require that _specific_ relationships have linkage. And, on that front, the full linkage requirement as it's currently specified still has the #624 loophole, just in reverse. That is, with my [example](https://github.com/json-api/json-api/issues/624#issue-77354760) `/companies/abc` resource, a request for `/companies/abc?include=owner` could respond with the included owner, but then linkage only on the `"ceo"` relationship, because that relationship's linkage happens to connect the primary data to the included resources. I think we're going to have to either: 1) decide that fixing loopholes like this is pedantry and leave it to implementors to understand what we mean, which they're likely to do; or 2) specify much more detailed rules that say explicitly that the _relationships_ picked out by the include path's segments **MUST** specify linkage unless there's a `fields` restriction. Option (2) is the road I went down at first, and it wasn't too hard to specify the requirements as a procedural algorithm, but it was much messier to do it as a set of declarative constraints, in the style of the rest of the spec.

Again, maybe someone else will have more success...

tkellen:
I'm going to take a look at this later today... just chiming in now to say that I hear your pain. I've tried to formally define this a few times too!

ethanresnick:
Sounds good, @tkellen! Who knew this "one small thing" was gonna cause such a headache :)

dgeb:
How about replacing the problematic sentence with a statement about the server's responsibility:

> If a server is unable to identify a relationship path or does not support inclusion of resources from a path, it **MUST** respond with `400 Bad Request`.

I agree with @ethanresnick's statement that "the concept of a relationship name is actually defined now in the Relationships section" and I think that any attempt to redefine relationship names just adds confusion.

tkellen:
I'm +1 to @dgeb's simplification here. I feel like trying to nail this edge case down with precise normative text is unnecessary. I don't think we'll leave any meaningful loopholes open with his resolution. What say you @ethanresnick ?

ethanresnick:
Fine with me! I've updated the text to match @dgeb's suggestion. The loopholes and the unspecified cases seem small enough for 1.0.

dgeb:
Thanks @ethanresnick :+1:

hhware:
@dgeb, an elegant eluding resolution!

Issue #536, opened 2015-04-02
Official extension proposal: Embedding related resources


dgeb:
The spec has been dancing around the concept of embedding related resources in create / update requests for some time (ref #131, #165, #202).

I don't favor "side-posting" (see #202) as a general solution because:
- it potentially merges "new" and "pre-existing" objects in a single representation of linkage
- it relies on client-generated IDs to provide linkages

Instead, the alternative that I've favored is the [JSON Patch extension](http://jsonapi.org/extensions/jsonpatch/). The order of operations explicitly requires related resources to be created before parent resources, so there's no ambiguity.

However, there are some problems with proposing the JSON Patch extension as a general solution:
- it is not additive to the base spec, and therefore requires a lot more work to implement
- it also relies on client-generated IDs to provide linkages

So, I'd like to propose an extension that has none of the limitations of side-posting or JSON Patch: an official "embedded" extension. Here's how I envision it:

> A link object could contain an `embedded` member that would include representations of _new_ resources. The `embedded` member could be included alongside the `linkage` member, so that both pre-existing and new resources could be referenced.

I believe this would be very useful for creating resources along with a POST or PATCH operation to a parent resource.

Here's an example that links a new article to two pre-existing tags and creates two new ones:

```
POST /articles
Content-Type: application/vnd.api+json; ext=embedded
Accept: application/vnd.api+json; ext=embedded

{
  "data": {
    "type": "articles",
    "title": "Flask is à la carte",
    "links": {
      "tags": {
        "linkage": [
          {"type": "tags", "id": "1"},
          {"type": "tags", "id": "2"}
        ],
        "embedded": [
          {"type": "tags", "label": "JSON"},
          {"type": "tags", "label": "REST"}
        ]
      }
    }
  }
}
```

Here's an example that replaces an existing article's tags with two new ones:

```
PATCH /articles/1
Content-Type: application/vnd.api+json; ext=embedded
Accept: application/vnd.api+json; ext=embedded

{
  "data": {
    "type": "articles",
    "id": "1",
    "links": {
      "tags": {
        "embedded": [
          {"type": "tags", "label": "JSON"},
          {"type": "tags", "label": "REST"}
        ]
      }
    }
  }
}
```

A few considerations:
- As an additive extension, this is not a blocker to 1.0.
- I don't believe that we need to support embedding at relationship URLs, which only accept linkage data.
[...]

PR #586, opened 2015-04-29
Ordering of relationships

hhware:
It looks like opinions differ on this topic (e.g., https://github.com/json-api/json-api/pull/473#issuecomment-88008211, https://github.com/json-api/json-api/issues/536#issuecomment-95626658, https://github.com/json-api/json-api/issues/536#issuecomment-96447547), so I think it would be useful to state it explicitly in the spec.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -291,6 +291,12 @@ relationship URLs.
 If present, a *related resource URL* **MUST** be a valid URL, even if the
 relationship isn't currently associated with any target resources.
 
+> Note: The spec does not impart meaning to order of resource identifier 
+objects in linkage arrays of to-many relationships, although implementations
+may do that. Arrays of resource identifier objects may represent ordered
+or unordered relationships, and both types can be mixed in one response
+object.
+
 For example, the following article is associated with an `author`:
 
 ```javascript
@@ -1510,9 +1516,9 @@ Accept: application/vnd.api+json
 ```
 
 If a client makes a `POST` request to a *relationship URL*, the server
-**MUST** append the specified members to the relationship using set
-semantics. This means that if a given `type` and `id` is already in the
-relationship, the server **MUST NOT** add it again.
+**MUST** add the specified members to the relationship unless they are 
+already present. If a given `type` and `id` is already in the relationship,
+the server **MUST NOT** add it again.
 
 > Note: This matches the semantics of databases that use foreign keys for
 has-many relationships. Document-based storage should check the has-many


bintoro:
Why not let it be up to the API? A set can be ordered, and it's still a set. There's no conflict between performing set operations and guaranteeing that the items are processed in a particular order.

dgeb:
I agree with @bintoro here and am :-1: on this.

Ordering may be significant for some relationships. For instance, an implementation might allow for the updating of `tags` on an `article` and want to preserve the specified order.

hhware:
> Why not let it be up to the API? A set can be ordered, and it's still a set.

I agree, but it is likely that any given API that contains ordered relationships, also contains unordered ones (just because order makes no sense for some relationships, and it is an unnecessary burden on the server to require all relationships be ordered if at least one of them is ordered: it requires facilities to store that ordering information). 

How the client can tell the difference between them? Is this detail supposed to be described in the API docs, and response format alone is not enough to define the semantics? How can HATEOAS approach be achieved by the API designer wrt ordering relationships? I might be missing something, but the only idea which comes to mind is to specify in `meta` that `links.this_relationship` is ordered, but `links.that_relationship` is unordered.

IMHO, at the very least, it should be clarified in the spec that any given relationship can be either a _set_ or a _totally ordered set_, to avoid misunderstanding.

bintoro:
> How the client can tell the difference between them?

Why must the client  be able to tell the difference? If the order doesn't have any meaning to the client, it simply doesn't have to care, even if the server internally manages the linkages in an ordered fashion.

I'm not convinced this warrants any mention in the spec. Nowhere is it implied that relationships maintain or do not maintain some kind of order, so obviously if an API needs to make or demand such guarantees, the API should explain this in its documentation.

hhware:
> Why must the client be able to tell the difference?

Suppose someone wants to use JSON API to build a HATEOAS API. Things like authentication mechanism, allowed characters, etc., must indeed be specified in the documentation. However, when the client is browsing the API in the HATEOAS sense, it may want to be able to tell whether a given relationship is ordered or not, e.g., to decide whether it makes sense to allow reordering control on it in the UI, etc. If both ordered and unordered relationships are allowed to appear in the same API, the format should distinguish between them.

> Nowhere is it implied that relationships maintain or do not maintain some kind of order

IMHO, it is not only implied, but stated directly: "the server MUST append the specified members to the relationship using set semantics". AFAIK, sets are unordered by definition: [wikipedia](http://en.wikipedia.org/wiki/Set_%28abstract_data_type%29), [wolfram](http://mathworld.wolfram.com/Set.html).

> I'm not convinced this warrants any mention in the spec.

I think it does. @bintoro, you think the spec allows both ordered and unordered relationships. @ethanresnick [expressed](https://github.com/json-api/json-api/pull/473#issuecomment-88008211) a different view. The spec's size is relatively small, and if two active contributors understand it differently, is not it a clear indicator that the spec itself needs improvement?

I am not saying that the statement which this PR makes is correct (in fact, my personal preference is totally different: https://github.com/json-api/json-api/pull/431#issuecomment-90378980). This PR is only trying to establish consistence with the statement which already exists in the spec: "the server MUST append the specified members to the relationship using set semantics". I am ready to update this PR with whatever is decided on this issue, I just want to see it through.

ethanresnick:
I think we may end up giving each api freedom here, as @bintoro and @dgeb suggest, but I'm with @hhware in that I think we _really_ should think this point through more and the spec should take a better-defined position. 

Otherwise, we're going to end up with a situation like javascript object property enumeration: objects were nominally supposed to be unordered, but then people started relying on the the enumeration order of implementations, to the point that ES6 now defines a deterministic enumeration order. In JSON API terms, the analogy would be that poor clients start relying on server processing order, such that an API couldn't make the relationships truly unordered even if it wanted to (which it might to support some databases etc).

hhware:
@tkellen, @dgeb, is any development planned here for 1.0?

tkellen:
I don't think we need normative text for this. Per @ethanresnick's suggestion, I'm open to a small non normative note to clarify our stance. I'm removing the 1.0 milestone from this issue but I'd be happy to review that addition any time if you're willing to update @hhware!

hhware:
@tkellen, I have updated the PR. I strongly feel the clarification, whatever it is, is needed in 1.0.

ethanresnick:
@hhware I think your non-normative note is great here! For your changes in the base text, what if we just did something _really_ minimal, like this:
- Change the word "append" (in the phrase "the server **MUST** append the specified members") to "add", since "append" has a strong connotation of sticking the new thing onto the end, which implies an order.
- Then, we'd drop the phrase "set semantics" entirely, since all we really mean by it is "test for membership", so that we don't have to go into ordered vs unordered sets, etc. 

So the change would be to rewrite this:

> If a client makes a `POST` request to a _relationship URL_, the server **MUST** append the specified members to the relationship using set semantics. This means that if a given `type` and `id` is already in the relationship, the server **MUST NOT** add it again.

to:

> If a client makes a `POST` request to a _relationship URL_, the server **MUST** add the specified members to the relationship if they are not already present. If a given `type` and `id` is already in the relationship, the server **MUST NOT** add it again.

Thoughts?

dgeb:
@hhware I like your non-normative note and prefer @ethanresnick's suggestion. Do you mind updating?

hhware:
@dgeb, updated with a slight change in wording, hopefully it is OK.

dgeb:
@hhware thanks - LGTM :+1:

Issue #573, opened 2015-04-24
proposal to reserve "attributes" member at top level of resource objects


tkellen:
At the moment, we reserve `id`, `type`, `links`, and `meta` members in resource objects. We feel pretty dang sure these are the only ones we will ever need. However, the possibility remains that at some point in the future, we may wish to reserve another. Doing so would be additive/backwards-compatible with the spec, but it would not be backwards compatible with implementations if they are currently using that member as an attribute in their data.

There has been quite a bit of discussion around the prefixing or not-prefixing of reserved members since the inception of the spec (e.g. #313, #362, #558, #569) and we've always landed on not prefixing. I personally agree with this approach, but I'd like to propose an alternate representation for resource objects that should completely eliminate these concerns by walling off implementation-specific attributes under their own namespace.

``` json
{
  "id": "1",
  "type": "book",
  "attributes": {
    "name": "J. R. R. Tolkien",
    "date_of_birth": "1892-01-03",
    "date_of_death": "1973-09-02"
  },
  "links": {
    "chapters": {
      "self": "/books/1/links/chapters",
      "related": "/books/1/chapters"
    }
  }
}
```

Not only does this form prevent future conflicts, it would serve to simplify the normative language of the specification--terms like `members` and `attributes` would be reflected in the data structure itself.

Pending positive feedback from people currently implementing servers and clients I would like to make a PR for this change.

cziegenberg:
I like your idea to wrap the attributes!

I also thought about this way to handle the problem, but decided against suggesting it, because of the additional wrapping elements... but when I see your example, it's much more readable than my suggestion (#569), but offers the same new flexibility: 

The reserved member names don't need to be reserved anymore (for attributes). One could discuss if it is a good practice or not, but I think many API designers would love to get the "type" back. :)

```
{
  "id": "1",
  "type": "book",
  "attributes": {
    "name": "J. R. R. Tolkien",
    "type": "hardcover",
    "date_of_birth": "1892-01-03",
    "date_of_death": "1973-09-02",
    "links": {
      "amazon": "http://www.amazon.com/dp/B0079KT81G/"
    }
  },
  "links": {
    "chapters": {
      "self": "/books/1/links/chapters",
      "related": "/books/1/chapters"
    }
  }
}
```

What I have to add to this suggestion is an additional (optional) top level member "extensions", which enables extensions to add additional data without creating conflicts (as described in #569).

One example:

[...]

Issue #574, opened 2015-04-24
Proposal to reserve "extensions" member at the top level of documents/resource objects


cziegenberg:
In addition to #573, I suggest to introduce a new top level member "extensions", which enables extensions to add additional data without creating conflicts (as previously described in #569).

One example:

I thought about an extension to sign requests and responses, therefore I would have to add signature information to the document. To avoid conflicts (with attributes, other top level members and other extensions), every extension needs its own namespace.

So I would suggest to wrap them (following the suggested structure in #573) like in the following example (example name of the extension is "signature"):

```
{
  "id": "1",
  "type": "person",
  "attributes": {
    "firstname": "Justin",
    "lastname": "Timberlake",
    "signature": "http://commons.wikimedia.org/wiki/File:Justin_Timberlake_signature.svg",
  },
  "extensions": {
    "signature": {
      "value": "...",
      "algorithm" : "...",
      "canonicalization": "...",
      "key": "..."
    }
  }
}
```

As you can see, "signature" is defined as an attribute. The extension also uses this a member of this name, but the namespace prevents a conflict.

Please note: 

This is just a simplified example. An extension as described above would normally add data at the document level (to sign the whole document), not at the resource object level - but other extensions could add resource object specific data. So I request to add this member to both.

hhware:
Would not this be covered by top-level `meta`?

tkellen:
I don't think so. We've walled off `meta` for implementations specific data. If we started mandating usage there for extensions we'd run the same risk of collision that I'm trying to prevent with #573 (would love your feedback there @hhware).

That said, I don't think this addition is something we should discuss in the abstract. I'm cautiously interested in the concept, but given that it is additive, and that we don't have a concrete need for it (yet), I don't think this should be put in the spec now.

ethanresnick:
> I'm cautiously interested in the concept, but given that it is additive, and that we don't have a concrete need for it (yet), I don't think this should be put in the spec now.

My thoughts exactly!

lgebhardt:
If `attributes` is reserved at the top level then there won't be name collisions between attribute names and any keys extensions choose to use. So I see this as unneeded at this point. An extension could freely define any keyword in this space to use. For example the extension you define that wants to sign a response could safely use the `signature` key at the top. No real need to nest it.

hhware:
> So I request to add this member to both.

[...]

Issue #432, opened 2015-03-08
Revisiting versioning strategy of the JSON API spec


hhware:
**(1)** AFAIK, versioning strategy of JSON API spec is intentionally left [undefined](https://github.com/json-api/json-api/issues/46#issuecomment-17656083). IMHO, since the spec does have versions (1.0, RC2, etc. are being widely referenced), it would be useful to specify how these labels are constructed.

**Proposal:** add _"JSON API is versioned according to the specification of [Semantic Versioning](http://semver.org/)"_ somewhere in the Introduction. If SemVer is adopted, 1.0 should obviously become 1.0.0.

**(2)** Even though JSON API will not be introducing backward-incompatible changes after 1.0, nothing prevents it from introducing significant changes in a backward-compatible manner, possibly multiple times over the years. When a client-side library implementing JSON API 1.N is processing a response from a server implementing JSON API 1.M, where N is much higher than M, that client should be able to guess availability of features available on the server from the response format. However, it could be argued that knowing M first-hand from the server could simplify the job of the client.

In addition to that, if it so happens that JSON API actually decides for some reason to introduce backward-incompatible changes in the future, despite the current intentions not to ever do that, transition will be much easier for clients if version of the format is always explicitly communicated. IMHO, it does not hurt to be cautious, especially if the only expense in this case is to add a few characters to headers...

**Proposal**: require that both client and server advertise JSON API version they are using via media type parameter, for instance:

```
Content-Type: application/vnd.api+json; json-api-version=1.0.0
```

**(3)** If (2) is accepted, it could be immediately useful for those who have already started using JSON API: the draft was out in the wild for almost two years! For example, the project which I am involved in uses pre-RC2 JSON API spec and advertises `Content-Type: application/vnd.api+json`. Since RC2, our system is formally incompatible with JSON API, but we cannot even change content type now because of the existing integrations with clients, etc. This would quickly solve this petty problem of ours:

```
Content-Type: application/vnd.api+json; json-api-version=1.0.0-rc.1
```

I would be glad to submit a PR for any/all of this.

dgeb:
Marked for discussion before 1.0.

eneuhauser:
[...]

PR #588, opened 2015-05-01
reserve attributes in top level of resource object

tkellen:
This addresses the discussion in #573.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -154,7 +154,9 @@ resource collections
   "data": {
     "type": "articles",
     "id": "1",
-    // ... attributes of this article
+    "attributes": {
+      // ... attributes of this article
+    }
   }
 }
 ```
@@ -188,6 +190,7 @@ the client and represents a new resource to be created on the server.
 
 In addition, a resource object **MAY** contain any of these top-level members:
 
+* `"attributes"`: an "attributes object", providing information about a resource.
 * `"links"`: a "links object", providing information about a resource's
   relationships (described below).
 * `"meta"`: non-standard meta-information about a resource that can not be
@@ -200,7 +203,9 @@ Here's how an article (i.e. a resource of type "articles") might appear in a doc
 {
   "type": "articles",
   "id": "1",
-  "title": "Rails is Omakase",
+  "attributes": {
+    "title": "Rails is Omakase"
+  },
   "links": {
     "author": {
       "self": "/articles/1/links/author",
@@ -212,27 +217,29 @@ Here's how an article (i.e. a resource of type "articles") might appear in a doc
 // ...
 ```
 
+#### Attributes Object <a href="#document-structure-resource-object-attributes-object" id="document-structure-resource-object-attributes"></a>
+
+The value of the `"attributes"` key is a JSON object (an "attributes object")
+that represents information about the resource object it is contained within.
+
+The top level of this object shares a namespace with the members of `links` and
+**MUST** reserve the `id` and `type` keys. Apart from these restrictions this
+object can contain members keyed by any string valid for this specification.
+
 #### Attributes <a href="#document-structure-resource-object-attributes" id="document-structure-resource-object-attributes"></a>
 
-A resource object **MAY** contain additional top-level members. These members
-represent "[attributes]" and may contain any valid JSON value.
+All members which appear in an "attributes object" are considered attributes and
+may contain any valid JSON value.
 
 Although has-one foreign keys (e.g. `author_id`) are often stored internally
 alongside other information to be represented in a resource object, these keys
 **SHOULD NOT** appear as attributes. If relations are provided, they **MUST**
 be represented under the "links object".
 
-#### Complex Attributes <a href="#document-structure-resource-object-complex-attributes" id="document-structure-resource-object-complex-attributes"></a>
-
-"[Complex attributes]" are [attributes] whose value is an object or array with
-any level of nesting. An object that constitutes or is contained in a complex
-attribute **MUST** reserve the `id`, `type`, `links`, and `meta` members for future
-use.
-
 #### Fields <a href="#document-structure-resource-object-fields" id="document-structure-resource-object-fields" class="headerlink"></a>
 
-A resource object's [attributes] and relationships are collectively called its
-"[fields]". <a href="#document-structure-resource-object-fields" id="document-structure-resource-object-fields"></a>
+With the exception of the `self` link, a resource object's [attributes] and
+[links] are collectively called its "[fields]".
 
 #### Resource Identification <a href="#document-structure-resource-identification" id="document-structure-resource-identification" class="headerlink"></a>
 
@@ -263,12 +270,10 @@ be a string.
 
 The value of the `"links"` key is a JSON object (a "links object") that
 represents related resources, keyed by the name of each association.
-These associations share namespace with resource attributes:
-associations of a given resource **MUST** be named differently than
-attributes of that resource.
+These associations share a namespace with [attributes]. Associations of a
+given resource object **MUST** be named differently than its [attributes].
 
-The key `"self"` is reserved within the links object for the resource URL,
-as described below.
+The keys `"id"`, `"type"` and `"self"` are reserved within the links object.
 
 #### Resource URLs <a href="#document-structure-resource-urls" id="document-structure-resource-urls" class="headerlink"></a>
 
@@ -280,7 +285,9 @@ A resource object **MAY** include a URL in its links object, keyed by
 {
   "type": "articles",
   "id": "1",
-  "title": "Rails is Omakase",
+  "attributes": {
+    "title": "Rails is Omakase"
+  },
   "links": {
     "self": "http://example.com/articles/1"
   }
@@ -295,7 +302,7 @@ response that includes the resource as the primary data.
 
 A resource object **MAY** contain references to other resource objects
 ("relationships"). Relationships may be to-one or to-many. Relationships
-can be specified by including a member in a resource's links object.
+can be specified by including a member in a resource's [links] object.
 
 The name of the relationship declared in the key **SHALL NOT** be `"self"`.
 
@@ -339,7 +346,7 @@ Resource linkage **MUST** be represented as one of the following:
 
 A "linkage object" is an object that identifies an individual related resource.
 It **MUST** contain `type` and `id` members.
-A linkage object **MAY** include a `"meta"` member to contain non-standard 
+A linkage object **MAY** include a `"meta"` member to contain non-standard
 meta-information about linkage.
 
 > Note: Resource linkage in a compound document allows a client to link
@@ -356,7 +363,9 @@ For example, the following article is associated with an `author` and `comments`
 {
   "type": "articles",
   "id": "1",
-  "title": "Rails is Omakase",
+  "attributes": {
+    "title": "Rails is Omakase",
+  },
   "links": {
     "self": "http://example.com/articles/1",
     "author": {
@@ -394,7 +403,9 @@ A complete example document with multiple included relationships:
   "data": [{
     "type": "articles",
     "id": "1",
-    "title": "JSON API paints my bikeshed!",
+    "attributes": {
+      "title": "JSON API paints my bikeshed!"
+    },
     "links": {
       "self": "http://example.com/articles/1",
       "author": {
@@ -415,23 +426,29 @@ A complete example document with multiple included relationships:
   "included": [{
     "type": "people",
     "id": "9",
-    "first-name": "Dan",
-    "last-name": "Gebhardt",
-    "twitter": "dgeb",
+    "attributes": {
+      "first-name": "Dan",
+      "last-name": "Gebhardt",
+      "twitter": "dgeb"
+    },
     "links": {
       "self": "http://example.com/people/9"
     }
   }, {
     "type": "comments",
     "id": "5",
-    "body": "First!",
+    "attributes": {
+      "body": "First!"
+    },
     "links": {
       "self": "http://example.com/comments/5"
     }
   }, {
     "type": "comments",
     "id": "12",
-    "body": "I like XML better",
+    "attributes": {
+      "body": "I like XML better"
+    },
     "links": {
       "self": "http://example.com/comments/12"
     }
@@ -555,11 +572,15 @@ Content-Type: application/vnd.api+json
   "data": [{
     "type": "articles",
     "id": "1",
-    "title": "JSON API paints my bikeshed!"
+    "attributes": {
+      "title": "JSON API paints my bikeshed!"
+    }
   }, {
     "type": "articles",
     "id": "2",
-    "title": "Rails is Omakase"
+    "attributes": {
+      "title": "Rails is Omakase"
+    }
   }]
 }
 ```
@@ -603,7 +624,9 @@ Content-Type: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
-    "title": "JSON API paints my bikeshed!",
+    "attributes": {
+      "title": "JSON API paints my bikeshed!"
+    },
     "links": {
       "author": {
         "related": "http://example.com/articles/1/author"
@@ -806,9 +829,6 @@ GET /articles/1?include=author,comments,comments.author
 A client **MAY** request that an endpoint return only specific [fields] in the
 response on a per-type basis by including a `fields[TYPE]` parameter.
 
-> Note: Only [fields] are affected; `type`, `id`, and (optionally) `self` are
-included as normal.
-
 The value of the `fields` parameter **MUST** be a comma-separated (U+002C
 COMMA, ",") list that refers to the name(s) of the fields to be returned.
 
@@ -958,8 +978,10 @@ Accept: application/vnd.api+json
 {
   "data": {
     "type": "photos",
-    "title": "Ember Hamster",
-    "src": "http://example.com/images/productivity.png",
+    "attributes": {
+      "title": "Ember Hamster",
+      "src": "http://example.com/images/productivity.png"
+    },
     "links": {
       "photographer": {
         "linkage": { "type": "people", "id": "9" }
@@ -998,8 +1020,10 @@ Accept: application/vnd.api+json
   "data": {
     "type": "photos",
     "id": "550e8400-e29b-41d4-a716-446655440000",
-    "title": "Ember Hamster",
-    "src": "http://example.com/images/productivity.png"
+    "attributes": {
+      "title": "Ember Hamster",
+      "src": "http://example.com/images/productivity.png"
+    }
   }
 }
 ```
@@ -1037,8 +1061,10 @@ Content-Type: application/vnd.api+json
   "data": {
     "type": "photos",
     "id": "550e8400-e29b-41d4-a716-446655440000",
-    "title": "Ember Hamster",
-    "src": "http://example.com/images/productivity.png",
+    "attributes": {
+      "title": "Ember Hamster",
+      "src": "http://example.com/images/productivity.png"
+    },
     "links": {
       "self": "http://example.com/photos/550e8400-e29b-41d4-a716-446655440000"
     }
@@ -1099,7 +1125,9 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
-    "title": "To TDD or Not"
+    "attributes": {
+      "title": "To TDD or Not"
+    }
   }
 }
 ```
@@ -1109,9 +1137,9 @@ Accept: application/vnd.api+json
 Any or all of a resource's attributes **MAY** be included in the resource
 object included in a `PATCH` request.
 
-If a request does not include all of the fields for a resource, the server
-**MUST** interpret the missing fields as if they were included together with
-their current values. It **MUST NOT** interpret them as `null` values.
+If a request does not include all of the [fields] for a resource, the server
+**MUST** interpret the missing [fields] as if they were included with their
+current values. It **MUST NOT** interpret them as `null` values.
 
 For example, the following `PATCH` request is interpreted as a request to
 update only the `title` and `text` attributes of an article:
@@ -1125,8 +1153,10 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
-    "title": "To TDD or Not",
-    "text": "TLDR; It's complicated... but check your test coverage regardless."
+    "attributes": {
+      "title": "To TDD or Not",
+      "text": "TLDR; It's complicated... but check your test coverage regardless."
+    }
   }
 }
 ```
@@ -1149,7 +1179,9 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
-    "title": "Rails is a Melting Pot",
+    "attributes": {
+      "title": "Rails is a Melting Pot"
+    },
     "links": {
       "author": {
         "linkage": { "type": "people", "id": "1" }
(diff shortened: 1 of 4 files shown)


tkellen:
@bintoro, would you mind taking a look at #573 and this when you have a minute?

bintoro:
Whoa, wait. Why are complex attributes being removed altogether? They're a widely-requested feature.

IMO the "subattributes" within complex attributes should absolutely not be enclosed in a further `attributes` object. It's kind of funny, then, that there might be a `links` member holding outbound relationships, but I think we're just going to have to live with that.

As for what member names should be reserved inside complex attributes, I've always been of the opinion that it would be best to reserve just `links`. If `type` and possibly `id` are going to be valid attribute names going forward, then the same should of course apply here.

Other than that, I'm fully on board with the main idea here, i.e. introducing `attributes` at the resource-object level.

tkellen:
@bintoro Just trying to work through the implications of this adjustment, please consider this a working draft.

Given that "complex attributes" reserve all the same top level keys as resource objects, I think they should be subject to the same constraints. The discontinuity created by not using the `attributes` key in "complex attributes" is pretty untenable to me.

I would rather discard this idea and accept the risks outlined in #573 if we all can't agree on this point.

bintoro:
>  The discontinuity created by not using the `attributes` key in "complex attributes" is pretty untenable to me.

Hmm. I don't think we have to regard `attributes` and `links` as strictly parallel members. The instinct to put `attributes` inside complex attributes probably stems from their superficial similarity to resource objects, but we all agree they are not comparable.

Think of it this way: all attributes appear in a single `attributes` container, whereas `links` can appear anywhere to represent relationships. It's a perfectly logical structure.

hhware:
> Think of it this way: all attributes appear in a single `attributes` container, whereas `links` can appear anywhere to represent relationships. It's a perfectly logical structure.

@bintoro, interesting! And the rule to form paths to those links,  would it be to drop `attributes.` from actual path only on resource level and drop `links.` from the actual path on any level? E.g., the actual path `<resource>.links.<link>` translates to `<resource>.<link>`, the actual path `<resource>.attributes.<complex_attribute>.links.<link>` translates to `<resource>.<complex_attribute>.<link>`, the actual path `<resource>.attributes.<complex>.<attribute>.links.<link>` translates to `<resource>.<complex>.<attribute>.<link>`, right?

bintoro:
@hhware Yes.

To put it another way, an attribute can be any valid JSON value, as promised by the spec, as long as any nested objects reserve `id` (?) and `links` members. This is a minimal restriction and would be easy to work with in the context of NoSQL datastores and existing JSON structures.

Introducing `attributes` inside complex attributes would essentially turn them into resource objects in their own right, which would defeat their purpose. In fact, complex attributes do not even have attributes! They just have _members_ since they're plain JSON.

`links` objects, on the other hand, are OK within complex attributes because, unlike the regular members, they're a JSON API -specific thing that's always going to need a backend-to-API transformation. I hope this clarifies why I feel `attributes` and `links` are not analogous.

hhware:
@bintoro, pretty smart! I agree, it could be useful to have `links` reserved anywhere in attributes.

> reserve `id` (?) 

I only suggested to reserve `attributes.id`, not `id` anywhere in attributes, and only because paths can include `id`, `<attribute>` and `<link>`, so if `attributes.id` were allowed, it would be unclear how to reference both `<resource>.attributes.id` and `<resource>.id` at the same time. I agree, there should be no restriction on internal structure of attributes besides these two things.

tkellen:
> Introducing attributes inside complex attributes would essentially turn them into resource objects in their own right, which would defeat their purpose. In fact, complex attributes do not even have attributes! They just have members since they're plain JSON.

If complex attributes are explicitly for plain JSON and not for addressable resources it feels pretty strange to me that they would include references to addressable resources via links. That said, given that I don't personally have a use-case for complex attributes (and if I did, I doubt I'd use links within them), I suppose I can look the other way.

@bintoro, would you mind taking a stab at what the revised definition of a complex attribute would be so I can re-introduce it in the document?

I'd really like to hear from @dgeb and @steveklabnik on this. I think the main point of the PR is really important but I'm not going to merge without broad consensus.

format/index.md
```diff
@@ -214,24 +219,18 @@ Here's how an article (i.e. a resource of type "articles") might appear in a doc
 
 #### Attributes <a href="#document-structure-resource-object-attributes" id="document-structure-resource-object-attributes"></a>
 
-A resource object **MAY** contain additional top-level members. These members
-represent "[attributes]" and may contain any valid JSON value.
+A resource object **MAY** contain an "[attributes]" object. If included, this
+object **MUST** reserve the `id` member at it's top level, but can otherwise
+contain any valid JSON value.
```
dgeb: I believe that `type` MUST be reserved as well as `id`. Otherwise, requests to sort or filter by `type` would be ambiguous.

Also: s/it's/its/

tkellen:
Is there any reason that links for complex attributes can't be represented in the top level `links` object via dot notation?

dgeb:
> Is there any reason that links for complex attributes can't be represented in the top level links object via dot notation?

This would be possible using URI templates, which could account for arrays. However, URI templates are no longer explicitly supported in the spec.

dgeb:
Even though `type` and `id` share a namespace with members of `links` and `attributes`, I see the explicit separation of `attributes` as important to future-proofing the spec. It clearly cordons off user-defined fields into objects which define their meaning, and allows for future meaning to be given to groups of fields through the introduction of a new sibling to `links` and `attributes`. This could happen in the base spec or simply in additive extensions.

I'm :+1: on merging this soon to allow implementations to adapt before May 21. I would like to open a separate milestoned issue to discuss defining complex attributes and reserving members within them.

@tkellen could you please update the home page example as well?

tkellen:
Updated. I'm prepared to merge and re-open the complex attributes issue assuming you're good with the current changes.

bintoro:
> If complex attributes are explicitly for plain JSON and not for addressable resources it feels pretty strange to me that they would include references to addressable resources via links.

There is clearly a use case for outbound links from complex attributes (e.g., #238).

Whether complex attributes should be addressable has been discussed multiple times (e.g., #383), and the majority opinion seems to be against. Doing so would essentially make them embedded resource objects, which goes against the linkage paradigm.

> Is there any reason that links for complex attributes can't be represented in the top level links object via dot notation?

This was discussed comprehensively in #383 and decided against. (It was actually about putting them at the resource-object level, but the idea is the same.)

> @bintoro, would you mind taking a stab at what the revised definition of a complex attribute would be so I can re-introduce it in the document?

But I'm not sure what needs to be done here. Does the definition of complex attributes have to change in any way apart from what members are reserved? I'm still puzzled as to why the section is being removed.

PR #504 is aimed at improving the explanation on complex attributes, so any amendments to that section can be implemented there.

tkellen:
Sorry for not being up to speed on that particular conversation @bintoro, I have obviously not considered complex attributes carefully enough (reasons stated above). I'll re-read those issues so I have the proper context. Assuming you're fine with rebasing #504 off of this, would you feel comfortable having me merge this as is?

bintoro:
> Assuming you're fine with rebasing #504 off of this, would you feel comfortable having me merge this as is?

Yeah, it's fine to recreate the removed section in #504. In fact, the definition of complex attributes is not even strictly necessary; it only exists to elaborate on the topic because questions about it have been raised repeatedly. JSON objects as attributes have always been allowed regardless.

However, there are a few other things I would like to discuss. Apologies for raising these issues at such a late stage, but I've been occupied with other stuff.
- There is no longer a definition for _attributes_, yet the spec refers to the term in a number of places. The _Attributes_ section only talks about an _attributes object_.
- > If included, this object [...] can otherwise contain any valid JSON value.
  
  This doesn't seem right. An _attribute_ can have any valid JSON value; the attributes object may contain any number of those.
- > MUST reserve the `id` and `type` members
  
  I didn't realize until now that `type` would continue to be reserved. In #573 it seemed that part of the support for the whole idea was based on the premise that `type` would become available as an attribute name.
  
  If resource-object-level scalar members and `attributes` will continue to share a namespace, the value of this proposal will be reduced quite a bit. If further special members are introduced in the future, we'll have a situation where some such members share a namespace and some don't.

hhware:
> If resource-object-level scalar members and attributes will continue to share a namespace, the value of this proposal will be reduced quite a bit.

I suggest to add a clear statement that the following resource members are addressable via namespace `<resource_name>.*` 
- specifically in the following places: `sort`, `fields[TYPE]` (anywhere else?):
  - `id`, 
  - `type` (if it is declared addressable, and if so, it should obviously be reserved in attributes and links), 
  - members of `<resource_name>.attributes` and their descendants via dot-separate paths, 
- specifically in the following places: `include` (anywhere else?):
  - members of `<resource_name>.links` and their immediate descendants.

This will 
- make the spec more readable, especially to newcomers (currently the only reference to namespace  `<resource_name>.*` seems to be in http://jsonapi.org/format/#document-structure-links),
- enable other query parameters introduced by extensions or implementations to address other resource member and their children via paths, e.g., to enable query parameters like `resource_members_enabled=meta.some_meta_member,some_resource_level_member` in the future.

tkellen:
No worries @bintoro. 

I was originally thinking `type` would remain available, but @dgeb makes a good point about the ambiguity for `sort` and `filter`. I suppose we could specify an order of preference for them to gain that back?

I see your point about the difference between attributes and the attributes object.  I'll push a fix for that shortly.

ethanresnick:
@bintoro 

> If resource-object-level scalar members and attributes will continue to share a namespace, the value of this proposal will be reduced quite a bit. If further special members are introduced in the future, we'll have a situation where some such members share a namespace and some don't.

That was my first reaction too when @dgeb mentioned that they'd share a namespace. Now I'm wondering though if having them be separate namespaces would also cause problems, like the ambiguity with `type` in sort that @hhware alluded to. @tkellen's suggestion to solve that with an order of preference could work for `type`, but what happens when we add a new sibling to `"attributes"` down the line, and someone has an attribute with the same name (as our new member) that they're using in `?sort` requests. Then, suddenly, our new resource-object-level field can't be sorted on.

I see a couple solutions:
1. We could distinguish in query parameters between top-level members (meaning, members at the top of the resource-object) and attributes/links. So, for concision, all top-level members could be referenced as `top.[key-name]`. So you'd have `?filter[top.id]=myJSONAPIID` distinct from `?filter[id]=valOfIdInAttributes`. That would offer real extensibility then, and any hing we reserved in the `"attributes"` object would be only to avoid confusion (e.g. it might be good to reserve `id`) or to enable other types of extensibility (like reserving `"links"` within complex attributes).
2. While I think Option 1 is a decent answer to the problem at hand, I still worry that we're going to run into cases down the line where we want types of extensibility that the current spec doesn't give us. For example, we might want to add a member at the document's top-level that conflicts with the behavior of a commonly-used extension. To solve this problem, I think we should consider the more conventional solution of a reserved-word list (which has saved javascript, e.g., as it's tried to grow without breaking bc). These reserved words might not apply to attribute or link names (since it's hard for people to change their data models), but they could apply to the names of members that extensions are allowed to add.

cziegenberg:
Doesn't a collection of resources that can be sorted/filtered always contain resource objects of the same resource type? It's not defined in the spec, but this is how I understand it - so IMHO sorting/filtering by (resource) type doesn't make sense here and only `id` needs to be reserved.

bintoro:
This may be beating a dead horse, but what was it about reserved-key prefixes that caused them not to catch on? I don't remember seeing any concrete objections apart from underscore prefixes being considered ugly, which I agree with. But there are other characters.

It seems to me we're going through a whole lot of trouble here to solve issues that wouldn't exist to begin with if we just used `@id`, `@type`, and so on.

hhware:
> We could distinguish in query parameters between top-level members (meaning, members at the top of the resource-object) and attributes/links. So, for concision, all top-level members could be referenced as `top.[key-name]`. So you'd have `?filter[top.id]=myJSONAPIID` distinct from `?filter[id]=valOfIdInAttributes`. 

Perhaps we could provide a different syntax to refer to resource-level member names? E.g., immediate children of resource can be referenced by enclosing it in `<>` or some other braces, and others are referenced directly? Examples:
- `id`: both `articles.<id>` `articles.id` refer to the same thing because `id` is reserved (_if_ it is reserved),
- `articles.<type>`, `type` is not reserved,
- `articles.<links>.author.linkage` could be used in requests to, e.g., enable linkage member of `author`, whereas `articles.author` could be used to refer to relationships,
- `articles.<meta>`, `articles.<new_resource_level_member>`, etc

> Doesn't a collection of resources that can be sorted/filtered always contain resource objects of the same resource type?

@ziege, no, heterogeneous collections are also supported

dgeb:
In the current spec, attributes and links already share a namespace (see http://jsonapi.org/format/#document-structure-links). Although it's not explicitly stated, this namespace is also shared by `type` and `id`.

I don't think this should change if we accept this PR. I am unclear on the value proposition of allowing an attribute named `type` in addition to a resource's `type`. Furthermore, it would be surprising to encounter a link named `contact` along with an attribute named `contact` in the same resource. These allowances would just lead to confusion.

Instead, I see the value proposition of the `attributes` object to be that it clearly walls off user-defined attributes from spec-defined members, such as `type`, `id`, and `links`. It allows for the introduction in the future of a sibling to `meta` or `links` without breaking backward compatibility. These future members could define fields with unique requirements and shared characteristics.

Let's say that an extension introduces the `embedded` member additively. It would live alongside `type`, `id`, `links`, `meta`, and `attributes`. It's very important to note that `embedded` would not conflict with an attribute already named `embedded` (say, a boolean particular to a resource). Inside the `embedded` object might reside other properties that share the same per-resource namespace as `type`, `id`, `links`, and `attributes`.

I hope this example illustrates the (somewhat subtle) advantages of the `attributes` member. By walling off user-defined attributes from spec-defined attributes completely, we can avoid custom prefixes while still keeping the spec relatively future-proof.

hhware:
> Furthermore, it would be surprising to encounter a link named `contact` along with an attribute named `contact` in the same resource. These allowances would just lead to confusion.

@dgeb, if you are referring to [my comment](https://github.com/json-api/json-api/pull/588#issuecomment-98780041), I only suggested to clearly state that they all share the same namespace `<resource_name>.*`, not to split it in multiple namespaces.

dgeb:
> if you are referring to my comment, I only suggested to clearly state that they all share the same namespace <resource_name>.*, not to split it in multiple namespaces.

@hhware I was not referring to a specific comment - just the concept of shared namespaces in general, as raised by many in this thread.

> It seems to me we're going through a whole lot of trouble here to solve issues that wouldn't exist to begin with if we just used @id, @type, and so on.

@bintoro I don't see these prefixes as a silver bullet. Yes, this in theory would allow `@id` to coexist with `id`, but then you've got another set of problems - how to differentiate between `@id` and `id` when specifying `fields` and `sort` params.

In short, I see nothing wrong with shared namespaces for fields, as well as a rich set of keywords (e.g. `attributes`, `links`) to contextualize those fields.

bintoro:
> I am unclear on the value proposition of allowing an attribute named `type` in addition to a resource's `type`.

`type` is an extremely common attribute name. If you look at #573, you'll find that most people supported this proposal precisely because it would lift the restriction on available attribute names.

> Furthermore, it would be surprising to encounter a link named contact along with an attribute named contact in the same resource.

I don't think anyone's suggesting that this should be allowed. All fields would continue to have a shared namespace, just as now.

Non-attribute `id` and `type` members are not fields and would therefore live in the resource object top-level namespace only.

> It's very important to note that `embedded` would not conflict with an attribute already named `embedded` (say, a boolean particular to a resource).

Yes. And any issues such as sorting troubles would affect `embedded` and `type` equally, so this argument would only seem to lend support for allowing `type` as well. What's the rationale for special-casing it?

(EDIT: Oh yeah, the other `embedded` was supposed to be a container, so the sorting situation doesn't apply. But if it were a scalar member, then you'd have a conflict.)

> [...] how to differentiate between `@id` and `id` when specifying fields and sort params

By including the prefix in the query string parameter.

dgeb:
Since JSON API has a concept of `type`, implementations should adapt to it or choose a different word.

I believe it would be a source of confusion to allow a resource's top-level `id` or `type` to coexist along with a separate attribute also named `id` or `type`. Similarly, I believe it would be a source of confusion to allow `@id` / `@type` to coexist with `id` / `type`.

bintoro:
If duplicate names are a source of confusion, then we shouldn't even be discussing a PR whose primary purpose is to make them possible through future additions.

The fact of the matter is that potential sorting etc. issues will exist regardless of what happens with `type`, unless this PR comes with a guarantee that `id` and `type` will forever be the only selectable/sortable members at the resource object level.

dgeb:
> If duplicate names are a source of confusion, then we shouldn't even be discussing a PR whose primary purpose is to make them possible through future additions.

This PR allows for new container objects, such as `embedded` [in my example](https://github.com/json-api/json-api/pull/588#issuecomment-98796687).

> The fact of the matter is that potential sorting etc. issues will exist regardless of what happens with type, unless this PR comes with a guarantee that id and type will forever be the only selectable/sortable members at the resource object level.

You're correct that it would not allow for other selectable/sortable members at the resource object's top level, according to my interpretation.

The only selectable members will remain "fields".

The only sortable members will remain "fields" + `type` + `id`.

The definition of "fields" can always be expanded to include other user-defined fields in other containers, such as the fictitious `embedded`.

I think this is still sufficient justification for the extra `attributes` object, but am glad to discuss broader interpretations.

ethanresnick:
First I just want to throw out that I'm unconditionally in favor of adding the `"attributes"` member, regardless of what we decide about all the scoping issues. I think it's worth adding simply because it makes the spec easier to follow, and it makes it much clearer what different types of things there are (and can be) in a resource object. When you consider things like @hhware's proposal (#590) for HATEOS-style actions, it's clear that something like that would be a mess without `"attributes"`.

But about the scoping issues:
1. If extensions can add resource-object-level members, then we still have an opportunity for conflict. Namely: the spec might want to add a member at that level that an extension is already using. Granted, this situation is way better than the pre `"attributes"` world, in that there will be far fewer extensions claiming these members than there will be APIs (with many attributes each) using the base spec. So maybe this potential for conflict doesn't matter, but it's worth noting that it would still exist. If we care about it, I see three solutions:
   1. Something like the `extensions` container [proposed earlier](https://github.com/json-api/json-api/issues/574).
   2. A prefix character (which I'm increasingly inclined toward)
   3. Or reserving specific member names that extensions can't use, which was my [second proposal](https://github.com/json-api/json-api/issues/574) above.
2. I don't like the idea of excluding the possibility for future resource-object-level members to be used in query parameters. Enabling that is what my first proposal [here](https://github.com/json-api/json-api/pull/588#issuecomment-98791042) was all about. But notice that that proposal works, essentially, by adding a prefix to the top-level keys just when they're referenced in the query parameter (the `top.` part), which makes one wonder: why not be consistent and just put a prefix in the payload too?

hhware:
> First I just want to throw out that I'm unconditionally in favor of adding the "attributes" member, regardless of what we decide about all the scoping issues

I support that!

> The only selectable members will remain "fields".
> 
> The only sortable members will remain "fields" + type + id.

IMHO, it is good for now, but what if more complex ways to address members of various sub-objects of a resource need to be supported in the future (an example is [mixing of resource and relationship attributes in one sorting request](https://rawgit.com/hhware/json-api/i415-p431-example/i415-p431.html#mixed-sorting))? I think it would be more robust wrt the unknown future development directions of the spec to do something like the following:
1. Define shared namespace for `id`, `type`, members of attributes (with dot-separated descendants), (immediate) members of links. Each of them is addressable directly via existing dot-separated notation (like `articles.type`). This seems to be enough for the current needs of the spec and for the future needs of many not-too-sophisticated implementations.
2. Devise a standard approach of reaching anything else in the resource object in a way _compatible_ with (1). It can be done via prefix, braces, reserved top-level "entrace member", prefix for spec-defined members, anything else. This may turn out to be useful for more complex implementations which may actually need to address those members, whether their names happen to conflict with attribute names or not.

If the spec limits itself to doing (1) now, and it turns out (2) is needed in the future, backward compatibility guarantees may not allow to introduce (2) in an elegant and non-conflicting way. On the other hand, if it turns out (2) is not needed in the future -- so what? The cost of unjustified reservation of a special symbol seems to be much lower than the cost of [minting a new media type](https://github.com/json-api/json-api/issues/432#issuecomment-79495877) and a mess associated with it.

format/index.md
```diff
@@ -212,26 +217,26 @@ Here's how an article (i.e. a resource of type "articles") might appear in a doc
 // ...
 ```
 
+#### Attributes Object <a href="#document-structure-resource-object-attributes-object" id="document-structure-resource-object-attributes"></a>
+
+The value of the `"attributes"` key is a JSON object (an "attributes object")
+that represents information about the resource object it is contained within.
+
+This object **MUST** reserve the `id` and `type` keys at its top level, but
+can otherwise contain any valid JSON value.
```
dgeb: How about:

> Members of this object share a namespace with `id`, `type`, and the members of `links`. Each member can contain any valid JSON value.

ethanresnick:
@bintoro Amidst all this we lost the section on complex attributes, and the members reserved in them. Would you be up for making a PR (or rebasing from your old ones/old text) to bring this back?

bintoro:
@ethanresnick Sure. But what to reserve? I'm inclined to reserve as few keys as possible. Do you foresee a need for anything besides "relationships"?

ethanresnick:
@bintoro Definitely `relationships`, but I'm also inclined to reserved `attributes` (to communicate that this isn't another resource object), and I might also reserve `links` just in case. But I don't think we need to reserve `type`/`id`, especially if we're reserving attributes.

bintoro:
> I'm also inclined to reserved `attributes` (to communicate that this isn't another resource object)

@ethanresnick Could you elaborate on this, please? I mean, how does reserving `attributes` help to draw a distinction between an object-as-an-attribute and a resource object?

Reserving `links` for now might be a good idea, just to be on the safe side, but if no one is able to foresee a use case for it, I'd be in favor of dropping that restriction eventually.

ethanresnick:
> @ethanresnick Could you elaborate on this, please? 

Sure. I said that because, when we were initially discussing the introduction of `attributes` at the resource object level with @tkellen, his first instinct (if I'm remembering correctly) was that complex attributes should have an `attributes` member within them. It quickly became clear that that wouldn't make sense, because complex attributes aren't embedded resources. But if Tyler was getting tripped up by this, I'm sure others will too. Everyone's in a hurry and will maybe only skim the section, whatever. So this restriction is just a simple way to emphasize the point; when people come across it, they'll be forced to have that "aha, so this is a different thing" moment.

bintoro:
@ethanresnick Interestingly I have the exact opposite view :smile: The less similarities there are between complex attributes and resource objects, the less opportunity for confusion.

In fact, I would like to get rid of the concept of "complex attribute" altogether and simply point out that an object or array is indeed a valid value for an attribute. Or perhaps retain the term but not have a separate section to discuss them.

ethanresnick:
> The less similarities there are between complex attributes and resource objects, the less opportunity for confusion.

Ohhhh, I see what you mean. You're thinking that users will assess similarities by looking at the definitions (so reserving `attributes` would make the definition more similar), whereas I was assuming they'd assess similarities by looking at example payloads, and preventing `attributes` would force the payloads to look more different. But now that I think it through, I'm not sure where they'd get those example payloads from, so your point of view probably makes more sense. Let's do what you're suggesting then: only reserve `relationships` and `links`, downplay complex attributes in terms of sections and then, perhaps, discreetly slip a complex attribute into one of the examples, and have that complex attribute happen to not use the `attributes` key.

PR #589, opened 2015-05-01
Immutables

### LINKED ISSUE #575: Proposal to reserve "virtuals" member at top level of resource objects
The spec knows "attributes" but not "virtual attributes" (or "derived attributes"). This type of attribute is usually calculated from the values of other attributes of the same resource (or of related resources). 

The problem with these attributes is, that they cannot be updated directly (only by updating the attributes they are derived from), so they are read-only and MUST NOT be used in CREATE/UPDATE requests.

Examples for this type of attribute:
- age (calculated from the date of birth and the current time)
- item total price (calculated from the number of items and the single price)
- order total price (calculated from the total prices of related order items)

The only way to handle this in the current version of the spec is to reply with a "409 Conflict" error, but this error is also used for other conflicts, e.g. with client generated ids. 

This conflict could be easily prevented by excluding the derived attributes from the request - but to be able to do that, the client needs to know which attributes are "virtual".

Suggestion:

Reserve a new member "virtuals" at the top level of resource objects, based on the suggested change #573. They MUST share the same namespace with the other resource attributes.

```
{
  "id": "1",
  "type": "person",
  "attributes": {
    "firstname": "Justin",
    "lastname": "Timberlake",
    "signature": "http://commons.wikimedia.org/wiki/File:Justin_Timberlake_signature.svg",
    "date-of-birth": "1981-01-31"
  },
  "virtuals": {
    "age": 34
  }
}
```

hhware:
Proposing support for members `immutable` in attempt to resolve #575.

I am cutting corners here: this PR in its current form assumes that #588 is accepted, and hence should not go ahead of #588. I want to throw it in anyway because not so much time is left before 1.0.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -30,6 +30,7 @@ for exchanging data.
   - [Compound Documents](#document-structure-compound-documents)
   - [Meta information](#document-structure-meta)
   - [Top-level Links](#document-structure-top-level-links)
+  - [Read-only attributes and relationships](#document-structure-readonly)
 - [Fetching Data](#fetching)
   - [Fetching Resources](#fetching-resources)
   - [Fetching Relationships](#fetching-relationships)
@@ -170,6 +171,8 @@ A document's top level **MAY** also have the following members:
 * `"links"`: URLs related to the primary data.
 * `"included"`: a list of resource objects that are related to the primary data
   and/or each other ("included resources").
+* `"readonly"`: a list, per resource type, of lists of attributes or 
+  relationships of resources of a given type, which cannot be modified by client.
 
 If any of these members appears in the top-level of a response, their values
 **MUST** comply with this specification.
@@ -192,6 +195,8 @@ In addition, a resource object **MAY** contain any of these top-level members:
   relationships (described below).
 * `"meta"`: non-standard meta-information about a resource that can not be
   represented as an attribute or relationship.
+* `"readonly"`: a list of attributes or relationships of the given resource, 
+  which cannot be modified by client.
 
 Here's how an article (i.e. a resource of type "articles") might appear in a document:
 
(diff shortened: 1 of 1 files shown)


hhware:
I think it would be better to use the term `unmodifiable`: their modification via API is not allowed, but their values can be of course changed by server...

cziegenberg:
I'd say that "immutable" is okay, although it doesn't describe the state of the resource object but of its attributes here.

And it's clear, that this is an information for the client only and that it doesn't limit the server in modifying the resource attributes.

bintoro:
There's actually a term for this: "read-only".

cziegenberg:
All words express the same and all still mean that the server can modify it anyway... :)

hhware:
Updated the PR:
- replaced `immutable` with `readonly`
- added 403

hhware:
Updated: added a clause about accuracy for client / resource / moment in time, similar to one in #590.

pixelhandler:
@hhware perhaps this additional spec can be achieved using the meta section of the payload, rather than defining additional structure. I had considered something similar for "protected-attributes" as a sibling of "attributes" but the more I thought about it I'm happy with expressing that level of detail as meta content, see https://github.com/json-api/json-api/issues/575#issuecomment-101510988

hhware:
@pixelhandler, this is what I initially thought about it (https://github.com/json-api/json-api/issues/575#issuecomment-96380083), but then I concluded that it could have broader audience than just one API: `meta` is intended for custom stuff, specific to a given API...

dgeb:
Given that both attributes and relationships may be read-only, I'm opposed to a common bucket that could group both together purely from a structural perspective.

I think it's also important to realize that immutability _can_ be temporal, and that attributes that are read-only at the time of one request may be mutable at a later time due to a resource's change in state.

hhware:
> Given that both attributes and relationships may be read-only, I'm opposed to a common bucket that could group both together purely from a structural perspective.

But on the other hand, they share the namespace. IMHO, this is strong enough commonality to allow them to be listed together in a common bucket. I would no mind separating them in different sub-buckets though, if that helps.

> I think it's also important to realize that immutability can be temporal, and that attributes that are read-only at the time of one request may be mutable at a later time due to a resource's change in state.

Is not it an argument in favor of having such mechanism? Many things about resource _can_ be temporal: its existence, its representation in any given response, immutability of some members. Temporal aspect can be handled via resource versioning.

jelhan:
I think this could be very well solved using an extension available in v1.1. An extension does not require changes to base spec. This should allow experiments and quicker iteration than trying to add it to main spec.

bednic:
I like this 👍

ghost (commented):
https://www.virustotal.com/graph/embed/g80fa9101a8454b6dbf8b1297f4fc474178b2a2fa1a5f4fcc8f0e5d5eb37289b6?theme=dark

Issue #159, opened 2013-11-12
Does not implement Web Linking


rkh:
JSON API currently does not implement [Web Linking](http://tools.ietf.org/html/rfc5988). More specific, it does not implement relations according to the RFC.

Judging from a Twitter discussion on Sunday, this might be intentional rather than by accident.

RFC 5988 divides into two kinds of relations types: Registered and extension types. I think the main reason this was not used is because extension types are supposed to be URIs (which are not meant to be dereferenced).

However:

> Note that while extension relation types are required to be URIs, a serialisation of links can specify that they are expressed in another form, as long as they can be converted to URIs.
## Why is this bad?
- The server cannot rely on general purpose clients inferring semantics from well defined, [registered relation types](http://www.iana.org/assignments/link-relations/link-relations.xhtml), like "next". A side effect of this is that a "profile" meta attribute had to be added, which could have otherwise been just a link.
- It is not possible to properly use extension types, making integration with other API standards, like [json-home](http://tools.ietf.org/html/draft-nottingham-json-home-03) a bit more problematic.
### Not fixing this

``` json
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author": "http://example.com/people/17",
      "comments": "http://example.com/comments/1"
    }
  }]
}
```

You could assume the standpoint that generic implementations of pagination and integration with json-home are not worth it, or not feasible and that a mapping from relation types to JSON API links should happen in the documentation, if at all.
### Fully implementing it

This would probably be the most radical approach and break compatibility with existing implementations.

An example of what this would look like:

``` json
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author": "http://example.com/people/17",
      "http://example.com/rel/comments": "http://example.com/comments/1"
    }
  }]
}
```

Or with a well defined shorthand (inspired by HAL):

``` json
{
  "links": {
    "curies": {
      "ex": "http://example.com/rel/{rel}",
      "acme": "http://docs.acme.com/relations/{rel}",
    }
  },
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
      "author": "http://example.com/people/17",
      "ex:comments": "http://example.com/comments/1"
    }
  }]
}
```

To simplify this, the `:` prefix could be defined as `://{host}/relations/{rel}` implicitly:

``` json
{
  "posts": [{
    "id": "1",
    "title": "Rails is Omakase",
    "links": {
[...]

PR #590, opened 2015-05-01
Proposal: discoverable actions

hhware:
IMHO, the suggestion in the FAQ to use `OPTIONS` to communicate a list of actions available on a resource is impractical. There are only 5 HTTP verbs in common use, it is clearly not enough for a complex API. The solution is to provide custom endpoints (such as `POST /articles/1/postprocess`). Essentially, these endpoints provide additional actions for the "parent" endpoint (in this case `/article/1`), but they are not discoverable by performing `OPTIONS` on the parent. 

This proposal is an attempt to provide a standardized solution to this problem: a resource-level member `actions`, which would make JSON API more suitable for APIs supporting HATEOAS. I hope that the proposed format could cover majority of use cases. For methods with complex interfaces, and where the server wishes to provide more information than fits in stadanrd members, `meta` can be used.

I would like to request to consider this for 1.0. It is additive feature, but I hope it is not too controversial, and I also think that the argument for consideration for 1.0 from section Milestoning of https://github.com/json-api/json-api/pull/431#issuecomment-90378980 applies here.

### DIFF
--- a/faq/index.md
+++ b/faq/index.md
@@ -39,7 +39,8 @@ interchange format itself.
 
 ### How to discover resource possible actions? <a href="#how-to-discover-resource-possible-actions" id="how-to-discover-resource-possible-actions" class="headerlink"></a>
 
-You should use the OPTIONS HTTP method to discover what can be done with a
+If resource-level member `"actions"` is not provided by the server,
+you should use the OPTIONS HTTP method to discover what can be done with a
 particular resource. The semantics of the methods returned by OPTIONS is defined
 by the JSON API standard.
 
--- a/format/index.md
+++ b/format/index.md
@@ -192,6 +192,8 @@ In addition, a resource object **MAY** contain any of these top-level members:
   relationships (described below).
 * `"meta"`: non-standard meta-information about a resource that can not be
   represented as an attribute or relationship.
+* `"actions"`: a list of actions, which the client can perform on 
+  the given resource.
 
 Here's how an article (i.e. a resource of type "articles") might appear in a document:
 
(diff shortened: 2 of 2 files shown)


ethanresnick:
I think some way to discover the available actions is needed, but the way to do it is far from obvious: the approaches range from the most granular (e.g. resource-level data like you're proposing) to schemas, schema documents, and specs like [json home](http://tools.ietf.org/html/draft-nottingham-json-home-02). So I think your proposal is a very good start to the discussion, but I suspect it will certainly trigger a heated debate, and so is probably something best left to consider after 1.0, which now has a (supposed) release date of May 21.

hhware:
> approaches range from the most granular (e.g. resource-level data like you're proposing) to schemas, schema documents, and specs like json home.

@ethanresnick, exactly, this PR is specifically about granular control, which could enable HATEOAS features in the APIs. It specifically requires the returned information to be valid for the given client on the given resource at the given moment of time. Another consideration is that it should not require additional trips to the server: this could be very important to some APIs.

Schemas are static. I think, they do not fit the bill.

Ashamed of my ignorance, but what is "schema document"? Could you provide a reference?

As for JSON Home, AFAIK, it is endpoint-centric, i.e. it allows to provide additional information about specific endpoint but does not allow to link custom endpoints to the "parent" endpoint, which is the primary goal of this proposal. Am I wrong?

Also, JSON Home documents are supposed to be provided at separate endpoints. The format of JSON Home messages could be potentially reused inside JSON API responses, but it is unclear whether it gives much of advantage without full conformance to JSON Home spec.

I am ready to rewrite this PR to comply with a standard which can accommodate the proposed change. Are there such standards?

> it will certainly trigger a heated debate, and so is probably something best left to consider after 1.0, which now has a (supposed) release date of May 21.

If it happens, it will inevitably have to wait until post 1.0, but it is not the case as of yet. Not that I am going to stop caring about JSON API after 1.0, but given how long 1.0 has taken, 1.1 may easily take another year, having in mind that backward compatibility guarantees will be in place and things will likely be considered with extra caution (the cost of [minting a new media type](https://github.com/json-api/json-api/issues/432#issuecomment-79495877) seems pretty high to me). As I tried to describe in section Milestoning of https://github.com/json-api/json-api/pull/431#issuecomment-90378980, there is a POV of the spec and libraries, and there is a POV of specific APIs. From the POV of specific APIs (the end users of the spec), there is no point to postpone _anything_ post 1.0 if it _can be handled_ before 1.0 (e.g., because it is clear and non-controversial), whether the change is additive or not. The spec is timeless "law of the land", the APIs are "people", who are born and die under that law, and they have no second chance if the law happened to be changed too late in their lifetime. Lawmakers should care about both abstract justice and people.

ethanresnick:
> Ashamed of my ignorance, but what is "schema document"? Could you provide a reference?

Sorry, my mistake using loose terminology. By schema vs. schema document, I was just trying to distinguish between proposals we've had to serve the schema at the top-level of each request, vs. to create a dedicated endpoint for it (the contents of that endpoint being what I was calling the "schema document"). But they're basically the same :)  Also, I think that some of those proposals tried to describe (at least) the valid https methods for the resources they were describing, which is why I put them alongside json-home.

> As for JSON Home, AFAIK, it is endpoint-centric, i.e. it allows to provide additional information about specific endpoint but does not allow to link custom endpoints to the "parent" endpoint, which is the primary goal of this proposal. Am I wrong?

Gotcha. You're looking for true HATEOS, as opposed to just some API-wide documentation (even if machine-readable) at a separate endpoint. I think this could be a good idea, though I'm not passionate either way.

All I'd suggest for this PR, then, is checking out the history [https://github.com/json-api/json-api/issues/159](here), which is applicable at least to the choice of the `"label"` keyword. Also check out the other efforts to support Web Linking in JSON, if you haven't seen them already. (Decent overview [here](https://www.mnot.net/blog/2011/11/25/linking_in_json).)

hhware:
@ethanresnick, thanks for the references!

> You're looking for true HATEOS

Not necessarily. I am actually looking for a flexible option, which can hopefully be used by "true HATEOAS" APIs as well as by API clients which do not really browse endpoints in a HATEOAS sense, but still can benefit from knowing the list of actions available on a given resource to a given user at a given moment of time. For example, this can allow web frontend to accurately display to the end user only the UI controls which are currently available to her on a given resource, which is orthogonal to HATEOAS.

hhware:
I have formulated another proposal to address the issue in a more general way. I am closing this PR in favor of #622.

PR #592, opened 2015-05-02
Add "collection" link

ethanresnick:
In the response to requests for single resources, this allows a top-level link that points to the collection from which the resource was drawn.

This is an additive change, and it's for a feature that hasn't been requested to date. But it's seems so obvious—the key matches an [existing link relation name](http://tools.ietf.org/html/rfc6573) (like all our other `"links"` keys), and it aids discovery, so why not have it?—that I thought I'd try to slip it in for 1.0. 

Let me know if I'm overlooking anything here!

### DIFF
--- a/implementations/index.md
+++ b/implementations/index.md
@@ -47,7 +47,7 @@ assembled to vet them.
 * [lode / jsonapi](https://github.com/lode/jsonapi) a simple and friendly library, easy to understand for people without knowledge of the specification.
 
 ### Node.js <a href="#server-libraries-node-js" id="server-libraries-node-js" class="headerlink"></a>
-* [Fortune.js](http://fortunejs.com) is a framework built to implement json-api.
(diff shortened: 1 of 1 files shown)


ethanresnick:
Note: one could argue that this should exist at the resource-level rather than at the top-level, which would allow it to also be specified in responses that return more than one resource (which might not all be from the same collection). I'd be totally fine with that, if it's what others prefer, I just wasn't sure whether you guys would want to reserve another key (i.e. besides `"self"`) in the resource-level `"links"`.

ethanresnick:
Update: This PR was originally to add a "collection" link at the document's top level. But, this key was always better-suited for the resource level; it's just that, prior to #618, it couldn't exist at that level without us reserving another key. So, once #618 is done, I'll rebase this PR to add `"collection"` as an optional, resource-level link.

PR #596, opened 2015-05-05
Change of 409 responses to comply with RFC 7231

cziegenberg:
Following RFC 7231 for the HTTP/1.1 protocol the server "SHOULD generate a payload that includes enough information for a user to recognize the source of the conflict" when sending a 409 status code:
https://tools.ietf.org/html/rfc7231#section-6.5.8


cziegenberg:
Can this be merged? I think the additional sentence clearly solves the inconsistent with RFC 7231.

tkellen:
:+1:

PR #598, opened 2015-05-05
Move original draft creation to beginning of history.

dgeb:


dgeb:
cc @wycats @wifelette

Issue #567, opened 2015-04-21
Allowed format/characters for member names


cziegenberg:
I thought about the format of member names and the following questions came into my mind:
- Is there a format for (top-level) member names? 
- Which characters are allowed (at special positions)? 
- Are member names case sensitive? 
- Can I use integers as members?

Suggestions:
- Periods (U+002E FULL-STOP, ".") aren't allowed, because it's used as the relationship path separator.
- Commas (U+002C COMMA, ",") aren't allowed, because it's used in the include parameter as a separator for multiple relationship paths.
- Members should not start with a digit ("0-9"), as this could be a problem when mapping API to the code, e.g. in PHP and C# object members of this type aren't allowed.
- Members should not contain a minus (U+002D HYPHEN MINUS, "-"), as this could be a problem when mapping API to the code, e.g. in PHP and C# object members of this type aren't allowed.
- Members should not contain a colon (U+003A COLON, ":"), as this could be a problem when mapping API to the code, e.g. in PHP and C# object members of this type aren't allowed. Also they could be reserved for later use in the include parameter?!
- When checking for duplicate members, members should be handled case insensitive, because they name a resource, which has special semantics, and this is independent from the spelling. When using member names in a request (e.g. with the include parameter), they should be handled case sensitive, because this would allow duplicate content for different spellings.
- Theoretically it isn't important if lower- or upper-case characters are used for the member names, but I'd suggest to always use lower-case, as this is the current format of the standard members and it would simplify the use by API clients.
- Underscores (U+005F LOW LINE, "_") should be allowed, as I don't know problems with underscores in member variables of any language. They could be used as the first character, but IMHO that wouldn't be good for readability.

Resulting format:
^[a-z]+[a-z0-9_]*$

passcod:
Are you talking about what is covered here? http://jsonapi.org/recommendations/#naming Also probably see #571.

dgeb:
I think it's probably a good idea for us to restrict use of periods and commas in names, since both could lead to ambiguity in interpreting certain query params.

I'm also tempted to not allow hyphens at the beginning of names, and go back to the assumption that default sort criteria is ascending (instead of requiring a plus prefix). This has proven somewhat annoying to implement due to URL encoding.

However, there's no reason to limit the names used in a serialization format based on language requirements. Every language has its own (unique) requirements, and every implementation will need a serialization / deserialization layer.

Marking for consideration by 1.0.

cziegenberg:
[...]

PR #601, opened 2015-05-08
Allowed format/characters for member names

cziegenberg:
This patch contains the changes discussed in #567, to reach the following goals:
- Define basic MUSTs for naming members that were missing in the spec (including support for NON-ASCII characters)
- Give API designers the freedom to use another naming scheme if required
- Define recommendations to standardize the naming

This suggestion includes several considerations
- Existing naming of reserved members
- URL mapping / URL safe characters ([RFC 3986](http://tools.ietf.org/html/rfc3986))
- Readability and usability, common spelling
- Reduction of potential implementation problems with some languages
- Not breaking the previous naming recommendations

### DIFF
--- a/recommendations/index.md
+++ b/recommendations/index.md
@@ -9,11 +9,14 @@ are beyond the scope of the base JSON API specification.
 
 ## Recommendations for Naming <a href="#naming" id="naming" class="headerlink"></a>
 
-The allowed characters for naming members are defined in the spec, but it's recommended to only use the lower-case ASCII characters a-z and 0-9, and to start the name with a-z. Furthormore the hypen minus (U+002D HYPHEN-MINUS, "-") should be used as a word separator.
+The allowed and recommended characters for an URL safe naming of members are defined in the format spec. In addition, to standardize member names, the following restrictions are recommended:
+- Member names **SHOULD** be all lower-case.
+- Member names **SHOULD NOT** start with a digit (U+0030 to U+0039, "0-9")
+- Member names cointaining multiple words **SHOULD** use a hyphen minus (U+002D HYPHEN-MINUS, "-") as word separator.
 
-All these recommended characters are URL safe following [RFC 3986](http://tools.ietf.org/html/rfc3986), so member names can be mapped to URLs without translation (see below for details).
-
-These recommendations help to standardize the naming of members and to avoid potential problems.
+Together with the recommendations in the format spec this results in:
+- Member names **SHOULD** start and end with the characters "a-z" (U+0061 to U+007A)
+- Member names **SHOULD** contain only the characters "a-z" (U+0061 to U+007A), "0-9" (U+0030 to U+0039) and "-" (U+002D HYPHEN-MINUS)
 
 ## Recommendations for URL Design <a href="#urls" id="urls" class="headerlink"></a>
 
--- a/recommendations/index.md
+++ b/recommendations/index.md
@@ -9,14 +9,9 @@ are beyond the scope of the base JSON API specification.
 
 ## Recommendations for Naming <a href="#naming" id="naming" class="headerlink"></a>
 
-The allowed and recommended characters for an URL safe naming of members are defined in the format spec. In addition, to standardize member names, the following restrictions are recommended:
-- Member names **SHOULD** be all lower-case.
-- Member names **SHOULD NOT** start with a digit (U+0030 to U+0039, "0-9")
-- Member names cointaining multiple words **SHOULD** use a hyphen minus (U+002D HYPHEN-MINUS, "-") as word separator.
-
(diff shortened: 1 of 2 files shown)


dgeb:
@ziege thanks for your work here and in #567.

I believe that many of these restrictions are sound but some go too far. It is important to allow for public discussion of each restriction. If you don't mind making a lot of tweaks, we can comment on them inline.

To start, I'd prefer that the recommendations marked **SHOULD** be moved exclusively to the recommendations page.

Also, there needs to be a clarification of the term ASCII. We are discussing non-control characters 32-126 only, correct? I believe that extended ASCII codes (>= 128) should probably be covered in the "non-ASCII" category.

cziegenberg:
> If you don't mind making a lot of tweaks, we can comment on them inline.

Let's try it...

> I'd prefer that the recommendations marked SHOULD be moved exclusively to the recommendations page.

I also thought about placing it there, but didn't because these recommendations are relatively important. I also thought about linking the recommendation page but wanted to keep the relevant information in one document.

> We are discussing non-control characters 32-126 only, correct?

Yes, the allowed ASCII characters are defined in detail in the sub sections of the added "Member names" section.

ziege:
Sorry, one big comment... Didn't work on my mobile.

dgeb:
> Yes, the allowed ASCII characters are defined in detail in the sub sections of the added "Member names" section.

The "Member names" section does not define what is meant by ASCII characters. It calls out specific ASCII characters, but does not define the whole set. Therefore, it is also unclear what is meant by "non-ASCII" characters.

dgeb:
> I also thought about placing it there, but didn't because these recommendations are relatively important. I also thought about linking the recommendation page but wanted to keep the relevant information in one document.

To tell you the truth, I am reluctant to even include these as recommendations. Although the specification is in English and uses English reserved keys, non-English characters are perfectly valid for implementation-specific keys.

cziegenberg:
Ah, okay. First we called it UNICODE characters, but as this can overlap with ASCII, I changed it to NON-ASCII...

I will change it. Perhaps "UNICODE except ASCII (code 0 - 127)" is clearer. Or do you have a better suggestion?

dgeb:
Although it's a little awkward, I'm not opposed to the term "non-ASCII", but believe it needs to be defined. I think that "UNICODE except ASCII (code 0 - 127)" is the right definition.

cziegenberg:
I did the following changes:
- Removed the word "ASCII" as it wasn't clearly defined
- Divided the recommendations in two parts:
  - General recommendation in the format spec to use URL safe characters, which is relevant for URL mapping.
  - Further recommendations to standardize the member name format in the recommendations page only.
- Avoided duplications

recommendations/index.md
```diff
@@ -9,13 +9,14 @@ are beyond the scope of the base JSON API specification.
...
-pluralized resource types can be used as URL segments without translation,
-as discussed below.
+The allowed and recommended characters for an URL safe naming of members are defined in the format spec. In addition, to standardize member names, the following restrictions are recommended:
+- Member names **SHOULD** be all lower-case.
+- Member names **SHOULD NOT** start with a digit (U+0030 to U+0039, "0-9")
+- Member names cointaining multiple words **SHOULD** use a hyphen minus (U+002D HYPHEN-MINUS, "-") as word separator.
+
+Together with the recommendations in the format spec this results in:
+- Member names **SHOULD** start and end with the characters "a-z" (U+0061 to U+007A)
+- Member names **SHOULD** contain only the characters "a-z" (U+0061 to U+007A), "0-9" (U+0030 to U+0039) and "-" (U+002D HYPHEN-MINUS)
```
passcod: Why not just discard all that is in this PR and specify only those two rules, then? It seems simpler and much clearer (the long lists in the spec especially are really tedious).

cziegenberg:
@passcod: I reduced it to the two resulting rules.

ethanresnick:
@ziege Thanks for fighting for all these changes! This PR really looks great, at least from where I'm standing. In particular, reserving the punctuation that we're already using (for sorting etc.) seems so crucial that I can't believe it didn't come up before, and reserving other punctuation will, I suspect, be very helpful down the line for extending the query parameter syntax. (In particular, I think it'll be important for allowing us to reference new members at the top-level of a resource object, besides `type` and `id`.) So again, just great work!

cziegenberg:
Thanks for the positive feedback @ethanresnick!

tkellen:
:+1:

dgeb:
Thanks again very much, @ziege! :shipit:

cziegenberg:
Great! Thanks to you too!

jonkoops:
Thanks @ziege, these are some important additions.

PR #605, opened 2015-05-08
Allow missing `Accept` header

ethanresnick:
The current text says:

> Servers **MUST** return a `406 Not Acceptable` status code if [the `Accept`] header is missing

But this is a violation of the HTTP spec, which [says](https://tools.ietf.org/html/rfc7231#section-5.3.2) that a missing `Accept` header "implies that the user agent will accept any media type in response". Therefore, `406` can never be an appropriate response here. 

So that left two options: keep the requirement for an `Accept` header—even when specifying it wouldn't be necessary for HTTP content-negotiation—and return a `400` when it's missing; or, just allow a missing `Accept` header and treat it according to HTTP's semantics (i.e. as implying `*/*`). Since I don't think we want to mess with HTTP, the second option seemed better.

So this PR both removes the "is missing" part from the 406 conditions and updates the first sentence to say that the `Accept` header isn't required per se; what's required is that the client indicates (per HTTP) that it can accept the JSON-API media type.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -510,10 +510,10 @@ The top-level links object **MAY** contain the following members:
 Data, including resources and relationships, can be fetched by sending a
 `GET` request to an endpoint.
 
-JSON API requests **MUST** include an `Accept` header specifying the JSON
-API media type. This header value **MUST** also include media type
-extensions relevant to the request. Servers **MUST** return a `406 Not
-Acceptable` status code if this header is missing or specifies an
+Clients **MUST** indicate that they can accept the JSON API media type, per
+the semantics of the HTTP `Accept` header. If present, this header value **MUST**
+also include any media type extensions relevant to the request. Servers **MUST**
+return a `406 Not Acceptable` status code if this header specifies an
 unsupported media type.
 
 > Note: Servers may support multiple media types at any endpoint. For example,


dgeb:
Thanks @ethanresnick :+1:

neomerx:
Does it mean the following headers are valid?

```
Accept: text/*, text/plain, text/plain;format=flowed, */*
Accept: text/*;q=0.3, text/html;q=0.7, text/html;level=1, text/html;level=2;q=0.4, */*;q=0.5
```

Which means if we don't have any extensions for 'text' we should return JSON API as the last option? 

It looks not a small change from simple requirement 'header must be json api + ?ext' to parsing various types with priorities.

ethanresnick:
@neomerx Yes, those headers would be valid...as they well should be. The client's said it'll accept anything, and JSON API is an anything. Fwiw, PHP appears to have a [built-in function](http://php.net/manual/en/function.http-negotiate-content-type.php) for content negotiation (which is pretty damn cool, actually) and there are also [libraries](http://williamdurand.fr/Negotiation/) too.

**EDIT** The PHP function isn't built-in, but it is part of a standard extension. And the library works too :)

neomerx:
@ethanresnick People generally avoid installing PECL unless it's absolutely inevitable and both 'solutions' have issues on HHVM.

Anyways API clients are closely coupled to servers which obviously very different from browsers. Thus such header parsing in 99.(9)% cases will be redundant. For fast APIs that work with memcached/solr/other fast data sources it can require substantial resources (my guess 10+% loss in latency and requests per second).

Is it possible to reflect this 'coupling' fact in the spec and make parsing arbitrary headers optional?

PR #606, opened 2015-05-09
Add katharsis -  Java implementation of json-api

masterspambot:
I would like to add another language to list of implementation to JSON:API. Currently katharsis library is on finishing/polishing stage so it will be available fully-documented within 2-3 weeks.


dgeb:
@masterspambot great to see a Java implementation - thanks!

Issue #608, opened 2015-05-11
Handling multiple errors


ethanresnick:
There are cases in which the current spec cannot be interpreted literally because it requires simultaneously responding two different status codes, which is obviously impossible. Consider the following request to a server that doesn't support client-generated ids:

```
POST /people
Content-Type: application/vnd.api+json

{
  "data": {
    "type": "photos",
    "id": 550e8400-e29b-41d4-a716-446655440000",
    "attributes": {}
  }
}
```

The spec simultaneously requires here that the server:
1. "**MUST** return 403 Forbidden in response to an unsupported request to create a resource with a client-generated ID." and
2. "**MUST** return 409 Conflict when processing a POST request in which the resource's type does not match the server's endpoint."

So what's the appropriate behavior here?

My inclination is that we should say here what we're already saying in the JSON-PATCH extension (and actually that we should move the JSON-PATCH copy into the base spec). Namely, the server should have the choice _either_ to stop processing the request and respond with the appropriate status code for whichever error it happened to encounter first, _or_ to process the request all the way through, pick a generic status code, and return an array of errors.

dgeb:
> My inclination is that we should say here what we're already saying in the JSON-PATCH extension (and actually that we should move the JSON-PATCH copy into the base spec). Namely, the server should have the choice either to stop processing the request and respond with the appropriate status code for whichever error it happened to encounter first, or to process the request all the way through, pick a generic status code, and return an array of errors.

:+1:

tkellen:
@ethanresnick Have the bandwidth to tackle a PR for this?

ethanresnick:
@tkellen I'll try later tonight, but no guarantees. Like I said, all my family's in town, so I don't have much time, and I'm much more focused on trying to write up something for #614 (which will take a ton of copy, but it ultimately seems like the most important issue, since it determines what we can get wrong now without too many consequences).

tkellen:
Resolved in #655.

PR #609, opened 2015-05-11
Don't require `Accept` validation before `Content-Type` validation

ethanresnick:
The current text implicitly requires a server to check the validity of the `Accept` header (and respond 406 if it's invalid) before checking the validity of the `Content-Type` header. This PR allows the server to check the headers in either order, which will be more convenient for implementors. (This was quite an arbitrary restriction anyway; if anything, it would make sense to check the format of incoming data, i.e. `Content-Type`, first.)

I think the only reason that the text currently reads as it does is to prevent the kind of incompatibility discussed in #608, in which the spec can be simultaneously mandating two different response codes (which it would indeed be doing under my revised text if _both_ the `Accept` and the `Content-Type` headers are invalid). But, as discussed in #608, this type of contradiction can already happen in other places, so we need a generic way to handle it. And it seems unlikely that the approach used here—namely specifying an implicit processing order—is the one we want to use, as doing that for every possible cause of multiple errors would unduly constrain implementations.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -91,9 +91,8 @@ name in the `ext` media type parameter with the `Accept` header. Servers
 that do not support a requested extension or combination of extensions
 **MUST** return a `406 Not Acceptable` status code.
 
-If the media type in the `Accept` header is supported by a server but the
-media type in the `Content-Type` header is unsupported, the server
-**MUST** return a `415 Unsupported Media Type` status code.
+If the media type in the client's `Content-Type` header is unsupported, the
+server **MUST** return a `415 Unsupported Media Type` status code.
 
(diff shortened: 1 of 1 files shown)


tkellen:
:+1:

dgeb:
This seems good @ethanresnick. Could you please re-target this for the [Extensions](http://jsonapi.org/extensions/) page?

I think we have similar ordering logic in the base spec's new [media type negotiation](http://jsonapi.org/format/#media-type-negotiation) section that could be simplified similarly.

tkellen:
I believe this was resolved in #703.

dgeb:
Yes - this has been resolved.

Issue #604, opened 2015-05-08
Error Object Extensibility


ethanresnick:
Right now, every spec-defined object is declared as not allowing extra members (unless an extension is in use), except for errors objects, which say that the user MAY add extra members. Is there a reason for this exception? It seems like the rationale for preventing users from adding members to other objects (namely, to preserve base spec extensibility), applies just as much to errors, no? We’ve already seen [one instance](https://github.com/json-api/json-api/issues/584) of someone wanting to add another property to the errors object, so our wanting to extend this object in the future doesn’t seem impossible…

My suggestion would be to keep error objects non-extensible, just like everything else.

(For reference, btw, when I say that all other spec-defined objects are declared as inextensible, I’m referencing this clause: “Unless otherwise noted, objects defined by this specification MUST NOT contain any additional members. Client and server implementations MUST ignore object members not recognized by this specification.”)

ethanresnick:
@dgeb @tkellen thoughts on this?

tkellen:
Agreed! Would you mind opening a PR to make this change?

ethanresnick:
@tkellen [Done](https://github.com/json-api/json-api/pull/610)!

tkellen:
Cool! I will review this with @dgeb in the AM.

ethanresnick:
This can be closed now too! :)

PR #610, opened 2015-05-11
Don't allow additional members in error objects by default

ethanresnick:
This is a PR for #604

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -1511,6 +1511,8 @@ An error object **MAY** have the following members:
   associated resource(s) in the request document. Each path **MUST** be relative
   to the resource path(s) expressed in the error object's `"links"` member
   [e.g. `["/first-name", "/last-name"]` to reference a couple attributes].
+* `"meta"` - to contain non-standard meta-information about the error.
+
 
 [attributes]: #document-structure-resource-object-attributes
 [links]: #document-structure-resource-links
--- a/format/index.md
+++ b/format/index.md
@@ -1512,8 +1512,6 @@ An error object **MAY** have the following members:
   to the resource path(s) expressed in the error object's `"links"` member
   [e.g. `["/first-name", "/last-name"]` to reference a couple attributes].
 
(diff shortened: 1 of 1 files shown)


csantero:
My implementation customizes the error object quite a bit. There is a fair amount of useful information that I'm not sure I could shoehorn into the allowed fields. At the very least, can we allow `meta` on error objects?

ethanresnick:
@csantero What are the extra fields you've added? Maybe we could include them in base spec. Or, if some of them really apply just to your api, they could always be declared in an extension.

csantero:
@ethanresnick The additional keys I include in JSONAPI.NET at the moment are:
- `stackTrace` for unhandled server exceptions
- `inner`, which recursively serializes Exception.InnerException properties.

Eventually I'd like to support pluggable formatting on a per-exception type basis, to provide richer exception details. A user of the library could define a custom Exception class with 10 different fields, and there's no way those would map to any-predefined members on the json-api spec. A `meta` key would be perfect for that.

ethanresnick:
Ok, I've updated the PR to allow `"meta"` in errors, so the net effect of this is to continue to allow users to add properties to errors (which there's clearly a demand for), but require them to do it in `"meta"`, so that we can standardize some of those property names down the line without conflict.

format/index.md
```diff
@@ -1511,8 +1511,8 @@ An error object **MAY** have the following members:
   associated resource(s) in the request document. Each path **MUST** be relative
   to the resource path(s) expressed in the error object's `"links"` member
   [e.g. `["/first-name", "/last-name"]` to reference a couple attributes].
+ * `"meta"` - to contain non-standard meta-information about error.
```
csantero: The indentation for this item looks off in the rendered markdown. Also it might be better to say `the error` instead of `error`.

Otherwise :+1:

ethanresnick: Fixed!

dgeb:
:+1:

Issue #618, opened 2015-05-13
Proposal for `"relationships"` member at resource object top-level


ethanresnick:
Here's what I'm proposing. Instead of resource objects looking like:

```
{
  attributes: {...}
  links: {
   self: "",
   tags: {...}
  }
}
```

We'd have:

```
{
  attributes: {...},
  links: {
   self: ""
  },
  relationships: {
   tags: {...}
  }
}
```

That is, the self link would go in a separate `"links"` object, and everything currently in the resource-level `"links"` would go in a new container (re)named `"relationships"`. (`"tags"` is just an example relationship in the code sample above.)

There were two things motivating this proposal:
1. Extensibility of resource-level `"links"`. It's quite likely that we may want to add other `"links"` in the future, and we can't do this post 1.0 if `"links"` and relationships share a namespace. Even pre 1.0, we can't do this without putting further restrictions on the name of a user's relationships, which I don't think we want to do. As examples of keys we might want to add to resource-level `"links"`, consider: `"collection"` (the url for the collection the resource is from), `"latest-version"`, `"alternate"`, etc. all of which are [standard](www.iana.org/assignments/link-relations.html) link-relation names that could be quite useful. I'm also thinking about how we set ourselves up to be able to implement web linking later if a common json format for it ever arises, and having a separate property bag where every official relation name is a valid key, and where the value is always a url, seems like a prerequisite.
2. Clarity, in a few different senses. First, the spec talks about "relationships" and "related resources", so having this terminology reflected in the json (much like we did for attributes), would make the spec easier to follow. Second, the `"self"` link really does seem to be a different kind of thing from a relationship, in that it can't have linkage and that a request to, e.g. `/people/1/links/self` doesn't make sense, so putting it in a different container also seems appropriate. Moreover, we'd get consistency between the format of the `"links"` object at the resource level and the top-level. (Shout out to @bintoro for making most of these points about terminology in #380. The key difference between my proposal and the discussion there is that I'm proposing to have the relationships and the links into two separate objects, where as the prior discussion left them in the same object but debated that object's name.)

Here are the effects that my proposal would have on terminology:
1. The object holding the relationships be called the "relationships" object. Each "link object" it contains would become a "relationship object". The resource-level `"links"` object, which still exists, would continue to be referred to as such.
[...]

PR #625, opened 2015-05-17
Relationships object

ethanresnick:
A PR for #618. See the inline comments for some new issues that arose while I was creating it.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -283,21 +283,8 @@ Relationships may be to-one or to-many. Relationships can be specified by
 including a member in a resource's relationship's object. The name of the
 relationship is its key in the the relationship object.
 
-The value of a relationship **MUST** be one of the following:
-
-* A URL for the related resource(s) (a "related resource URL"). When fetched, it
-  returns the related resource object(s) as the response's primary data. For
-  example, an `article`'s `comments` relationship could specify a URL that
-  returns a list of comment resource objects when retrieved through a `GET`
-  request. A related resource URL **MUST** remain constant even when the
-  relationship (the set of referenced resources) mutates. That is, the response
-  from a related resource URL always reflects the current state of the
-  relationship.
-
-* An object (a "relationship object").
-
-If a relationship is provided as a relationship object, it **MUST** contain at
-least one of the following:
+The value of a relationship **MUST** be an object (a "relationship object"),
+which **MUST** contain at least one of the following:
 
 * A `"links"` member that contains at least one of the following:
   * A `"self"` member, whose value is a URL for the relationship itself (a
@@ -335,7 +322,7 @@ relationship URLs.
 If present, a *related resource URL* **MUST** be a valid URL, even if the
 relationship isn't currently associated with any target resources.
 
-For example, the following article is associated with an `author` and `comments`:
+For example, the following article is associated with an `author`:
 
 ```javascript
 // ...
@@ -352,8 +339,7 @@ For example, the following article is associated with an `author` and `comments`
         "related": "http://example.com/articles/1/author"
       },
       "linkage": { "type": "people", "id": "9" }
-    },
-    "comments": "http://example.com/articles/1/comments"
+    }
   },
   "links": {
     "self": "http://example.com/articles/1"
@@ -365,10 +351,6 @@ For example, the following article is associated with an `author` and `comments`
 The `author` relationship includes a URL for the relationship itself (which
 allows the client to change the related author directly), a related resource URL
 to fetch the resource objects, and linkage information.
-
-The `comments` relationship is simpler: it just provides a related resource URL
-to fetch the comments. The URL can therefore be specified directly as the
-attribute value.
 
 #### Resource Links <a href="#document-structure-structure-resource-object-links" id="document-structure-resource-object-links" class="headerlink"></a>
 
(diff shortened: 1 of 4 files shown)


format/index.md
```diff
@@ -266,45 +269,19 @@ consistently throughout an implementation.
...
-("relationships"). Relationships may be to-one or to-many. Relationships
-can be specified by including a member in a resource's [links] object.
+The keys `"id"` and `"type"` are not allowed within the relationships object.
 
-The name of the relationship declared in the key **SHALL NOT** be `"self"`.
+Relationships may be to-one or to-many. Relationships can be specified by
+including a member in a resource's relationship's object. The name of the
+relationship is its key in the the relationship object.
 
 The value of a relationship **MUST** be one of the following:
```
ethanresnick: @dgeb @tkellen How would you guys feel now about getting rid of the `"name": "related resource url"` representation of a relationship? There was an [argument for](https://github.com/json-api/json-api/pull/453) removing it before, and I think that argument's much stronger now, as this would be the only instance in the spec of a URL occuring outside of a `"links"` object.

format/index.md
```diff
@@ -472,14 +488,14 @@ multiple times.
 
 As discussed above, the document **MAY** be extended to include
 meta-information as `"meta"` members in several locations: at the top-level,
-within resource objects, within link objects, and within linkage objects.
+within resource objects, within relationship objects, and within linkage objects.
```
ethanresnick, line 473: As I mentioned off-handedly in #618, I think there's an argument for renaming "linkage objects" to "resource identifier objects". This is a bit longer, but it emphasises very nicely the spec’s concept that type/id is used to uniquely identify a resource, and it allows us to (more sensibly) refer to objects of this form where we’re using them in contexts other than the `"linkage"` key, such as for bulk deletion in the bulk extension.

If you guys agree, I can make a separate PR for that, since I don't want this one to get too unwieldy.

format/index.md
dgeb: @ethanresnick I agree that representation should be removed from this proposal, since it no longer merges the concept of links and relationships.

format/index.md
dgeb, line 473: I agree with this point as well. I'm not thrilled with the name, but the simplifying concept is solid.

dgeb:
@ethanresnick thanks for all your hard work on this! @tkellen and I will review and discuss first thing tomorrow.

format/index.md
ethanresnick: Done!

format/index.md
ethanresnick, line 473: Cool. I'm not thrilled with the name either, so let me know if you think of something better!

ethanresnick:
@dgeb Sounds good!

tkellen:
Late night :+1: on this. Excited to take a more considered look tomorrow AM. Thanks for proposing this @ethanresnick!

format/index.md
hhware, line 473: I would use something like "relationship tuple" or "relationship linkage tuple", and the key might be just `tuple`...

ethanresnick:
Thanks for the proofreading, @hhware! Much appreciated!

dgeb:
@ethanresnick at this point, @tkellen and I think that `data` makes more sense than `linkage` in a relationship object. This could be termed "relationship data" and would of course be accessible at "relationship URLs". What do you think?

dgeb:
To expand on that thought - it establishes `data` as the generic key to represent the data behind any kind of custom element - such as a relationship object. When elements, such as relationship objects, are retrieved at their `self` links, their `data` will be accessible as the top-level `data` member. I think it's a simplifying move that provides a good convention for future extension.

ethanresnick:
@dgeb Yup! I really like that! Should I add it to this PR, or should we merge this and open a second one?

**EDIT**: I've gotta run an errand now, but I'll put together PRs tonight for all the issues you and @tkellen +1d

dgeb:
@ethanresnick sounds great - thanks in advance!

format/index.md
ethanresnick, line 473: Anything with the word "relationship" in it misses the point for me, since I think we want to use these objects in contexts other than relationships to identify single resources. Also, I still think most people don't know what a "tuple" is and that it's kind of a scary word. (Moreover, linkage objects aren't actually tuples, since the keys aren't ordered and there can't be repeated keys.)

Maybe "resource pointer object"? Or just "resource pointer", as in "A 'resource pointer' is an object containing a "type" and "id" key, which are used to uniquely identify a resource."?

dgeb, line 473: I agree that the core concept to get across is resource identification, so let's just go with "resource identifier object" for now. This kind of terminology can always be changed later if necessary.

dgeb:
@ethanresnick I'm going to rebase and merge this along with the `data` change I mentioned above.

dgeb:
Rebased and merged (see https://github.com/json-api/json-api/commits/gh-pages). I'll open a separate PR for the `linkage` -> `data` change mentioned above.

ethanresnick:
Thanks @dgeb! Sounds good, and sorry I wasn't able to open that last night.

tkellen:
I'll have a PR in shortly with some minor revisions for cleanup/clarity and then I'm out for the day.

PR #626, opened 2015-05-18
Clarify response for empty collections

### LINKED ISSUE #331: Status Codes for simple GET requests
I realize that status codes are usually pretty much "check [the official spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html)," but given that certain cases tend to engender discourse, I'm thinking that "your anti-bikeshedding weapon" should give certain basic guidelines on proper status codes for GET requests.

The particular example where I'm currently feeling some pain is a case where a collection resource is being requested with particular filters that yield an empty collection, e.g. `/posts/487/comments` or `/comments?post=487` or `/posts/487/comments?page=10` (in cases where post 487 exists).  I've seen some [rather](https://groups.google.com/forum/#!topic/api-craft/wngl_ZKONyk) [lively](http://programmers.stackexchange.com/questions/203492/when-to-use-http-status-code-404-in-an-api) [debate](http://programmers.stackexchange.com/questions/199742/restful-api-design-what-should-i-return-if-there-are-no-rows) and [strong opinion](https://www.brightpearl.com/developer/latest/concept/empty-list-response-code.html) out there on whether these should return a 200, 204, or 404, and have been engaged in some (friendly) bikeshedding about this issue.

I am inclined to support 200 for the described situations based on my interpretation of the HTTP spec, but regardless, I think it would make sense to address the issue not as a specific question but as part of generally specifying common cases of GET request status codes.  We don't need to copy the HTTP spec, but I think specifying the most common situations and questionable cases would be very helpful.

ethanresnick:
Closes #331

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -163,7 +163,7 @@ resource collections
 ```
 
 A logical collection of resources (e.g., the target of a to-many relationship)
-**MUST** be represented as an array, even if it only contains one item.
+**MUST** be represented as an array, even if it only contains one item or is empty.
 
 Error objects **MUST** appear under a top-level key named `"errors"`.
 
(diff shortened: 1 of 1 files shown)


tkellen:
:+1:

PR #627, opened 2015-05-18
Recommendation for supporting `PATCH`less clients

### LINKED ISSUE #621: Standard Incompatible with IE8
HTTP PATCH Method not supported :(

http://bugs.jquery.com/ticket/13240

ethanresnick:
Note that this only recommends _how_ to support PATCHless clients if the server wants to support them at all. It _doesn’t_ go so far as to recommend that a server should support them.

Closes #621.

### DIFF
--- a/recommendations/index.md
+++ b/recommendations/index.md
@@ -132,3 +132,11 @@ Furthermore, multiple filters can be applied to a single request:
 ```http
 GET /comments?filter[post]=1,2&filter[author]=12
 ```
+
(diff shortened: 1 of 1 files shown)


dgeb:
I'd like to see an extension that could indicate support for this, but this seems fine for now. :+1:

tkellen:
:+1:

steveklabnik:
Thanks so much for writing this up!

It's sorta unfortunate that it starts with `X-`, but it seems to be what people use, so that's better.

PR #631, opened 2015-05-20
remove + prefix for ascending sorts

tkellen:
it's weird to support due to urlencoding issues, and is no longer
desirable given that we have clear restrictions on member keys.


dgeb:
Glad to be able to return to this with our new naming restrictions. URL encoding of `+` was messy. :+1:

neomerx:
Moreover Symfony for example removed pluses from parameters anyway

PR #632, opened 2015-05-20
Introduce "resource identifier objects"

dgeb:
Instead of referring to "linkage objects" that identify resources with
`type` and `id`, this concept is generalized with a new definition:
"resource identifier objects".

Using this term throughout the spec provides simplification and
consistency, as it can be used not only for relationships but also
for bulk deletes.

Furthermore, with the introduction of the relationships object, we no
longer need to force the term "linkage" into the document structure.
Within a relationships object, this member has been replaced simply with
`"data"`, which represents the relationship data. This has a nice
parallel for relationship URLs in which relationship data _is_ the
primary data.

Credit to @ethanresnick for the term "resource identifier object".

See #618 and #625 for more history and discussion.

### DIFF
--- a/extensions/bulk/index.md
+++ b/extensions/bulk/index.md
@@ -28,7 +28,7 @@ server **MUST NOT** be changed by a request if any individual operation fails.
 Multiple resources can be created by sending a `POST` request to a URL that
 represents a collection of resources. The request **MUST** include an array
 of resource objects as primary data. Each resource object **MUST** contain
-at least a `type` member.
+at least a `"type"` member.
 
 For instance, multiple photos might be created with the following request:
 
@@ -60,7 +60,7 @@ Accept: application/vnd.api+json; ext=bulk
 Multiple resources can be updated by sending a `PATCH` request to a URL that
 represents a collection of resources to which they all belong. The request
 **MUST** include an array of resource objects as primary data. Each resource
-object **MUST** contain at least `type` and `id` members.
+object **MUST** contain at least `"type"` and `"id"` members.
 
 For example:
 
--- a/extensions/bulk/index.md
+++ b/extensions/bulk/index.md
@@ -91,8 +91,8 @@ Accept: application/vnd.api+json; ext=bulk
 Multiple resources can be deleted by sending a `DELETE` request to a URL that
 represents a collection of resources to which they all belong.
 
-The body of the request **MUST** contain a `data` member whose value is an
-an array of objects that each contain a `type` and `id`.
+The body of the request **MUST** contain a `"data"` member whose value is an
+an array of resource identifier objects.
 
 For example:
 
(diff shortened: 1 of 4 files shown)


extensions/bulk/index.md
```diff
@@ -91,8 +91,8 @@ Accept: application/vnd.api+json; ext=bulk
 Multiple resources can be deleted by sending a `DELETE` request to a URL that
 represents a collection of resources to which they all belong.
 
-The body of the request **MUST** contain a `data` member whose value is an
-an array of objects that each contain a `type` and `id`.
+The body of the request **MUST** contain a `"data"` member whose value is an
+an array of resource identifier objects.
```
tkellen, line 95: I would link this to the resource identifier objects section (same for all references).

format/index.md
```diff
@@ -220,18 +237,16 @@ its "[fields]".
 
 #### Resource Identification <a href="#document-structure-resource-identification" id="document-structure-resource-identification" class="headerlink"></a>
 
-Every JSON API resource object is uniquely identified by a combination of
-its `type` and `id` members. This identification is used for linkage in
-compound documents and batch operations that modify multiple resources at a
-time.
+Every resource object is uniquely identified by a combination of its `"type"`
```
tkellen: by _the_ combination of ?

format/index.md
```diff
@@ -1201,7 +1222,7 @@ Accept: application/vnd.api+json
 #### Updating a Resource's Relationships <a href="#crud-updating-resource-relationships" id="crud-updating-resource-relationships" class="headerlink"></a>
 
 If a relationship is provided in the `links` section of a resource object in a
```
tkellen: should this be the `relationships` section?

tkellen:
:+1: with those minor nits addressed. I'm okay with waiting on the inter-document linking, if you think it'll be too arduous given our current timeframe.

format/index.md
dgeb: Good catch - we should have replaced this when introducing relationships.

format/index.md
dgeb: agreed - "the" would be better here

extensions/bulk/index.md
dgeb, line 95: Agreed - might as well add these references now.

dgeb:
@tkellen Thanks for the review. I just made all those changes with the exception of the inter-document links. We'll need a thorough review to catch all of those. Plus I'm not sure if there's a better strategy than just hard-coding the URLs in each document.

tkellen:
:+1:

ethanresnick:
+1. Thanks for actually writing this up, @dgeb. My family's in town today and tomorrow and I'm graduating, so I didn't have a chance to do this.

dgeb:
Wow, congratulations @ethanresnick! Hopefully our little spec will graduate along with you :)

ethanresnick:
Thanks @dgeb! It's been quite a crazy day.

neomerx:
@ethanresnick for someone crazy day only begins :wink: 
I've just closed 'relationships' change and got another one

PR #636, opened 2015-05-20
Fix type.  Missed a couple of links->relationships

ndrochak:
Resource object "links" were renamed to "relationships".  A couple of places in the spec that change was missed (I think).

HTH,
Nick

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -1032,7 +1032,7 @@ Accept: application/vnd.api+json
 }
 ```
 
-If a relationship is provided in the `links` section of the resource object, its
+If a relationship is provided in the `relationships` section of the resource object, its
 value **MUST** be a relationship object with a `linkage` member. The value of this key
 represents the linkage the new resource is to have.
 
@@ -1207,7 +1207,7 @@ Accept: application/vnd.api+json
 
 #### Updating a Resource's Relationships <a href="#crud-updating-resource-relationships" id="crud-updating-resource-relationships" class="headerlink"></a>
 
-If a relationship is provided in the `links` section of a resource object in a
+If a relationship is provided in the `relationships` section of a resource object in a
 `PATCH` request, its value **MUST** be a relationship object with a `linkage` member.
 The relationship's value will be replaced with the value specified in this member.
 


steveklabnik:
:+1: 

now, let's see why github says i can't merge

ndrochak:
Oh, maybe it was fixed already.  I was too slow.

steveklabnik:
So, the reason is that since you've forked, the output has changed a bit. It needs to be updated per current master.  Do you know how to rebase it? I can walk you through the steps if not.

ndrochak:
no clue how to rebase.  i can barely spell it :)

steveklabnik:
Okay. :) Here's what you need to do. If you don't want to do any of this, let me know and I can just take care of it, but I figured you may want to learn :)

First, start off in your json-api checkout, make sure you're on your `ndrochak-fix-links` branch:

```
$ git checkout ndrochak-fix-links
```

The next step is to grab the upstream changes. The quickest way in terms of commands is to use `git fetch`:

```
$ git fetch https://github.com/json-api/json-api gh-pages
remote: Counting objects: 4, done.
remote: Compressing objects: 100% (3/3), done.
remote: Total 4 (delta 2), reused 2 (delta 0), pack-reused 0
Unpacking objects: 100% (4/4), done.
From https://github.com/json-api/json-api
 * branch            gh-pages -> FETCH_HEAD
```

What this last bit of output means is that the contents of the `gh-pages` branch are now in a special ref called `FETCH_HEAD`. We'll use that next:

```
$ git rebase FETCH_HEAD
First, rewinding head to replay your work on top of it...
Applying: Fix type.  Missed a couple of links->relationships
Using index info to reconstruct a base tree...
M   format/index.md
Falling back to patching base and 3-way merge...
Auto-merging format/index.md
CONFLICT (content): Merge conflict in format/index.md
Failed to merge in the changes.
Patch failed at 0001 Fix type.  Missed a couple of links->relationships
The copy of the patch that failed is found in:
   /home/steve/src/json-api/.git/rebase-apply/patch

When you have resolved this problem, run "git rebase --continue".
If you prefer to skip this patch, run "git rebase --skip" instead.
To check out the original branch and stop rebasing, run "git rebase --abort".
```

This is why GitHub says that it can't figure out what's up. There's some kind of conflict here. You can run `git status` to see what's up:

```
$ git status
rebase in progress; onto c91461c
You are currently rebasing branch 'ndrochak-fix-links' on 'c91461c'.
  (fix conflicts and then run "git rebase --continue")
  (use "git rebase --skip" to skip this patch)
  (use "git rebase --abort" to check out the original branch)

Unmerged paths:
  (use "git reset HEAD <file>..." to unstage)
  (use "git add <file>..." to mark resolution)

    both modified:   format/index.md
```

This implies that `format/index.md` is where the problem is, which makes sense, as you edited it. Opening up the file, we can look for `=====` to see where the issue is:

```
<<<<<<< HEAD
If a relationship is provided in the `"relationships"` member of the
resource object, its value **MUST** be a relationship object with a `"data"`
member. The value of this key represents the linkage the new resource is to
have.
=======
If a relationship is provided in the `relationships` section of the resource object, its
value **MUST** be a relationship object with a `linkage` member. The value of this key
represents the linkage the new resource is to have.
>>>>>>> Fix type.  Missed a couple of links->relationships
```

The bits between `<<<` and `===` and `===` and `>>>` are the two sides. So what we can see here is that 'relationships' was fixed here, but github didn't know what to do because of the `"`s. What' we'd do here is make the text be correct, which means choosing the `"` version. So you'd delete the `<<<` and `>>>` lines, and the wrong paragraph.

Looking below, there's another section:

```
<<<<<<< HEAD
If a relationship is provided in the `"relationships"` member of a resource
object in a `PATCH` request, its value **MUST** be a relationship object
with a `"data"` member. The relationship's value will be replaced with the
value specified in this member.
=======
If a relationship is provided in the `relationships` section of a resource object in a
`PATCH` request, its value **MUST** be a relationship object with a `linkage` member.
The relationship's value will be replaced with the value specified in this member.
>>>>>>> Fix type.  Missed a couple of links->relationships
```

Same deal. So it looks like this was actually fixed, which means that this patch isn't actually needed. So sorry for not checking on that before I had you do this :( Hopefully this has been instructive anyway. If there had been some need for a change, we would have fixed these two places, then run this:

``` bash
$ git add format/index.md # tell git we're happy with these changes
$ git rebase --continue  # continue our rebase
$ git push origin -f # push our changes back up. Since we've rewritten history, we must force push
```

Does that make sense?

ndrochak:
Perfect sense.  Thanks for the tutorial.  I'm familiar w/ SVN, but not Git
yet, but that was perfectly clear.

Thanks,
Nick

Issue #624, opened 2015-05-17
More `include`-related vaguery


ethanresnick:
This issue may be dependent on the resolution to #497, but is distinct from it. The issue is the following line:

> If a link object refers to resource objects included in the same compound document, it **MUST** include resource linkage to those resource objects.

Imagine that the resource at `/companies/abc` has two relationships: `"owner"` and "`ceo"`, and that these relationships point to the same person resource. I'm concerned that, as written, the line above would require that, on a request for `/companies/abc?include=owner` would require that the link object for the _`"ceo"` relationship_ include linkage, not just a `"related"` or `"self"` url. That seems like an unintended consequence that the language should be clarified to remove.

tkellen:
I see what you mean. This seems like a bit of a reach, but you're technically correct. If there is a way to alleviate this concern without making this clause unnecessarily complex, I'd be +1 to a PR.

dgeb:
Yeah, this has bothered me as well. This definitely needs rewording or removal.

dgeb:
Closed via #637.

PR #637, opened 2015-05-20
Require "full linkage" in compound documents.

### LINKED ISSUE #497: Explain how "in-between" resources for the "include" parameter should be handled

dgeb:
Define full linkage in the "compound documents" section and expand upon
its implications in the "inclusion of related resources" section.

[Closes #497 and #624]

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -288,9 +288,6 @@ which **MUST** contain at least one of the following:
 A relationship object that represents a to-many relationship **MAY** also contain
 pagination links under the `"links"` member, as described below.
 
-If a relationship object refers to resource objects included in the same compound
-document, it **MUST** include resource linkage to those resource objects.
-
 Resource linkage **MUST** be represented as one of the following:
 
 * `null` for empty to-one relationships.
@@ -380,6 +377,16 @@ responses are called "compound documents".
 
 In a compound document, all included resources **MUST** be represented as an
 array of resource objects in a top-level `"included"` member.
+
+Compound documents require "full linkage", meaning that every included
+resource **MUST** be identified by at least one [resource identifier object]
+in the same document. These resource identifier objects could either be
+primary data or represent resource linkage contained within primary or
+included resources.
+
+> Note: Full linkage ensures that included resources are related to either
+the primary data (which could be resource objects or resource identifier
+objects) or to each other.
 
 A complete example document with multiple included relationships:
 
@@ -853,7 +860,7 @@ section of the compound document.
 The value of the `include` parameter **MUST** be a comma-separated (U+002C
 COMMA, ",") list of relationship paths. A relationship path is a dot-separated
 (U+002E FULL-STOP, ".") list of relationship names. Each relationship name
-**MUST** be identical to the key in the `links` section of its parent
+**MUST** be identical to the key in the `"relationships"` member of its parent
 resource object.
 
 > Note: For example, a relationship path could be `comments.author`, where
@@ -873,15 +880,22 @@ for each relationship name can be specified:
 GET /articles/1?include=comments.author
 ```
 
-> Note: A request for `comments.author` should not automatically also
-include `comments` in the response. This can happen if the client already
-has the `comments` locally, and now wants to fetch the associated authors
-without fetching the comments again.
+> Note: Because compound documents require full linkage, intermediate
+resources in a multi-part path must be returned along with the leaf nodes.
+For example, a response to a request for `comments.author` should
+automatically include `comments` as well as the `author` of each of those
+`comments`.
+
+> Note: A server may choose to expose a deeply nested relationship such as
+`comments.author` as a direct relationship with an alias such as
+`comment-authors`. By abstracting the underlying relationship, the server
+can still provide full linkage in compound documents without including
+potentially unwanted intermediate resources.
 
 Multiple related resources can be requested in a comma-separated list:
 
 ```http
-GET /articles/1?include=author,comments,comments.author
+GET /articles/1?include=author,comments.author
 ```
 
 ### Sparse Fieldsets <a href="#fetching-sparse-fieldsets" id="fetching-sparse-fieldsets" class="headerlink"></a>


tkellen:
:+1: Bummed we couldn't come up with a better solution for 1.0 but this does seem like the correct approach for now.

format/index.md
```diff
@@ -873,15 +880,22 @@ for each relationship name can be specified:
...
-include `comments` in the response. This can happen if the client already
-has the `comments` locally, and now wants to fetch the associated authors
-without fetching the comments again.
+> Note: Because compound documents require full linkage, intermediate
+resources in a multi-part path must be returned along with the leaf nodes.
+For example, a response to a request for `comments.author` should
+automatically include `comments` as well as the `author` of each of those
+`comments`.
+
+> Note: A server may choose to expose a deeply nested relationship such as
```
ethanresnick, line 889: Is this sentence consistent with the requirement above that "each relationship name **MUST** be identical to the key in the `"relationships"` member of its parent resource object"? That would seem to imply that, in a request for `?include=comments.author`, the relationship names are `"comments"` and `"author"`, but not `"comment-authors"`, so a `"comment-authors"` key would be non-compliant. Also, allowing a key like `"comment-authors"` at all creates an ugly read-write asymmetry.

I suppose you could say `?include=comment-authors` instead, and then you'd be in the clear. (Though the note should be updated to reflect that.)

dgeb, line 889: > I suppose you could say ?include=comment-authors instead, and then you'd be in the clear. 

That was the intention of the note, that the alias would be used instead of the full path. I should clarify it since it wasn't obvious to you.

ethanresnick:
Also, I know I'm jumping in late here...I didn't have a chance to look at #497 carefully until now. But let me just make a naive proposal that might balance @dgeb's tradeoff from #497 about full linkage vs. adhoc queries:

What if we imagine—just for a mental model; the final spec text needn't take this form—that every listed `include` path implicitly starts with "primary", and that `"primary"` is automatically a member of the `include`d paths list. So a request for `/articles?include=comments,comments.author` would implicitly have _three_ include paths: "primary", "primary.comments" and "primary.comments.author". Then, we'd define a rule that says: 

"For any path in the list of `include`d paths, if there is a second path in that list that matches the first path up until its last segment (i.e. up until its last period), then the relationship objects picked out by the second path _must_ have linkage at their `data` member."

By this I mean: the presence of _both_ `primary.comments.author` and `primary.comments` as include paths triggers a requirement that the `"authors"` relationship in each comment have linkage. This is triggered because `primary.comments` matches `primary.comments.author` up until its last segment.

I haven't thought about how to specify this requirement concisely, but it does seem like it would give us the best of all worlds. Namely:
1. It would mandate that, in requests like `/articles?include=comments`, the `comments` relationship always has linkage, because there would implicitly be two included paths (`"primary"` and `"primary.comments"`), and these match up until their last path segment, which requires the `"comments"` relationship to have linkage.
2. But, it allows that, in requests like `/articles?include=comments.author`, the author resources don't need to be connected.
3. Meanwhile, a client that _wants_ the linkage can 100% ensure that it gets it by simply requiring the intermediate resources explicitly with `/articles?include=comments,comments.author`.

Thoughts?

nevson:
@dgeb Awesome, thanks this helped me a lot.

@ethanresnick So 2.) `articles?include=comments.author` would return a compound document like the following where comments and author aren't connected

``` json
{
  "data": [{
    "type": "articles",
    "id": "1",
    "attributes": {
    },
    "links": {
    },
    "relationships": {
      "author": {
        "links": {
        },
        "data": { "type": "people", "id": "9" }
      },
      "comments": {
        "links": {
        },
        "data": [
        ]
      }
    }
  }],
  "included": [{
    "type": "people",
    "id": "9",
    "attributes": {
      "first-name": "Dan",
      "last-name": "Gebhardt",
      "twitter": "dgeb"
    },
    "links": {
    }
  }, {
    "type": "comments",
    "id": "5",
    "attributes": {
      "body": "First!"
    },
    "links": {
      "self": "http://example.com/comments/5"
    }
  }, {
    "type": "comments",
    "id": "12",
    "attributes": {
      "body": "I like XML better"
    },
    "links": {
      "self": "http://example.com/comments/12"
    }
  }]
}
```

and 3.) `articles?include=comments,comments.author` would then connect the comments with the author so that you know which author belongs to which comments? How would the response look like? In the example above this relationship is missing so far.

neomerx:
Have 'resource linkage' was replaced with 'resource identifier object'? If so this PR needs update.

dgeb:
> Have 'resource linkage' was replaced with 'resource identifier object'? If so this PR needs update.

Resource linkage is still a concept that pertains to relationships. It can be expressed with "resource identifier objects". See the relationships section for resource objects which explains valid expressions of resource linkage. 

And of course, let me know if I've missed anything or if it is confusing.

neomerx:
@dgeb thanks

dgeb:
FWIW the concept of "resource identifier objects" was introduced as a general construct to identify resources. There's nothing structurely different about the spec in this regard though. It just allows us to use this term in places outside of relationships.

neomerx:
@dgeb I think I got it. I'm just less terminology but more technology

format/index.md
dgeb, line 889: I have changed that note to read:

> Note: A server may choose to expose a deeply nested relationship such as
> `comments.author` as a direct relationship with an alias such as
> `comment-authors`. This would allow a client to request
> `/articles/1?include=comment-authors` instead of
> `/articles/1?include=comments.author`. By abstracting the nested
> relationship with an alias, the server can still provide full linkage in
> compound documents without including potentially unwanted intermediate
> resources.

dgeb:
@neomerx :)

dgeb:
@ethanresnick I think that the simplicity in the new requirement for "full linkage" comes about because it is related to _all_ compound documents, whether they are constructed at the client's request or the server's preference (when no `include` params are specified). You seem to be adding rules for linkage based client requests, which is only part of this problem. I think that without the full linkage requirement, it is very difficult to fully spec out what implementations need to support.

ethanresnick:
> You seem to be adding rules for linkage based client requests, which is only part of this problem.

Are you worried about servers that support default includes and yet _don't_ give the client the ability to provide an `include` parameter? (In which case we want to make sure the default includes that server picks don't seem broken.) In that case, couldn't we just extend my proposal to say: when a server is including resources by default and it doesn't support the `include` parameter, it must offer "full linkage" for that request.

Would that leave anything uncovered?

format/index.md
ethanresnick, line 889: Ok, +1. That text makes perfect sense now! (And it probably made sense before, I just misread it.)

dgeb:
@ethanresnick Your proposal is intriguing.

Another alternative that I discussed with @tkellen, possibly for > 1.0, is that servers MAY support a special case of `include` parameters that specify omitted intermediate linkages. For instance, `?include=!comments.authors` or `?include=-comments.authors`. This would allow the normal case to still specify full linkage easily.

But these proposals would all require exceptions to the full linkage requirement that's so clean in the compound documents section.

ethanresnick:
Oh interesting; that proposal you discussed with @tkellen is pretty intriguing too! Since neither of these options require changes to the document's structure, though, it seems like it'd be ok to work them out/whether to use them over the next couple days, now that we have some 1.0 breathing room. So I'm gonna sleep on it, and chime back in tomorrow!

dgeb:
> So I'm gonna sleep on it, and chime back in tomorrow!

Thanks in advance @ethanresnick. Enjoy your well deserved rest!

tkellen:
I'm +1 to considering a way to get this back if clients are clearly/obviously opting into a "broken" document. I made a few attempts at normative language for this and it was a mess. Please don't forget to consider nesting more than 2 levels deep. Excited to see what you come up with @ethanresnick!

ethanresnick:
@dgeb Thanks! I really am going to sleep now :)

tkellen:
Another thing I think we need is a way to link back to the primary document if intermediate records are skipped. For example, `?include=-comments.authors` should have some form where linkage directly to the authors appears in the primary data (but not as a dot-notated key because that is normatively disallowed).

bintoro:
> For example, `?include=-comments.authors` should have some form where linkage directly to the authors appears in the primary data (but not as a dot-notated key because that is normatively disallowed).

But isn't the dot notation disallowed precisely to reserve it for a particular purpose such as this one?

In cases where intermediate records are excluded because they are big, the linkage question could potentially be solved by putting the excluded records in `included` anyway but only delivering the relevant linkage information (no attributes etc.).

ksimka:
:+1: as I've been just searchng for the example of a partial include. Looks good now.

dgeb:
Let's continue the linkage discussion at #666. 

@ethanresnick - please let me know if you're still up for submitting a PR for this (assuming you agree with the details).

ethanresnick:
@dgeb I'll read through #666 and #670 in a few hours and chime in there! I'd be happy to write up a PR when we settle on something (or if we have already)

PR #638, opened 2015-05-20
add source key to error object for indicating source of problem

tkellen:


ethanresnick:
+1

lode:
The documentation is not clear to me about what the `source` object can contain. Can it **only** (optionally) contain `pointer` and `parameter`? Or are you free to use other keys there? And is it intended to use `parameter` only for query strings or are parameters in POST body's also okay?

hhware:
> Can it only (optionally) contain

@lode, section Document Structure says: "Unless otherwise noted, objects defined by this specification MUST NOT contain any additional members."

> only for query strings or are parameters in POST body's also okay?

IMHO, since it is written "query parameter", POST body cannot be included.

lode:
Thanks @hhware!

Issue #634, opened 2015-05-20
List in Recommendations/Naming incorrectly displayed


cziegenberg:
After the merge of #601 the list under "Naming" doesn't display correctly, see http://jsonapi.org/recommendations/

I wanted to correct it, but in the preview the list is there... Does anybody know how to fix this?

dgeb:
Looks like a line break is needed before the list.

cziegenberg:
Created an PR for it.

PR #640, opened 2015-05-20
Try to fix listing in naming section

cziegenberg:
Added a line break before the list, as suggested in #634


dgeb:
That did the trick - thanks @ziege!

PR #641, opened 2015-05-20
Clarify applicability of inclusion of related resources and sparse fieldsets.

### LINKED ISSUE #562: Including related resources in CRUD requests
Can we include related resources in a response to a CRUD operation?

Example: users can create posts, posts can be liked by users, like is a resource too. User A has post P, user B likes P creating a new like L. But creating a like causes an increment of "liked_num" field of a post. So in order to reduce the number of requests we want to return a new version of P with created L in a single response.

Should it looks like this?

`POST /likes?include=post`

``` js
{
    "data": {
        "id": 100,
        "type": "like",
        "user_a_id": 3,
        // ...
        "links": {
            "post": {
                "linkage": {"id": 50, "type": "post"}
            }
        }
    },
    "included": [
        {
            "id": 50,
            "type": "post",
            // ...
            "likes_num": 42 
        }
    ]
}
```

If yes, can someone please add an example of such response to the CRUD section of the spec?

dgeb:
Clarify that inclusion of related resources can apply to relationship
endpoints, and to request types other than GET.

Clarify that sparse fieldsets can apply to any endpoint that responds with
resources as primary or included data, regardless of the request type.

[Fixes #562]

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -995,6 +995,9 @@ parameter `sort`, the server **MUST** return elements of the top-level
 The server **MAY** apply default sorting rules to top-level `data` if
 request parameter `sort` is not specified.
 
+> Note: This section applies to any endpoint that responds with a resource
+collection as primary data, regardless of the request type.
+
 ### Pagination <a href="#fetching-pagination" id="fetching-pagination" class="headerlink"></a>
 
 A server **MAY** choose to limit the number of resources returned in a response
@@ -1032,6 +1035,9 @@ be used as a basis for any of these strategies. For example, a page-based
 strategy might use query parameters such as `page[number]` and `page[size]`,
 an offset-based strategy might use `page[offset]` and `page[limit]`, while a
 cursor-based strategy might use `page[cursor]`.
+
+> Note: This section applies to any endpoint that responds with a resource
+collection as primary data, regardless of the request type.
 
 ### Filtering <a href="#fetching-filtering" id="fetching-filtering" class="headerlink"></a>
 
--- a/format/index.md
+++ b/format/index.md
@@ -911,6 +911,21 @@ Multiple related resources can be requested in a comma-separated list:
 GET /articles/1?include=author,comments.author
 ```
 
+Furthermore, related resources can be requested from a relationship endpoint:
+
+```http
+GET /articles/1/relationships/comments?include=comments.author
+```
+
+In this case, the primary data would be a collection of [resource identifier
+objects] that represent linkage to comments for an article, while the full
+comments and comment authors would be returned as included data.
+
+> Note: This section applies to any endpoint that responds with primary
+data, regardless of the request type. For instance, a server could support
+the inclusion of related resources along with a `POST` request to create a
+resource or relationship.
+
 ### Sparse Fieldsets <a href="#fetching-sparse-fieldsets" id="fetching-sparse-fieldsets" class="headerlink"></a>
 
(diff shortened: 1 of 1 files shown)


tkellen:
I know it is pretty redundant, but shouldn't the clause for "This section applies to any endpoint..." belong in the sorting and pagination sections as well?

format/index.md
```diff
@@ -901,6 +901,21 @@ Multiple related resources can be requested in a comma-separated list:
 GET /articles/1?include=author,comments.author
 ```
 
+Furthermore, related resources can be requested from a relationship endpoint:
+
+```http
+GET /articles/1/relationships/comments?include=comments.author
```
tkellen, line 917: Wouldn't the include here be relative to the relationship (comments), so `?include=author`?

dgeb, line 917: There needs to be a way to include the comments too. `GET /articles/1/relationships/comments?include=comments` should return the relationship data as well as the related comments (in `included`).

dgeb:
> I know it is pretty redundant, but shouldn't the clause for "This section applies to any endpoint..." belong in the sorting and pagination sections as well?

Yeah, good catch. Those sections should probably specify "This section applies to any endpoint that returns a collection of resources ..."

format/index.md
tkellen, line 917: Gotcha. I haven't implemented support for this yet. Makes sense, I think.

dgeb:
@tkellen ok, clarified applicability for pagination and sorting via 49a8755fa1d9d180cb909fc6e165555a54d6e5a5

ethanresnick:
:+1:

tkellen:
:+1:

ethanresnick:
@dgeb I was thinking about this more just now... Does it really make sense for `include` to apply to relationship endpoints? Because `/articles/1/relationships/comments?include=comments` is identical functionally to `/articles/1/comments` (that is, including the primary linked resources is the same as using the related resource URL). And, for any nested relationships, like comments.author, doing `/articles/1/comments?include=author` seems to make more sense than doing `/articles/1/relationships/comments?include=comments.author`...it's much more consistent with the path naming for `include` on standard collections.

dgeb:
@ethanresnick I think the distinction between relationship data from related resource data is important. Relationship data can have a distinct order specified by the server, be paginated in a unique way, and each resource identifier object can contain metadata. I don't want to lose this distinction, yet I want to allow `include` to be used at relationship endpoints to minimize the number of requests.

ethanresnick:
>  I think the distinction between relationship data from related resource data is important. Relationship data can have a distinct order specified by the server, be paginated in a unique way, and each resource identifier object can contain metadata.

Couldn't all of these features be implemented at the related URL too, though? That is, if the relationship data at `/articles/1/relationships/comments` was sorted or paginated, the primary data at `/articles/1/comments` could be sorted and paginated in an analogous way. Likewise, any meaningful metadata that would be part of the resource identifier objects at the relationship URI could be added to the resource objects at the related URI.

I'm not _strongly_ opposed to continuing to allow `include` at relationship URIs and I'm going to write up a version of #580 now that keeps this capability. But it does seem like it's making our lives more complicated in terms of specifying `include` behavior and the full linkage requirement (as you'll see); that it's making the meaning of the `include` parameter's value less clear to users (especially if a URI isn't visually identifiable as a relationship URI); and that it could be added seamlessly post 1.0 if a compelling use case arises (i.e. requests to relationship urls with an include parameter would 400 now and 200 later).

PR #642, opened 2015-05-21
Clarify status codes for responses of all types.

### LINKED ISSUE #534: revisit status code MUSTs
Are these necessary? We may wish to do recommendations here, or use MUST for the status code range vs mandating specifics.

dgeb:
For all responses, clearly state that servers must respond according to HTTP semantics.

Allow `202 Accepted` responses as appropriate.

[Closes #534]

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -1144,15 +1144,12 @@ to create a resource with a client-generated ID.
 
 ##### 201 Created <a href="#crud-creating-responses-201" id="crud-creating-responses-201" class="headerlink"></a>
 
-A server **MUST** respond to a successful resource creation request according to
-[`HTTP semantics`](http://tools.ietf.org/html/rfc7231#section-6.3).
+If a `POST` request did not include a [Client-Generated
+ID](#crud-creating-client-ids) and the requested resource has been created
+successfully, the server **MUST** return a `201 Created` status code.
 
 The response **SHOULD** include a `Location` header identifying the location
 of the newly created resource.
-
-If a `POST` request did not include a [Client-Generated
-ID](#crud-creating-client-ids), and a resource has been created, the server
-**MUST** return a `201 Created` status code.
 
 The response **MUST** also include a document that contains the primary
 resource created.
--- a/format/index.md
+++ b/format/index.md
@@ -1187,9 +1187,10 @@ server **MUST** return a `202 Accepted` status code.
 ##### 204 No Content <a href="#crud-creating-responses-204" id="crud-creating-responses-204" class="headerlink"></a>
 
 If a `POST` request *did* include a [Client-Generated
-ID](#crud-creating-client-ids), the server **MUST** return either a `201
-Created` status code and response document (as described above) or a `204 No
-Content` status code with no response document.
+ID](#crud-creating-client-ids) and the requested resource has been created
+successfully, the server **MUST** return either a `201 Created` status code
+and response document (as described above) or a `204 No Content` status code
+with no response document.
 
 > Note: If a `204` response is received the client should consider the resource
 object sent in the request to be accepted by the server, as if the server
--- a/format/index.md
+++ b/format/index.md
@@ -727,9 +727,13 @@ fetch a single resource that does not exist, except when the request warrants a
 
 ##### Other Responses <a href="#fetching-resources-responses-other" id="fetching-resources-responses-other" class="headerlink"></a>
 
-Servers **MAY** use other HTTP error codes to represent errors. Clients
-**MUST** interpret those errors in accordance with HTTP semantics. Error
-details **MAY** also be returned, as discussed below.
+A server **MAY** respond with other HTTP status codes.
+
+A server **MAY** include [error details] with error responses.
+
+A server **MUST** prepare responses, and a client **MUST** interpret
+responses, in accordance with
+[`HTTP semantics`](http://tools.ietf.org/html/rfc7231).
 
 ### Fetching Relationships <a href="#fetching-relationships" id="fetching-relationships" class="headerlink"></a>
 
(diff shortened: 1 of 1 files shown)


tkellen:
I think we need to change all the status code MUSTS to SHOULDS or this could potentially be contradictory, no?

dgeb:
Obviously, the most important requirement is that servers obey HTTP semantics. However, does this mean that other response requirements be downgraded from MUSTs to SHOULDs? Or is it possible for the spec to walk this tightrope with two layers of MUSTs? It _seems_ possible if we can can guarantee that these MUSTs obey HTTP semantics.

@steveklabnik @wycats any opinions on this? I suspect @tkellen might be right that SHOULDs are appropriate.

ethanresnick:
I mean, there's always the option of text like: "**MUST** return a 200 OK or another code that complies with HTTP semantics." For 1.0, that might just be the best bandaid fix, since it would allow for  202 Accepted, 200 OK after PATCH, etc. And we could always refine that later. With this language in place, a PR for #608 seems like it would get much simpler, and would be something I could take on.

tkellen:
+1 to _MUST return a XXX or another code that complies with HTTP semantics..._ on all of these.  I think that covers the bases we're trying to cover.

dgeb:
"MUST return a XXX or another code that complies with HTTP semantics..." sounds a lot like SHOULD (which might be the right answer here).

ethanresnick:
@dgeb Yup, they're pretty similar. The longer form might technically limit further the valid codes, but a SHOULD would be fine.

bintoro:
> However, does this mean that other response requirements be downgraded from MUSTs to SHOULDs?

I'd say it's perfectly fine to continue to specify MUST requirements for specific scenarios.

"or another code that complies..." together with the main rule seems ~~entirely~~ equivalent to SHOULD. (EDIT: In fact, SHOULD can be seen as a stronger recommendation than a MUST that simply reiterates the main rule and mentions one status code as if it was just an example.)

ethanresnick:
@bintoro I say that the "**MUST** be Code X or HTTP compliant" form is stronger only because of the following. Imagine a request that warrants a `202 Accepted`. If we say it "**MUST** be 200 Ok or a compliant code", then we're fine. If we say it "**SHOULD be 200 Ok"`, then we've got a recommendation that the person will ignore and, once they do that, our spec leaves them totally unconstrained. So they could then return a `404` theoretically and they'd be compliant from our spec's POV (though not from HTTP). The "**MUST or __" form says explicitly if our constraint doesn't fit, you at least need to follow HTTP.

But all this is _really_ small semantic details, so it doesn't matter to me either way.

bintoro:
@ethanresnick I was under the assumption that the subsection-level main rules ("A server MUST respond to a request to [perform an operation] according to HTTP semantics") would remain in place, in which case the HTTP constraint would still apply, should the implementation choose not to follow the more specific SHOULD.

ethanresnick:
@bintoro If I'm reading right, that clause only exists under the creating resources section right now. In the Updating Resources section, for example, the only other responses allowed are for error codes, so [this issue](https://github.com/json-api/json-api/pull/529) would still apply. That said, the constraint you quote is pretty well-stated, so I'd be happy with **SHOULD**s and then having that constraint as a fallback, assuming its moved to be high enough that it applies to all sections. In fact, doing it that way is _much_ neater than adding "or...." to every single status code mention! So +1

bintoro:
> If I'm reading right, that clause only exists under the creating resources section right now.

But this very PR inserts the same clause at the top of every section dealing with responses, so I think we're good.

Of course, one might ask if all those repetitive clauses could be consolidated into one statement placed after the _Creating, Updating and Deleting Resources_ heading.

ethanresnick:
@bintoro D'oh. You're totally right (again); I forgot this was a PR and not an issue. I'm deleting my comments above just to save @dgeb et al some reading time in the morning, which I'm sure will be scarce. SHOULD it is.

dgeb:
Thank you all for your review. I have looked at this with fresh eyes and pushed a new set of changes.

I am reluctant to weaken the MUSTs for specific appropriate status codes. These are all in accordance with HTTP semantics, so we shouldn't need to weaken them. @wycats and I worked hard to eliminate the wiggle room provided by the SHOULDs in previous versions of the spec.

However, what is needed is some wiggle room in the "Other Responses" sections. I have reworded these sections to be more general and less error-specific:

> A server **MAY** respond with other HTTP status codes.
> 
> A server **MAY** include [error details] with error responses.
> 
> A server **MUST** prepare responses, and a client **MUST** interpret responses, in accordance with
> [`HTTP semantics`](http://tools.ietf.org/html/rfc7231).

These catch-all sections should allow for any other valid responses when the conditions mentioned for specific status codes do not apply.

ethanresnick:
@dgeb I really like this approach in general! However, I don't think these changes yet carve out the particular exception I brought up originally in #529. Can we add an allowance for 200 responses to address that?

dgeb:
> Can we add an allowance for 200 responses to address that?

@ethanresnick That would be consistent with creates and I would be fine with it.

format/index.md
```diff
@@ -1183,9 +1195,13 @@ recognize the source of the conflict.
 
 ##### Other Responses <a href="#crud-creating-responses-other" id="crud-creating-responses-other" class="headerlink"></a>
```
tkellen, line 1216: I think this header should be ####.

format/index.md
```diff
@@ -1558,16 +1590,26 @@ DELETE /photos/1
...
+return a `202 Accepted` status code.
+
 ##### 204 No Content <a href="#crud-deleting-responses-204" id="crud-deleting-responses-204" class="headerlink"></a>
 
-A server **MUST** return a `204 No Content` status code if a delete request is
-successful.
+A server **MUST** return a `204 No Content` status code if a deletion
+request is successful.
 
 ##### Other Responses <a href="#crud-deleting-responses-other" id="crud-deleting-responses-other" class="headerlink"></a>
```
tkellen, line 1630: Whoops, I think this header should be #### and the previous (duplicated) section should be removed.

ethanresnick:
@dgeb Ok! I've opened a PR here! https://github.com/json-api/json-api/pull/651 (That's just merging into your status-codes branch, though, not gh-pages.)

dgeb:
@ethanresnick - merged - thanks!

tkellen:
I'd like to see the "other responses" section stand on its own vs appearing multiple times, but I'm :+1: to merge if we want to deal with that at a later date. The normative text is good.

Issue #622, opened 2015-05-15
Links, relationships, discoverable actions, mapping to RFC5988


hhware:
This proposal is a successor to #590. I would like to request to consider this for 1.0 because it won't make much sense afterwards.

@ethanresnick has submitted #618 while I was thinking about this. Even though the content of #618 is a part of what is proposed here, the scope of this issue is different from #618, so I am not hijacking #618...
##### Motivation
1. Five HTTP verbs are not enough to describe all possible actions available for a resource in a complex API. The solution is to add custom endpoints. There should be a way to associate those custom endpoints with the primary endpoint of the resource.
2. It would be great if JSON API allowed to build HATEOAS-compatible APIs. This requires ability to dynamically determine which actions can be taken on a given resource and its relationships, as well as what are the starting points to explore the API without much prior knowledge about it. It is often important for the clients to save trips to the server, so such information should be returned as part of a response, not on a separate endpoint.
3. Bring consistency to top-level and resource-level `links` and to relationships. A lot has been said on this. One thing I am trying to solve here is that each relationship provides multiple links, whereas members of top-level `links` and resource-level `self` are singular links, so they are not similar at all, one is a actually a component of another. Another aspect of it is making all `links` have the same structure, which can also be considered as a step towards the concept of "`links` can appear anywhere" (https://github.com/json-api/json-api/pull/588#issuecomment-98261811).
4. It would be useful to allow implementations to provide various links relevant to the given resource or response object, whether "verbs" or "nouns", whether representing actions on the endpoints of resource/its relationships/custom endpoints, or different aspects/views/versions/etc of the given resource, or anything else related to the given resource. Currently, only a very limited set of links is allowed (relationships, pagination, self), and `meta` is the only option for those who need more. It is workable, but it will drive fragmentation in the emerging JSON API ecosystem.
5. Provide an implementation of [linking in JSON](https://www.mnot.net/blog/2011/11/25/linking_in_json): allow association of [RFC5988 Link Relations](https://www.iana.org/assignments/link-relations/link-relations.xml) to _all_ links. This must be optional in order not to affect adoption of the spec.
6. Make links extendable: they are currently defined as strings, which does not allow to add any customization/context to them, either in basic spec in the future or via extensions today.
7. While achieving all this, the subset of the spec which defines minimally required data structures should not become much more complicated than it currently is, so that adoption is not affected.
##### Proposed changes
[...]

Issue #613, opened 2015-05-12
related languages


relvao:
I need to identify related resources in different languages. I'm not sure how it would be done. Any ideas?

We consider that an article in 2 different locales is in fact 2 different documents that have a relationship. They can have different properties, eg. diferrent tags. 

Returning all the data in the same request wouldn't be an option. Furtermore we have several versions of each document, up to 100, as we have several versions in the same language depending on the country. 

Example URLs:
/en/GB/articles/my-test
/pt/PT/artigos/o-meu-teste

In HTML we have rel="alternate" hreflang="x". Google advices the use of the link header for non html pages - Link: http://es.example.com/; rel="alternate"; hreflang="es" but I would rather have that information in the response body. Do we have something comparable in the spec?

Thanks, 
Tiago

gr0uch:
According to HTTP semantics, the `Accept-Language` header should be used to negotiate the language of the response, and `Content-Language` should indicate the language of the response.
http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

relvao:
Hi @daliwali 

Thanks for the input. We are aware of those headers and content-language is properly applied in our API. Due to a combination of factors we chose to use the url to indicate the language being requested. 

The fact is we do have different URLs for localised versions of the page, as well as a lot of websites out there. Does anyone now how to best represent this using json-api?

TL;DR;
We chose to use a URL to identify the locale to make our API easier to consume.

From google:
"Keep the content for each language on separate URLs. Don’t use cookies to show translated versions of the page. Consider cross-linking each language version of a page. That way, a French user who lands on the German version of your page can get to the right language version with a single click.

Avoid automatic redirection based on the user’s perceived language. These redirections could prevent users (and search engines) from viewing all the versions of your site."

There is some information about the issue in these links:
- http://webmasters.stackexchange.com/questions/403/how-should-i-structure-my-urls-for-both-seo-and-localization
- https://support.google.com/webmasters/answer/182192?hl=en#2

ethanresnick:
If your articles are different documents with, for example, different tags, then they can simply be different resources in the json-api sense, with different ids and URLs, but all part of the same `"articles"` collection. So it sounds like the only issue is how to link an article to its counterparts in another language. My suggestion would be to just add an `"alternates"` key to ~~`"links"`~~ `"relationships"` for this purpose.

So I'd have something like:

```
GET /en/GB/articles/my-test
Accept: application/vnd.api+json

{
  "data": {
[...]

Issue #571, opened 2015-04-24
Recommendation for hierarchical types?


SFEley:
Today's version of the [Recommendations](http://jsonapi.org/recommendations/) page suggests an extremely limited set of characters for names:

> It is recommended that resource types, attribute names, and association names be "dasherized"; i.e. consist of only lower case alphanumeric characters and dashes (U+002D HYPHEN-MINUS, "-").

I can see the virtue of this for attribute names, where any other punctuation might collide with the multitude of ways of specifying paths (dot-nesting, URIs, JSON Path, et cetera).  But I'm having trouble with it when trying to map type names to their concrete representations in other languages — in particular, it gives no room for namespaces or inherited subtypes.

My real world example is this: we have a Ruby abstract class called **Segment** which represents a set of customers. By itself it makes sense, and if no specialization were needed I could call the JSON API type `"segments"` and be done with it.  But we have different _kinds_ of segments.  Some segments are broken out by demographics or purchase history.  Some of them represent lists in third-party service providers such as MailChimp. These usually have additional properties and behavior — classic inheritance, in other words.

For reasons that are debatable but out of scope, we're often using namespaces to indicate the class relationships.  So there's a **Segment::AgeRange** and a **Segment::NeverPurchased** and a **Segment::MailChimpList** and so forth.  It'd be swell if the JSON API type alone were sufficient to inform the client what kind of object to construct, and the fully qualified name is necessary.  (We have other kinds of MailChimpList in our third-party service wrapper; we're not just namespacing for fun.)

So the question is, what to call the types?  Off the top of my head I can think of a few options: 
1. Ignore the problem. Call the types `"segment-never-purchased"` and so forth, and rely on external knowledge to map that to the actual class. This presents some maintainability issues, to put it mildly.
2. Overload the hyphen, allowing multiple hyphens in a row to represent a namespace separator. This has a couple of problems:
   1. The name `"segment--never-purchased"` looks ugly to me. And while that isn't a technical issue, I'm the kind of programmer who gives out _The Pragmatic Programmer_ as gifts and believes in my gut that aesthetics matter. Stop caring about elegance and you'll stop noticing broken windows.
   2. The double-dash has other kinds of implicit meaning in some of our representations, e.g. Markdown documentation, where it often gets automatically converted to a long em `—`. We're using [Apiary](https://apiary.io) and [API Blueprint](https://apiblueprint.org), so inconsistent Markdown behavior _is_ a technical issue.
[...]

PR #644, opened 2015-05-21
Bare minimum use-case for links

hhware:
See https://github.com/json-api/json-api/issues/622#issuecomment-103734652

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -463,8 +463,8 @@ multiple times.
 
 As discussed above, a JSON API document **MAY** be extended to include
 meta-information as `"meta"` members in several locations: at the top-level,
-within resource objects, within relationship objects, and within [resource
-identifier objects].
+within resource objects, within relationship objects, within [resource
+identifier objects], and within link objects.
 
 All `"meta"` members **MUST** have an object as a value, the contents of which
 can be used for custom extensions.
@@ -506,6 +506,36 @@ the following members:
 * `"related"` - a related resource URL (as defined above) when the primary
   data represents a resource relationship.
 * Pagination links for the primary data (as described below).
+
+For links currently defined by the spec (`self`, `related`, `prev`, `next`, 
+`first`, `last`), the value of each member of a `links` object can be either 
+a string containing the link URL or a "link object", which can contain
+the following members:
+
+* `"href"` - the string containing the link URL,
+* `"meta"` - non-standard meta-information about the link.
+
+Examples of the supported formats:
+
+```
+"links": {
+  "self": "http://example.com/posts",
+}
+```
+```
+"links": {
+  "self": {
+    "href": "http://example.com/posts",
+    "meta": {}
+  }
+}
+```
+
+> Note: For links which can be defined or allowed by the spec 
+in the future, value of each member of a `links` object is not 
+constrained by the current version of the spec. It should be 
+assumed that anything can be allowed in the future for such 
+values: object, array, scalar.
 
 ### Member Names <a href="#document-structure-member-names" id="document-structure-member-names" class="headerlink"></a>
 
@@ -1008,10 +1038,10 @@ to a subset ("page") of the whole set available.
 A server **MAY** provide links to traverse a paginated data set ("pagination
 links").
 
-Pagination links **MUST** appear in the link object that corresponds to a
+Pagination links **MUST** appear in the links object that corresponds to a
 collection. To paginate the primary data, supply pagination links in the
 top-level `links` object. To paginate an included collection returned in
-a compound document, supply pagination links in the corresponding link
+a compound document, supply pagination links in the corresponding links
 object.
 
 The following keys **MUST** be used for pagination links:


format/index.md
```diff
@@ -513,6 +513,39 @@ the following members:
   data represents a resource relationship.
 * Pagination links for the primary data (as described below).
 
+Value of each member of a `links` object can be either a string
+containing the link URL or a "link object", which can contain
+the following members:
+
+* `"href"` - the string containing the link URL or an array containing multiple strings containing the link URLs,
```
dgeb: I'm not in favor of the array option here.

tkellen: I want the array option for the future but I don't think we can normatively map it to the current behavior without hamstringing ourselves later. I suspect we'll wind up using the `data` key if we need multiple entries in the future.

tkellen:
I'm going to update this so we can get it merged now.

format/index.md
ethanresnick: I wouldn't bring this up if I didn't feel very strongly about it, since I know I'm beating a dead horse. But the more I think about it, the more convinced I am that having an array of target URIs for the same relation is something we're _going_ to want post 1.0—it's fundamental—such that I _really, really_ think we should include the array form in 1.0 to make the transition easier. Let me give two examples of the array form's value from **real issues that have already been raised**, that I felt I had to answer inadequately because the array form doesn't exist.
1. The first is the issue of [communicating hierarchical types](https://github.com/json-api/json-api/issues/571). That is, imagine an API with hierarchical collections for `products` > `appliances` > `cooking appliances` > `microwaves`, and a client requests `GET /products`. One of the resource objects returned from `/products` is a microwave, so its resource object has `"type": "microwaves"`, allowing the client to easily read any microwave-specific fields or apply chair-specific presentation. But, for the client to check whether the resource is also an appliance or a cooking appliance (e.g. to offer the user filters in the UI), I [suggested](https://github.com/json-api/json-api/issues/571#issuecomment-97376061) relying on information in `meta`. However, since hierarchical types are relatively common, it'd be better if this could be exposed to generic clients. And a `collection` link that could hold an array of ([optionally named](https://github.com/json-api/json-api/issues/622#user-content-named-collections)) targets would be a _perfect_, standards compliant way to state those `is-a` relationships.
2. The second example is this request for [linking to alternate versions of a resource in different languages](https://github.com/json-api/json-api/issues/613). I suggested making an `alternates` relationship but, of course, that's just a poor-man's version of an `alternate` link relation with an array of targets!

If those aren't convincing, then I'll drop it forever. But I just wanted to make one last stab at this with some real world, already present issues.

tkellen: The only way I can see to support this with our current behavior would be to normatively state that the first index of an array version of `href` is the primary. That would allow us to retain the current behavior while allowing future expansion. If @dgeb is amenable to this, I will attempt to revise this PR later today.

ethanresnick: @tkellen What about the [first option discussed here](https://github.com/json-api/json-api/issues/622#issuecomment-103729996), where `href` isn't an array, but the value at the key in `links` can optionally be? (That's the option I was imagining anyway when I made my comment. An array in `href` works, I guess, but it's actually less flexible in that all the other keys in the object then have to apply to all the uris in `href`, as opposed to each URI getting its own set of options)

tkellen: That's fine. We'd still need to normatively state that the first item is the "primary" link.

ethanresnick: Ohhh, I think I get what you're saying. I.e., that we've only had one link to date, so if a client want's to know _the_ link to use, it should use the first one, right?

tkellen: Correct. If we allow them, we need to specify how they should be used in the context of the current spec. A generic client would need to know which one to fetch given that we have no semantics around array-type links currently.

ethanresnick: Oh. Another exhausted ah ha moment. I forgot that all the `links` keys/relations we currently have defined _only make sense_ with a single URI. I've been thinking so much about hierarchical types and extensions that I keep thinking that we already have `collection` and `profile` keys. Sorry!

Can we just say that each key in `links` may accept a single link (given as a URI or object + `href`) or an array of such links, but then we specify that `self`, `related`, etc. only take one link? That way, the array form doesn't apply to them, but clients know that arrays may be there in the future (and even in 1.0 for `profile`, if #650 gets merged). Then, for the keys where arrays do make sense, there doesn't need to be a "primary", which is good since those relations often don't have any inherent order.

tkellen: Okay, so, I think what we need to do is abort the concept of the entire links object being an array. We can say specific link _relations_ will support an array form filled with either strings or objects any time in the future, though.

Since we are only reserving `self` and `related` currently, and neither of those require an array, we don't need to speak normatively about this, and the window is still open for supporting arrays on new key later. Sound good @ethanresnick / @hhware ?

ethanresnick: > Since we are only reserving self and related currently, and neither of those require an array, we don't need to speak normatively about this, and the window is still open for supporting arrays on new key later. Sound good @ethanresnick / @hhware ?

Yeah, that works for me for 1.0! Thanks @tkellen, and sorry for my scatterbrainedness on this :)

hhware:
Off-topic: the amount of traffic these days is crazy, I have no time to even read all those posts. Is it necessary to have a release today? Not that my participation is important, but this is not a commercial project with a deadline, why does it have to be done this way?

dgeb:
@hhware Indeed. I'm used to being able to block out the world with a push to a deadline, but that is obviously not possible (or desirable) here. Traffic has indeed been off the charts and yet we need to be very careful with so much at stake. We will have an update before the EOD.

format/index.md
hhware: We are currently reserving `self`, `related` and 4 pagination links. I support the complete openness of the spec to the future in this regard, please review my updated PR.

format/index.md
```diff
@@ -463,8 +463,8 @@ multiple times.
 
 As discussed above, a JSON API document **MAY** be extended to include
 meta-information as `"meta"` members in several locations: at the top-level,
-within resource objects, within relationship objects, and within [resource
-identifier objects].
+within resource objects, within relationship objects, within [resource
+identifier objects], and within link object.
```
dgeb: s/link object/link objects/

hhware:
We are currently reserving `self`, `related` and 4 pagination links. I support the complete openness of the spec to the future in this regard, please review my updated PR.

format/index.md
```diff
@@ -507,6 +507,37 @@ the following members:
   data represents a resource relationship.
 * Pagination links for the primary data (as described below).
 
+For links currently defined by the spec (`self`, `related`, `prev`, `next`, 
+`first`, `last`), value of each member of a `links` object can be either
```
dgeb: s/value/the value/

format/index.md
```diff
@@ -507,6 +507,37 @@ the following members:
...
+```
+"links": {
+  "self": {
+    "href": "http://example.com/posts",
+    "meta": {}
+  }
+}
+```
+
+For links which can be defined or allowed by the spec in the future,
```
tkellen: I believe this should be with the non-normative note below.

hhware:
> I believe this should be with the non-normative note below.

I had a different feeling, but I guess it is not possible to formally define the future :) 

Accounted for all 3 comments, thanks!

dgeb:
Merged via 93e903028e81cc68ce20a4f61c57f45cd087fd43 (a rebase was needed).

PR #646, opened 2015-05-21
Reorganize Resource Object Section

ethanresnick:
More stylistic changes designed to simplify the Resource Object section, which was getting quite unwieldy. Specifically, I:
- Put the “Fields” section right after the Attributes and Relationships sections, where it clearly belongs, rather than after the Links section.
- Merged the “Attributes Object” and “Attributes” sections into one, for parallelism with the treatment of relationships, which doesn’t start with a special section describing the “relationships object”. This also makes the hierarchy simpler.
- Put the Resource Identification section above the Attributes section, to match the structure of the example payload and the opening of the “Resource Objects” section.
- But simplified the “Resource Identification” section and its hierarchy to make the whole block much more compact, so that readers don’t get bogged down with it now that it's higher up. Also, I removed some redundancies. For example, the text before said: “Every resource object is uniquely identified by the combination of its `"type"` and `"id"` members. A resource object's `"type"` and `"id"` pair **MUST** refer to a single, unique resource.”, but these two sentences both said exactly the same thing.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -210,8 +210,23 @@ Here's how an article (i.e. a resource of type "articles") might appear in a doc
 }
 // ...
 ```
+#### Resource Identification <a href="#document-structure-resource-identification" id="document-structure-resource-identification" class="headerlink"></a>
 
-#### Attributes Object <a href="#document-structure-resource-attributes-object" id="document-structure-resource-attributes-object" class="headerlink"></a>
+Every resource object **MUST** contain an `"id"` member and a `"type"` member.
+The value of each of these members **MUST** be a string.
+
+Each resource object's `"type"` and `"id"` pair **MUST** identify a single,
+unique resource.
+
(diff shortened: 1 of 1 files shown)


bintoro:
I like this. Combining the _Attributes Object_ and _Attributes_ sections is something I've wanted to do as well.

Can we please get this merged ASAP, so I can submit a PR for the complex attribute stuff that's now missing?

tkellen:
:+1: can you please review for intra-document links that may have been broken by this?

dgeb:
Let's fix any broken links in separate PRs / commits. Thanks @ethanresnick - I've been wanting to merge attributes into a single section myself.

ethanresnick:
Thanks @dgeb! Also, I just did a search for the ids of the sections I removed, and we appear to be in the clear links-wise :)

dgeb:
Great - thanks for checking @ethanresnick

PR #649, opened 2015-05-21
Mark extensions as WIPs.

### LINKED ISSUE #643: Mark official extensions as preliminary in advance of base spec reaching 1.0
The bulk and jsonpatch extensions need thorough review, which may not be possible before the base spec is tagged 1.0.

dgeb:
[Closes #643]


tkellen:
:+1:

Issue #614, opened 2015-05-12
Rethink extension negotiation


gr0uch:
I had a bad feeling about implementing the official extensions, or really any extensions at all to the spec, since extensions may completely break the base specification. 

In [a comment regarding content negotiation](https://github.com/jshttp/negotiator/issues/35#issuecomment-101079112), Roy Fielding writes:

> it is a waste of time to negotiate on the basis of obscure parameters while using a meaningless media type like vnd.api+json. Define separate media types (preferably with meaning) if you think the differences will matter to a client. You will note that the examples in the spec are artificial, since there isn't any real-world need to negotiate on parameter names.

"meaningless media type like vnd.api+json"... ouch ;)

steveklabnik:
Where did fielding make that comment? I didn't even know he was aware.

gr0uch:
I accidentally copied and pasted the wrong link when I posted it, should be fixed in the OP now.

steveklabnik:
ah interesting. I kinda agree with @royfielding here, though I obviously disagree with 'meaningless'

dgeb:
Taken from the same comment:

> It is assumed that a server capable of meaningful negotiation based on a media type parameter is also going to know what the parameter means (or provide a ranked configuration interface for the resource owner to specify how to handle them). Naturally, this might be media-type specific and is certainly server implementation-specific.

So this is the approach we are taking.

He goes on to state:

> In any case, commas are not allowed inside of parameter values unless the value is enclosed in DQUOTEs, so several of those examples are syntactically invalid.

He's correct about this and this is also [clearly stated in json-api's format section](http://jsonapi.org/format/#extending). Some of the examples in that issue _are_ invalid.

As for the "meaningless" comment: I suspect it was made because the media type is in the vendor tree and not the standards tree. It may have been made without even connecting `vnd.api+json` to this effort. Regardless, there is clearly meaning to this spec, even if it is not an official standard.

ethanresnick:
Yup, I've been worried this too. (I asked Roy on twitter about the other issue, which is what prompted his comment.) My tentative feeling now is that:
1. Extensions shouldn't be allowed to break compatibility with the base spec. If they need to do this, they're just their own media type. Calling something an "extension" when it's allowed to totally rewrite the rules is nonsensical.
[...]

PR #650, opened 2015-05-21
Extensions Overhaul

ethanresnick:
Ok guys. Here's the PR that I've been alluding to for #614. It's a lot of new stuff, but I think it's really important to get extensions right for 1.0. 

Our current system let's any extension do anything, which means we're giving up the opportunity to signal that whole classes of extensions are composable and to encourage people to design composable extensions. Making extensions composable is useful for generic clients, client and server libraries, and the prospect of extension reuse/developing an extension ecosystem. 

We're also allowing extensions that contradict the base spec (as opposed to just adding to it), which opens the door to a lot of confusion, and to the possibility that the media type will become pretty meaningless.

Finally, the current system has both the `ext`/`supported-ext` approach _and_ the profile option, which is just plain bizarre.

This PR is an attempt to unify and add rigor to the extension mechanisms to get the best of all worlds. The thinking behind most of these changes is already covered in depth in #614, but I'll try to summarize it here:
1. Dividing the extensions into profile extensions and comprehensive extensions comes from [this comment](https://github.com/json-api/json-api/issues/614#issuecomment-103717409). The idea is to guarantee composability where possible, while also having a more powerful type of extension that can make non-composable changes.
2. But restricting even the comprehensive extensions to honoring the base spec is aimed at making sure they're truly additive. I discussed this idea [here](https://github.com/json-api/json-api/issues/614#issuecomment-103726921).
3. The requirement that profile extensions used prefixed member names is based on the fact that we need _some_ way to make them forward compatible, and doing it this way: 1) [let's us](https://github.com/json-api/json-api/issues/614#issuecomment-103733776) define JSON-LD as a profile extension and 2) offers an approach that can be carried over to query parameters to consistently and visually delineate userland members from spec-defined ones.
4. The new advertising mechanism with options is based on the fact that `supported-ext` sucked in at least four ways:
   - It makes the protocol chattier than necessary, since the list of extensions has to be sent on every response, even after negotiation's done.
   - It's an abuse of the media type, since `supported-ext` doesn't actually tell you anything about the content. Meanwhile, `OPTIONS` is designed precisely for things like this.
   - The `supported-ext` list can't signal which _combinations_ of the listed extensions are supported. The OPTIONS format doesn't right now either (since I just wanted to get this in), but it could easily be extended to later.
   - The `supported-ext` header can't signal whether the listed extensions apply to other endpoints too (since not every extension will apply to every endpoint). But, again, the `OPTIONS` response could be extended in the future to specify which extensions apply beyond the current resource, to save future requests.

Note that I haven't written the last chunk of the negotiation section yet, but I just wanted to get this in asap so you guys could see it.

### DIFF
--- a/extensions/index.md
+++ b/extensions/index.md
@@ -6,15 +6,17 @@ show_sidebar: true
 
 JSON API can be extended in several ways:
 
-* The `supported-ext` and `ext` media type parameters can be used to negotiate
-  support for extensions,
-  [as discussed in the base specification](/format#extending).
-  Official and custom extensions to the specification are discussed below.
+* [Meta information](/format/#document-structure-meta) can be included throughout
+  request and response documents, to communicate extra data or enable additional
+  behaviors. Meta information is non-standard, and so should be used only for
+  customizations that only apply to a single API or a small set of APIs.
 
-* Meta information can be included in several places in a document,
-  [as discussed in the base specification](/format/#document-structure-meta).
+* API creators can also author [extensions](/format#extending) if they need to
+  make more radical changes to the base specification than is possible using
+  `meta`. Extensions also allow individuals to define a format for supporting
+  a new capability that can then be published and shared for use by others.
+  Two such published extensions are the official extensions listed below.
 
-* A profile can be specified in the top-level `meta` object, as discussed below.
 
 ## Official Extensions <a href="#official-extensions" id="official-extensions" class="headerlink"></a>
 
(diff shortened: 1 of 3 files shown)


dgeb:
@ethanresnick I don't mean to be discouraging but I should caution that this is probably too much for the spec to absorb at this late date.

cziegenberg:
I also think it's important to get some basic rules in to the 1.0 version to avoid serious problems with extensions in future versions.

I think the suggestion goes into the right direction, but I'd suggest to clearly define only one prefix character, preferably the @ (for added?), followed by the extension name and a hyphen minus.

Example: "@myext-myattribute"

Advantage:

You will never get conflicts when using multiple extensions at the same time, also when they use the "same" attribute name. 

This is what I tried to reach in my suggestion to add an extension member to wrap these attributes, but prefixing is also okay I think - and better when thinking about query parameters added by extesions: they should be prefixed the same way.

Only when multiple extensions use the same extension name, a conflict would be possible, but this is a general problem that could only be solved by a centralized extension registration.

ethanresnick:
@dgeb That's what I feared, but I also wasn't sure what the alternative was...

@ziege Thanks! By the way, the reason I allowed two prefixes is so that extension keys that aren't for JSON-LD (which uses the "@") don't conflict with keys that JSON-LD adds in the future. Basically, the idea (not made explicit for lack of time) was that JSON-LD would use @-prefixed keys and everything else would use the underscore.

hhware:
@ethanresnick, interesting concept! 

I do not quite understand, is there a real need in listing profiles in `links` in particular and in the body in general? What is the advantage of it compared to having all action related to extensions (whether comprehensive or profile ones) happening in headers? E.g., RFC6906 recommends to use `profile` media type parameter to specify profiles. Maybe I am just missing the point, but what is the justification for having this in both body and headers?

ethanresnick:
> @ethanresnick, interesting concept!

Thanks @hhware! I'm still fleshing the proposal out, and hopefully we'll all get to have a thorough discussion of it post 1.0. 

> I do not quite understand, is there a real need in listing profiles in links in particular and in the body in general? What is the advantage of it compared to having all action related to extensions (whether comprehensive or profile ones) happening in headers? 

The conventional reasoning given for including some seemingly-redundant links like the profiles (but also "self" etc) in the payload is that the payload data could become separated from the HTTP headers, e.g., if the data alone is stored in a client's cache for later use, or it's passed around between different consumers, etc. Admittedly, these are relatively rare cases, but its still considered best practice to include the `profiles` in the payload data to account for them. It makes the system somewhat more resilient. It also makes parsing simpler, since a parser can just operate on the payload data, without having to account for the HTTP conversation.

RFC 6906 mentions this best practice when it says:

>  Representations using a 'profile' media type parameter still SHOULD include that value in the representation using the 'profile' link relation, since the media type label of a representation can easily get lost when it is taken out of its conversational context.

As far as why put these in `links`, rather than somewhere else in the body: that goes back to my hope from our conversation on discoverable actions that the keys in `"links"` will eventually map to RFC 5988 relations, and so a `"profile"` link there in links be a standards-compliant way for the document to assert that it complies with the linked profile.

hhware:
> It makes the system somewhat more resilient. It also makes parsing simpler, since a parser can just operate on the payload data, without having to account for the HTTP conversation.

@ethanresnick, why is it done only for profile extensions then? IMHO, this consideration is equally (if not more) applicable to comprehensive extensions. Obviously, RFC does not talk about comprehensive extensions because it does not have them. Also, obviously, it is problematic to specify comprehensive extensions in the body because of chicken and egg problem. Perhaps it would be best to skip this "resiliency" consideration and make parser take into account the HTTP conversation?

> As far as why put these in `links`

Sorry, that part of my question did not have much sense, the answer was obvious...

ethanresnick:
> why is it done only for profile extensions then? IMHO, this consideration is equally (if not more) applicable to comprehensive extensions.

You're right: it probably should apply to comprehensive extensions too, and there's no reason we can't mandate that. It just means we'll have to say that at least one part of the payload, i.e. the part listing the comprehensive extensions, can't be changed by the extension. That seems reasonable to me.

The big question will be _how_ to specify the comprehensive extensions in the body. That will depend somewhat on the resolution to #684, but my inclination is that we could mint an extended relation that means "this is a comprehensive extension applying to the context document", and then list the comprehensive extensions in `"links"`, keyed with that custom relation, just like we do for `profile`s.

ethanresnick:
Closing in favor of #915.

PR #653, opened 2015-05-21
Explain complex attributes

bintoro:
Adds a paragraph that explicitly allows complex attributes. Note that no such term is formally defined anymore, since it's not really necessary and may just be confusing.


dgeb:
:+1:

tkellen:
:+1: thanks @bintoro

ethanresnick:
Thanks @bintoro !

PR #656, opened 2015-05-21
change references from standard to specification

tkellen:
JSON-API is not yet a standard--we should use the terminology
"specification" until it becomes one.


dgeb:
Agreed - this could certainly rub some people the wrong way.

PR #657, opened 2015-05-21
Fix 652: Meta alone as a valid top level member

lgebhardt:
Allow meta to be a lone top level member for all responses.
Allow 200 OK responses when meta is returned as a lone top level member.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -109,10 +109,13 @@ evolve through additive changes.
 A JSON object **MUST** be at the root of every JSON API response containing
 data. This object defines a document's "top level".
 
-A document **MUST** contain one of the following top-level members:
+A document **MUST** contain at least one of the following top-level members:
 
 * `"data"`, containing the document's "primary data"
 * `"errors"`, containing an array of [error objects](#errors)
+* `"meta"`, non-standard meta-information.
+
+The members `"data"` and `"errors"` may not coexist in the same document.
 
 The document's "primary data" is a representation of the resource or collection
 of resources targeted by a request.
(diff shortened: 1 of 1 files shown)


format/index.md
```diff
@@ -109,10 +109,13 @@ evolve through additive changes.
...
 data. This object defines a document's "top level".
 
-A document **MUST** contain one of the following top-level members:
+A document **MUST** contain at least one of the following top-level members:
 
 * `"data"`, containing the document's "primary data"
 * `"errors"`, containing an array of [error objects](#errors)
+* `"meta"`, non-standard meta-information.
+
+The members `"data"` and `"errors"` may not coexist in the same document.
```
dgeb: **MUST NOT** coexist

format/index.md
```diff
@@ -1361,10 +1363,14 @@ return a `202 Accepted` status code.
 
 If a server accepts an update but also changes the resource(s) in other ways
```
tkellen: ...in ways other than those specified..

dgeb:
LGTM :+1:

format/index.md
```diff
@@ -1361,10 +1363,14 @@ return a `202 Accepted` status code.
...
 than those specified by the request (for example, updating the `updated-at`
-attribute or a computed `sha`), it **MUST** return a `200 OK` response.
+attribute or a computed `sha`), it **MUST** return a `200 OK` response. The
+response document **MUST** include a representation of the updated resource(s)
+as if a `GET` request was made to the request URL.
 
-The response document for a `200 OK` **MUST** include a representation of
-the updated resource(s) as if a `GET` request was made to the request URL.
+A server **MUST** return a `200 OK` status code if an update is successful,
```
tkellen, line 1370: I would remove this section and add a special case in 204 stating that if meta is included, 200 must be used in lieu of 204.

tkellen:
So, this is normatively fine, which probably is what matters most today, but I think this could generally be simplified by using more or less what we had previously and just noting in the 204 sections that 200 should be used if `meta` is present in what would be an otherwise empty response.

PR #658, opened 2015-05-21
reformat list of places meta can appear

csantero:
`meta` can appear in error objects now, so this section needed updating. I felt that with the extra item it made more sense to format the list of places `meta` can appear as a bulleted list.


dgeb:
Agreed - thanks @csantero!

tkellen:
fastest pr merge evar

PR #660, opened 2015-05-21
Remove extension mechanism from base spec.

dgeb:
Instead of rushing the extension mechanism into the 1.0 release, we are
marking it as experimental and removing it from the base spec.

This allows us to solidify the extension concepts for a future release.

### DIFF
--- a/extensions/index.md
+++ b/extensions/index.md
@@ -4,17 +4,83 @@ title: Extensions
 show_sidebar: true
 ---
 
+## Status
+
+**Extensions are an experimental feature** and should be considered a work
+in progress. There is no official support for extensions in the base JSON
+API specification.
+
+## Extending JSON API <a href="#extending-json-api" id="extending-json-api" class="headerlink"></a>
+
 JSON API can be extended in several ways:
 
-* The `supported-ext` and `ext` media type parameters can be used to negotiate
-  support for extensions,
-  [as discussed in the base specification](/format#extending).
-  Official and custom extensions to the specification are discussed below.
+* [Official](#official-extensions) and [custom](#custom-extensions) extensions
+  are in development. The `supported-ext` and `ext` media type parameters can
+  be used to [negotiate support for extensions](#extension-negotiation).
 
 * Meta information can be included in several places in a document,
   [as discussed in the base specification](/format/#document-structure-meta).
 
-* A profile can be specified in the top-level `meta` object, as discussed below.
+* A profile can be specified in the top-level `"meta"` object, as discussed
+  below.
+
+## Extension Negotiation <a href="#extension-negotiation" id="extension-negotiation" class="headerlink"></a>
+
+The JSON API specification **MAY** be extended to support additional
+capabilities.
+
+An extension **MAY** make changes to and deviate from the requirements of the
+base specification apart from this section, which remains binding.
+
+Servers that support one or more extensions to JSON API **MUST** return
+those extensions in every response in the `supported-ext` media type
+parameter of the `Content-Type` header. The value of the `supported-ext`
+parameter **MUST** be a comma-separated (U+002C COMMA, ",") list of
+extension names.
+
+For example: a response that includes the header `Content-Type:
+application/vnd.api+json; supported-ext="bulk,jsonpatch"` indicates that the
+server supports both the "bulk" and "jsonpatch" extensions.
+
+If an extension is used to form a particular request or response document,
+then it **MUST** be specified by including its name in the `ext` media type
+parameter with the `Content-Type` header. The value of the `ext` media type
+parameter **MUST** be formatted as a comma-separated (U+002C COMMA, ",")
+list of extension names and **MUST** be limited to a subset of the
+extensions supported by the server, which are listed in `supported-ext`
+of every response.
+
+For example: a response that includes the header `Content-Type:
+application/vnd.api+json; ext="ext1,ext2"; supported-ext="ext1,ext2,ext3"`
+indicates that the response document is formatted according to the
+extensions "ext1" and "ext2". Another example: a request that includes
+the header `Content-Type: application/vnd.api+json; ext="ext1,ext2"`
+indicates that the request document is formatted according to the
+extensions "ext1" and "ext2".
+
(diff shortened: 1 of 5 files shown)


bintoro:
Okay, so will extensions use a completely different media type eventually, or how does this work in terms of 1.0 implementations and their lack of awareness of the `ext` parameter?

dgeb:
1.0 client implementations won't be aware of the `ext` parameter or any media type parameters.

We should probably explicitly declare that media type parameters must be ignored. I was debating where to place such a statement.

dgeb:
I am working on wording in the base spec regarding the `ext` parameter to ensure that it's future compatible. Will update soon...

dgeb:
I've introduced a media type negotiation section in the base spec - please review carefully 85d83863ed55da8c1e97981664591e5c6334aec8

extensions/index.md
```diff
@@ -4,17 +4,83 @@ title: Extensions
...
+## Extending JSON API <a href="#extending-json-api" id="extending-json-api" class="headerlink"></a>
+
 JSON API can be extended in several ways:
 
-* The `supported-ext` and `ext` media type parameters can be used to negotiate
-  support for extensions,
-  [as discussed in the base specification](/format#extending).
-  Official and custom extensions to the specification are discussed below.
+* [Official](#official-extensions) and [custom](#custom-extensions) extensions
+  are in development. The `supported-ext` and `ext` media type parameters can
```
ethanresnick, line 18: Can we start the sentence that says "The `supported-ext` and `ext`..." with the clause "Currently, ", to suggest that the parameters may change?

format/index.md
```diff
@@ -26,64 +26,24 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+Clients **MUST** send all JSON API data in request documents with
+the header `Content-Type: application/vnd.api+json`.
+
+Clients **MUST** ignore all parameters for the `application/vnd.api+json`
+media type received in the `Content-Type` header of response documents.
+
+Servers **MUST** send all JSON API data in response documents with
+the header `Content-Type: application/vnd.api+json`.
+
+Servers **MUST** return a `406 Not Acceptable` status code if the
```
ethanresnick, line 40: I'd rewrite this fourth paragraph to be something like:

> Methods for extending this specification will be added at a later date. Servers that implement this specification before those extension mechanisms are defined, or that don't implement those mechanisms once they are defined, **MUST** respond with `400 Bad Request` to any request that includes a media type parameter in the `Content-Type` header. These servers also **MUST** respond with a `400 Bad Request` for any request that, if it lists the JSON API media type in its `Accept` header, only lists parameterized forms of it.

The key thing going on here is that we're making requests with media type parameters explicit `400` errors, rather than letting the server ignore those parameters or respond with `406` or `415`. Requiring an explicit error will make it easier, when we add extensions in the future, for clients to distinguish between _extensions not being supported at all_ vs. _a particular extension not being supported_. 

The other thing I tried to do was clarify that the `Accept` header can take multiple values, e.g. `Accept: application/vnd.api+json;ext=bulk, application/vnd.api+json`, and that the header is only problematic if **every** JSON-API option it lists contains parameters. If only some of the options contain parameters (as in the prior example), then the requests should succeed—and, in fact, future clients may want to format their `Accept` header like this to work with older servers.

(Additionally, we could add that, in the 400 response, the server **SHOULD** (must?) return an error object with a particular name and/or code.)

ethanresnick:
@dgeb Thanks for punting on extensions for now, so that we can both get the mechanism right and still get 1.0 out tonight! That seems like the right call. 

A couple other things:
1. Any chance that punting on extensions means we have the bandwidth to further consider the attempted-"best of all worlds" proposal I made for `include` rules? I [responded](https://github.com/json-api/json-api/pull/637#issuecomment-104102464) to your last comment earlier. If you agree with the solution in concept, then I can work up the PR!
2. Once we tag 1.0 tonight, what are your thoughts on having a delay of a week or so before doing any big 1.0 promo campaigns? That would give client libraries, which I suspect are almost all be broken now, given the rapid changes over the last two days, a chance to catch up.

dgeb:
> Any chance that punting on extensions means we have the bandwidth to further consider the attempted-"best of all worlds" proposal I made for include rules?

[Commented](https://github.com/json-api/json-api/pull/637#issuecomment-104457219)

> what are your thoughts on having a delay of a week or so before doing any big 1.0 promo campaigns?

Lines up well with #661 :)

tkellen:
Nicely done @dgeb! I'm really glad we've decided to take this approach. :+1: from me.

format/index.md
dgeb, line 40: @ethanresnick I'd welcome some refinements to this language in the next week in a separate focused PR.

ethanresnick:
Thanks @dgeb! Also, any comments on https://github.com/json-api/json-api/pull/660#discussion_r30853844?

dgeb:
@ethanresnick see https://github.com/json-api/json-api/pull/660#discussion_r30863742

We now have a little time to ensure the correctness and readability of this section.

bintoro:
The new section on media type negotiation distinguishes between `ext` and other parameters. Is this mainly to allow proprietary parameters for versioning and such?

dgeb:
The intent is to reserve the `ext` parameter as a way to request an extension. 1.0 servers won't support extensions and therefore need to reject all such requests (which may come from 1.1 clients that use `ext`).

bintoro:
@dgeb Sorry, I phrased that poorly. What I mean is, why not put a blanket ban on all parameters? I'm assuming this is because we want to allow APIs to use the same header for their own stuff, too. This is fine, but it also implies that JSON API will never use anything besides `ext`.

dgeb:
> This is fine, but it also implies that JSON API will never use anything besides ext.

@bintoro Given the current requirements on the server, I think this could more accurately be said "JSON API will never use another media type parameter besides `ext` to negotiate an extension".

Since this PR is now merged, please open an issue to discuss this further if you like. I'm not sure it's a major problem, but we should be clear-eyed about the implications and alternatives.

PR #661, opened 2015-05-21
JSON API reaches final review phase.

dgeb:
Due to the massive number of changes to the specification in the past
two weeks, we are providing an extra week for final review before
tagging 1.0.

This release candidate, technically RC4, exists solely to provide time
for the specification to be carefully reviewed for inconsistencies and
defects. Do not expect any changes to the specified document structure.

Thanks to the supportive (and prolific) group of contributors who have
helped us reach this milestone. Let's ensure that the final spec is
as good as can be!

---

_This should be merged by the end of today._

### DIFF
--- a/_includes/status.md
+++ b/_includes/status.md
@@ -1,4 +1,5 @@
 ## Status <a href="#status" id="status" class="headerlink"></a>
 
-**This document is a work in progress** and will change as implementation work
-progresses. See the [Status](/status) page for more information.
+**This document is a work in progress** and is in a final review phase
+before the specification reaches 1.0 on May 28, 2015. See the
+[Status](/status) page for more information.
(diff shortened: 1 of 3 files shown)


ethanresnick:
Yay! I know 1.0’s taken forever, but its also astonishing how much more rigorous the design has become lately. Let's remember that, less than four months ago, we were still using PUT; we had the option of responses without a `data` key and resource objects without a `type`; we recommended URIs that used comma-separated ids; the equivalent of the related resource url typically pointed to the resources linked at the time the client received the URI, rather than at the time it requested it; we didn't even have a first draft of extension negotiation; and we had no attributes, relationships, or resource identifier objects. Great work everyone!

hhware:
Yeah, a couple more release candidates, and we are there! :)

steveklabnik:
:confetti_ball:

tkellen:
:shipit:

rilian:
:+1:

PR #663, opened 2015-05-22
Optimize allowed characters for version 1.0

cziegenberg:
As the extensions have been removed from version 1.0 of the spec and it's not clear how they will be realized, I suggest to also reserve the two currently allowed ASCII accent characters for later use.

They don't seem to be important (as these characters are normally used only in combination with other characters and then would form a UNICODE character, which is still allowed), and it will always be possible to allow them in a future version (but not to disallow them later).


dgeb:
I am fine reserving these :+1:

tkellen:
Same!

PR #665, opened 2015-05-22
Add YAYSON to the list of implementations

jede:
Hi!

First of all, thanks for all the hard work towards 1.0!

We made an isomorphic JS library for reading and serializing JSON API data called [YAYSON](https://github.com/confetti/yayson). Would love to add it to the list of implementations! We use it in production at [Confetti.events](http://confetti.events). Right now it supports the basics of RC4.


dgeb:
YAY! :smile:

jede:
Hooray! :confetti_ball:

PR #667, opened 2015-05-22
Added example for "Sparse Fieldsets"

ksimka:
Example demonstrating how `fields` param works.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -955,8 +955,84 @@ COMMA, ",") list that refers to the name(s) of the fields to be returned.
 If a client requests a restricted set of [fields], an endpoint **MUST NOT**
 include additional [fields] in the response.
 
+Basic request:
+
+```http
+GET /articles?include=author
+```
+
+```text
+HTTP/1.1 200 OK
+Content-Type: application/vnd.api+json
+
+{
+  "data": {
+    "type": "articles",
+    "id": "1",
+    "attributes": {
+      "title": "JSON API paints my bikeshed!",
+      "body": "The shortest article. Ever.",
+      "created": 1432306588,
+      "updated": 1432306589
+    },
+    "relationships": {
+      "author": {
+        "linkage": {
+          "id": 42, "type": "people"
+        }
+      }
+    }
+  },
+  "included": [
+    {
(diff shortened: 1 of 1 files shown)


dgeb:
Thanks for this, @ksimka!

We may ask you to retarget this to a new top-level examples page we're planning in #503.

tkellen:
@ksimka We've just added the page @dgeb mentioned--would you mind refactoring?

ksimka:
Yes, but I'm not sure that I understand #670 correctly. The response in the example depends on it. Could you please clarify if we've already resolved the confusion between `include` and `fields`?

tkellen:
@dgeb is about to submit a PR that puts this issue to bed. Let's revisit after?

ksimka:
> Let's revisit after?

Ok, I'll refactor right after #670 is closed.

tkellen:
Just closed via #688.

ksimka:
Sorry, I feel stupid, but still not sure if I get it right )

If `include=a.b.c&fields[A]=d,e` where `a` consists of `A`, `b` — `B` and `c` — `C`, so `A` will not have linkage to objects `B` (that will be in `included`), and we can't link them properly. Or does that mean to not include `b` at all? Then `C` objects can't be linked to anything, they're useless. Or should server in this case provide straight linkage from `a` to `c`? Or am I missing something obviuos again?)

format/index.md
```diff
@@ -955,10 +955,86 @@ COMMA, ",") list that refers to the name(s) of the fields to be returned.
...
+      }
+    }
+  ]
+}
+```
+
+Request with `fields` parameter:
+
 ```http
 GET /articles?include=author&fields[articles]=title,body&fields[people]=name
```
dgeb, line 1003: Relationships must also be included in `fields`, even if they're specified via `include`. Therefore, the `author` field would have to be included in `fields[articles]` to get the response below.

In other words, the request would need to be:

``` http
GET /articles?include=author&fields[articles]=title,body,author&fields[people]=name
```

ksimka:
Thanks, @dgeb. Closing in favour of #691.

PR #672, opened 2015-05-22
Typo fix

ethanresnick:
Obviously, a relationship's name is its key in the _relationship__s_*\* object, not the _relationship_ object.


dgeb:
Good catch @ethanresnick!

PR #673, opened 2015-05-22
Media type negotiation cleanup

ethanresnick:
This does a few things:
1. It doesn't let the server silently ignore media type parameters other than `ext`. Instead, it requires the server to explicitly return an error. We could add all sorts of parameters down the line, and post 1.0–clients that use these parameters may need to know whether they've been honored or not; silent ignoring could be quite unacceptable. By contrast, if we add any parameters down the line that can be simply ignored by old servers, then the post 1.0–client that gets the explicit error can simply retry its request without the parameter. (Of course, none of this effects 1.0 clients, since they're not allowed to include any parameters And, if they do, it's probably a bug, so an explicit error is more helpful for them too.)
2. It applies the same treatment to `ext` as it does to every other media type parameter—namely, it requires the server to send the error mentioned in (1). This is better than using 406/415 here now, which will make it harder, when we add extensions in the future, for clients to distinguish between extensions not being supported at all vs. a particular extension not being supported. Also, there's no guarantee that we'll stick only to the `ext` parameter for required extension negotiation. (My proposal in #650, for example, will, when I finish the last section, have an `ext` and a `required-profile` parameter.)
3. It accounts for the fact that the Accept header can take multiple values, e.g. `Accept: application/vnd.api+json;ext=bulk, application/vnd.api+json`, and that the header is only problematic if _every_ JSON-API option it lists contains parameters. If only some of the options contain parameters (as in the prior example), then the requests should succeed—and, in fact, future clients may want to format their `Accept` header like this to work with older servers.
4. **Update** It also moves the remaining stray note about content negotiation that was in the "Fetching Data" section into this single top section. And it renames the section to "Content Negotiation", which is what everyone else calls this process, so we should too. 
5. **Update**: Finally, it reworks the requirements that were in "Fetching Data" to be more consistent with my understanding of our intentions. See my inline comments

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -28,26 +28,31 @@ interpreted as described in RFC 2119
 
 ## Content Negotiation <a href="#content-negotiation" id="content-negotiation" class="headerlink"></a>
 
-Clients **MUST** send all JSON API data in request documents with
-the header `Content-Type: application/vnd.api+json`.
+Clients **MUST** send all JSON API data in request documents with the header
+`Content-Type: application/vnd.api+json`, without any media type parameters.
 
-Clients **MUST** ignore all parameters for the `application/vnd.api+json`
+If a client violates this requirement, servers **MUST** respond with a
+`415 Unsupported Media Type` status code.
+
+Clients that include the JSON API media type in their `Accept` header, **MUST**
+specify the media type there at least once without any media type parameters.
+
+If a client violates this requirement, and its `Accept` header doesn't contain
+any other media types that the server can produce, the server **MUST** respond
+with a `406 Not Acceptable` status code.
+
+Servers **MUST** send all JSON API data in response documents with the header
+`Content-Type: application/vnd.api+json`, without any media type parameters.
+
+Servers **MUST NOT** send responses with the `application/vnd.api+json` media
+type if the client has not indicated, using its `Accept` header, that it can
+understand this format.
+
+Clients **MUST** ignore any parameters for the `application/vnd.api+json`
 media type received in the `Content-Type` header of response documents.
 
-Servers **MUST** send all JSON API data in response documents with
-the header `Content-Type: application/vnd.api+json`.
-
-Servers **MUST** return a `406 Not Acceptable` status code if the
-`application/vnd.api+json` media type is modified by the `ext` parameter in
-the `Accept` header of a request. Otherwise, servers **MUST** return a `415
-Unsupported Media Type` status code if the `application/vnd.api+json` media
-type is modified by the `ext` parameter in the `Content-Type` header of a
-request. Servers **MUST** ignore all other parameters for the
-`application/vnd.api+json` media type in `Accept` and `Content-Type`
-headers.
-
-> Note: These requirements may allow future versions of this specification
-to support an extension mechanism based upon the `ext` media type parameter.
+> Note: These requirements exist to allow future versions of this specification
+to use media type parameters for extension negotiation and versioning.
 
 ## Document Structure <a href="#document-structure" id="document-structure" class="headerlink"></a>
 
@@ -592,16 +597,6 @@ The following characters **MUST NOT** be used in member names:
 Data, including resources and relationships, can be fetched by sending a
 `GET` request to an endpoint.
 
-Clients **MUST** indicate that they can accept the JSON API media type, per
-the semantics of the HTTP `Accept` header. If present, this header value **MUST**
-also include any media type extensions relevant to the request. Servers **MUST**
-return a `406 Not Acceptable` status code if this header specifies an
-unsupported media type.
-
-> Note: Servers may support multiple media types at any endpoint. For example,
-a server may choose to support `text/html` in order to simplify viewing content
-via a web browser.
-
 Responses can be further refined with the optional features described below.
 
 ### Fetching Resources <a href="#fetching-resources" id="fetching-resources" class="headerlink"></a>
@@ -1070,10 +1065,6 @@ strategies.
 
 A server **MAY** allow resources of a given type to be created. It **MAY**
 also allow existing resources to be modified or deleted.
-
-Any requests that contain content **MUST** include a `Content-Type` header
-whose value is `application/vnd.api+json`. This header value **MUST** also
-include media type extensions relevant to the request.
 
 A request **MUST** completely succeed or fail (in a single "transaction"). No
 partial updates are allowed.
--- a/format/index.md
+++ b/format/index.md
@@ -48,6 +48,13 @@ Servers **MUST NOT** send responses with the `application/vnd.api+json` media
 type if the client has not indicated, using its `Accept` header, that it can
 understand this format.
 
+> Note: If a client's `Accept` header indicates that it accepts standard JSON
+but not the JSON API media type, the server may choose serve it a JSON API
+document identified as `application/json`. If the server does so, is not
+considered to be sending "JSON API data" for the purposes of the requirement
+above, but rather simple JSON data, and the client cannot rely on any of the
+rules laid out in this specification.
+
 Clients **MUST** ignore any parameters for the `application/vnd.api+json`
 media type received in the `Content-Type` header of response documents.
 
--- a/format/index.md
+++ b/format/index.md
@@ -26,7 +26,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 interpreted as described in RFC 2119
 [[RFC2119](http://tools.ietf.org/html/rfc2119)].
 
-## Media Type Negotiation <a href="#media-type-negotiation" id="media-type-negotiation" class="headerlink"></a>
+## Content Negotiation <a href="#content-negotiation" id="content-negotiation" class="headerlink"></a>
 
 Clients **MUST** send all JSON API data in request documents with
 the header `Content-Type: application/vnd.api+json`.


format/index.md
```diff
@@ -34,20 +34,21 @@ the header `Content-Type: application/vnd.api+json`.
 Clients **MUST** ignore all parameters for the `application/vnd.api+json`
 media type received in the `Content-Type` header of response documents.
```
bintoro, line 36: I'm thinking this paragraph could perhaps be deleted.

format/index.md
```diff
@@ -34,20 +34,21 @@ the header `Content-Type: application/vnd.api+json`.
...
 
+Clients that include the JSON API media type in their `Accept` header, **MUST**
+specify the media type there at least once without any media type parameters.
+
 Servers **MUST** send all JSON API data in response documents with
 the header `Content-Type: application/vnd.api+json`.
```
bintoro, line 46: I would add "without any media type parameters" here as well.

format/index.md
```diff
@@ -34,20 +34,21 @@ the header `Content-Type: application/vnd.api+json`.
...
+either of the requirements above. The server's response to such violations
+**MUST** contain a JSON API document with an [error object](#error-objects) and
+the `"href"` member of that object **MUST** be
+`"http://jsonapi.org/errors/invalid-media-type-parameter"`.
 
-> Note: These requirements may allow future versions of this specification
-to support an extension mechanism based upon the `ext` media type parameter.
+> Note: These requirements may be loosened in the future to allow subsequent
+versions of this specification to support an extension mechanism in which
+media type parameters are used for extension negotiation.
```
bintoro: Suggestion: "These requirements exist to allow subsequent versions of this specification to use media type parameters for extension negotiation and versioning."

format/index.md
ethanresnick, line 36: I thought that too for a while. But I decided to leave it because we might, in the future, want to let the server to automatically opt the client in to some parameters (e.g. `profile`, which just adds _extra_ semantics) and, if we do that, we don't want old clients to blow up.

format/index.md
ethanresnick: Much better! Updating now.

format/index.md
```diff
@@ -593,10 +594,9 @@ Data, including resources and relationships, can be fetched by sending a
 `GET` request to an endpoint.
 
 Clients **MUST** indicate that they can accept the JSON API media type, per
-the semantics of the HTTP `Accept` header. If present, this header value **MUST**
-also include any media type extensions relevant to the request. Servers **MUST**
-return a `406 Not Acceptable` status code if this header specifies an
-unsupported media type.
+the semantics of the HTTP `Accept` header. Servers **MUST** return a
+`406 Not Acceptable` status code if this header specifies only unsupported
+media types.
```
bintoro: Perhaps it would now be better to cover this in the _Media Type Negotiation_ section because it's part of the same logic:
- If the media type is wrong, return 406 Not Acceptable.
- If the media type is right but there's a parameter, return 400 Bad Request.

One might also ask if the rules governing Content-Type should work similarly:
- If the media type is wrong, return 415 Unsupported Media Type.
- If the media type is right but there's a parameter, return 400 Bad Request.

format/index.md
ethanresnick, line 46: Ok, I added this both here and in the first paragraph. I'm not 100% sure it's necessarily, but it certainly can't hurt.

format/index.md
ethanresnick: I'm indifferent; up to @dgeb et al.

format/index.md
bintoro, line 36: Yeah, obviously clients are not permitted to blow up in the face of unexpected parameters. I just thought that would be... well, obvious. Maybe it's indeed better to leave this intact just in case.

However, right now the paragraph is right between the two requirements to which "either of the requirements above" is intended to refer, so you may want to consider moving it to the very end or explicitly mentioning "request requirements above" or something like that.

ethanresnick, line 36: Ok, I reordered the paragraphs a bit. Let me know what you think, and thanks always for your close readings!

bintoro:
Some API providers like to supply version information as a media type parameter, so let's be mindful of the fact that such a practice will now be disallowed unless a provision is made for it in these requirements.

format/index.md
bintoro, line 36: I think it reads much better now :+1:

ethanresnick:
> Some API providers like to supply version information as a media type parameter, so let's be mindful of the fact that such a practice will now be disallowed.

True. Are you saying we should have a non-normative note somewhere pointing this out to people, or make an exception for a version parameter?

bintoro:
> True. Are you saying we should have a non-normative note somewhere pointing this out to people, or make an exception for a version parameter?

The latter. Not saying we necessarily should, but _if_ we want to accommodate the use case, I think it would be a reasonable provision to reserve something like `api-version`, or maybe `api-*`, for private use.

format/index.md
ethanresnick: **Edit**: I've now moved this into the top section, as you suggested. About the other logic, seem my line notes on the new diff.

format/index.md
```diff
@@ -592,16 +602,6 @@ The following characters **MUST NOT** be used in member names:
 Data, including resources and relationships, can be fetched by sending a
 `GET` request to an endpoint.
 
-Clients **MUST** indicate that they can accept the JSON API media type, per
```
ethanresnick, line 595: Rather than saying that "Clients **MUST** indicate that they can accept the JSON API media type", I'm saying above that "Servers **MUST NOT** send responses with the `application/vnd.api+json` media type if the client has not indicated...it can understand this format." This seems to get at what we want—namely, you must ask for JSON API data in order to get it—while giving clients the freedom to ask for the other representations that, as our non-normative note alluded to, can exist at the same endpoint.

format/index.md
```diff
@@ -592,16 +602,6 @@ The following characters **MUST NOT** be used in member names:
...
-the semantics of the HTTP `Accept` header. If present, this header value **MUST**
-also include any media type extensions relevant to the request. Servers **MUST**
-return a `406 Not Acceptable` status code if this header specifies an
```
ethanresnick, line 598: I removed this line about "Servers **MUST** return a `406 Not Acceptable` status code if this header specifies an unsupported media type" because it seems unnecessarily strict. 

The HTTP spec explicitly says that if "none of the available representations for the response have a media type that is listed as acceptable, the origin server can either honor the header field by sending a 406 (Not Acceptable) response or disregard the header field by treating the response as if it is not subject to content negotiation.” Why should we take away this freedom to return some representation, which might be better than an unhelpful 406, if HTTP explicitly allows it? 

Again, that fallback representation couldn't be the JSON API media type because of the new line I mentioned earlier: "Servers MUST NOT send responses with the application/vnd.api+json media type if the client has not indicated...it can understand this format." So removing this line leaves us the guarantees we want, but keeps us out of specifying things that are unnecessary and/or beyond the prerogative of the spec.

format/index.md
```diff
@@ -26,28 +26,38 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
 
-Clients **MUST** ignore all parameters for the `application/vnd.api+json`
+Clients that include the JSON API media type in their `Accept` header, **MUST**
+specify the media type there at least once without any media type parameters.
+
+Servers **MUST** return a `400 Bad Request` status code if clients violate
+either of the requirements above. The server's response to such violations
+**MUST** contain a JSON API document with an [error object](#error-objects).
+The `"href"` member of that object **MUST** be
+`"http://jsonapi.org/errors/invalid-media-type-parameter"`.
```
dgeb: This is a more specific requirement for errors than anywhere else in the spec. I think we should remove this sentence about error object requirements. I'd prefer to prepare some detailed recommendations for using the error object in different scenarios.

format/index.md
```diff
@@ -26,28 +26,38 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+type if the client has not indicated, using its `Accept` header, that it can
+understand this format.
 
-> Note: These requirements may allow future versions of this specification
-to support an extension mechanism based upon the `ext` media type parameter.
+However, if a client's `Accept` header indicates that it accepts standard JSON
+but not the JSON API media type, the server **MAY** serve it a JSON API document
+identified as `application/json`. If the server does so, is not considered to
+be sending "JSON API data", but rather simple JSON data, and the client cannot
+rely on any of the rules laid out in this specification.
```
dgeb: I'm not sure that anything here should be normative.

format/index.md
```diff
@@ -26,28 +26,38 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
-`application/vnd.api+json` media type is modified by the `ext` parameter in
-the `Accept` header of a request. Otherwise, servers **MUST** return a `415
-Unsupported Media Type` status code if the `application/vnd.api+json` media
-type is modified by the `ext` parameter in the `Content-Type` header of a
-request. Servers **MUST** ignore all other parameters for the
-`application/vnd.api+json` media type in `Accept` and `Content-Type`
-headers.
+Servers **MUST NOT** send responses with the `application/vnd.api+json` media
+type if the client has not indicated, using its `Accept` header, that it can
+understand this format.
```
dgeb: I'd still prefer to specify situations in which 406 and 415 are appropriate. We can avoid any mention of the `ext` parameter and replace it with mention of _any_ media type parameter.

format/index.md
ethanresnick: @dgeb ~~What if we went the alternate route and added equally specific errors for other cases too? The idea of being able to identify an error by a URI seems really powerful to me. For clients, it helps them know exactly which error occurred (without us having to partition the error object's `code` values) and, for users, we could actually put up a page at http://jsonapi.org/errors that contains short, human-readable descriptions of each error, so it could be a huge usability win too.~~

~~If you're on board with the idea, I could work up a PR.~~

**EDIT** I'll remove the detail from here so we can get this merged, and then just open a separate PR for the other stuff.

format/index.md
ethanresnick: Ok! My biggest concern with this was about the identifiability of the cause of the error. But I'll switch to 415/406 now and then the identifiability issue can be addressed in the PR about error object hrefs.

ethanresnick:
@dgeb Ok, I removed the detailed error href, made the note about returning json api as json non-normative, and added `406`/`415` cases!

dgeb:
@ethanresnick thanks! 

Would you mind squashing these commits down to one, or perhaps a few logical ones. I'd prefer to keep the history sane and free of additions that are removed in the same PR.

ethanresnick:
@dgeb Done!

format/index.md
```diff
@@ -26,28 +26,40 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
 
-> Note: These requirements may allow future versions of this specification
-to support an extension mechanism based upon the `ext` media type parameter.
+> Note: If a client's `Accept` header indicates that it accepts standard JSON
```
tkellen, line 51: I don't think we should encourage this. I believe that this minor wheel greasing will actually be a disservice to clients. If they are required to provide the media type they accept, I suspect they will be more likely to find out what it is before consuming the data from a json-api service blindly.

tkellen, line 51: For context, I intend to provide a formatter for text/html in Endpoints that will return the response in a browser friendly format (tables, hyperlinks, etc).

tkellen:
:+1: from me, without commit a6a051cd61a5f080ce0629a830ce94f178e8974e

tkellen:
Closing in favor of #703. Thanks for your work here @ethanresnick!

PR #674, opened 2015-05-23
Require 400 for invalid include path

ethanresnick:
If we don't say this explicitly, a server might simply ignore invalid include paths. That would be bad because, if we expand the set of valid include paths down the line (e.g. with `-comments.author` style paths), we want the client to know whether those paths were honored or not.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -895,6 +895,10 @@ resource object.
 `comments` is a relationship listed under a `articles` resource object, and
 `author` is a relationship listed under a `comments` resource object.
 
+If any segment in any relationship path listed in the `include` parameter
+contains a relationship name that is not a [valid member name](#document-structure-member-names),
+the server **MUST** respond with `400 Bad Request`.
+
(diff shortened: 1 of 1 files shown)


ethanresnick:
**Edit**: After thinking about this more, I realized that an "invalid include path" is a pretty hard thing to define, because:
1. Some apis might not have schemas for each resource and, even for those that do, 
2. An include path can look for a relationship on a heterogeneous collection (e.g. `/my-heterogeneous-items?include=x`), so some of the `type`s' schema's in that response might have a relationship named "x" while others might not, and it's not clear that we want that to count as an invalid include path.

So, I updated this PR to say simply that if the include path segments aren't all valid member names, then the server must 400. This satisfies the PR's original goal (preserving our ability to extend include reliably for excluded paths etc), while not wading into the murky definitional issues mentioned above.

tkellen:
Closing in favor of #580, where @dgeb has suggested a general resolution to this and a different issue.

Issue #578, opened 2015-04-27
Resource identification by combination of its type and id members - in what scope?


mBread:
The section [Resource Identification](http://jsonapi.org/format/#document-structure-resource-identification) says "Every JSON API resource object is uniquely identified by a combination of its type and id members." but doesn't say what scope this applies to. Obviously, that cannot be true globally unless there is a registry of "type" values.

I expect the common inference is that it is "per API", but that is not said explicitly, and that concept is not present in the spec (at least not up to this point). Also, there do not appear to be any restrictions preventing one API (one implementation) from including links to another API (another implementation) - so just by following links you may find another resource with the same type and ID.

Later on [Compound Documents](http://jsonapi.org/format/#document-structure-compound-documents) says : "Note: In a single document, you can think of the type and id as a composite key that uniquely references resource objects in another part of the document." Is that what was meant in "Resource Identification" or not?

bintoro:
> I expect the common inference is that it is "per API"

Yes, this is the intent.

> there do not appear to be any restrictions preventing one API (one implementation) from including links to another API (another implementation)

As I understand it, this is not supposed to happen — at least not via the standard linkage mechanism.

tkellen:
Resolved in #676

PR #676, opened 2015-05-23
Clarify scope of type/id uniqueness

ethanresnick:
This is a PR to clarify/close the issue raised in #578.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -181,8 +181,9 @@ Here's how an article (i.e. a resource of type "articles") might appear in a doc
 Every resource object **MUST** contain an `"id"` member and a `"type"` member.
 The value of each of these members **MUST** be a string.
 
-Each resource object's `"type"` and `"id"` pair **MUST** identify a single,
-unique resource.
+Within a given API, each resource object's `"type"` and `"id"` pair **MUST**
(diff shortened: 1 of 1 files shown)


dgeb:
This may seem a bit pedantic, but I don't think that the spec normatively defines an "API". There are concepts that imply such a definition, such as a "server" needing to respond to `self` and `related` links.

Maybe this speaks to needing to merge what's implied by "server" and "API"?

ethanresnick:
Good point. Especially if we start thinking about relating resources between APIs, it'll only become more important to be explicit about these definitions somewhere.

Would a sentence like "The set of URIs controlled by a server, or multiple servers acting as one, constitute an API." suffice?

tkellen:
Yes, I think so @ethanresnick.

ethanresnick:
@tkellen Done!

dgeb:
> @tkellen Done!

@ethanresnick not seeing the change - did you forget to push?

ethanresnick:
D'oh. I accidentally pushed to a branch I have going for something else. Fixed.

tkellen:
:+1: from me

dgeb:
Thanks @ethanresnick!

PR #679, opened 2015-05-24
Fix broken reference links

jede:
Found some broken links on the format page, so I corrected the section headers.


tkellen:
This slipped under my radar and was fixed in #682, sorry we didn't get this merged. Thanks for taking the time to submit this anyway!

jede:
No worries :)

PR #680, opened 2015-05-26
Data and errors

hhware:
The removed text is an inconsistent duplication of the following:

> The members `"data"` and `"errors"` **MUST NOT** coexist in the same document.


dgeb:
:+1:

PR #682, opened 2015-05-26
Fixed href for attributes, relationships, and 'resource relationships'

joananeves:
Just a really minor fix because links were not working on Format page.


tkellen:
Thank you @joananeves!

PR #685, opened 2015-05-27
Updated formatting for consistent highlighting

jameswatts:


tkellen:
Thanks for this @jameswatts but I think we're going to go with #702 (if anything at all).

jameswatts:
@tkellen no problem, any improvement is good :+1:

Issue #659, opened 2015-05-21
Document Cleanup


tkellen:
Before we mark 1.0, we need to review the specification and do some cleanup to address the now scattered nature of the document. Below, a checklist of things we want to review:
- [x] clean up definition of meta and semantics around use
- [x] clean up definition of links and the semantics around use
- [x] replace references to URLs with the more general "links" (which can be a string or link object)
- [x] clean up bulleted lists to use the same style of prose everywhere
- [x] review use of "" and *\* around keywords/links
- [x] clean up usages of `-` and `:` after member names
- [x] review use of quoted vs. unquoted member names (e.g. `id` vs. `"id"`)
- [x] remove Content-Type from examples without content (e.g. GET, DELETE)

cziegenberg:
I just created PR #663, to avoid conflicts with future versions and the implementation of deferred features (like the extensions), This should be merged before 1.0.

tkellen:
boom.

dgeb:
:tada:

Issue #46, opened 2013-05-07
JSON API spec is not versioned


b:
The JSON API spec itself needs version information. When one of those mythical changes for which the current spec needs "flexibility", the version would be changed.

lukfugl:
:+1:

steveklabnik:
Yes, when we have something that's reasonably "final" we can talk about this.

b:
When the design is done the work to add any version information and behavior to the design will happen?

steveklabnik:
The version information is currently there:

> This document is a work in progress and will likely change over the next month as implementation work progresses. Implementors should be aware that this specification is not stable.

I'm not comfortable with adding anything further than this at this time. This is a draft, a request for feedback. Things get official version numbers when they're actually final in some form.

b:
Think you're misunderstanding. A JSON API speaker needs a way to signal which version of the protocol it supports, regardless of what the current version number on the spec is. The mechanism needs to be part of the spec, even if the current document doesn't have a version.

steveklabnik:
Yes, I did misunderstand, thank you.

My inclination is that most media types do not present an internal version, because the commitment to backwards compatibility is so great. That said, no official decision has been made about this, so I'll re-open it until we discuss it formally.

b:
Version information can be expressed and negotiated in numerous ways, media type being one. I'm not pushing a specific mechanism, only urging strongly that there be one, and soon. Versioning of APIs and versioning of resources are also missing or underspecified, but are being handled in other issues.

steveklabnik:
After a discussion with @wycats, we plan on doing the usual media-type thing: never remove, only add. Therefore we don't need explicit versioning after we declare it stable, as it will _always_ be backwards compatible.

I've updated #1 to add things around versioning your entire API.

PR #687, opened 2015-05-27
add support for top level jsonapi member

tkellen:


dgeb:
This is the solution to the versioning issue, discussed exhaustively in https://github.com/json-api/json-api/issues/432, that @tkellen, @steveklabnik, and I arrived at after further discussion. By allowing implementations to place a version in the document, they can advertise their _possible_ level of support for different additive features introduced at different versions of the spec. Furthermore, the top-level `jsonapi` member can allow for other spec-specific information without encroaching on the wholly user-defined `meta` members.

:+1:

joananeves:
Shouldn't [FAQ](http://jsonapi.org/faq#why-is-json-api-not-versioned) refer to this new top level member?
It only refers to #46.

format/index.md
```diff
@@ -527,6 +529,28 @@ the value of each member of a `links` object is not constrained by the
 current version of the spec. It should be assumed that anything can be
 allowed in the future for such values: object, array, or scalar.
 
+### JSON API Object <a href="#document-structure-jsonapi-object" id="document-structure-jsonapi-object" class="headerlink"></a>
+
+A JSON API document **MAY** include information about its implementation under
+a top level `"jsonapi"` member. If present, it **MUST** be an object containing
+a `version` member whose value is a string indicating the highest JSON API
+version supported. Other than `"meta"`, servers **MUST NOT** include any
```
ethanresnick, line 537: @dgeb Possible nitpick: should we say: "a `version` member... indicating the highest JSON API version at least partially supported"? Because I'm imagining that a server could support some optional features from 1.1 without fully supporting 1.0's optional features, and the reference to "partial support" would seem to make it less clearer that the server can return 1.1 even though it hasn't completely implemented 1.0.

ethanresnick:
:+1:

format/index.md
tkellen, line 537: I believe that's covered in my note below:

> Note: Because JSON API is committed to making additive changes only, the version string primarily indicates which new features a server may support.

format/index.md
```diff
@@ -451,6 +452,7 @@ As discussed above, a JSON API document **MAY** be extended to include
 meta-information as `"meta"` members in several locations:
 
 * at the top-level
+* within the top level jsonapi object.
```
steveklabnik, line 455: `jsonapi` should probably be in backticks here

tkellen, line 455: Done. A lot of cleanup like this needs to happen document-wide. If you see any general cleanup issues, could you add them to #659?

PR #688, opened 2015-05-27
Include exception to full linkage requirement.

### LINKED ISSUE #670: Provide exception for compound document "full linkage"
Allow an exclusion in the full linkage requirement for compound documents when `fields` excludes a particular relationship that would otherwise contain linkage.

dgeb:
Full linkage is not required when relationship fields are
excluded via sparse fieldsets.

[Closes #670]


tkellen:
Works for me. :+1:

PR #690, opened 2015-05-27
Links in errors

hhware:
Completing the `links` invasion by eradicating the last remaining `href` outside of `links`.

@ethanresnick, I think `about` is a better relation name for this.


tkellen:
:+1: from me

dgeb:
Good call @hhware :+1:

PR #691, opened 2015-05-27
Added example for "Sparse Fieldsets" (examples page)

ksimka:
It's instead of #667


dgeb:
Looks good to me. Thanks @ksimka :+1:

PR #694, opened 2015-05-27
disallow links and included if data is missing

### LINKED ISSUE #689: some top-level keys should not be allowed unless primary data is present
The spec currently says:

```
A document's top level **MAY** also have the following members:

* `"jsonapi"`: an object describing the server's implementation
* `"links"`: URLs related to the primary data.
* `"included"`: an array of resource objects that are related to the primary
  data and/or each other ("included resources").

If any of these members appears in the top-level of a response, their values
**MUST** comply with this specification.
```

In cases where `"data"` is absent (i.e. the response contains errors or just plain `"meta"`), then I believe `"included"` and maybe also `"links"` should be disallowed from the top-level. I wasn't able to find any language that stated this explicitly, though maybe I missed something.

csantero:
This closes #689. I would request that this be considered for review before 1.0, since this change will not be possible afterward.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -134,6 +134,8 @@ A document's top level **MAY** also have the following members:
 If any of these members appears in the top-level of a response, their values
 **MUST** comply with this specification.
 
+If a document does not contain a top-level `data` key, then the `included` key **MUST NOT** be present either.
+
 ### Resource Objects <a href="#document-structure-resource-objects" id="document-structure-resource-objects" class="headerlink"></a>
 
 "Resource objects" appear in a JSON API document to represent resources.


dgeb:
I'm fine with this requirement. However, I think it should be moved below the next sentence / paragraph so that the phrase "any of these members" in that sentence clearly apply to the full list. 

@csantero - would you mind making this change please?

lode:
Why is a top-level `links` not allowed? Shouldn't that always contain a self link which resulted in this error?

dgeb:
@lode that's a good point - thanks for raising it. Let's leave `links` out of this requirement.

csantero:
@dgeb Updated the diff to only cover `included`, and moved the paragraph.

dgeb:
Great - thanks @csantero!

PR #696, opened 2015-05-27
type constraint

tkellen:


dgeb:
I think we want to specify that type values MUST follow the same restrictions that apply to member names.

And we could add this recommendation for type values to http://jsonapi.org/recommendations/#naming.

tkellen:
Updated.

dgeb:
Thanks :+1:

PR #698, opened 2015-05-27
Update sample HTTP request styling

nonsensery:
#### What's this PR do?

Fixes some inconsistencies in the styling (and content) of the sample HTTP requests on the "Format" page.

I made these three changes:
1. Use `http` for the code-block language.
2. Add missing `HTTP/1.1` to the request line.
3. Include the `Accept` and `Content-Type` headers.

Changes (1) and (2) mentioned above enable proper syntax highlighting for these blocks.

Regarding change (3), I can imagine some people objecting because it "adds noise" to the sample requests. However, those headers are **required** for a valid request, so it seems reasonable to include them. (In fact, _some_ of the sample requests already included these headers. I only added them where they were missing.)
#### Screenshots (if appropriate)

Before:
screen shot 2015-05-27 at 1 46 48 pm

After:
screen shot 2015-05-27 at 1 47 06 pm

Before:
screen shot 2015-05-27 at 1 47 30 pm

After:
screen shot 2015-05-27 at 1 48 08 pm

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -645,19 +645,25 @@ A server **MUST** support fetching resource data for every URL provided as:
 For example, the following request fetches a collection of articles:
 
 ```http
-GET /articles
+GET /articles HTTP/1.1
+Content-Type: application/vnd.api+json
+Accept: application/vnd.api+json
 ```
 
 The following request fetches an article:
 
 ```http
-GET /articles/1
+GET /articles/1 HTTP/1.1
+Content-Type: application/vnd.api+json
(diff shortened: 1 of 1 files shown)


steveklabnik:
Thanks so much! this is great!

dgeb:
Thanks @nonsensery. I'm :+1: to most of this - but specifying `Content-Type` for requests without content (e.g. GET requests) is inappropriate. Would you mind removing these in a separate PR please?

PR #700, opened 2015-05-27
Remove `Content-Type` header from GET requests

nonsensery:
#### What's this PR do?

Removes the `Content-Type` header from the sample HTTP requests on the "Format" page.

This is a follow-up to PR #698 that added these headers (among other things), requested by @dgeb.

It seems that the header _is_ allowed for GET requests (that is, it's not _forbidden_), but it's a bit odd since there is no content.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -651,7 +651,6 @@ For example, the following request fetches a collection of articles:
 
 ```http
 GET /articles HTTP/1.1
-Content-Type: application/vnd.api+json
 Accept: application/vnd.api+json
 ```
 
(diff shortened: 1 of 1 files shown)


nonsensery:
Somewhat unrelated to whether this PR gets merged, thinking about this made me wonder whether it might actually be a _good_ thing to specify the JSON API content type for GET requests. In general, the `Content-Type` header specifies the format of the request and the `Accept` header specifies the expected format of response. Since JSON API is an entire protocol built atop HTTP, not just a content type, maybe _every_ JSON API request should indicate "Hey, I'm speaking JSON API."

ethanresnick:
:+1: I get what the examples are trying to show/signal here, but let's just stick with HTTP semantics, for which content-type on a bodyless request is non-sensical

dgeb:
Thanks for removing these headers @nonsensery :+1:

nonsensery:
:+1:

PR #702, opened 2015-05-27
Convert all JSON code blocks to valid JSON

nonsensery:
#### What's this PR do?

Updates the JSON code blocks so that they are properly highlighted.

In this branch, I've changed the JSON code blocks so that they're all tagged as `json` (previously, some were `javascript`). Since JSON is more restrictive than JavaScript, I made some changes to the content of the code blocks so that they are all valid JSON now.

Unfortunately, this is a situation with no "right" answer. In order to make the code blocks into valid JSON data structures, I had to make some changes to their content. (For example, comments are not allowed in JSON data.) I've tried to find a compromise that allows the syntax to be valid while still communicating the essential points.

In general, I'm not crazy about compromising on the content just to make the syntax highlighter happy, so I'm interested in iterating here. One way or another, there will be a comprise here. Currently, the compromise is inconsistent syntax highlighting.
#### Alternatives worth considering?

Another option would be to just use to `javascript` for all of these code blocks, since that would allow the inline comments to stay.

The JS syntax highlighter seems to be more forgiving of partial data structures as well, for example:

`javascript`:
screen shot 2015-05-27 at 3 18 32 pm

`json`:
screen shot 2015-05-27 at 3 18 48 pm

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -93,16 +93,22 @@ Primary data **MUST** be either:
 
 For example, the following primary data is a single resource object:
 
-```javascript
+```json
 {
   "data": {
     "type": "articles",
     "id": "1",
     "attributes": {
-      // ... this article's attributes
+      "title": "JSON API paints my bikeshed!"
     },
     "relationships": {
-      // ... this article's relationships
+      "author": {
+        "links": {
+          "self": "/articles/1/relationships/author",
+          "related": "/articles/1/author"
+        },
+        "data": { "type": "people", "id": "9" }
+      }
     }
   }
 }
@@ -111,7 +117,7 @@ For example, the following primary data is a single resource object:
 The following primary data is a single [resource identifier object] that
 references the same resource:
 
-```javascript
+```json
 {
   "data": {
     "type": "articles",
@@ -162,25 +168,23 @@ In addition, a resource object **MAY** contain any of these top-level members:
 
 Here's how an article (i.e. a resource of type "articles") might appear in a document:
 
-```javascript
-// ...
-{
-  "type": "articles",
-  "id": "1",
-  "attributes": {
-    "title": "Rails is Omakase"
-  },
-  "relationships": {
-    "author": {
-      "links": {
-        "self": "/articles/1/relationships/author",
-        "related": "/articles/1/author"
-      },
-      "data": { "type": "people", "id": "9" }
+```json
+  {
+    "type": "articles",
+    "id": "1",
+    "attributes": {
+      "title": "JSON API paints my bikeshed!"
+    },
+    "relationships": {
+      "author": {
+        "links": {
+          "self": "/articles/1/relationships/author",
+          "related": "/articles/1/author"
+        },
+        "data": { "type": "people", "id": "9" }
+      }
     }
   }
-}
-// ...
 ```
 #### Resource Identification <a href="#document-structure-resource-identification" id="document-structure-resource-identification" class="headerlink"></a>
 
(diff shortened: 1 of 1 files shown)


nonsensery:
This PR would supersede #685.

format/index.md
```diff
@@ -93,16 +93,22 @@ Primary data **MUST** be either:
...
 
-```javascript
+```json
 {
   "data": {
     "type": "articles",
     "id": "1",
     "attributes": {
-      // ... this article's attributes
+      "title": "JSON API paints my bikeshed!"
```
nonsensery, line 102: Since comments are not allowed in JSON, I've switched to including the entire object.

format/index.md
```diff
@@ -162,25 +168,23 @@ In addition, a resource object **MAY** contain any of these top-level members:
...
-  },
-  "relationships": {
-    "author": {
-      "links": {
-        "self": "/articles/1/relationships/author",
-        "related": "/articles/1/author"
-      },
-      "data": { "type": "people", "id": "9" }
+```json
+  {
```
nonsensery, line 172: In lieu of `// ...` comments to indicate that this is only part of a larger JSON payload, I've increased the indentation of this section. (Same thing on lines 293-311 and 333-342.)

format/index.md
```diff
@@ -162,25 +168,23 @@ In addition, a resource object **MAY** contain any of these top-level members:
...
+    "type": "articles",
+    "id": "1",
+    "attributes": {
+      "title": "JSON API paints my bikeshed!"
```
nonsensery, line 176: Oh yeah, I also changed the `title` to be consistent with the rest of the page ("Rails is Omakase" is article 2).
(Same thing on line 297.)

format/index.md
```diff
@@ -480,9 +480,6 @@ For example:
       "Steve Klabnik",
       "Dan Gebhardt"
     ]
-  },
-  "data": {
-    // ...
```
nonsensery, line 485: Here, I've removed the `"data"` section, since the `// ...` comment inside is not legal. Maybe it would make more sense to keep it and replace the `...` with `"type"` and `"id"` instead?

format/index.md
```diff
@@ -517,16 +514,21 @@ the following members:
 
 Examples of the supported formats:
 
-```
-"links": {
-  "self": "http://example.com/posts",
+```json
+{
```
nonsensery, line 518: The JSON syntax highlighter requires a fully-formed JSON data structure, so I've added the `{` and `}` around the sample `"links"` section. (Same thing on lines 526-533.)

ethanresnick:
I think the solution here is to just highlight everything as javascript, since:
1. I don't think we should compromise by getting rid of the comments;
2. We're already highlighting the vast majority of our examples as javascript; and 
3. the javascript highlighting is more colorful (which is helpful for making the examples stand out).

@nonsensery Would you rebase this to highlight as js instead?

tkellen:
Can we also add a FAQ entry in this PR that says we're not changing these again? :p

nonsensery:
Many of the changes in this branch are to turn the code blocks into valid JSON. Rebasing/adapting this branch would require undoing those. It would probably be easier to just make a new branch/PR for the switch to JS highlighting. I have some free time today, so I can take a look at it.

nonsensery:
I've put together a new PR (#928) that switches to JS highlighting without modifying the content at all.

ethanresnick:
Closing in favor of #928

Issue #518, opened 2015-03-29
406 Not Acceptable


e-oz:
From http://jsonapi.org/format/#fetching :

> JSON API requests MUST include an Accept header specifying the JSON API media type. This header value MUST also include media type extensions relevant to the request. Servers MUST return a 406 Not Acceptable status code if this header is missing or specifies an unsupported media type.

2 reasons against:  
1. Not always you can control http headers in request (especially CORS). It should not prevent you to make a request.
2. This limitation should not be for the sake of limitation. Why should we increase number of failed requests? Just because "standard" wants it? In real world, sometimes projects [uses POST instead of GET](https://blogs.dropbox.com/developers/2015/03/limitations-of-the-get-method-in-http/), what is much greater violation, but.. Standard should not add limitation just for the sake of limitation - there are recommendations for it.

e-oz:
I'm not going to implement this proposal of standard because of your position about "PATCH" verb and "id" field in responses, so I don't see reasons to argue and closing all my issues.

PR #703, opened 2015-05-27
Content negotiation redux

dgeb:
This is a refinement of @ethanresnick's work in #673. It separates the client and server responsibilities for clarity and corrects a few aspects. 

I won't attempt to summarize the tweaks. It's important to review the full text carefully.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -28,26 +28,31 @@ interpreted as described in RFC 2119
 
 ## Content Negotiation <a href="#content-negotiation" id="content-negotiation" class="headerlink"></a>
 
-Clients **MUST** send all JSON API data in request documents with
-the header `Content-Type: application/vnd.api+json`.
+Clients **MUST** send all JSON API data in request documents with the header
+`Content-Type: application/vnd.api+json`, without any media type parameters.
 
-Clients **MUST** ignore all parameters for the `application/vnd.api+json`
+If a client violates this requirement, servers **MUST** respond with a
+`415 Unsupported Media Type` status code.
+
+Clients that include the JSON API media type in their `Accept` header, **MUST**
+specify the media type there at least once without any media type parameters.
+
+If a client violates this requirement, and its `Accept` header doesn't contain
+any other media types that the server can produce, the server **MUST** respond
+with a `406 Not Acceptable` status code.
+
+Servers **MUST** send all JSON API data in response documents with the header
+`Content-Type: application/vnd.api+json`, without any media type parameters.
+
+Servers **MUST NOT** send responses with the `application/vnd.api+json` media
+type if the client has not indicated, using its `Accept` header, that it can
+understand this format.
+
+Clients **MUST** ignore any parameters for the `application/vnd.api+json`
 media type received in the `Content-Type` header of response documents.
 
-Servers **MUST** send all JSON API data in response documents with
-the header `Content-Type: application/vnd.api+json`.
-
-Servers **MUST** return a `406 Not Acceptable` status code if the
-`application/vnd.api+json` media type is modified by the `ext` parameter in
-the `Accept` header of a request. Otherwise, servers **MUST** return a `415
-Unsupported Media Type` status code if the `application/vnd.api+json` media
-type is modified by the `ext` parameter in the `Content-Type` header of a
-request. Servers **MUST** ignore all other parameters for the
-`application/vnd.api+json` media type in `Accept` and `Content-Type`
-headers.
-
-> Note: These requirements may allow future versions of this specification
-to support an extension mechanism based upon the `ext` media type parameter.
+> Note: These requirements exist to allow future versions of this specification
+to use media type parameters for extension negotiation and versioning.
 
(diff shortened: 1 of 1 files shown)


tkellen:
:+1:

ethanresnick:
Thanks for working on these refinements, @dgeb! I'm excited to get this text in too!

:+1: On dividing client and server responsibilities, and on some of your rewordings, which made things far less awkward.

:-1: On the sentence "Servers **MUST** respond with a `406 Not Acceptable` status code if a request's `Accept` header doesn't contain any media types that the server can produce." Again, this is us being more prescriptive [than HTTP itself](https://github.com/json-api/json-api/pull/673#discussion_r30962383), and feels like stepping out of bounds to me.

:-1: On removing the non-normative note about serving JSON API as JSON. I added that because a good server _should_ support this, to make it easier for developers to make requests using generic clients (like curl) without having to go to the (totally unnecessary) trouble of setting a custom `Accept` header. Responding with JSON as a fallback also just generally makes the API more robust.

Of course, you guys have the final call on all this, but that's my .02

tkellen:
Okay, the curl example has convinced me. I think you're right RE: fallbacks @ethanresnick.

ethanresnick:
@tkellen Cool! :)

Credit where it's due, though: the idea for the note originally goes back to https://github.com/json-api/json-api/issues/518.

dgeb:
If we remove this statement: "Servers MUST respond with a 406 Not Acceptable status code if a request's Accept header doesn't contain any media types that the server can produce."

Then we remove the ability for clients / servers to negotiate based on media type parameters in the future.

dgeb:
This is why I specifically called out the `ext` parameter in the current section on media type negotiation: http://jsonapi.org/format/#media-type-negotiation

dgeb:
This is awkwardly worded, but provides the allowance we need while (I think) respecting @ethanresnick's concerns about overstepping ourbounds:

> Servers MUST respond with a `406 Not Acceptable` status code if a request's `Accept` header contains the JSON API media type and the only instances of that media type are modified with media type parameters.

ethanresnick:
@dgeb Your reworded clause works for me! 

Also, sorry, I should've been clearer originally. I was imagining that, if we removed the sentence, extension negotiation would still work because the client would check the `Content-Type` of the response to know if the contract has been established successfully. If it got JSON-API content back (eventually, with the extension parameters), things worked; if it got another content type, or a 406, then negotiation has failed.

The theoretical advantage of this would have been that it preserves the flexibility that HTTP allows for responding to a request like:

``` http
GET /articles
Accept: text/plain
```

With:

``` http
200 OK
Content-Type: application/json
{
   "data": { //... }
}
```

Rather than a 406.

Also, my thinking was that clients would have to inspect `Content-Type` in some cases anyway (and not just rely on status codes) to determine the result of negotiation—as when they say they'll support multiple (combinations of) extensions and they need to know which (combination) was chosen—so it wouldn't be too painful to make them always check that field.

That said, if we can give status code indicators, I think we should, and your new clause seems to let us do that while keeping all of HTTP's allowed wiggle room

dgeb:
@ethanresnick thanks for confirming! I've pushed the reworded clause.

Let's try to get something non-normative in tomorrow that will scratch your itch regarding working with generic clients.

format/index.md
```diff
@@ -26,28 +26,35 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
 
-Servers **MUST** return a `406 Not Acceptable` status code if the
-`application/vnd.api+json` media type is modified by the `ext` parameter in
-the `Accept` header of a request. Otherwise, servers **MUST** return a `415
-Unsupported Media Type` status code if the `application/vnd.api+json` media
-type is modified by the `ext` parameter in the `Content-Type` header of a
-request. Servers **MUST** ignore all other parameters for the
-`application/vnd.api+json` media type in `Accept` and `Content-Type`
-headers.
+Servers **MUST** respond with a `406 Not Acceptable` status code if a
```
tkellen, line 51: How about

> Servers **MUST** respond with a `406 Not Acceptable` status code when a request's `Accept` header contains the JSON API media type and **all** instances of that media type are modified with media type parameters.

tkellen:
Thanks for re-working this @dgeb. I'm :+1: to merge with that minor change.

format/index.md
dgeb, line 51: :+1:

dgeb:
Thanks for the careful review @ethanresnick and @tkellen!

ethanresnick:
@dgeb Absolutely!

Btw, as far as adding the non-normative note about serving `application/json` goes: let me know you want me to take another stab at that and, if so, whether there's anything in particular that should change from my prior version beyond general concision. Otherwise, if it's easier for you to just write something up yourself, that's fine too.

dgeb:
@ethanresnick Thanks - I'll definitely review your non-normative note again now that the normative changes have been merged.

PR #705, opened 2015-05-28
Clean up descriptions of attributes and relationships.

dgeb:
Clean up descriptions of attributes and relationships:
- Move the common namespace requirement to Fields
- Provide symmetry and conciseness in descriptions.

Also, removes some unnecessary commas from the content negotiation section in a separate commit.


tkellen:
Beat me to it. Huge :+1: to this simplification.

PR #706, opened 2015-05-28
Remove odd relationships requirement under attributes.

dgeb:
I don't believe that this requirement needs to be normative. Furthermore,
it seems to limit the future possibilities of the `relationships` members
reserved under complex attributes.


tkellen:
:+1: this was my addition and I agree it's out of place now.

PR #707, opened 2015-05-28
style code tags in lists as well as paragraphs

### LINKED ISSUE #704: CSS for code tags doesn't let them stand out inside lists
When code tags appear in paragraphs, they get the shaded background. This CSS doesn't apply inside lists. Currently, the word "about" in the second item of the before screenshot doesn't really stand out. Applying the paragraph `code` CSS to all `code` tags makes "about" stand out, but now the subject of each list item is also shaded, which may or may not be the desired effect. What do people think?
### Before:

jsonapi code blocks

tkellen:
closes gh-704


dgeb:
Thanks - that was bugging me!

PR #708, opened 2015-05-28
suffix identifiers in backticks with colons consistently

tkellen:


format/index.md
```diff
@@ -261,9 +261,9 @@ relationship isn't currently associated with any target resources.
 
 Resource linkage **MUST** be represented as one of the following:
 
-* `null` for empty to-one relationships.
-* a [resource identifier object] for non-empty to-one relationships.
-* an empty array (`[]`) for empty to-many relationships.
+* `null`: for empty to-one relationships.
```
dgeb: I don't think we want a `:` here, since we're not defining anything.

format/index.md
```diff
@@ -261,9 +261,9 @@ relationship isn't currently associated with any target resources.
...
+* `[]`: for empty to-many relationships.
```
dgeb: Same here - probably ok to stick with the old wording about "an empty array (`[]`)"

dgeb:
Looks good except for those two nits.

tkellen:
updated

dgeb:
thanks @tkellen!

PR #709, opened 2015-05-28
simplify normative text for fetching data

tkellen:


ethanresnick:
This seems to be saying that a server can't have _any_ `POST` only URIs. Is that intentional? What about if it has a collection that you can add too but not list (i.e. `POST` but not `GET`)?

I think the point of the old wording, even though it was necessarily more verbose, was that it was only constraining URIs the server advertised in particular `links`, like `self`, `related`, etc.

dgeb:
@ethanresnick thanks for catching this oversight. @tkellen is going to rephrase in terms of links.

tkellen:
The spirit of this change was to remove the explicit listing for each link because it would be silly to have to add to this list every time we add a new supported link relation. Honestly, I'm not convinced we need this normative phrase at all. It seems incredibly tacit to me that an API should allow fetching resources it links to. Additionally, the previous clause normatively disallowed error responses.

tkellen:
@ethanresnick What do you think about removing this normative requirement entirely?

ethanresnick:
@tkellen, I like the _idea_ of a normative requirement, but the thing we'd be requiring here does seem blindingly obvious. So I suppose I'd be ok with removing the text, provided @dgeb is too.

dgeb:
We've reverted because I remember that @wycats and I specifically came up with these conditions to ensure that GETs are supported at these endpoints.

By reverting the old verbiage we're not closing the door on `POST` only URLs while guaranteeing that these endpoints can still be fetched.

ethanresnick:
@dgeb That sounds good to me. Also, supporting a `GET` request doesn't mean returning a 2xx, it just means not returning `405 Method Not Allowed` (right?). So if there's an authentication error at one of the linked endpoints, returning 401/403, still seems like it would be fair game. :+1:

PR #719, opened 2015-05-28
JSON API version MAY

ethanresnick:
It's optional for a server to provide the JSON API version it's using in its response document, in that its optional for it to include a `jsonapi` member at all. Yet, the current construction means that if the server does include a `jsonapi` member, then it must include a version. That seems a little odd, since, even now, the server could be including the `jsonapi` member for non-version-related reasons, like to take advantage of the `meta` container there. Also, if we add more keys in `jsonapi` servers could again want to add that object (with those new keys) without being compelled to add a version, which I think we're trying to keep optional.

This PR accomplishes that by giving the version key a MAY even if the `jsonapi` object is present.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -531,7 +531,7 @@ array of values, whereas a `self` link does not).
 
 A JSON API document **MAY** include information about its implementation
 under a top level `jsonapi` member. If present, its value **MUST** be an
-object (a "jsonapi object") that **MUST** contain a `version` member whose
+object (a "jsonapi object") that **MAY** contain a `version` member whose
 value is a string indicating the highest JSON API version supported. This
 object **MAY** also contain a `meta` member, whose value is a [meta] object
 that contains non-standard meta-information.
@@ -544,8 +544,8 @@ that contains non-standard meta-information.
 }
 ```
 
-If this member is not present, clients should assume the server implements
-version 1.0 of the specification.
+If the `version` member is not present, clients should make assume the server
+implements version 1.0 of the specification.
 
 > Note: Because JSON API is committed to making additive changes only, the
 version string primarily indicates which new features a server may support.


tkellen:
I'm curious why do you find it compelling to allow this. If an implementation goes through the trouble to use the top level `jsonapi` member, why wouldn't it also indicate a version?

ethanresnick:
I don't find including `version` very compelling, but I think it's possible that we'll add something down the line that _would_ be compelling (e.g. a supported features list, an endpoints-style manifest). So, it just seems like a server should be able to take advantage of those members without also being forced into providing a `version` that would otherwise be optional; making it such that using a new key triggers a requirement for a different key seems like a bit of a nonsequiter.

tkellen:
Fair enough. If @dgeb agrees, I'm game to merge this.

dgeb:
Seems fine. But maybe we should say "If the `version` member is not present" instead of "If this member is not present"

ethanresnick:
@dgeb Oh wow, I totally missed that phrase until now. I changed it to say "the `version` member" as you suggested. 

But, should we also change it to say: "If the `version` member is not present, clients should make assume the server implements **at least** version 1.0 of the specification." That way, the client can test support for other potential features, rather than assuming they simply aren't supported?

dgeb:
> That way, the client can test support for other potential features, rather than assuming they simply aren't supported?

"at least version 1.0" makes sense to me. :+1:

tkellen:
I'm :+1: to merging after that update too.

format/index.md
```diff
@@ -544,8 +544,8 @@ that contains non-standard meta-information.
 }
 ```
 
-If this member is not present, clients should assume the server implements
-version 1.0 of the specification.
+If the `version` member is not present, clients should make assume the server
```
dgeb, line 547: "clients should make assume" -> "clients should assume"

dgeb:
ping @ethanresnick :)

are you free to make the tweak above?

dgeb:
I'm going to take this PR over and make the tweak - thanks @ethanresnick for bringing this up!

dgeb:
Merged via 5b6ece144b55b3d8b554c45541f2b3cc1a3444a3

ethanresnick:
Thanks for taking this over @dgeb! Sorry I couldn't be available synchronously for the past hour or so...been in a meeting thing.

PR #723, opened 2015-05-29
add a few intra-document links

tkellen:
I would just push this to master but I don't trust myself after this long editing. @dgeb?


dgeb:
LGTM :+1:

PR #724, opened 2015-05-29
400 for unknown, valid spec query parameters

ethanresnick:
If we add a new query parameter later that complies with the base spec naming rules (i.e. no hyphens, underscores, etc), servers that don't support that query parameter need to return 400 if they encounter it. (Though servers are free to ignore implementation-specific query parameter names if they want.) This PR ensures that.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -1704,6 +1704,10 @@ least one non a-z character (U+0061 to U+007A). It is **RECOMMENDED** that a
 U+002D HYPHEN-MINUS, "-", U+005F LOW LINE, "_", or capital letter is used
 (e.g. camelCasing).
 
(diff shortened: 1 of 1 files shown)


format/index.md
```diff
@@ -1704,6 +1704,10 @@ least one non a-z character (U+0061 to U+007A). It is **RECOMMENDED** that a
 U+002D HYPHEN-MINUS, "-", U+005F LOW LINE, "_", or capital letter is used
 (e.g. camelCasing).
 
+If a server encounters a query parameter that does not follow the naming
```
tkellen, line 1707: does not **know** how

dgeb:
Looks good now :+1:

PR #730, opened 2015-05-29
Styles: typography improvements

ethanresnick:
Tweaks:
1. Use more generous leading. This is helpful given how long our line lengths are by conventional typographic standards. It also helps considering the (padded) gray background behind inline code snippets, which need a bit of breathing room.
2. Don't line break within inline code snippets. This makes them a bit easier to read, by making sure, e.g. that "400" and "Bad Request" stay on the same line in `400 Bad Request`.

Note: this doesn't recompile the sass to update the `all.css` and `all.css.map` files, since I assume compilation is handled by a build step somewhere.

### DIFF
--- a/stylesheets/all.sass
+++ b/stylesheets/all.sass
@@ -26,7 +26,7 @@ $border-radius-main: 4px
 body
   font-family: "Helvetica Neue", "Helvetica", "Open Sans", sans-serif
   font-size: 15px
-  line-height: 1.5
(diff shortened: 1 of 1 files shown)


dgeb:
@ethanresnick could you push the css and css.map files as well actually - we're just relying on github pages for now, with no build step.

tkellen:
I'm going to do this and land it now, actually.

PR #732, opened 2015-05-29
Fix duplicate `[links]` definitions

nonsensery:
#### What's this PR do?

This is a small change to fix some `[links]` URLs.

There were two entries in the Markdown URLs section (at the end of the document) for `[links]`: one pointing at the top-level **Links** section, and one pointing at the **Links** sub-section of the **Resource Objects** section. This branch renames the latter to `[resource links]`, and updates the places were `[links]` was being used incorrectly.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -255,7 +255,7 @@ A relationship object that represents a to-many relationship **MAY** also contai
 
 #### Related Resource Links <a href="#document-resource-object-related-resource-links" id="document-resource-object-related-resource-links" class="headerlink"></a>
 
-A "related resource link" provides access to resource objects [linked][links]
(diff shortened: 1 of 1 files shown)


nonsensery:
Congratulations on 1.0, BTW!

dgeb:
> Congratulations on 1.0, BTW!

Thanks @nonsensery - and thanks for all your help in the final days!

So I've merged a rebased and partial form of this PR in 81e7797692dd6f2045911a1c04401165685a42ea. Thanks for fixing the reference duplication and clarifying the "Resource Links" heading. I've stuck with using the general "links" references instead of your changes, since the resource-level links object can just contain `self` links now, and those references were more general.

PR #734, opened 2015-05-31
Update description of Fortune.js

gr0uch:
We library now. Also there's [tests](https://github.com/fortunejs/fortune/blob/master/test/integration/serializers/json_api.js) for compatibility with JSON API 1.0.


tkellen:
Rad!

PR #735, opened 2015-06-01
move fields definition above attributes and relationships

tkellen:
hopefully this will make it more clear that id and type are
reserved in both the attributes and relationships containers.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -212,6 +212,16 @@ The values of `type` members **MUST** adhere to the same constraints as
 can be either plural or singular. However, the same value should be used
 consistently throughout an implementation.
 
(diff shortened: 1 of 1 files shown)


bmac:
:+1:

tkellen:
@bmac just asked me to confirm if `type` was reserved under attributes. I immediately linked him to the attributes section of the spec, only to remember this reservation is not listed there, but rather under fields. This allows a single normative statement to cover both the relationships and attributes containers, but is a bit confusing given how much further into the document you have to go to find it. In addition to moving this, I believe we should add a small non-normative note to both the attributes and relationships section.

@dgeb?

dgeb:
That flows better - thanks @tkellen!

PR #740, opened 2015-06-02
Add quotes around JSON Pointer example strings

dgeb:
We should probably clarify that the JSON Pointer examples in the Errors section are strings.


tkellen:
:+1: those were lost in the general cleanup we did around other identifiers unnecessarily having quotes.

PR #751, opened 2015-06-05
Reindent json properly

amenophis:


steveklabnik:
Which definition of 'proper' is this? Ie, is this some sort of globally recognized style for formatting?

amenophis:
I just make it more readable. Use Sublime Text indentation.
you could take it or not ;)

steveklabnik:
Fair enough! I was just curious :)

Let's see what others think. We haven't made an explicit judgment about how to format these things.

tkellen:
Six of this half a dozen of the other to me. If we're going to update one we should make them all consistent though.

tkellen:
@dgeb @ethanresnick? If either of you want this, this PR should be updated to include all other examples. I'm indifferent.

ethanresnick:
I'm indifferent too. Maybe not worth the hassle.

tkellen:
I appreciate the effort @amenophis, but I think we're going to pass on this for now.

Issue #647, opened 2015-05-21
Sparse field sets examples not RFC 3986 compliant


jaspervz:
The current examples for sparse field sets use non-RFC 3986 compliant URLs: [ and ] are reserved characters and must be percent encoded.

The RFC 3986 compliant version of the example is: GET /articles?include=author&fields%5Barticles%5D=title,body&fields%5Bpeople%5D=name

Should [ and ] be used to denote the type if the client needs to encode these? At least the example needs to be updated because it isn't RFC 3986 compliant.

See: https://www.ietf.org/rfc/rfc3986.txt
3.4 Query: 
query       = *( pchar / "/" / "?" )

And 
pchar = unreserved / pct-encoded / sub-delims / ":" / "@"
unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
reserved    = gen-delims / sub-delims
gen-delims  = ":" / "/" / "?" / "#" / "[" / "]" / "@"
sub-delims  = "!" / "$" / "&" / "'" / "(" / ")"
                  / "*" / "+" / "," / ";" / "="

So [ and ] must be percent encoded.

Proposal: perhaps the dot could work instead, the example becomes than:
GET /articles?include=author&fields.articles=title,body&fields.people=name

tkellen:
ref #544

ethanresnick:
@tkellen If our spec were the only thing in the world, I'd be in favor of using `(` and `)`, or similar, instead of `[` and `]` to avoid this percent encoding trouble, but the `[]` syntax has become so common for expressing nested query parameters that I can't imagine we'd move away from it. So we just have to mention in our examples somewhere that the characters should be percent encoded (maybe even if we don't render them that way for usability?).

tkellen:
Precisely my thinking as well @ethanresnick. Would you mind working up a PR?

ethanresnick:
@tkellen Sure!

tkellen:
Resolved!

PR #756, opened 2015-06-07
Note that fields[TYPE] brackets must be percent encoded

ethanresnick:
This is simply making explicit a requirement from the URI spec that was
already implicitly in effect, so it’s not a BC break.

Addresses #647

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -993,6 +993,10 @@ GET /articles?include=author&fields[articles]=title,body&fields[people]=name HTT
 Accept: application/vnd.api+json
 ```
 
+> Note: The above example URI shows unencoded `[` and `]` characters simply for
+readability. In practice, these characters must be percent-encoded, per the
+requirements [in RFC 3986](http://tools.ietf.org/html/rfc3986#section-3.4).
+
 > Note: This section applies to any endpoint that responds with resources as
 primary or included data, regardless of the request type. For instance, a
 server could support sparse fieldsets along with a `POST` request to create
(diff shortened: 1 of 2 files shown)


tkellen:
Thanks @ethanresnick!

tkellen:
@dgeb what do you think about adding this MUST post 1.0?

tkellen:
ping @dgeb

format/index.md
```diff
@@ -988,11 +988,18 @@ COMMA, ",") list that refers to the name(s) of the fields to be returned.
 If a client requests a restricted set of [fields], an endpoint **MUST NOT**
 include additional [fields] in the response.
 
+The `[` and `]` characters in a `fields[TYPE]` parameter **MUST** be
+percent-encoded, per the URI formatting requirements
+[detailed in RFC 3986](http://tools.ietf.org/html/rfc3986#section-3.4).
+
```
dgeb: I think we can eliminate this normative addition and leave the non-normative notes to clarify usage.

ethanresnick:
Ok, updated!

There’s still an implicit requirement that these characters be percent-encoded, of course—the URI wouldn’t be valid under the URI spec if they weren’t—but I can understand us not wanting to specify that at the JSON API level (to avoid duplication/over-specification or in case the format for URIs evolves).

dgeb:
>  I can understand us not wanting to specify that at the JSON API level (to avoid duplication/over-specification or in case the format for URIs evolves).

My thinking exactly. Thanks for updating @ethanresnick!

@tkellen LGTM now :+1:

PR #757, opened 2015-06-07
Add print style sheet

jayeff:
This is a first version of the print style sheet. Additional testing on different browsers is still needed (currently only tested in Chrome for Mac).


stylesheets/print.scss
```diff
@@ -0,0 +1,56 @@
...
+  widows: 2;
+}
+
+// hanging punctation
+ul {
+  list-style: disc outside none;
+  padding: 0;
+}
+
+// link handling
```
jayeff, line 66: I did not yet make up my mind if the following rules are an improvement or not.

The first two rules print the URL target after the anchor tag. Most pages only have few links to other internal or external pages (except the implementation page). Therefore I'm not sure if this behavior improves the printed pages or not.

The specification has a lot of internal links, which have no benefit on printed paper as tapping a piece of paper does almost nothing :sunglasses: The last rule hides all fragment links but now I have no indication about cross-reference anymore.

jayeff:
Hi @tkellen,

I just tested the stylesheet, did some minor changes and rebased my branch.

Tested the print stylesheet in Chrome 44 beta on OSX, Firefox 38 on OSX, Safari 8 on OSX, IE11 on Win8.1, Chrome 43 on Win8.1 and Firefox 39 on Win8.1. The print style does look good for me in these browsers.

One issue is that IE11 and Firefox (both on Win+OSX) will not print the `{json:api}` Logo/Background-Image on the start page. Is this an issue to be fixed before merging or only a minor inconvenience?

ethanresnick:
@jayeff Hey! Sorry for the delay with this. It looks great!

I'd love to get it merged now if you're up for making a few small tweaks. I'll leave comments inline.

stylesheets/print.scss
```diff
@@ -0,0 +1,52 @@
+@page {
+  margin-top: 1.5cm;
+  margin-right: 1cm;
+  margin-left: 1cm;
+  margin-bottom: 2cm;
+}
+
+body {
+  font: 11pt Georgia, "Times New Roman", Times, serif;
```
ethanresnick: By using the `font` shorthand here, I think you're inadvertently overriding some styles from the base (all.css) stylesheet, like `line-height`. As a result, bullet points are a bit too close together (and even overlapping when they have inline code). Can you replace this with separate `font-size` and `font-family` declarations.

stylesheets/print.scss
```diff
@@ -0,0 +1,52 @@
...
+}
+
+nav, footer, div.sidebar {
```
ethanresnick: Given that the specification is _so long_ (40 pages when I try to print it), I think it would be quite helpful to have a table of contents at the top. This would help orient new readers to the scope of the spec, but also give readers who want to jump to a specific section a way to see roughly how far in the document their desired section is and what the sections around it are called. In others words, I'm saying I think I'd leave `div.sidebar` visible, and maybe format it to be as table of contents-y as possible.

~~If you wanted to go the extra mile, you could follow the headings in the `div.sidebar` table of contents with their page number! (See [here](http://stackoverflow.com/questions/13465747/how-do-i-add-a-page-number-to-a-link-to-an-anchor).) That would be incredibly useful.~~ Edit: so browsers don't support the page numbers yet, but having the sections listed at the top would still be useful, I think.

stylesheets/print.scss
ethanresnick, line 66: This seems good to me!

jayeff:
Hi Ethan,

sorry for the late reply—I must have missed the notifications from August 28th. It's good that you temporarily closed this PR last weekend so that it popped up on my radar again :smile: 

The branch is rebased on the latest `gh_pages`, I fixed the line-height issue (nicely spotted btw) and added a commit which adds the table of contents (unfortunately no page-numbers). If everything is fine I can squash the commits later.

Johannes

ethanresnick:
Thanks Johannes!

Just one thing: it looks like your compiled stylesheet has a stray rule:

```
div.sidebar nav.document-outline:before {
    content: "111";
    font-size: 18px; }
```

which I'm guessing is left over from some testing you were doing, since it only shows up in the compiled version and not the raw scss file.

If you remove that (or just recompile) and then squash the commits down, I think this is ready to merge!

jayeff:
Jep, missed the last recompilation. Branch is updated.

ethanresnick:
Awesome, merged!

jayeff:
Cool. Thanks.

PR #761, opened 2015-06-10
put normative statements into json-api-formatted file

bobholt:
Per @tkellen, this is a json-api-formatted file consisting of all normative statements within http://jsonapi.org/format/

### DIFF
--- a/normative-statements.json
+++ b/normative-statements.json
@@ -1,0 +1,2695 @@
+{
+  "jsonapi": {
+    "version": "1.0"
+  },
+  "data": [
+    {
+      "id": "content-negotiation",
+      "type": "sections",
+      "attributes": {
+        "title": "Content Negotiation"
+      },
+      "links": {
+        "self": "http://jsonapi.org/format/#content-negotiation"
+      },
+      "relationships": {
+        "statements": {
+          "data": [
+            { "id": "request-content-type", "type": "normative-statements" },
+            { "id": "request-accept", "type": "normative-statements" },
+            { "id": "response-ignore-parameters", "type": "normative-statements" },
+            { "id": "response-content-type", "type": "normative-statements"},
+            { "id": "response-unsupported-media-type", "type": "normative-statements"},
+            { "id": "response-not-acceptable", "type": "normative-statements"}
+          ]
+        }
+      }
+    },
+    {
+      "id": "document-structure",
+      "type": "sections",
+      "attributes": {
+        "title": "Document Structure"
+      },
+      "links": {
+        "self": "http://jsonapi.org/format/#document-structure"
+      },
+      "relationships": {
+        "statements": {
+          "data": [
+            { "id": "additional-members", "type": "normative-statements" },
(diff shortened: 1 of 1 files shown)


normative-statements.json
```diff
@@ -0,0 +1,2319 @@
...
+      "id": "content-negotiation",
+      "type": "sections",
+      "attributes": {
+        "title": "Content Negotiation"
+      },
+      "links": {
+        "self": "http://jsonapi.org/format/#content-negotiation"
+      },
+      "relationships": {
+        "statements": [
```
ksimka: > A "relationship object" MUST contain at least one of the following: links, data, meta

Missing "data" field.

normative-statements.json
```diff
@@ -0,0 +1,2319 @@
...
+    }
+  ],
+  "included": [
+    {
+      "id": "request-content-type",
+      "type": "normative-statements",
+      "attributes": {
+        "level": "MUST",
+        "description": "Clients **MUST** send all JSON API data in request documents with the header `Content-Type: application/vnd.api+json` without any media type parameters.",
+        "relationships": {
```
ksimka: Take "relationships" (here and further) outside of "attributes".

bobholt:
Nice catches, @ksimka.

ethanresnick:
+1. Awesome work @bobholt!

dgeb:
Thanks so much @bobholt!

I wonder if we should try to add non-normative statements and examples as well, so that the entire spec could be generated from this kind of data. Otherwise, I am concerned that either this data or the generated spec could get out of sync.

(To be clear: I know you've worked really hard on this already @bobholt - and I'm not suggesting that you take this on as well).

bobholt:
@dgeb: @tkellen and I have talked about this, too. While I originally was concerned that doing something like that might come out sounding like it was written by a space alien, in this form, it might make sense, though the example resources might still be gross. 

Other extensions to this could include: 
- sub-sections that have a `parent` relationship pointing to the parent section (e.g. a "Top Level" section would have a parent relationship pointing to the "Document Structure"). This raises the question about whether the sections, subsections, or both should have relationships with the statements within the document
- categories or tags such as "Request," "Response," "Document Structure," "Server Processing". I originally started with "client-requirements," "server-requirements" types, but there was too much overlap. A taxonomy like this would aid implementers in grouping like requirements across sections of the specification.

tkellen:
My current thinking is that we can parse this into a hash that can be used as template strings in format.md. I am buried right now, but I'd like to take a stab at this soon. In the interim, I'd like to land this, commit to keeping it in sync, and add some mention to it on the implementations page so people can use it until we're done with our generalized integration suite.

bobholt:
This has been updated based on #763.

tkellen:
Going to get this templated into the format doc tonight.

tkellen:
Working through this this morning. Your attention to detail is incredible @bobholt. I'm halfway through and haven't found a single error.

PR #763, opened 2015-06-10
make normative statements more useful in isolation

bobholt:
There is probably more than can be done, but here are a few nits I picked.
- remove pronouns that become ambiguous when statements are used in isolation
- break up compound statements


tkellen:
:+1: thank you @bobholt. Will you be updating the normative statements (in the json file) as well?

bobholt:
I will, should either land @tkellen.

PR #765, opened 2015-06-11
Examples for Sparse Fieldsets are targeting '/articles' collection, an array must be returned

joananeves:


ksimka:
:+1:

dgeb:
Good catch - thanks @joananeves!

PR #768, opened 2015-06-12
break up statements on acceptable 200 responses for resources

bobholt:
Break theses statements up for ease of understanding and testing.

Consider moving the `null` case to fetching relationships, as that seems to be the only place it could occur.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -709,15 +709,6 @@ Content-Type: application/vnd.api+json
 If an individual resource exists, a server **MUST** respond to a successful
 request to fetch that resource with a [resource object][resource objects].
 
-If the requested URL is one that might correspond to a single resource but
-does not currently, a server **MUST** respond to a successful request to fetch
-that resource with `null` provided as primary data for the response document.
-
-> Note: Consider, for example, a request to fetch a to-one related resource link.
-This request would respond with `null` when the relationship is empty (such that
-the link is corresponding to no resources) but with the single related resource's
-resource object otherwise.
-
 For example, a `GET` request to an individual article could return:
 
 ```http
@@ -742,21 +733,6 @@ Content-Type: application/vnd.api+json
       }
     }
   }
-}
-```
-
-If the above article's author is missing, then a `GET` request to that related
-resource would return:
-
-```http
-HTTP/1.1 200 OK
-Content-Type: application/vnd.api+json
-
-{
-  "links": {
-    "self": "http://example.com/articles/1/author"
-  },
-  "data": null
 }
 ```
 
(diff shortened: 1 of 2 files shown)


tkellen:
I'm :+1: to moving that null case into the fetch relationships section.

bobholt:
Note there is an existing statement [there](http://jsonapi.org/format/#fetching-relationships-responses-404) that could be replaced:

> If a relationship link URL exists but the relationship is empty, then `200 OK` **MUST** be returned, as described above.

tkellen:
double +1 to repurposing that.

bobholt:
Took a swing at it. Feel free to tweak/trash.

dgeb:
I feel like we lose clarity by moving the `null` case to the 404 section, especially since a 404 is not appropriate for the `null` case.

I'd rather turn the MUST in the 404 section into a non-normative note to eliminate the repetition of the same normative requirement.

tkellen:
Mind updating this so we can land @bobholt?

bobholt:
I guess I'm unsure how to proceed. I moved the `null` case to the relationship section, but left it in `200 OK`. I added the collection empty array case explicitly in the relationship section for symmetry. I left the `404` section alone.

I'll make the existing `404` statement non-normative, and we can take any further changes from there.

tkellen:
I think this looks good. @dgeb can you confirm? I think you might've been looking at a collapsed diff when you made your last comment.

format/index.md
```diff
@@ -809,6 +785,16 @@ value for [resource linkage], as described above for
...
+If the requested URL corresponds to a to-one relationship, but a resource does
+not currently exist for that relationship, a server **MUST** respond to a
+successful fetch request with `null` provided as primary data for the response
+document.
+
+Likewise, if the requested URL corresponds to a to-many relationship, but no
+resources currently exist for that relationship, a server **MUST** respond to
+a successful fetch request with an empty array (`[]`) provided as primary data
+for the response document.
+
```
dgeb, line 797: These two paragraphs are covered entirely by the paragraph above that reads:

>  The primary data in the response document **MUST** match the appropriate value for [resource linkage], as described above for [relationship objects][relationships].

format/index.md
```diff
@@ -887,8 +873,8 @@ a relationship link URL that does not exist.
 does not exist. For example, when `/articles/1` does not exist, request to
 `/articles/1/relationships/tags` returns `404 Not Found`.
 
-If a relationship link URL exists but the relationship is empty, then
-`200 OK` **MUST** be returned, as described above.
+> Note: If a relationship link URL exists but the relationship is empty, then
+`200 OK` will be returned, as described above.
```
dgeb, line 877: I'm good with making this non-normative so that the normative statement is in one place :+1:

format/index.md
```diff
@@ -745,21 +736,6 @@ Content-Type: application/vnd.api+json
...
-Content-Type: application/vnd.api+json
-
-{
-  "links": {
-    "self": "http://example.com/articles/1/author"
-  },
-  "data": null
-}
-```
-
```
dgeb, line 762: I believe that the normative mention of the possible `null` response here is appropriate. It is not redundant with the "fetching relationships" section, since this section covers "fetching resources". These sections clarify that `null` is an appropriate response at a related resource URL (e.g. `/articles/1/author`).

PR #769, opened 2015-06-12
remove unnecessary section attribute

bobholt:
- `section` attribute was left over from previous version
- replaced by `section` relationship


tkellen:
:+1:

PR #770, opened 2015-06-12
Reserve 'group' keyword for grouping

masterspambot:
The keyword `group` should be reserved for server side grouping strategies. We also should start discussion about grouping response format or leave it standard agnostic.


tkellen:
Hey @masterspambot! I'm open to adding something like this, but I don't think we'll see it landing in the very near future. You can see our roadmap [here](http://discuss.jsonapi.org/t/json-api-weekly-meeting-june-22nd-2015/23).

PR #783, opened 2015-06-22
add discourse forum to about page

tkellen:


dgeb:
Also removes Google Group link :+1:

ethanresnick:
@dgeb @tkellen Now that we've removed the Google Groups link from the site, we should probably turn off posting there (after adding a final post linking to discourse), right? 

cc @steveklabnik

steveklabnik:
yeah, in theory. I mean, people haven't been posting there, so I don't see the reason to change the setting anyway.

Issue #772, opened 2015-06-17
Outline cropped on smaller viewports


jgwhite:
The fixed-position document outline is super useful but unfortunately gets cut off on smaller viewports.

![screen_shot_2015-06-17_at_17_58_52](https://cloud.githubusercontent.com/assets/34030/8213571/a8d8ca5c-151a-11e5-8c21-3e972818ac46.png)

Tried introducing `overflow: auto` when the sidebar is fixed but this has some unfortunate side effects such as double-scrolling and over-running the footer.

@wvteijlingen any workarounds spring to mind?

wvteijlingen:
Ah yes, this has been pointed out before. The quickest solution that springs to mind right now is to create a breakpoint on viewport height, and forego the fixed position below a certain height. Otherwise we could fix it with scrolling overflow, but as you said, that creates some ugly scrolling.

I'm a bit busy now but I'll see if I can whip up a fix in the next couple of days.

tkellen:
Thanks for the heads up! I'm closing in favor of #785.

PR #785, opened 2015-06-24
Added fix for sidebar overflowing small viewports

wvteijlingen:
This works around the issue in #772. When the window is too small to contain the sidebar, it is not fixed and stays at the top of the page.

It’s not the prettiest solution I must admit, but it does the job of preventing the overflow on small viewports.

### DIFF
--- a/javascripts/all.js
+++ b/javascripts/all.js
@@ -17,18 +17,30 @@ $(document).ready(function() {
 function fixElement(element, offset) {
     if(element.length == 0) return;
 
+    var $window = $(window);
+    var elementBottom = element.height() + offset;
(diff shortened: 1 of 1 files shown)


tkellen:
On my end this just makes the sidebar disappear at small sizes?

wvteijlingen:
Hmm, are you sure it's not just pinned to the top? You get the effect of it disappearing when you scroll to halfway and then resize the viewport, but really it just jumps to the top.

It's not the most beautiful solution, that's for sure...

tkellen:
Thanks @wvteijlingen, I'm going to merge #856 in favor of this.

Issue #745, opened 2015-06-04
Links: a relation for actions


hhware:
I suggest to register with IANA a relation to identify actions, as AFAIK there is no existing relation for this yet. The motivation is that an API client should have a way to extract all actions from the list of links for a given resource, whether they are standardized or custom. 

Suggestion for relation name: `action`.

Ref: (8) of #684

ethanresnick:
Hey @hhware! Thanks for opening this. I'm sorry I'm just responding to it now. I've been working on other projects post 1.0, so I haven't been able to give much time to these conversations, but I do want to discuss this. 

First, I _really_ like the goal that you're trying to achieve here, which I take to be making all the non-standard actions that a resource supports discoverable. However, I'm not sure `links` is the right place for this.

Before explaining my position, though, I want to clarify how I'm using the term "resource" below, since I know it has (at least) two meanings—the JSON API one and the HTTP one—and our understandings of each meaning may be different. I apologize in advance if my use of this terminology is already familiar to you (or others), or if I'm misunderstanding the term's meaning, but I just want to make sure we're on the same page before we start.

Ok, so, when I talk about an HTTP resource, I mean a _concept_ (e.g. "the 1988 baseball world series winner" or "the movies Brad Pitt has starred in"). This concept is always a noun and is identified by a URI. So, the first resource might be identified by `http://example.com/word-series-winners/1988` and the second by `http://example.com/movies?actor=Brad%20Pitt`. The second resource would, in JSON API speak, be a "collection", since it represents a set of movies. But, from the HTTP perspective, that multiple movies are in the set doesn't matter; the set of movies is is still a single resource. Then, because these URIs/resources represent nouns, the verbs (GET, POST, etc.) are used to retrieve or act on them.

Now, back to an `action` link relation. Basically, I think:

First, that many "non-standard" actions can actually be reformulated as standard actions on new HTTP resources, if the resources are designed properly. Then, these actions can be advertised at the protocol level (i.e. with the `Allow` header) and with just a standard link to the new resource (on which the action exists), rather than a custom "action" link.

[...]

PR #792, opened 2015-06-29
Add recommendations for asynchronous processing

benoittgt:
Started a discussion on discourse about it  : http://discuss.jsonapi.org/t/long-running-jobs-or-asynchronous-processing/26

Maybe the status response have too much infos for this example.
Maybe "actions" should not be there.

Any feedbacks more than welcome.

### DIFF
--- a/recommendations/index.md
+++ b/recommendations/index.md
@@ -149,3 +149,36 @@ Although JSON API does not specify the format of date and time fields, it is
 recommended that servers align with ISO 8601. [This W3C
 NOTE](http://www.w3.org/TR/NOTE-datetime) provides an overview of the
 recommended formats.
+=======
+
+## Asynchronous Processing <a href="#asynchronous-processing" id="asynchronous-processing" class="headerlink"></a>
+
+Consider a situation when you need to create a resource and the operation takes long time to complete.
+
+The request **SHOULD** return a status `202 Accepted` with a link in Content-Location.
+
+```text
+HTTP/1.1 202 Accepted
+Content-Type: application/vnd.api+json
+Content-Location: https://example.com/photos/queue/5234
+```
+
+```json
+{
+   "data":{
+      "type":"queue job",
+      "id":"5234",
+      "attributes": { "status": "Pending request, waiting other process" },
+      "links": { "self": "/photos/queue/5234"}
+      }
+   }
+}
+```
+
+When job process is done the location given earlier **SHOULD** return `303 See other` with a link in Location.
+
+```text
+HTTP/1.1 303 See other
+Location: https://example.com/photos/4577
+Content-Type: application/vnd.api+json
+```


benoittgt:
Do I need to add some change ?

ryansch:
At first glance, this is really good stuff.

recommendations/index.md
```diff
@@ -149,3 +149,53 @@ Although JSON API does not specify the format of date and time fields, it is
...
+      },
+      {
+        "code": "invalid_format",
+        "field": "/photos/file01"
+      }
+      ]
+    },
+    "actions": {
+        "cancel": "/photos/queue/5234/cancel",
+        "delete": "/photos/queue/5234/delete"
```
ryansch: What would be the difference between `cancel` and `delete`?

ethanresnick:
@benoittgt Coming back to this... I'd like to see this merged, but it needs a few updates. I'll comment inline. Let me know if you're up for bringing this up to date.

recommendations/index.md
```diff
@@ -149,3 +149,53 @@ Although JSON API does not specify the format of date and time fields, it is
...
+        "code": "too_long",
+        "field": "/photos/title"
+      },
+      {
+        "code": "invalid_format",
+        "field": "/photos/file01"
+      }
+      ]
+    },
+    "actions": {
```
ethanresnick: First, this `"actions"` container should (temporarily) be removed. We're working on a standard way to support actions (see, e.g., #745) but, until then, I don't want to recommend something that will likely contradict our final solution.

recommendations/index.md
```diff
@@ -149,3 +149,53 @@ Although JSON API does not specify the format of date and time fields, it is
...
+
+## Asynchronous Processing <a href="#asynchronous-processing" id="asynchronous-processing" class="headerlink"></a>
+
+Consider a situation when you need to create a resource and the operation takes long time to complete.
+
+The request **SHOULD** return a status `202 Accepted` with a link in Header/Location.
+
+```text
+HTTP/1.1 202 Accepted
+Location: https://example.com/photos/queue/5234
```
ethanresnick: Fwiw, HTTP [does not define](https://tools.ietf.org/html/rfc7231#section-7.1.2) the meaning of the `Location` header for a `202` response, so it's not clear that this is appropriate here. (It only defines it for 201 and 3xx.) A more appropriate header might be `Link` with some relation. I'll need to think about this. 

//cc @steveklabnik

recommendations/index.md
```diff
@@ -149,3 +149,53 @@ Although JSON API does not specify the format of date and time fields, it is
...
+
+```json
+{
+  "links": {
+    "self": "/photos/queue/5234"
+  },
+    "data": {
+      "type": "queue job",
+      "id": "5234",
+      "status": "Process failed, see errors",
```
ethanresnick: `"status"` and `"errors"` need to be in an `"attributes"` bucket to comply with JSON API.

recommendations/index.md
benoittgt: A job canceled can be restarted. But it's ok for me if a job is only deleted.

recommendations/index.md
benoittgt: I'm ok with that

recommendations/index.md
benoittgt: Yep. Thanks.

recommendations/index.md
benoittgt: Yes you right. Maybe the link in `links` in enough ?

recommendations/index.md
```diff
@@ -149,3 +149,51 @@ Although JSON API does not specify the format of date and time fields, it is
...
+{
+   "links":{
+      "self":"/photos/queue/5234"
+   },
+   "data":{
+      "type":"queue job",
+      "id":"5234",
+      "attributes":{
+         "status":"Process failed, see errors",
+         "errors":[
```
ethanresnick: These errors shouldn't be here, I don't think, because if the server already knows that the request has errored, it should send back the error (not a 200).

recommendations/index.md
ethanresnick: Yes, a link in `links` could be enough, but I don't think a top-level `self` link is the right link. The top-level `"self"` is defined as the URI the request was made to. So, the top-level `self` (if present at all—and, in fact, it probably doesn't make sense to include in response to a POST) would be the URI the user sent the job to, not the URI of the queued job itself.

A resource-level `self` link could work, though. So the request would be `POST /tasks` and the response would be:

``` text
HTTP/1.1 202 Accepted
```

``` js
{
   "data": {
      "type": "queued",
      "id": "1234",
      "attributes": { "status": "pending..." },
      "links": { "self": "/photos/queue/5234"} // <- the link would go here
   }
}
```

Does that work for you?

It seems to very nicely comply with HTTP, which says that:

> The representation sent with [a 202] response ought to describe the request's current status and point to (or embed) a status monitor that can provide the user with an estimate of when the request will be fulfilled.

ethanresnick: Ok, I figured it out! The 202 response should contain a `Content-Location` header that points to the URI of the queue/status resource. From the [HTTP spec](https://tools.ietf.org/html/rfc7231#section-3.1.4.2):

> If Content-Location is included in a 2xx (Successful) response... and its field-value refers to a URI that differs from the effective request URI, then the origin server claims that the URI is an identifier for a different resource corresponding to the enclosed representation... [For a POST request, among others], such a Content-Location indicates that this payload is a representation reporting on the requested action's status and that the same report is available (for future access with GET) at the given URI."

Everything above I said about using a resource-level self link still applies, but `Content-Location` should be sent as well.

recommendations/index.md
benoittgt: Ok. Managing errors seems not related to this recommendation. Errors where discussed here https://github.com/json-api/json-api/pull/828. Seems enough for me.

recommendations/index.md
benoittgt: More than ok for me.

recommendations/index.md
```diff
@@ -149,3 +149,41 @@ Although JSON API does not specify the format of date and time fields, it is
 recommended that servers align with ISO 8601. [This W3C
 NOTE](http://www.w3.org/TR/NOTE-datetime) provides an overview of the
 recommended formats.
+=======
+
+## Asynchronous Processing <a href="#asynchronous-processing" id="asynchronous-processing" class="headerlink"></a>
+
+Consider a situation when you need to create a resource and the operation takes long time to complete.
+
+The request **SHOULD** return a status `202 Accepted` with a link in Header/Location.
```
ethanresnick: This should now read "... with a URI in the `Content-Location` header pointing to the resource describing the operation's  status".

recommendations/index.md
```diff
@@ -149,3 +149,41 @@ Although JSON API does not specify the format of date and time fields, it is
...
+Content-Type: application/vnd.api+json
+Content-Location: https://example.com/photos/queue/5234
+```
+
+The location gave **SHOULD** return `200 Ok` with the status of the selected job. The body **MAY** contains `"actions"` to manage the job.
+
+```json
+{
+   "links":{
+      "self":"/photos/queue/5234"
```
ethanresnick: As discussed, this top-level `self` link should go away

recommendations/index.md
```diff
@@ -149,3 +149,41 @@ Although JSON API does not specify the format of date and time fields, it is
...
+
+The request **SHOULD** return a status `202 Accepted` with a link in Header/Location.
+
+```text
+HTTP/1.1 202 Accepted
+Content-Type: application/vnd.api+json
+Content-Location: https://example.com/photos/queue/5234
+```
+
+The location gave **SHOULD** return `200 Ok` with the status of the selected job. The body **MAY** contains `"actions"` to manage the job.
```
ethanresnick: This text can now be removed too :)

recommendations/index.md
```diff
@@ -149,3 +149,41 @@ Although JSON API does not specify the format of date and time fields, it is
...
+      "type":"queue job",
+      "id":"5234",
+      "attributes": { "status": "Pending request, waiting other process" },
+      "links": { "self": "/photos/queue/5234"}
+      }
+   }
+}
+```
+
+When job process is done the location gave **SHOULD** return `303 See other` with the location of the data in Header/Location.
```
ethanresnick: s/gave/given earlier/

s/Header\/Location/its `Location` header.

ethanresnick:
@benoittgt Awesome! I just noted some small grammatical corrections too. Can you please make those, squash this all down to one commit, and then I'll merge it?

benoittgt:
Thanks Ethan for the time you took to review my PR. :sake:

ethanresnick:
Happy to :) This looks great now. Merged!

krainboltgreene:
Should this also mention that further requests should return `102`?

PR #808, opened 2015-07-13
Add shwoodard/jsonapi golang lib for serializing and deserializing

shwoodard:
Add goland jsonapi lib to list of implementations.  `shwoodard/jsonapi` uses struct tags rather than interfaces to construct and destruct [jsonapi.org](http://jsonapi.org) payloads.  The idea for adding a second library to the mix is to minimize the amount of code needed to read and produce jsonapi formatted json payloads.

See,

https://github.com/shwoodard/jsonapi
http://godoc.org/github.com/shwoodard/jsonapi


dgeb:
Great - thanks Sam! :+1:

PR #812, opened 2015-07-20
Update index.md with katharsis-servlet as a Java implementation

woonsan:
Hi,

I've cut a release of katharsis-servlet which is an adapter module for katharsis-core, so I think it might be good to add this module info as an implementation in the page.
This adapter module provides a generic adapter for traditional servlet/filter based Java web applications. This module is based on katharsis-core like katharsis-rs is, but katharsis-servlet provides servlet/filter bridges while katharsis-rs provides JAX-RS container filter bridge.

Please let me know if you have any questions.

Regards,

Woonsan


steveklabnik:
Thank you!

PR #813, opened 2015-07-24
Corrects typos in reserved characters

jfinkels:
Also clarifies difference between comma and period by using parallel syntax.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -589,9 +589,9 @@ first or last character:
 The following characters **MUST NOT** be used in member names:
 
 - U+002B PLUS SIGN, "+" _(used for ordering)_
-- U+002C COMMA, "," _(used separator for multiple relationship paths)_
-- U+002E PERIOD, "." _(used as relationship path separators)_
-- U+005B LEFT SQUARE BRACKET, "[" _(use in sparse fieldsets)_
+- U+002C COMMA, "," _(used as a separator between relationship paths)_
+- U+002E PERIOD, "." _(used as a separator within relationship paths)_
+- U+005B LEFT SQUARE BRACKET, "[" _(used in sparse fieldsets)_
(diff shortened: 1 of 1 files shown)


ethanresnick:
Thanks @jfinkels! 

I agree we should clean this language up. If we're on this path though, we might as well go all the way, so I've made an inline comment for how to make the language even shorter/clearer. Let me know what you think of that!

format/index.md
```diff
@@ -589,9 +589,9 @@ first or last character:
 The following characters **MUST NOT** be used in member names:
 
 - U+002B PLUS SIGN, "+" _(used for ordering)_
-- U+002C COMMA, "," _(used separator for multiple relationship paths)_
-- U+002E PERIOD, "." _(used as relationship path separators)_
-- U+005B LEFT SQUARE BRACKET, "[" _(use in sparse fieldsets)_
+- U+002C COMMA, "," _(used as a separator between relationship paths)_
```
ethanresnick, line 592: How about just "used to separate relationship paths"?

format/index.md
```diff
@@ -589,9 +589,9 @@ first or last character:
...
+- U+002E PERIOD, "." _(used as a separator within relationship paths)_
```
ethanresnick: How about "used as a separator within **a** relationship path"

format/index.md
jfinkels, line 592: To me, I think that's still ambiguous about separating _within_ versus separating _between_. If you do "used to separate relationship paths" for comma, I guess you should do "used to separate relationship names within relationship paths" for period.

ethanresnick, line 592: > To me, I think that's still ambiguous about separating within versus separating between.

Ok, fair. I've merged your language as is :)

jfinkels:
Cool!

PR #814, opened 2015-07-25
Change example dates to ISO8601

mitchlloyd:
When people are studying the specification, they tend to primarily focus on the examples because they are the quickest way to begin understanding the format. Although JSON API makes no recommendation about the format of the dates, using the Unix epoch format (seconds since 1970) in the examples can appear to be a recommendation.

These timestamps are not an ideal format for dates and times in JSON APIs. The numbers are ambiguous, especially considering that Javascript will parse _miliseconds_ since 1970 correctly when initializing a date, but will show the wrong date for seconds.

Here is some [StackOverflow discussion on the topic](http://stackoverflow.com/questions/10286204/the-right-json-date-format).

Even though the spec does not cover date formats, I'm proposing a change to this example so that if people do interpret the examples as a recommendation for date formatting, at least they will end up with a reasonable standard.

### DIFF
--- a/examples/index.md
+++ b/examples/index.md
@@ -26,8 +26,8 @@ Content-Type: application/vnd.api+json
     "attributes": {
       "title": "JSON API paints my bikeshed!",
       "body": "The shortest article. Ever.",
-      "created": 1432306588,
-      "updated": 1432306589
+      "created": "2015-05-22T14:56:29.000Z",
+      "updated": "2015-05-22T14:56:28.000Z"
     },
     "relationships": {
       "author": {


ethanresnick:
I 1000% agree with this. I think we should even go further and add a recommendation that dates be in ISO8601. (That could be on the recommendation page, if not a SHOULD in the text itself.)

I tweeted out JSON API the other day and [one response](https://twitter.com/JamesMoger/status/623828840457072640) I got was bemoaning this lack of date format standardization. Not standardizing doesn't seem to provide much, while it opens the door to all the client inconsistencies and bikeshedding that this spec is designed to prevent.

Thoughts @dgeb @tkellen @steveklabnik?

steveklabnik:
I am not totally opposed, I guess I'm just not sure how it's supposed to be
supported, since we can't invent a new data type... like, how do you
distinguish what is supposed to be a date?

Regardless, that's offtopic for this pr. I'm :+1:

tkellen:
:+1: to adding this as a recommendation.

mitchlloyd:
@steveklabnik I was thinking about the issue of Dates vs Date/Times. Looking at the [W3 specification for date and time formats](http://www.w3.org/TR/NOTE-datetime) it appears that ISO8601 does have formats for different specificity of dates:

```
Year:
      YYYY (eg 1997)
   Year and month:
      YYYY-MM (eg 1997-07)
   Complete date:
      YYYY-MM-DD (eg 1997-07-16)
   Complete date plus hours and minutes:
      YYYY-MM-DDThh:mmTZD (eg 1997-07-16T19:20+01:00)
   Complete date plus hours, minutes and seconds:
      YYYY-MM-DDThh:mm:ssTZD (eg 1997-07-16T19:20:30+01:00)
   Complete date plus hours, minutes, seconds and a decimal fraction of a
second
      YYYY-MM-DDThh:mm:ss.sTZD (eg 1997-07-16T19:20:30.45+01:00)
```

Perhaps the recommendation could be to follow ISO8601 for all date and date/time fields.

dgeb:
> Perhaps the recommendation could be to follow ISO8601 for all date and date/time fields.

That sounds good to me :+1:

tkellen:
@mitchlloyd would you mind opening a PR adding this?

mitchlloyd:
Will do! Happy to contribute :)

PR #819, opened 2015-07-27
Add JsonApiNet to implementation list for .NET

l8nite:
I wrote a library for easy deserialization of JSON API documents into C# classes with support for entity remapping, included resources, complex type parsing, etc. It is intended to be used in client applications written in C#.

https://github.com/l8nite/JsonApiNet 

This PR adds the library to the list of implementations for .NET


tkellen:
Thanks @l8nite!

PR #820, opened 2015-07-27
Add date and time format recommendation

mitchlloyd:
As requested by @tkellen in https://github.com/json-api/json-api/pull/814, here is a pull request adding a recommendation for date and time formats.

### DIFF
--- a/recommendations/index.md
+++ b/recommendations/index.md
@@ -142,3 +142,10 @@ that wish to support these clients are recommended to treat `POST` requests as
 `PATCH` requests if the client includes the `X-HTTP-Method-Override: PATCH`
 header. This allows clients that lack `PATCH` support to have their update
 requests honored, simply by adding the header.
+
+## Formatting Date and Time Fields <a href="#date-and-time-fields" id="date-and-time-fields" class="headerlink"></a>
+
+Although JSON API does not specify the format of date and time fields, it is
+recommended that servers align with ISO 8601. [This W3C
+NOTE](http://www.w3.org/TR/NOTE-datetime) provides an overview of the
+recommended formats.


recommendations/index.md
```diff
@@ -142,3 +142,10 @@ that wish to support these clients are recommended to treat `POST` requests as
 `PATCH` requests if the client includes the `X-HTTP-Method-Override: PATCH`
 header. This allows clients that lack `PATCH` support to have their update
 requests honored, simply by adding the header.
+
+## Formatting Date and Time Fields <a href="#date-and-time-fields" id="date-and-time-fields" class="headerlink"></a>
+
+Although JSON API does not specify the format of date and time fields, it is
+recommended that servers align with ISO 8601.  [This W3C
```
steveklabnik: the smallest of nits: one space after the period

mitchlloyd: My lower school English instruction failed me. I always try to use one space, but the muscle-memory is deep.

dgeb:
Looks good - thanks @mitchlloyd!

recommendations/index.md
steveklabnik: Hehe, I had to un-learn it too, I'm _just_ old enough to have learned to type on a typewriter...

ethanresnick:
I've been thinking more about how this recommendation should be worded over the past few days, and I think the current formulation has some problems. (Sorry I didn't have a chance to comment on the other issue before this was written up!)

Basically, the issue is that ISO 8601 allows a huge number of format variants, so, if we don't offer more precise guidance, I don't think we get the benefits that making a recommendation is designed to achieve.

ISO8601 supports:
- A "basic" and "extended" version, where the key difference is whether the representation includes separators (e.g. `19850412` vs `1985-04-12`)
- "Extended years", i.e. those that can't be represented by four integers, because they fall outside the range (-9999, 9999). But these are only supported by prior agreement of the parties. E.g. `+022000`, for my scifi game set in the year 22,000.
- Specifying a date at the precision of the week (as opposed to just year, month, day precisions), e.g. `1985-W15`, the 15th week of 1985.
- Specifying particular dates using in alternate formats. E.g. `2015-W01-2` refers to the second day of the first week of 2015. (Confusingly, that's actually a date _in 2014_ because the first calendar week of 2015 starts back in 2014; it's equivalent to `2014-12-30`.) Also, ISO8601 allows formats like `1985-102`, which denotes the 102nd day of 1985.
- Picking out times with _decimal_ fractions. So, `23,3`, which can also be written `23.3`, means 23 hours and 3 tenths of an hour (i.e. `23:18`). This can also apply at the minute or second level (e.g. `23:30,4` is equivalent to `23:30:24`, because the `,4` denotes 40% of a minute). The number of decimal digits used is to be agreed between the parties.
- Intervals, recurring intervals, etc.

As I see it, the point of specifying the format of date strings in JSON API would be that the user of a client library would only have to say "This attribute holds a date" and then the client library would automatically know how to convert the date string it gets from the server to an instance of the client language's `Date` object. (Think, in Ember Data, the use of `DS.attr('date')`.)

But, if we just say in JSON API that the date string has to be ISO8601 compliant, the client has to figure out which allowed ISO8601 format is in use before it can parse the string. This should be possible for it to do this without user intervention, though I'm not 100% sure. If it's not, every client would have to be explicitly configured for the API it's using—ick. 

But even if automatically parsing any ISO8601-compliant string is possible, it's not something most languages support out of the box. For example, Javascript's `Date.parse` supports only a ["simplified" version](https://people.mozilla.org/~jorendorff/es6-draft.html#sec-date-time-string-format) of the ISO8601 Extended format. PHP's `strtotime` can detect and parse most of the allowed ISO8601 formats, but not some, like the `YYYY-DDD` form. Java can parse any form, but only if the user specifies in advance which [formatting pattern](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html#ISO_LOCAL_DATE) the parser should use, which defeats the point.

So, I'm thinking a few things:
1. This one is pretty obvious, but I just want to make it explicit: we should be clear that our recommendation is only about specifying an _instant_ in time, not a time interval. So, even if we support something like the simple string `"2015"`, which ISO8601 allows, that isn't picking out the entire year 2015 (a time interval), but rather a single point in time that happens to be specified at a low precision; thus, the client should be able to substitute a more precise representation of that date without consequence (e.g. `2015-01-01T00:00:00`, which is what the default "convert to timestamp" would likely yield, but also something arbitrary like `2015-02-09`). If the user really wants an interval, they need a start and end date (or one date and a duration), which our recommendation shouldn't get into the formatting for.
2. We probably should pick a subset of the ISO 8601 formats to allow, per the discussion above.
3. We have to decide what max level of precision we want to support (e.g. are nanoseconds representable?) and how far into the future and past we want to support (i.e., how many integer places we allow for extended years). My feeling here is that the server, as the definitive source of data, should be able to use as high a precision as it wants, and be able to use as many extended year places as it wants. But we have to balance that with easy, out-of-the-box parsing, which, again, is a big part of the point.

I'll comment here later with more concrete ideas about how we might address the above issues when I have a sec (and when I've had time to look into date parsing in more languages). But, for now, I just wanted to throw these issues out there.

tkellen:
Thanks for the thorough review @ethanresnick. I'm supportive of being a bit more prescriptive in our recommendation. Looking forward to reading your PR!

steveklabnik:
I am not super sure that a subset of ISO 8601 is important, as I'm guessing your library would be able to parse of all of them. Maybe that's an incorrect assumption. IIRC, in Ruby, you can 

```
require 'time'
Time.iso8601("some date goes here")
```

dgeb:
I think we should examine the "basic" vs "extended" formats if they are redundant. At first glance, it seems like we should favor the extended format, which looks to be much more common and easy on the eyes.

I'm not sure that we should go much further than that with our ISO 8601 recommendation, although I'm open to discussion.

> we should be clear that our recommendation is only about specifying an instant in time, not a time interval.

If an API needs to represent time intervals, wouldn't it be consistent to recommend ISO 8601 for time intervals?

> We have to decide what max level of precision we want to support (e.g. are nanoseconds representable?) and how far into the future and past we want to support (i.e., how many integer places we allow for extended years).

This sounds like an implementation-specific concern. For instance, a scientific application might want to represent nanoseconds.

ethanresnick:
> I am not super sure that a subset of ISO 8601 is important, as I'm guessing your library would be able to parse of all of them.

That would be ideal, for sure. But even Ruby barfs on a few valid formats. E.g. any time with fractional hour or minute parts, whether separated by a `,` or a `.` doesn't seem to work with `DateTime.parse` or `DateTime.iso8601`. So the strings `"2001-04-04T23,5"` and `"2001-04-04T23.5"` both throw errors with `DateTime.iso8601` and are parsed incorrectly with `DateTime.parse`.

Moreover, Ruby seems to be one of the better languages for parsing ISO 8601 out of the box; most are much more limited, as I alluded to above. (E.g. Python, like Java, seems to require an [explicit format string](https://docs.python.org/2/library/datetime.html#strftime-and-strptime-behavior) unless you use a library.) Not that this should dictate our behavior too much, but if we can simplify the lives of client developers, then why not.

> At first glance, it seems like we should favor the extended format, which looks to be much more common and easy on the eyes.

:+1: 

> This sounds like an implementation-specific concern. For instance, a scientific application might want to represent nanoseconds.

Right. The problem is that ISO 8601 says that these things (i.e. the precision on fractional seconds and the number of places in extended years) have to be agreed on by the parties in advance. Otherwise, parsing and comparing may not work as expected. That's why JS, for instance, says that the extended years are alway six digits and the fractional seconds are always milliseconds.

ethanresnick:
> If an API needs to represent time intervals, wouldn't it be consistent to recommend ISO 8601 for time intervals?

Yeah, that would make some sense. The only reason I'm hesitant is that I've never actually seen an iso8601-formatted interval is used in practice, and I have no clue what the real world interoperability is like. Have you guys seen people do this?

ethanresnick:
I'm gonna reopen this, just so we remember to come back to it.

ethanresnick:
Jumping back to this: I think we should just recommend Javascript's format (a simplified version of ISO8601's extended format). That format addresses all of my [earlier concerns](https://github.com/json-api/json-api/pull/820#issuecomment-125433740). It:
- removes some of the weirder ways for specifying dates, like with the week indicator (`2015-W01`) and the day-of-year indicator (`2015-102`). Therefore, the remaining formats should be parseable by every common language out of the box.
- uses milliseconds as the fractional second unit and 6 digits as the size of extended years. Again, even though these are implementation-specific concerns, ISO 8601 requires the parties to agree on these values in advance to ensure that both parties can parse/store the provided values. So I think we should recommend some values by default, but then specify an extension that the client can negotiate for that allows the server to specify more precise dates/bigger extended years.
- doesn't make a recommendation for intervals

**EDIT**: Note that Javascript's subset of ISO8601 also appears to be the syntax used by [XSD](http://www.w3.org/TR/xmlschema11-2/#dateTime), which is the schema commonly-used to annotate JSON-API dates. However, XSD allows infinite precision on fractional seconds and infinite size on extended years, but then says that a "minimally conformant" implementation [need only support](http://www.w3.org/TR/xmlschema11-2/#partial-implementation) milliseconds and six-digit fractional years. So that's another approach we could take.

I'm also thinking we should move this into the base specification with a **SHOULD**, rather than just making it a recommendation, as I do want clients to be able to depend on it.

Thoughts?

ethanresnick:
Ref https://tools.ietf.org/html/rfc3339

wimleers:
+1 for making a recommendation.

See https://www.drupal.org/node/2768651 for what Drupal core's REST API did and its rationale, and https://www.drupal.org/node/2955581 for date-only fields. 

Basically: RFC3339, except for date-only fields (because those are not covered by RFC3339): those use ISO8601.

This is also what Drupal's JSON API module uses as of https://www.drupal.org/project/jsonapi/issues/2929932.

Issue #795, opened 2015-06-29
Support for multiple operations in a single request


tkellen:
It has been widely requested that JSON-API support creating/editing/updating/deleting multiple records in a single request (#202, #205, #753, #536). We have attempted to paint this bikeshed repeatedly over the years. Providing an "official" solution, or solutions, to this problem is one of the primary goals for v1.1 of this specification.

There are several requirements for a solution:
- It should allow mixed actions in a single request (e.g. creating + updating).
- It should work for multiple types of resources (e.g. articles + tags) that would otherwise be accessed at different endpoints.
- Every request should be transactional (i.e. all operations should succeed or fail together).
- Additive to the base spec (probably as an extension).

Here are a couple use-cases that test these requirements:
- Create two articles, add one existing tag to both, create a new tag and add to both.
- Change the title of an article and update its tags: add one existing tag, create another, and remove another (without replacing the whole tag set).

**We would like the community to propose ideas for how to solve these use cases (or chime in to support solutions provided in the responses to follow).**

Assume the following data is present

```
GET /articles

{
  "data": [{
    "id": "article-1",
    "type": "articles",
    "attributes": {
      "title": "the first article",
      "content": "some content for an article",
    },
    "relationships": {
      "tags": {
        "data": [
          { "id": "tag-1", "type": "tags" }
        ]
      }
    }
  }]
}
```

```
GET /tags

{
  "data": [{
    "id": "tag-1",
    "type": "tags"
    "attributes": {
      "name": "one"
    }
  }, {
    "id": "tag-2",
    "type": "tags"
    "attributes": {
      "name": "two"
    }
  }]
}
```

dgeb:
## Potential Solution: Operations Extension

One solution to this problem is to represent each operation as an entity in a request document. Operations could be transmitted in an array, since ordering is often significant. Similarly, responses to each operation could be returned in an array that parallels the operations in a request.

This solution is very similar to the experimental [JSON Patch extension](http://jsonapi.org/extensions/jsonpatch/) with the exception that it is additive to the base spec.

Operations could include the following members:
- `path` - the path to the applicable resource, if different from the request path
- `action` - `add` / `remove` / `update` actions that map to HTTP methods
- `data` - the data normally sent as top-level `data`

Operations would be sent in a top-level `operations` array.

Responses could also be returned in a top-level `operations` array, in which the `status` member could be added to each operation.

Requests would be sent with the `PATCH` method together with the TBD extension requirement.

[...]

PR #559, opened 2015-04-18
Meta in linkages

hhware:
Implements suggestion from https://github.com/json-api/json-api/pull/431#issuecomment-92938203. Please see #415 and #431 for more details.

I would like to request to consider this for 1.0.

PR #824, opened 2015-07-28
Revision Added as a Top-Level Member of Resource

scott-mcdonald:
…0 for versioning of resources.

Signed-off-by: Scott McDonald smcdonald@movietickets.com

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -159,6 +159,7 @@ the client and represents a new resource to be created on the server.
 
 In addition, a resource object **MAY** contain any of these top-level members:
 
+* `revision`: a [revision string][revision] representing a specific version of an individual resource uniquely identified by `type` and `id`.
 * `attributes`: an [attributes object][attributes] representing some of the resource's data.
 * `relationships`: a [relationships object][relationships] describing relationships between
  the resource and other JSON API resources.
@@ -173,6 +174,7 @@ Here's how an article (i.e. a resource of type "articles") might appear in a doc
 {
   "type": "articles",
   "id": "1",
+  "revision": "1111111111",
   "attributes": {
     "title": "Rails is Omakase"
   },
@@ -208,7 +210,33 @@ The values of `type` members **MUST** adhere to the same constraints as
 can be either plural or singular. However, the same value should be used
 consistently throughout an implementation.
 
-#### <a href="#document-resource-object-fields" id="document-resource-object-fields" class="headerlink"></a> Fields
+#### Revision <a href="#document-resource-object-revision" id="document-resource-object-revision" class="headerlink"></a>
+
+The **OPTIONAL** `revision` member is an *opaque* string identifier assigned by
+the web server to represent a specific version of a resource. This mechanism
+enables clients or servers to determine if two resources with the same `type`
+and `id` values are the same or different versions by comparing the `revision`
+value for equality.
+
+The purpose of the `revision` member is to enable *optimistic concurrency control*
+on resource updates. Therefore if a server supplies a `revision` member in a `GET`
+request, a client **SHOULD** pass the same `revision` member in a `PATCH` request
+when updating the resource.
+
+Another purpose of the `revision` member is to enable *caching* at the
+*fine-grained resource level* for clients or servers if desired. These caching
+mechanisms can leverage the ability to determine if two resources with the same
+`type` and `id` values are the same version by comparing the respective `revision`
+values for equality and apply caching policy as needed.
+
+The `revision` value of `"*"` is reserved for future use.
+
+> Note: JSON API is agnostic about the `revision` implementation used by a server. 
+> Effective revision implementation strategies include (but are not limited to):
+> using a hash function, using an incrementing counter, using a timestamp,
+> or a database solution like SQL Server's `rowversion` column type.
+
+#### Fields <a href="#document-resource-object-fields" id="document-resource-object-fields" class="headerlink"></a>
 
 A resource object's [attributes] and its [relationships] are collectively called
 its "[fields]".
@@ -1297,6 +1325,8 @@ primary data, the same request URL can be used for updates.
 
 The `PATCH` request **MUST** include a single [resource object][resource objects] as primary data.
 The [resource object][resource objects] **MUST** contain `type` and `id` members.
+The [resource object][resource objects] **SHOULD** contain `revision` member
+if the server supplied the **OPTIONAL** `revision` member in a `GET` request.
 
 For example:
 
@@ -1309,6 +1339,7 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
+    "revision": "1111111111",
     "attributes": {
       "title": "To TDD or Not"
     }
@@ -1338,6 +1369,7 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
+    "revision": "1111111111",
     "attributes": {
       "title": "To TDD or Not",
       "text": "TLDR; It's complicated... but check your test coverage regardless."
@@ -1371,6 +1403,7 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
+    "revision": "1111111111",
     "relationships": {
       "author": {
         "data": { "type": "people", "id": "1" }
@@ -1392,6 +1425,7 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
+    "revision": "1111111111",
     "relationships": {
       "tags": {
         "data": [
@@ -1462,6 +1496,13 @@ constraints (such as a uniqueness constraint on a property other than `id`).
 
 A server **MUST** return `409 Conflict` when processing a `PATCH` request in
 which the resource object's `type` and `id` do not match the server's endpoint.
+
+A server **SHOULD** return `409 Conflict` when processing a `PATCH` request to
+update a resource if there is a mismatched `revision` between client and server
+where the server was *unable to resolve* the `revision` mismatch on resource update.
+
+> Note: Only applicable if a server supplied the **OPTIONAL** `revision` member
+> in the response of a `GET` request.
 
 A server **SHOULD** include error details and provide enough information to
 recognize the source of the conflict.
@@ -1781,6 +1822,7 @@ An error object **MAY** have the following members:
   error.
 
 [resource objects]: #document-resource-objects
+[revision]: #document-resource-object-revision
 [attributes]: #document-resource-object-attributes
 [relationships]: #document-resource-object-relationships
 [fields]: #document-resource-object-fields
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -159,7 +159,6 @@ the client and represents a new resource to be created on the server.
 
 In addition, a resource object **MAY** contain any of these top-level members:
 
-* `revision`: a [revision string][revision] representing a specific version of an individual resource uniquely identified by `type` and `id`.
 * `attributes`: an [attributes object][attributes] representing some of the resource's data.
 * `relationships`: a [relationships object][relationships] describing relationships between
  the resource and other JSON API resources.
@@ -174,7 +173,6 @@ Here's how an article (i.e. a resource of type "articles") might appear in a doc
 {
   "type": "articles",
   "id": "1",
-  "revision": "1111111111",
   "attributes": {
     "title": "Rails is Omakase"
   },
@@ -210,27 +208,7 @@ The values of `type` members **MUST** adhere to the same constraints as
 can be either plural or singular. However, the same value should be used
 consistently throughout an implementation.
 
-#### Revision <a href="#document-resource-object-revision" id="document-resource-object-revision" class="headerlink"></a>
-
-The **OPTIONAL** `revision` member is an *opaque* string identifier assigned by
-the web server to represent a specific version of a resource. If the resource
-is *updated*, a new and different `revision` is assigned. This mechanism
-enables clients or servers to determine if two resources with the same `type`
-and `id` values are the same or different versions by comparing the `revision`
-value for equality.
-
-The purpose of the `revision` member is to enable *optimistic concurrency control*
-on resource updates. Therefore if a server supplies a `revision` member in a `GET`
-request, a client **SHOULD** pass the same `revision` member in a `PATCH` request
-when updating the resource.
-
-Another purpose of the `revision` member is to enable *caching* at the
-*fine-grained resource level* for clients or servers if desired. These caching
-mechanisms can leverage the ability to determine if two resources with the same
-`type` and `id` values are the same version by comparing the respective `revision`
-values for equality and apply caching policy as needed.
-
-#### Fields <a href="#document-resource-object-fields" id="document-resource-object-fields" class="headerlink"></a>
+#### <a href="#document-resource-object-fields" id="document-resource-object-fields" class="headerlink"></a> Fields
 
 A resource object's [attributes] and its [relationships] are collectively called
 its "[fields]".
@@ -1319,8 +1297,6 @@ primary data, the same request URL can be used for updates.
 
 The `PATCH` request **MUST** include a single [resource object][resource objects] as primary data.
 The [resource object][resource objects] **MUST** contain `type` and `id` members.
-The [resource object][resource objects] **SHOULD** contain `revision` member
-if the server supplied the **OPTIONAL** `revision` member in a `GET` request.
 
 For example:
 
@@ -1333,7 +1309,6 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
-    "revision": "1111111111",
     "attributes": {
       "title": "To TDD or Not"
     }
@@ -1363,7 +1338,6 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
-    "revision": "1111111111",
     "attributes": {
       "title": "To TDD or Not",
       "text": "TLDR; It's complicated... but check your test coverage regardless."
@@ -1397,7 +1371,6 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
-    "revision": "1111111111",
     "relationships": {
       "author": {
         "data": { "type": "people", "id": "1" }
@@ -1419,7 +1392,6 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
-    "revision": "1111111111",
     "relationships": {
       "tags": {
         "data": [
@@ -1490,14 +1462,6 @@ constraints (such as a uniqueness constraint on a property other than `id`).
 
 A server **MUST** return `409 Conflict` when processing a `PATCH` request in
 which the resource object's `type` and `id` do not match the server's endpoint.
-
-A server **SHOULD** return `409 Conflict` when processing a `PATCH` request in
-which the resource object's `revision` does not match the server's endpoint
-and the server was *unable to resolve* the `revision` mismatch for updating
-resources with the same `type` and `id` but mismatched `revision` members.
-
-> Note: Only applicable if a server supplied the **OPTIONAL** `revision` member
-> in the response of a `GET` request.
 
 A server **SHOULD** include error details and provide enough information to
 recognize the source of the conflict.
@@ -1817,7 +1781,6 @@ An error object **MAY** have the following members:
   error.
 
 [resource objects]: #document-resource-objects
-[revision]: #document-resource-object-revision
 [attributes]: #document-resource-object-attributes
 [relationships]: #document-resource-object-relationships
 [fields]: #document-resource-object-fields
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -159,6 +159,7 @@ the client and represents a new resource to be created on the server.
 
 In addition, a resource object **MAY** contain any of these top-level members:
 
+* `revision`: a [revision string][revision] representing a specific version of an individual resource uniquely identified by `type` and `id`.
 * `attributes`: an [attributes object][attributes] representing some of the resource's data.
 * `relationships`: a [relationships object][relationships] describing relationships between
  the resource and other JSON API resources.
@@ -173,6 +174,7 @@ Here's how an article (i.e. a resource of type "articles") might appear in a doc
 {
   "type": "articles",
   "id": "1",
+  "revision": "1111111111",
   "attributes": {
     "title": "Rails is Omakase"
   },
@@ -208,7 +210,27 @@ The values of `type` members **MUST** adhere to the same constraints as
 can be either plural or singular. However, the same value should be used
 consistently throughout an implementation.
 
-#### <a href="#document-resource-object-fields" id="document-resource-object-fields" class="headerlink"></a> Fields
+#### Revision <a href="#document-resource-object-revision" id="document-resource-object-revision" class="headerlink"></a>
+
+The **OPTIONAL** `revision` member is an *opaque* string identifier assigned by
+the web server to represent a specific version of a resource. If the resource
+is *updated*, a new and different `revision` is assigned. This mechanism
+enables clients or servers to determine if two resources with the same `type`
+and `id` values are the same or different versions by comparing the `revision`
+value for equality.
+
+The purpose of the `revision` member is to enable *optimistic concurrency control*
+on resource updates. Therefore if a server supplies a `revision` member in a `GET`
+request, a client **SHOULD** pass the same `revision` member in a `PATCH` request
+when updating the resource.
+
+Another purpose of the `revision` member is to enable *caching* at the
+*fine-grained resource level* for clients or servers if desired. These caching
+mechanisms can leverage the ability to determine if two resources with the same
+`type` and `id` values are the same version by comparing the respective `revision`
+values for equality and apply caching policy as needed.
+
+#### Fields <a href="#document-resource-object-fields" id="document-resource-object-fields" class="headerlink"></a>
 
 A resource object's [attributes] and its [relationships] are collectively called
 its "[fields]".
@@ -1297,6 +1319,8 @@ primary data, the same request URL can be used for updates.
 
 The `PATCH` request **MUST** include a single [resource object][resource objects] as primary data.
 The [resource object][resource objects] **MUST** contain `type` and `id` members.
+The [resource object][resource objects] **SHOULD** contain `revision` member
+if the server supplied the **OPTIONAL** `revision` member in a `GET` request.
 
 For example:
 
@@ -1309,6 +1333,7 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
+    "revision": "1111111111",
     "attributes": {
       "title": "To TDD or Not"
     }
@@ -1338,6 +1363,7 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
+    "revision": "1111111111",
     "attributes": {
       "title": "To TDD or Not",
       "text": "TLDR; It's complicated... but check your test coverage regardless."
@@ -1371,6 +1397,7 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
+    "revision": "1111111111",
     "relationships": {
       "author": {
         "data": { "type": "people", "id": "1" }
@@ -1392,6 +1419,7 @@ Accept: application/vnd.api+json
   "data": {
     "type": "articles",
     "id": "1",
+    "revision": "1111111111",
     "relationships": {
       "tags": {
         "data": [
@@ -1462,6 +1490,14 @@ constraints (such as a uniqueness constraint on a property other than `id`).
 
 A server **MUST** return `409 Conflict` when processing a `PATCH` request in
 which the resource object's `type` and `id` do not match the server's endpoint.
+
+A server **SHOULD** return `409 Conflict` when processing a `PATCH` request in
+which the resource object's `revision` does not match the server's endpoint
+and the server was *unable to resolve* the `revision` mismatch for updating
+resources with the same `type` and `id` but mismatched `revision` members.
+
+> Note: Only applicable if a server supplied the **OPTIONAL** `revision` member
+> in the response of a `GET` request.
 
 A server **SHOULD** include error details and provide enough information to
 recognize the source of the conflict.
@@ -1781,6 +1817,7 @@ An error object **MAY** have the following members:
   error.
 
 [resource objects]: #document-resource-objects
+[revision]: #document-resource-object-revision
 [attributes]: #document-resource-object-attributes
 [relationships]: #document-resource-object-relationships
 [fields]: #document-resource-object-fields
--- a/index.md
+++ b/index.md
@@ -27,6 +27,7 @@ Here's an example response from a blog that implements JSON API:
   "data": [{
     "type": "articles",
     "id": "1",
+    "revision": "1234567890",
     "attributes": {
       "title": "JSON API paints my bikeshed!"
     },
@@ -56,6 +57,7 @@ Here's an example response from a blog that implements JSON API:
   "included": [{
     "type": "people",
     "id": "9",
+    "revision": "9999999999",
     "attributes": {
       "first-name": "Dan",
       "last-name": "Gebhardt",
@@ -67,6 +69,7 @@ Here's an example response from a blog that implements JSON API:
   }, {
     "type": "comments",
     "id": "5",
+    "revision": "5555555555",
     "attributes": {
       "body": "First!"
     },
@@ -81,6 +84,7 @@ Here's an example response from a blog that implements JSON API:
   }, {
     "type": "comments",
     "id": "12",
+    "revision": "1212121212",
     "attributes": {
       "body": "I like XML better"
     },


cziegenberg:
I'd say that the client MUST send the revision for updating a resource if sent with the GET.

The server uses the revision to ensure that updates are send for the right resource revision, so IMHO it's required in all following updates - otherwise it doesn't make sense at all to use revisions.

Anither point is what's the content of the 409 response - the current or a merged version of the resource?

scott-mcdonald:
@ziege I agree with you in that originally I wanted to say the something like: If a `revision` is supplied in a `GET` request, then `revision` **MUST** be passed in a `PATCH` or `PUT` request. However @tkellen pointed out the following:

> we cannot say that PATCH or PUT MUST include revision--it would be backwards incompatible with 1.0 of the spec. This change must be wholly additive.

I agree if the next version is 1.1 it should be backwards compatible with 1.0 so I changed **MUST** to either **MAY** or **SHOULD**. If and when this standard goes to 2.0 then we should change the above to **MUST**.

To what the content of the response should be for a `409 Conflict` I will leave open to discussion. Currently the specification states the following for the `409 Conflict`:

> A server SHOULD include error details and provide enough information to recognize the source of the conflict.

I read that as the specification is agnostic on what to return as a response.

> Sidebar: My particular implementation detail for this scenario is if the server returns a `409 Conflict`, the client must then do a **new** `GET` request, reapply modifications, and try the `PATCH` or `PUT` request again up to a maximum number of attempts.

Another sidebar: Can someone explain the following line in the specification in the _Responses_ section for updating resources:

> Quote: A server MUST return 409 Conflict when processing a PATCH request in which the resource object's type and id do not match the server's endpoint.

I don't understand that statement. The `type` and `id` are just strings and if they do not exist in the system shouldn't the response be a `404 Not Found`. Or did I misinterpret?

ethanresnick:
> I agree if the next version is 1.1 it should be backwards compatible with 1.0 so I changed MUST to either MAY or SHOULD. If and when this standard goes to 2.0 then we should change the above to MUST.

By my reading, a MUST here actually would be backwards compatible, because it's a MUST that's _conditional_ on getting a `revision` key in the `GET`, which no 1.0 servers will return.  But I may be over looking something…

I have a different concern with the current wording of the PR, though, which is that I don't think it's accurate to say that `revision` is part of the identifying characteristics of a resource in the way `(type, id)` are. (Therefore, among other things, I wouldn't put the description of `revision` in the identity section of the spec.)

Consider a request to a relationship endpoint. If `revision` were part of resource identity, I'd expect the client return something like the below:

```
GET /people/1/relationships/photos

{
  "data": [{
    "type": "photos", "id": "4", "revision": "abcd"
  }]
}
```

Likewise, I'd imagine that a `DELETE` request would have the same body as the above, with `revision` included. But I don't think that's what we want, even if the targeted resources are `revision`ed. Because the resources are in or should be removed from the relationship regardless of their present `revision`, right? Likewise, it doesn't seem like we want to expose a mechanism to `GET` the various `revisions` of a resource, or `DELETE` a particular revision.

So, really, the identifier is still `(type, id)` and `revision` is just an attribute summarizing the state at the time its retrieved. To make the HTTP analogy: Etag isn't part of resource identity; identity is only based on the URI, and Etag only matter for updates/caching.

I also still wonder if there's a way to make Etag and `revision` match up. For example, we could say that every revisioned resource's `self` uri has to include no other resources by default (or at least support a `?include=` option to remove any default includes) and has to give empty top-level `meta`. That might be overkill...but it at least seems possible.

dgeb:
>  I don't think it's accurate to say that revision is part of the identifying characteristics of a resource in the way (type, id) are. (Therefore, among other things, I wouldn't put the description of revision in the identity section of the spec.)

I agree with @ethanresnick here and share the concerns he's raised.

If we allow `revision` in resource identifier objects that represent relationships, then it MUST represent a revision of the relationship itself as opposed to a revision of the related resource. I believe the exception is when a related resource represents a relationship (say, a "membership" resource). 

There is a lot of subtlety and potential for confusion here.

scott-mcdonald:
So rethinking having `revision` defined as part of the resource identifier object:
- Resource: `type`="people" `id`="jdoe@acme.com" `revision`="1111"
- Resource: `type`="people" `id`="jdoe@acme.com" `revision`="2222"

Are they the same resource? Yes and No. They are the same resource but at different versions. For identity purposes though it is just `type` and `id` isn't it.

The final affirmation for me is for resource linkage purposes it doesn't make sense to have `revision` which is why I did not add `revision` in various resource examples in the specification.

Therefore I have come to the same conclusion as you guys and to quote @dgeb:

> There is a lot of subtlety and potential for confusion here.

Do not want that now do we. Therefore `revision` should be primarily documented in the resource object section not the resource identifier object section.

@dgeb What was your take on the verbiage of if the optional `revision` is supplied in a `GET` request, it **MUST** be passed in a PATCH or PUT request that was talked about in this thread. I prefer **MUST**, I like @ethanresnick thought process that this does not break 1.0 backwards compatible because it is optional in the `GET` request.

tkellen:
After further reflection, I'm cautiously supportive of the **MUST** requirement for including `revision` if the server is providing revisions. I agree with @dgeb that there is a lot of opportunity for confusion here. I believe we should move the non-normative explanations to the FAQ. Alternatively, I think we should have more normative statements around how revision should be used in conjunction with etags.

dgeb:
> @dgeb What was your take on the verbiage of if the optional revision is supplied in a GET request, it MUST be passed in a PATCH or PUT request that was talked about in this thread. I prefer MUST, I like @ethanresnick thought process that this does not break 1.0 backwards compatible because it is optional in the GET request.

I actually favor a MAY for including revision with PATCH. First of all, I don't think a GET should be a pre-requisite for PATCH. Also, some clients will want to force a change (say an update to an attribute) regardless of the resource's revision. By forcing the inclusion of `revision`, we'd also be forcing servers to reject PATCHes if that field doesn't match.

On that same note, the one MUST I would say should be in place is that the server MUST reject a PATCH request that includes `revision` if it doesn't match the current revision of the resource.

cziegenberg:
> First of all, I don't think a GET should be a pre-requisite for PATCH

Isn't there always a GET before you can PATCH anything?

> Also, some clients will want to force a change (say an update to an attribute) regardless of the resource's revision

I'd never allow a client to bypass the revision check. Do you have an example when this could be a requirement?

> On that same note, the one MUST I would say should be in place is that the server MUST reject a PATCH request that includes revision if it doesn't match the current revision of the resource.

Instead of a MUST I'd allow the server to decide if the request is acceptable or not although the revision doesn't match - this could be useful if you send changes that do not conflict with changes made in the meantime. A conflict doens't necessarily mean an update of the same property, the server can also define other conditions (e.g. properties that depend on each other and therefore cannot be merged).

dgeb:
I think our differences come down to the question of what is expected of a server when passed a revision that isn't up to date. My assumption has been that the server should always outright reject such a request because the very point of passing in the revision is to ensure that the client and server are working with the same version of the resource. If that's not true, then it implies a sophisticated server that maintains multiple versions of each resource and is able to discern whether there's a conflict based upon a review of the history of changes between versions.

scott-mcdonald:
> I actually favor a MAY for including revision with PATCH. First of all, I don't think a GET should be a pre-requisite for PATCH. Also, some clients will want to force a change (say an update to an attribute) regardless of the resource's revision. By forcing the inclusion of revision, we'd also be forcing servers to reject PATCHes if that field doesn't match.

I want to cautiously disagree with this statement. If a server is returning `revision` values with resources on GET requests it is for a _reason and purpose_ by the server meaning the client should not be ignoring `revision` on PATCH requests and hence why I prefer **MUST** be passed on PATCH if supplied on GET. This is the behavior I believe most developers would expect. But am I being too conservative or heavy handed in this thinking. What if a web server wants to be more liberal and support a scenario where it allows an update to always succeed, aka a force update or overwrite feature like you mentioned. I have to admit this is more in line with the Robustness principle _"Be conservative in what you do, be liberal in what you accept from others"_ so I am torn now. Would be interested in what others think.

I **currently** have text basically saying the following:

If a server supplies a `revision` member in a `GET` request, a client **SHOULD** pass the same `revision` member in a `PATCH` request when updating the resource. SHOULD is in between MAY and MUST. Maybe this is exactly where it needs to be...

> On that same note, the one MUST I would say should be in place is that the server MUST reject a PATCH request that includes revision if it doesn't match the current revision of the resource.

Excellent point and should be added to the specification.

@tkellen To your point about etag and revision. I purposely left this out because they are not the same thing: etag is a version of the document as a whole, revision is a version of a specific, individual resource. To @dgeb point in a previous email thread, even if the document is just a single resource, with `Sparse Fieldsets` documents with the same resource and version but with different representations due to different fields being included would have different etags.

scott-mcdonald:
Added a 2nd draft, to summarize:
- Removed `revision` as part of a resource identifier object altogether. `Revision` is now part of resource object.
- Added a new child `revision` section under the resource object parent section with `revision` description there.
- Changed the normative text on when to pass `revision` to basically the following: If a server supplies a `revision` value on a `GET` request, a client **MUST** pass the same `revision` value in a `PATCH` request on update.
  - I changed to **MUST** for now because I believe it is what most developers would expect and want. But this can be changed based on further debate if needed, or changed by the owners when adding the pull-request.
- Added/Clarified that on updating resources on `revision` mismatch then the server **MUST** reject the update with a `409 Conflict`.
- I did not add any statements around how `revision` and `etag` can be used in conjunction for now. I suppose this can be added later but thinking the specification should remain agnostic on this topic. Not sure what to put for this.

format/index.md
```diff
@@ -1453,6 +1470,9 @@ constraints (such as a uniqueness constraint on a property other than `id`).
 A server **MUST** return `409 Conflict` when processing a `PATCH` request in
 which the resource object's `type` and `id` do not match the server's endpoint.
 
+A server **MUST** return `409 Conflict` when processing a `PATCH` request in
```
tkellen: I think this is too prescriptive. A server may be able to resolve a conflict and accept it.

tkellen: Also, this normative text is redundant with the text on line 221.

tkellen:
> I think our differences come down to the question of what is expected of a server when passed a revision that isn't up to date. My assumption has been that the server should always outright reject such a request because the very point of passing in the revision is to ensure that the client and server are working with the same version of the resource. If that's not true, then it implies a sophisticated server that maintains multiple versions of each resource and is able to discern whether there's a conflict based upon a review of the history of changes between versions.

I agree that most servers will likely outright reject mis-matching revisions, but I don't think we should normatively close the door on the possibility of a sophisticated server using JSON-API. For example, databases like datomic _do_ have a full revision history of every change.

scott-mcdonald:
@tkellen Thanks and excellent points, will add to a 3rd draft early next week based on other comments.

tkellen:
Sure thing @scott-mcdonald! I'm super excited to see this land and to implement it for [endpoints](http://github.com/endpoints).

tkellen:
One other thing to note, @bobholt assembled a json-api representation of all the normative statements in the spec (to be used for generating test suites). I'm working on making it possible to automatically generate that file, but it is currently still a manual process. If you could include any new normative statements in [this file](https://github.com/json-api/json-api/blob/gh-pages/normative-statements.json) for the next revision that would be awesome:

scott-mcdonald:
Added a 3rd draft based on feedback from @tkellen in this thread.

format/index.md
```diff
@@ -1287,6 +1309,8 @@ primary data, the same request URL can be used for updates.
 
 The `PATCH` request **MUST** include a single resource object as primary data.
 The resource object **MUST** contain `type` and `id` members.
+The resource object **MUST** contain `revision` member if the server supplied
+the `revision` member in a `GET` request.
```
dgeb: We can't make a **MUST** level requirement of clients based on the server implementation. If a 1.1 server includes `revision`, then a 1.0 client should still work without including `revision`.

scott-mcdonald:
Added a 4th draft based on feedback from @dgeb. Basically changed **MUST** to **SHOULD** so the specification is backwards compatible when a 1.0 client calls a 1.1 or greater server. Used **SHOULD** instead of **MAY** to stress the importance of properly handling updates when `revision` is involved.

tkellen:
This seems great @scott-mcdonald! Unfortunately, we can't land this until we make some infrastructural changes to the site to support multiple versions. This would be a part of 1.1-alpha. Also, can you please add your new normative statements to [this file](https://github.com/json-api/json-api/blob/gh-pages/normative-statements.json)?

Thanks!

ethanresnick:
How would people feel about waiting until we have extensions figured out before landing this? I'm thinking that, if we make this an extension, we might be able to turn some of those SHOULDs into MUSTs or otherwise make stronger guarantees in way that's backwards compatible than we can if we try to add this to the base spec.

tkellen:
As long as you lead the charge to getting extensions figured out I am game
to consider this. Both @dgeb and I have been too busy to really give
JSON-API the attention it deserves lately.

ethanresnick:
@tkellen Works for me!

scott-mcdonald:
Gentlemen,

I like @ethanresnick train of thought with respect to the **SHOULD** really being **MUST** for strong guarantees in terms of _optimistic concurrency control_ at the _resource_ level. This is what most developers would think is the _"normative"_ use case, IMHO.

Again IMHO, I feel it is a stretch to think it should be an extension though. To me an extension is something a client would negotiate FOR because it is looking for an additional feature but why would a client negotiate for revision at the resource level; this is something a service would want to impose without negotiation. The Bulk extension makes perfect sense as an extension but resource version does not feel right. I guess I would have to see how this works before making a final judgment.

From my perspective resource version is so fundamental it should be part of the base specification. The issue is with the backwards compatibility with 1.0 forever edict. By the way does that mean if the specification was at 4.2 it would always have to be backwards compatible to 1.0?

Again IMHO, if you increment the major version number it is no longer backwards to the previous version, this is what most developers would think is the _"normative"_ use case I believe. Under this pretext I would do the following if I was an owner: Keep the specification as it is currently written for 1.1 so it is backwards compatible to 1.0 with the **SHOULD** instead of the **MUST**. But if the specification ever goes to 2.0 change the **SHOULD** to **MUST**, etc.

The 1.0 specification has the "jsonapi" object that has the version in question (1.0, 1.1, 2.0) so I don't see this as a big deal. But I am sure you gentlemen have already discussed this forever backwards compatibility to 1.0 "ad nauseam" so the previous statements are probably not going to fly...

@tkellen as far as if this being a part of 1.1 alpha and being delayed I am in no big hurry so take your time. I currently have "version" in the attributes of a resource where needed for optimistic concurrency control, etc. When this did "land" was going to refactor to "revision" at the top level of resources, etc.

BTW, I do see "caching" at the resource level leveraging `revision` being some kind of extension?

tkellen:
A quick clarifying response from a skim of your note @scott-mcdonald.

JSON-API will never bump major. If it does, it becomes a new project with a new media type. Backwards compatibility is a contract we refuse to break.

tkellen:
Upon further consideration, I do agree with @scott-mcdonald that this should be a part of the base specification.

ethanresnick:
@scott-mcdonald What you said makes a lot of sense, so I'll revise my position too: I'm not necessarily against specifying this feature (or most of it) in the base spec. What was really motivating my earlier comment was a feeling that I just want to make sure we don't rush this, which I was worried might happen if we try to work on it concurrently with our focus on extensions and #795, which are the two issues marked for 1.1. As @dgeb and @tkellen have also said: there's a fair amount of subtlety here.

For example, here are some outstanding things I'm wondering about:
1. <a id="q1"></a>The exact semantics of the `"revision"` member. Namely, if a resource’s attribute is updated from value A to B, and then set back to A, are we saying that its `revision`: a)  MUST be different than before the value was originally changed to B; b) MAY be different; or MUST be the same. (Note that the last option would be incompatible with an incrementing row version integer.) Which approach does Etag take? What are the relative pros/cons?
2. Whether `"revision"` MUST be included in a PATCH (which would always require a GET and prevent forced updates) or merely SHOULD, or whether this is something that the server has discretion over. I don't have a clear feeling one way or the other from our discussion thus far and, again, I haven't had much time to think about it.
3. Whether, as @ziege alluded to, we want to/can define this feature such that sophisticated servers have more options than simply rejecting requests that don't match the `"revision"`. This seems like something we should enable if we can do so in an additive way (i.e. simple servers can still just reject the request), but I'm not sure what it would require.
4. If we delay the resource-level caching stuff for an extension, as you suggest, what do we need to work out in advance to ensure that the extension system will actually support adding in those constraints later.

If we can work all these details out easily, then awesome; I am excited to see this land! But I do want to make sure we're able to work them out first.

scott-mcdonald:
Hi @ethanresnick,

Understood about not wanting to rush. IMHO, I feel this topic has been highly vetted to this point reflecting in high quality of what has been produced to date (4 drafts for example) which is a good thing. Before answering your specific points let me provide an _Abstract_ from my point of view as a foundation: 

**Abstract**

The primary use case for a `revision` member is for _optimistic concurrency control_ purposes. The `revision` member is an opaque identifier and is a server-side implementation detail on how it is generated. If a server is providing the `revision` member in a read it should be passed back in any update. On updates, the server compares the client supplied `revision` to the actual current `revision` and if they match the update proceeds. If they do not match, the server can try to resolve the difference but if it can not then return a 409-Conflict. On a 409-Conflict a client should get the latest version of the resource, reapply updates, and try the update again if needed.

**Your Questions**

> The exact semantics of the "revision" member. Namely, if a resource’s attribute is updated from value A to B, and then set back to A, are we saying that its revision: a) MUST be different than before the value was originally changed to B; b) MAY be different; or MUST be the same. (Note that the last option would be incompatible with an incrementing row version integer.) Which approach does Etag take? What are the relative pros/cons?

The `revision` member is an opaque identifier and is a server-side implementation detail. Therefore the base specification **SHOULD NOT** state any implementation details on the semantics of the value itself. Fundamentally all a client and/or service wants to know is if two resources with the same type/id are the same version or not by comparing `revision` for equality. Etag values are semantically exactly what I have described here except applies for the response as a whole, see [https://en.wikipedia.org/wiki/HTTP_ETag](https://en.wikipedia.org/wiki/HTTP_ETag) as an example. 

What I like about this approach is in the base specification revision is defined as an opaque identifier. This opens up further refinement in all kinds of extensions that could go further on the actual semantics of the revision value, i.e. resource level caching.

> Whether "revision" MUST be included in a PATCH (which would always require a GET and prevent forced updates) or merely SHOULD, or whether this is something that the server has discretion over. I don't have a clear feeling one way or the other from our discussion thus far and, again, I haven't had much time to think about it.

This is up for debate. I just felt if a service is supplying `revision` in a **GET** it expects it in a **PATCH** otherwise why supply it all. But for backwards compatibility with 1.0 I set this to **SHOULD**.

> Whether, as @ziege alluded to, we want to/can define this feature such that sophisticated servers have more options than simply rejecting requests that don't match the "revision". This seems like something we should enable if we can do so in an additive way (i.e. simple servers can still just reject the request), but I'm not sure what it would require.

Agreed and this is why I used the following verbiage in the Update section (please note the _unable to resolve_ text):

A server **SHOULD** return `409 Conflict` when processing a `PATCH` request in
which the resource object's `revision` does not match the server's endpoint
and the server was _unable to resolve_ the `revision` mismatch for updating
resources with the same `type` and `id` but mismatched `revision` members.

> If we delay the resource-level caching stuff for an extension, as you suggest, what do we need to work out in advance to ensure that the extension system will actually support adding in those constraints later.

I have not given this much thought and quite frankly should be it's own ISSUE and deployable item. I do know that `revision` is going to play a key role just like Etag does in HTTP level caching.

**Summary**

IMHO, I believe how revision is currently written primarily for _optimistic concurrency control_ is at the right level for the base specification. I believe this maximizes flexibility for later extensions as in the base specification it is defined as an opaque identifier and we are not being too restrictive.

I would release it as-is with it currently being on the more minimal side; let the community digest it and see what feedback, issues come up and then refactor as needed.

ethanresnick:
Hi @scott-mcdonald,

I've been thinking more about this recently and would like to get it merged as soon as possible. I agree with your answers to my second and third questions, and I think there are only three things left to settle:
1. I agree completely that the spec should not state any `revision` implementation details. But it should define the _meaning_ of the `revision` member very explicitly, and this meaning is what I was trying to nail down with my first question. From your answer, it sounds like you're advocating for [option b](#user-content-q1), and I agree with you that that makes the most sense. However, I think your PR as worded could be read to inadvertently endorse option a. In particular, it says that "If the resource is _updated_, a new and different `revision` is assigned." To me, that means that, when the resource's attribute starts at value A, its `revision` might be `"1"`. Then, when the attribute's value is set to B, the revision could be set to `"2"`. Finally, when the attribute is set back to A, the server must set the `revision` to something "new and different", so it sets it to `"3"`—whereas option b implies that it should be allowed to set it back to `"1"`. 
   
   So, if we're going with option b, we need to update the PR to reflect that. Preferably, the phrasing would give explicit, almost if–then, constraints. For example, option b could be captured precisely (but far too verbosely; this is just a rough example) by saying: "At all points in time at which a given JSON API resource has `revision` value X, the resource's fields must take on the same set of values, Y." (Note that this doesn't go the other way and specify that, whenever the field values are the same, the `revision` must match; that would be option c, which is probably too rigid.) Trying to give strict constraints, as in my provided wording, is probably impractical though, because, for example, it could inadvertently prevent implementations from using a hash function. (Hash collisions would violate my constraint above.) So we need something a bit looser; I'd suggest taking a look at how the HTTP RFCs word this.
2. The issue of trying to nail down revision's meaning above brought up a new question for me: how does resource-level `meta` fit into this picture? In the constraint I gave above, a change to `meta` wouldn't imply a change in `revision`. Is that what we want? I think this depends on whether users are supposed to be setting `meta` at all...which the current spec is pretty vague on...or what `meta` might hold. I'd love for @tkellen, @dgeb, and @steveklabnik to chime in here. Once we figure out `meta` semantics more clearly, I'll have a stronger position on the fourth question.
3. Lastly, there are some other things to consider from HTTP's implementation.
   - HTTP allows the client to send multiple etags in the same `If-(None-)Match` header. Do you think this is something we should support for `revision`? Personally, I'm leaning towards no, but I can imagine one place it might be useful. I _don't_ think it'd be useful/worth it just for conditional updates because, even in HTTP, sending multiple etags should only slightly increase the odds of a conditional update succeeding; in our case, that advanatage seems even less applicable given that we've redefined `revision`'s semantics from "If-Match" (i.e. if the `revision` is exactly the one provided) to "If-Compatible-With" (i.e., if the server can resolve the conflict). However, the one place where I'd imagine it might be useful is for a conditional GET over WebSockets, where the body of the message could look like this: `{ "data": { "revision": ["a", "b", "c"] } }`. That would allow the server to skip sending back a full representation if the current state is "a", "b", or "c", which could be a savings for resources that alternate between a small number of states (many of which the client has already seen).
   - Finally, HTTP reserves the "*" value as a wildcard in `If-(None-)Match` headers, which can be useful when applying a method that could create or update a resource. (E.g., PUT with If-Match: \* means PUT, but only if this resource has already been created.) I think we likewise reserve `"*"` as a revision value, even if we don't define semantics for it now, as it could come in handy re the second question in #419. It could also come in handy as a precondition to a DELETE in a batch operation request, to prevent the request from totally failing if the DELETE has already occurred. Plus, reserving stuff is just always good. (In that vein, I think we should also think about limiting the characters allowed in revision values to those allowed in Etags, in case we want to define a mapping between the two. However, that gets complicated with double quotes and the weak modifier...again, let's figure that out after we address `meta`'s semantics.)

ethanresnick:
Also, @scott-mcdonald, I apologize if it feels like I'm nitpicking; I don't mean to, and I know you've been trying to get this merged for a long time. Keep in mind, though, that once we add something to the base spec, we can never take it back. That's why I'm harping so much on making sure we have the meaning specified precisely; and why I'm trying to make sure that, as much as possible, we're complementing rather than duplicating Etags; and that we're designing the feature to give us a bit of extensibility for future use cases; etc.

scott-mcdonald:
Hi @ethanresnick 

No need to ever apologize as you are just trying to make the specification as perfect as possible, I think that is the makeup of software engineers in general and it is all good. I had a chance to read your previous comment this morning and would like to provide an initial response and see if we cannot get this nailed down over the next few comments.

As far as the first question about the _meaning_ of the `revision` member my opinion is the following: We do not want the specification to be constrained or strict as it reduces flexibility. Therefore with respect to _optimistic concurrency control_ specifically the base specification **SHOULD NOT** advocate either of the options you have specified (option A or option B in previous comments) but instead simple state `revision` is defined as resources with the same type/id but same/different `revision` values are the same/different version respectively. I would argue that changing the resource from A to B back to A and what the `revision` value should be based on your stated options A or B is actually an implementation detail of the service. If the service was doing some kind of hash function that implies option B, if it was some incremental integer that is option A but to me this is all an implementation detail of the service and not be in the specification.

BUT to your point I did some research into ETag in the RFC specifications for comparative purposes. Going to paste in the sections that seemed important from RFC 2616:

> 13.3.3 Weak and Strong Validators
> 
> Since both origin servers and caches will compare two validators to decide if they represent the same or different entities, one normally would expect that if the entity (the entity-body or any entity- headers) changes in any way, then the associated validator would change as well. If this is true, then we call this validator a "strong validator."
> 
> However, there might be cases when a server prefers to change the validator only on semantically significant changes, and not when insignificant aspects of the entity change. A validator that does not always change when the resource changes is a "weak validator."
> 
> Entity tags are normally "strong validators," but the protocol provides a mechanism to tag an entity tag as "weak." One can think of a strong validator as one that changes whenever the bits of an entity changes, while a weak value changes whenever the meaning of an entity changes. Alternatively, one can think of a strong validator as part of an identifier for a specific entity, while a weak validator is part of an identifier for a set of semantically equivalent entities.
> 
> ```
>   Note: One example of a strong validator is an integer that is
>   incremented in stable storage every time an entity is changed.
>   An entity's modification time, if represented with one-second
>   resolution, could be a weak validator, since it is possible that
>   the resource might be modified twice during a single second.
>   Support for weak validators is optional. However, weak validators
>   allow for more efficient caching of equivalent objects; for
>   example, a hit counter on a site is probably good enough if it is
>   updated every few days or weeks, and any value during that period
>   is likely "good enough" to be equivalent.
> ```
> 
> A "use" of a validator is either when a client generates a request and includes the validator in a validating header field, or when a server compares two validators.
> 
> Strong validators are usable in any context. Weak validators are only usable in contexts that do not depend on exact equality of an entity. For example, either kind is usable for a conditional GET of a full entity. However, only a strong validator is usable for a sub-range retrieval, since otherwise the client might end up with an internally inconsistent entity.
> 
> Clients MAY issue simple (non-subrange) GET requests with either weak validators or strong validators. Clients MUST NOT use weak validators in other forms of request.
> 
> The only function that the HTTP/1.1 protocol defines on validators is comparison. There are two validator comparison functions, depending on whether the comparison context allows the use of weak validators or not:
> 
> ```
>   - The strong comparison function: in order to be considered equal,
>     both validators MUST be identical in every way, and both MUST
>     NOT be weak.
>   - The weak comparison function: in order to be considered equal,
>     both validators MUST be identical in every way, but either or
>     both of them MAY be tagged as "weak" without affecting the
>     result.
> ```
> 
> An entity tag is strong unless it is explicitly tagged as weak. Section 3.11 gives the syntax for entity tags.
> 
> If a client wishes to perform a sub-range retrieval on a value for which it has only a Last-Modified time and no opaque validator, it MAY do this only if the Last-Modified time is strong in the sense described here.
> 
> A cache or origin server receiving a conditional request, other than a full-body GET request, MUST use the strong comparison function to evaluate the condition.
> 
> These rules allow HTTP/1.1 caches and clients to safely perform sub- range retrievals on values that have been obtained from HTTP/1.0 servers.
> 
> > 3.11 Entity Tags
> 
> Entity tags are used for comparing two or more entities from the same requested resource. HTTP/1.1 uses entity tags in the ETag (section 14.19), If-Match (section 14.24), If-None-Match (section 14.26), and If-Range (section 14.27) header fields. The definition of how they are used and compared as cache validators is in section 13.3.3. An entity tag consists of an opaque quoted string, possibly prefixed by a weakness indicator.
> 
> ```
>   entity-tag = [ weak ] opaque-tag
>   weak       = "W/"
>   opaque-tag = quoted-string
> ```
> 
> A "strong entity tag" MAY be shared by two entities of a resource only if they are equivalent by octet equality.
> 
> A "weak entity tag," indicated by the "W/" prefix, MAY be shared by two entities of a resource only if the entities are equivalent and could be substituted for each other with no significant change in semantics. A weak entity tag can only be used for weak comparison.
> 
> An entity tag MUST be unique across all versions of all entities associated with a particular resource. A given entity tag value MAY be used for entities obtained by requests on different URIs. The use of the same entity tag value in conjunction with entities obtained by requests on different URIs does not imply the equivalence of those entities.
> 
> 14.19 ETag
> 
> The ETag response-header field provides the current value of the entity tag for the requested variant. The headers used with entity tags are described in sections 14.24, 14.26 and 14.44. The entity tag MAY be used for comparison with other entities from the same resource (see section 13.3.3).
> 
> ```
>   ETag = "ETag" ":" entity-tag
> ```
> 
> Examples:
> 
> ```
>   ETag: "xyzzy"
>   ETag: W/"xyzzy"
>   ETag: ""
> ```

So what does all this mean. The BIG question for me from reading this is `revision` and `etag` semantically the same thing - maybe? The RFC is primarily focused on caching while originally I was focusing on optimistic locking on resource updates. I can tell your mindset is coming from the HTTP caching perspective. I actually could not find any specifics on resource updates with PUT or PATCH and ETag in any RFC documents but it must exist somewhere?

So to your point we should do some forward thinking about `revision` and caching which is why you are bringing up the points you are making. So with that said we **COULD** adopt the same semantic meaning of "strong" and "weak" `revision` values as the ETag in the RFC specifications and thus use the same "verbiage" when describing the semantic meaning of revision. I thinks this meets both our needs. What do you think?

As for your question on `meta`, IMHO `meta` **SHOULD NOT** play any role in `revision`. IMHO `meta` should be strictly data describing the resource as a type, aka _structure metadata_ as to me this is the normative use case developers would expect but to your point meta as currently described is vague.

As for your question on multiple `revision` values per resource, I think we need to keep it the normative case of 1 `revision` value as the version of the resource for the base specification.

I do believe resource caching is a separate deployable item but I like your approach you are hinting at: When it comes time to add caching to the specification either in the base or as an extension use the some or all of the same HTTP header values and semantic meanings for caching like `If-Match`, `If-None-Match`, etc. Why reinvent the wheel plus it will make it easier to adopt by developers, etc.

I like the idea of reserving "*" as a `revision` value for future use in some capacity.

jeluard:
It might be worth looking at what [CouchDB](http://couchdb.org/) does in that regard. It has a `rev` concept mapped to HTTP ETag. The specific semantic of `rev` also highlights some advanced usage it would be great not to prevent.

Relevant links:
- http://guide.couchdb.org/draft/api.html#revisions
- http://guide.couchdb.org/draft/show.html#etags
- http://guide.couchdb.org/draft/conflicts.html#working

bf4:
Awesome! I'd like to see a top-level `revision` member as well, useful for identifying the app revision that generated the request, e.g. the git sha, which I believe is a common practice in 'microservices'.

ethanresnick:
Ok, finally coming back to this, and responding to @scott-mcdonald's comment:

>  If the service was doing some kind of hash function that implies option B, if it was some incremental integer that is option A but to me this is all an implementation detail of the service and not be in the specification

I think the meaning of my options got lost because I phrased them in legalese (sorry), so let me summarize them in more straightforward terms:
- Option A: server must use a hash function
- Option C: server must use an incrementing counter
- Option B: server can pick to use either option A or B

So, when I said I agree with you that we should be doing Option B, I meant that I agree that we shouldn't mandate hash vs. row version implementations. My point, though, was that the current wording of your PR seems to inadvertently _prevent_ the server from using a hash function because it says the `revision` value has to be "new and different" on each update. To me, that means that when the resource changes from state A to B and then back to A, the server can't reuse the original `revision` rather (from the first time the resource was at state A) because that value isn't "new and different".

So, updating the PR to not preclude hash implementations was the first "to do" in my prior comment.

> As for your question on multiple revision values per resource, I think we need to keep it the normative case of 1 revision value as the version of the resource for the base specification.

I wasn't suggesting that a resource could have multiple revisions; I was saying that a client could send multiple revisions for its (conditional get/update) request would be checked against. (In the same way that multiple Etags can be provided in an `If-Match` header.)

> When it comes time to add caching to the specification either in the base or as an extension use the some or all of the same HTTP header values and semantic meanings for caching like If-Match, If-None-Match, etc.

This is decidedly _not_ the approach I'm advocating. On the contrary, I think HTTP's header names are confusing and unnecessary, since each method is almost always used with only one header. The current approach, of just having a `revision` member, but treating it as an `If-None-Match` on GET requests and an `If-Match` on PATCH/DELETE requests is much better.

> I like the idea of reserving "*" as a revision value for future use in some capacity.

Awesome. Can you please add some text to your PR to this effect.

ethanresnick:
(@scott-mcdonald, I'll come back to your other points later, but above is a discussion of the small issues. Let's get those hammered out first.)

scott-mcdonald:
@ethanresnick Hi Ethan, I understand your points clearly now and I think we are approximately on the same page. Will get another "version" (pardon the pun) out in the next couple of days. BTW - Ethan is my son's name - nice name ;-)

hhware:
> I actually favor a MAY for including revision with PATCH. First of all, I don't think a GET should be a pre-requisite for PATCH. Also, some clients will want to force a change (say an update to an attribute) regardless of the resource's revision. By forcing the inclusion of revision, we'd also be forcing servers to reject PATCHes if that field doesn't match.

Fully support this! 

An example: a client is PATCHing a user record by updating its email. If the client knows for sure that the new email is correct (e.g., because it is _this client_ who is actually changing it), then it may not matter what the current value of email is. IMHO, it should be up to the API designer whether to allow clients to update attributes this way. The spec should not mandate pointless back-and-forth (GET before PATCH, possibly multiple times) unless it makes sense for the particular API.

> How would people feel about waiting until we have extensions figured out before landing this? I'm thinking that, if we make this an extension, we might be able to turn some of those SHOULDs into MUSTs or otherwise make stronger guarantees in way that's backwards compatible than we can if we try to add this to the base spec.

Accordingly, I do not support this proposal.

> My assumption has been that the server should always outright reject such a request because the very point of passing in the revision is to ensure that the client and server are working with the same version of the resource. If that's not true, then it implies a sophisticated server that maintains multiple versions of each resource and is able to discern whether there's a conflict based upon a review of the history of changes between versions.

Fully support that too! 

IMHO, If we are talking about an abstract sophisticated server, simple string `revision` may not be enough, the server may need to define its own attribute(s) to enable its sophisticated protocol. Why hijack `revision` for this purpose, if it does not guarantee to cover all possible cases? IMHO, `revision` should be kept straightforward. I would remove the statement about "_unable to resolve_ the `revision` mismatch" altogether.

> The exact semantics of the "revision" member. Namely, if a resource’s attribute is updated from value A to B, and then set back to A, are we saying that its revision: a) MUST be different than before the value was originally changed to B; b) MAY be different; or MUST be the same. (Note that the last option would be incompatible with an incrementing row version integer.) Which approach does Etag take? What are the relative pros/cons?

I think MUSTs are out of question here because they assume that the server keeps track of past revisions and is able to identify the past revision with the same value of the attributes, so the only option is MAY. Hence the question: why is it even necessary for the spec to be concerned with it? Perhaps I am missing something, but what would be broken if it is left undefined and even implemented inconsistently from one endpoint to another endpoint of a given API? The only consequence I see is possible cache miss in the situation where it might have been hit. 

> From your answer, it sounds like you're advocating for option b, and I agree with you that that makes the most sense. However, I think your PR as worded could be read to inadvertently endorse option a. In particular, it says that "If the resource is updated, a new and different revision is assigned."

Cannot it just say "If the resource is _updated_, value of its `revision` **MUST** change"? This also seems to take care of the item 2 in https://github.com/json-api/json-api/pull/824#issuecomment-135972581

> HTTP allows the client to send multiple etags in the same If-(None-)Match header. Do you think this is something we should support for revision?

By defining a dedicated header or by allowing to supply multiple revisions in `PATCH`?

> As for your question on meta, IMHO meta SHOULD NOT play any role in revision.

Why? IMHO, one use case for `meta` is to be a container for things not allowed by the spec, and it is hard to make a general statement whether such things play any role in revision. An example: https://github.com/json-api/json-api/pull/431#issuecomment-90766675 and https://github.com/json-api/json-api/issues/559. Again, I believe it should be up to the API designer to decide.

format/index.md
```diff
@@ -1453,6 +1481,14 @@ constraints (such as a uniqueness constraint on a property other than `id`).
 A server **MUST** return `409 Conflict` when processing a `PATCH` request in
 which the resource object's `type` and `id` do not match the server's endpoint.
 
+A server **SHOULD** return `409 Conflict` when processing a `PATCH` request in
+which the resource object's `revision` does not match the server's endpoint
+and the server was *unable to resolve* the `revision` mismatch for updating
```
hhware: I would remove this and the following lines, see https://github.com/json-api/json-api/pull/824#issuecomment-140977824

format/index.md
```diff
@@ -1453,6 +1481,14 @@ constraints (such as a uniqueness constraint on a property other than `id`).
...
+resources with the same `type` and `id` but mismatched `revision` members.
+
+> Note: Only applicable if a server supplied the **OPTIONAL** `revision` member
```
hhware: Since it is not mandatory for the client to supply `revision` in a `PATCH`, and since `GET` before `PATCH` is not mandatory, I believe this should be changed to 

> Note: Only applicable if a client supplied the **OPTIONAL** `revision` member in the `PATCH` request.

format/index.md
```diff
@@ -212,6 +214,26 @@ The values of `type` members **MUST** adhere to the same constraints as
...
+#### Revision <a href="#document-resource-object-revision" id="document-resource-object-revision" class="headerlink"></a>
+
+The **OPTIONAL** `revision` member is an *opaque* string identifier assigned by
+the web server to represent a specific version of a resource. If the resource
+is *updated*, a new and different `revision` is assigned. This mechanism
+enables clients or servers to determine if two resources with the same `type`
+and `id` values are the same or different versions by comparing the `revision`
+value for equality.
+
+The purpose of the `revision` member is to enable *optimistic concurrency control*
```
hhware: Why are this and the following paragraphs normative? Are not they just clarifications of intentions?

scott-mcdonald:
@ethanresnick Hi Ethan, I have done another revision for this story based on our exchange; when you get a chance take a look and see what you think. I noticed the introduction of "1.1" so I moved my revision member changes to the 1.1 _format/index.md file from the 1.0 version which I think is correct.

ethanresnick:
@scott-mcdonald Thanks for reworking this, Scott! On a quick glance, the current version looks much better! I'll do a careful review this weekend.

tkellen:
@ethanresnick Did you ever have a chance to review this?

ethanresnick:
@tkellen No, I didn't, and I want to apologize to @scott-mcdonald for that, since this is a very valuable contribution that's been sitting for far too long. I got about 80% through a review at one point, and found a couple problems, but then got focused on extensions. I'll try to go through this thread later this week and finish a review, so that we can get at least a beta version of this out ASAP.

_format/1.1/index.md
```diff
@@ -159,6 +159,7 @@ the client and represents a new resource to be created on the server.
 
 In addition, a resource object **MAY** contain any of these top-level members:
 
+* `revision`: a [revision string][revision] representing a specific version of an individual resource uniquely identified by `type` and `id`.
```
bf4, line 162: this could also be a recommendation as a [resource meta](http://jsonapi.org/format/#document-resource-objects)

raat1979:
Hi 
I only just started exploring the json-api, and was looking for a way to prevent parallel updates (optimistic locking) and found this discussion on revisions.
From my point of view revisions are different versions of the same document where it is possible to alter them individually. (wiki/git branches  etc.) 
If we are altering database records for example we would be much better of with a record timestamp or something which does not suggest there is some sort of version control.
In any case the timestamp does not match an error must be raised, for a revision this should work differently.

jelhan:
I think this could be very well addressed using an extension introduced by v1.1. This would also give room to experiment as such an extension does not require changes to the base spec. Overall it should speed up the process. We could later promote such an extension to an official extension, which should provide similar benefits as adding to the base spec.

PR #826, opened 2015-07-31
Added NJsonApi to the implementation list

jacek-gorgon:
[NJsonApi](https://github.com/jacek-gorgon/NJsonApi) is a .NET server implementation of the standard. It aims at good extensibility and performance while maintaining developer-friendliness with interchangable convenions and builder-style configuration.


steveklabnik:
Thank you!

PR #828, opened 2015-08-05
Example Error Objects

bf4:
Per discussion with @dgeb in https://github.com/rails-api/active_model_serializers/pull/1004

> Although it is still compliant to represent multiple validation errors in the same error object,
> it's not ideal. I would recommend creating a unique error object per validation error,
> so that fields such as code and title can be used consistently and expressively across errors.
> For instance, one code might indicate that a field is too short, and another might indicate
> that invalid characters were used. If you try to combine these errors into a single code,
> then it will probably end up being something more generic with less meaning, such as "invalid value".

References:
- http://jsonapi.org/format/#error-object
- Discussion in PRs
  - https://github.com/json-api/json-api/pull/237 [specifically](https://github.com/json-api/json-api/commit/6901143531c14f41e95818455fe36228cebf6625) and https://github.com/json-api/json-api/pull/638
  - https://github.com/json-api/json-api/pull/234
  - https://github.com/json-api/json-api/issues/7
  - https://github.com/json-api/json-api/pull/208
- Usage in Ember
  - http://emberjs.com/blog/2015/06/18/ember-data-1-13-released.html#toc_using-json-api-error-object-format
  - ember-data
    - [lib/adapters/errors.js](https://github.com/emberjs/data/blob/d842890357840700f0ee17f152415be0cba16030/packages/ember-data/lib/adapters/errors.js#L114-L132)
    - [lib/adapters/rest-adapter.js](https://github.com/emberjs/data/blob/3930e457f36a0bbdde0155c1b98ebe976e607358/packages/ember-data/lib/adapters/rest-adapter.js#L913-L925)
    - [lib/serializers/json-serializer.js](https://github.com/emberjs/data/blob/944f7e4b537cdc7cd2a80718d0447453c01de2db/packages/ember-data/lib/serializers/json-serializer.js#L1225-L1276)
    - [adapter-errors-test.js](https://github.com/emberjs/data/blob/21ebe9b61b33acd910c0d485e17926c89c51d9f6/packages/ember-data/tests/unit/adapter-errors-test.js#L32-L48)
    - [json-serializer-test.js](https://github.com/emberjs/data/blob/90a0011c8e5e1c41d1667c96a38c4a5e4a9c1e64/packages/ember-data/tests/integration/serializers/json-serializer-test.js#L566-L622)

### DIFF
--- a/examples/index.md
+++ b/examples/index.md
@@ -171,3 +171,354 @@ to indicate to clients the total number of pages in a collection (as opposed to
 the `"last"` link, which simply gives the URI of the last page). However, all
 `"meta"` values are implementation-specific, so you can call this member whatever
 you like (`"total"`, `"count"`, etc.) or not use it at all.
+
+## Error Objects
+
+Examples of how [error objects](http://jsonapi.org/format/#error-objects) work.
+
+### Responses to failed resource creation:
+
+Response with one error object, with attribute error:
+
+```http
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "title":  "Invalid Attribute",
+      "source": { "pointer": "data/attributes/first-name" },
+      "detail": "First name must contain at least three characters."
+    }
+  ]
+}
+```
+
+Notice that the attribute `first-name` is dasherized.  Dasherized attributes are preferred
+over underscored.
+
+Notice also that the `detail` is specific to this occurrence of the problem,
+whereas the `title` is more generic.
+
+Notice that the `source` points to the associated attribute with the error.
+
+Response with one error object, with relationship error:
+
+```http
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "status": "422",
+      "title":  "Missing Association",
+      "source": { "pointer": "data/relationships/author" },
+      "detail": "Post author cannot be blank."
+    }
+  ]
+}
+```
+
+Notice that the `source` points to the associated relationship with the error.
+
+Notice also that the error object includes a `status` code. Though optional,
+it can be useful when the code processing the JSON isn't directly aware of
+the HTTP response or even if the JSON is coming via non-HTTP protocols.
+
+Response with one error, using a `code`:
+
+```http
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "title": "Adapter Error",
+      "code": "ds.errors.invalid-error-expects-json-api-format",
+      "detail": "The adapter rejected the commit because it was not in the JSON API format."
+    }
+  ]
+}
+```
+
+Response with status:
+
+```http
+HTTP/1.1 400 Bad Request
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "status": "403",
+      "source": { "pointer": "data/attributes/secret-powers" },
+      "detail": "Access to secret powers not authorized on Sundays."
+    },
+    {
+      "status": "404",
+      "links": { "about": {
+        "href": "http://example.com/weapons/vorpal-blade",
+        },
+      },
+      "title": "Weapon not found."
+    },
+    {
+      "status": "422",
+      "source": { "pointer": "data/attributes/volume" },
+      "detail": "Volume does not, in fact, go to 11."
+    },
+    {
+      "status": "500",
+      "title": "The backend responded with an error",
+      "detail": "Reputation service not responding after three requests."
+    },
+    {
+      "status": "400",
+      "detail": "JSON parse error - Expecting property name enclosed in double quotes: line 1 column 2 (char 1)"
+    }
+  ]
+}
+```
+
+The only uniqueness constraint on error objects in a response is the `id` field.
+Thus, multiple errors on the same attribute may be described by an error object
+for each error.
+
+Notice that one of the error objects references an association that was not found using
+a `links` object containing an about `link`.
+
+Multiple errors on `first-name` attribute:
+
+```http
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "title": "Invalid Attribute",
+      "source": { "pointer": "data/attributes/first-name" },
+      "detail": "First name must contain at least three characters."
+    },
+    {
+      "title": "Invalid Attribute",
+      "source": { "pointer": "data/attributes/first-name" },
+      "detail": "First name must contain an emoji."
+    }
+  ]
+}
+```
+
+A "400 Bad Request" response would also be acceptable. See
+[http://stackoverflow.com/a/20215807/1261879]('400 vs 422 response to POST of data' on
+Stack Overflow). The JSON API doesn't take a position on 400 vs. 422.
+
+If the API docs specified:
+
+> | Error Code | Title                                                                             |
+> |------------|-----------------------------------------------------------------------------------|
+> |  11        | "Invalid Attribute"                                                               |
+> |  123       | "too short"                                                                       |
+> |  124       | "emoji missing"                                                                   |
+> |  225       | "password must contain a letter, number, space, neologism, and special character" |
+> |  226       | "passwords do not match"                                                          |
+> |  227       | "password cannot be one of last five passwords"                                   |
+
+Multiple errors on `password` attribute, with error `code`:
+
+```http
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "code":   "123",
+      "source": { "pointer": "data/attributes/first-name" },
+      "detail": "First name must contain at least three characters."
+    },
+    {
+      "code":   "225",
+      "source": { "pointer": "data/attributes/password" },
+      "detail": "Frobnicate is not a neologism."
+    },
+    {
+      "code":   "226",
+      "source": { "pointer": "data/attributes/password" },
+      "detail": "Password and password confirmation do not match."
+    }
+  ]
+}
+```
+
+Multiple errors on `first-name` attribute, with error `code`, and `title`:
+
+```http
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "jsonapi": { "version": "1.0" },
+  "errors": [
+    {
+      "code":   "123",
+      "title":  "too short",
+      "source": { "pointer": "data/attributes/first-name" },
+      "detail": "First name must contain at least three characters."
+    },
+    {
+      "code":   "124",
+      "title":  "emoji missing",
+      "source": { "pointer": "data/attributes/first-name" },
+      "detail": "First name must contain an emoji"
+    }
+  ]
+}
+```
+
+Notice that this response includes not only the `errors` top-level member,
+but the `jsonapi` top-level member. It may also have include the `meta` or `links`
+top-level members, but not the `data` member.
+
+```http
+PATCH /posts/1
+Content-Type: application/vnd.api+json
+Accept: application/vnd.api+json
+
+{ "datum": [ ] }
+```
+
+```http
+HTTP/1.1 422 Unprocesssable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "title":  "Missing Data Member",
+      "source": { "pointer": "/data" }
+    }
+  ]
+}
+```
+
+Notice that the `source` indicates the error was causes not by a missing attribute, but
+by missing `data`.
+
+### Responses to failed resource request:
+
+Response with one error object, with parameter error:
+
+```http
+GET /api/posts/1
+```
+
+```http
+HTTP/1.1 400 Bad Request
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "title":  "Missing Query Parameter",
+      "source": { "parameter": "auth_key" },
+      "detail": "The required parameter, auth_key, is missing."
+    }
+  ]
+}
+```
+
+Notice that the `source` indicates the URI query parameter which caused the error.
+This example with a missing `auth_key` parameter was chosen over an example with an invalid
+URI query parameter, since it may make sense to ignore invalid parameters in some cases, and
+have the request succeed rather than fail. As described in the [Bulk
+extension](http://jsonapi.org/extensions/bulk/#bulk-operations):
+
+> ...a request MUST completely succeed or fail (in a single "transaction").
+> ...any request that involves multiple operations MUST only succeed if all operations are performed
+successfully
+
+Examples of invalid parameters may be `?page[prev]=` (parameter present, but has no value),
+`?include=auther` (invalid parameter value, perhaps a typo), `?felds[author]=`,
+(invalid parameter key, perhaps a typo), `?redirect_to=http%3A%2F%2Fwww.owasp.org` (invalid
+parameter, in this case, a phishing attack), etc.
+
+### Using JSON API extensions
+
+Request and response using the [JSON PATCH
+Extension](http://jsonapi.org/extensions/jsonpatch/):
+
+```http
+PATCH /posts/1
+Content-Type: application/vnd.api+json; ext=jsonpatch
+Accept: application/vnd.api+json; ext=jsonpatch
+
+[ { "op": "remove" } ]
+```
+
+```http
+HTTP/1.1 422 Unprocesssable Entity
+Content-Type: application/vnd.api+json; ext=jsonpatch
+
+[
+  {
+    "errors": [
+      {
+        "status": "422",
+        "detail": "Operation missing required keys: 'path'"
+      }
+    ]
+  }
+]
+```
+
+Notice that JSON PATCH is a special semantics and does not mean an HTTP PATCH request
+containing JSON.
+
+Request and response using the [Bulk extension](http://jsonapi.org/extensions/bulk/)
+
+```http
+PATCH /photos
+Content-Type: application/vnd.api+json; ext=bulk
+Accept: application/vnd.api+json; ext=bulk
+
+{
+  "data": [{
+    "type": "articles",
+    "id": "1",
+    "attributes": {
+      "title": "To TDD or Not"
+    }
+  }, {
+    "type": "articles",
+    "id": "2",
+    "attributes": {
+      "title": "LOL Engineering"
+    }
+  }]
+}
+```
+
+```http
+HTTP/1.1 400 Bad Request
+Content-Type: application/vnd.api+json
+
+
+{
+  "errors": [
+    {
+      "status": "404",
+      "source": { "pointer": "/data/articles/1" },
+      "detail": "Article with id 1 could not be found."
+    }
+  ]
+}
+```
+
+Since requests are transactional, as the [Bulk
+extension](http://jsonapi.org/extensions/bulk/#bulk-operations) describes: *"The
+state of the server MUST NOT be changed by a request if any individual operation fails."*
+Thus, even though one of the objects to update was found, one was not, so none get updated.
--- a/examples/index.md
+++ b/examples/index.md
@@ -176,34 +176,11 @@ you like (`"total"`, `"count"`, etc.) or not use it at all.
 
 Examples of how [error objects](http://jsonapi.org/format/#error-objects) work.
 
-### Responses to failed resource creation:
+### A Basic Error Object
 
-Response with one error object, with attribute error:
-
-```http
-HTTP/1.1 422 Unprocessable Entity
-Content-Type: application/vnd.api+json
-
-{
-  "errors": [
-    {
-      "title":  "Invalid Attribute",
-      "source": { "pointer": "data/attributes/first-name" },
-      "detail": "First name must contain at least three characters."
-    }
-  ]
-}
-```
-
-Notice that the attribute `first-name` is dasherized.  Dasherized attributes are preferred
-over underscored.
-
-Notice also that the `detail` is specific to this occurrence of the problem,
-whereas the `title` is more generic.
-
-Notice that the `source` points to the associated attribute with the error.
-
-Response with one error object, with relationship error:
+In the response below, the server is indicating that it encountered an error
+while creating/updating the resource, and that this error was caused
+by an invalid `"first-name"` attribute:
 
 ```http
 HTTP/1.1 422 Unprocessable Entity
@@ -213,38 +190,34 @@ Content-Type: application/vnd.api+json
   "errors": [
     {
       "status": "422",
-      "title":  "Missing Association",
-      "source": { "pointer": "data/relationships/author" },
-      "detail": "Post author cannot be blank."
+      "source": { "pointer": "/data/attributes/first-name" },
+      "title":  "Invalid Attribute",
+      "detail": "First name must contain at least three characters."
     }
   ]
 }
 ```
 
-Notice that the `source` points to the associated relationship with the error.
+Every member in an error object is optional, but all help the client
+by providing extra details.
 
-Notice also that the error object includes a `status` code. Though optional,
-it can be useful when the code processing the JSON isn't directly aware of
-the HTTP response or even if the JSON is coming via non-HTTP protocols.
+The `source` member is used to [indicate](https://tools.ietf.org/html/rfc6901)
+which part of the request document caused the error.
 
-Response with one error, using a `code`:
+The `title` and `detail` members are similar, but `detail` is specific
+to this occurrence of the problem, whereas `title` is more generic.
 
-```http
-HTTP/1.1 422 Unprocessable Entity
-Content-Type: application/vnd.api+json
+The `status` member represents the HTTP status code associated with the problem.
+It's very helpful when multiple errors are returned at once (see below), as the
+HTTP response itself can only have one status code. However, it can also be
+useful for single errors, to save clients the trouble of consulting the HTTP
+headers, or for using JSON API over non-HTTP protocols, which may be officially
+supported in the near future.
 
-{
-  "errors": [
-    {
-      "title": "Adapter Error",
-      "code": "ds.errors.invalid-error-expects-json-api-format",
-      "detail": "The adapter rejected the commit because it was not in the JSON API format."
-    }
-  ]
-}
-```
+### Multiple Errors
 
-Response with status:
+When multiple errors occur in response to a single request, the server
+can simply add each error to the `errors` array:
 
 ```http
 HTTP/1.1 400 Bad Request
@@ -255,15 +228,7 @@ Content-Type: application/vnd.api+json
     {
       "status": "403",
       "source": { "pointer": "data/attributes/secret-powers" },
-      "detail": "Access to secret powers not authorized on Sundays."
-    },
-    {
-      "status": "404",
-      "links": { "about": {
-        "href": "http://example.com/weapons/vorpal-blade",
-        },
-      },
-      "title": "Weapon not found."
+      "detail": "Editing secret powers is not authorized on Sundays."
     },
     {
       "status": "422",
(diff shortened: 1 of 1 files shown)


examples/index.md
```diff
@@ -126,3 +126,119 @@ Content-Type: application/vnd.api+json
...
+Response with one error object:
+
+```http
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "source": { "pointer": "data/attributes/first_name" },
```
dgeb: Although underscores are compliant, I'd prefer to use dasherized names, such as `first-name`, in examples as per our recommendations.

steveklabnik:
Thanks for this! Good examples are so crucial to getting a spec adopted.

examples/index.md
```diff
@@ -126,3 +126,119 @@ Content-Type: application/vnd.api+json
...
+      "detail": "is too short"
+    },
+    {
+      "status": "422",
```
dgeb, line 234: Having `status` in isolation seems a bit odd. In this example, I'd expect it to be part of the first error object.

I think an example that shows multiple different errors with different `status` values would be great. For example, one error object could be a `404` and another `422`, and the status for the whole response could be `400`.

examples/index.md
```diff
@@ -126,3 +126,119 @@ Content-Type: application/vnd.api+json
...
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "code":   "123",
+      "title":  "too short",
+      "source": { "pointer": "data/attributes/first_name" },
+      "detail": "is too short"
```
dgeb: `detail` could be longer than `title` and less of a fragment (the spec also says "specific to this occurrence of the problem."). Perhaps `"First name must contain at least three characters."` for this example.

dgeb:
> Thanks for this! Good examples are so crucial to getting a spec adopted.

Agreed! Great start here, @bf4 :+1:

examples/index.md
bf4, line 234: Great, thanks!  I made it this way on purpose since I didn't understand when error objects might have different status attributes.

examples/index.md
bf4: Thanks.  I'm glad you caught that :)

bf4:
I've made some updates, there might be some code blindness going on here.. I'm not sure if I did 'parameter' right. It appears to have evolved from 'path'.

I'm also thinking about adding examples of an error object with
- a links object
- a meta object
- a source pointer to a 'primary' data object.

And a response with, in addition to an `errors` top-level member, the other valid top-level members
- meta: a meta object that contains non-standard meta-information.
- jsonapi: an object describing the server's implementation
- links: a links object related to the primary data. (?)

bf4:
@dgeb @steveklabnik Any more feedback on the changes? (I've made them in incremental commits)

examples/index.md
```diff
@@ -126,3 +126,206 @@ Content-Type: application/vnd.api+json
...
+whereas the title is more generic.
+
+```http
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "id": "ds.errors.invalid-error-expects-json-api-format",
```
dgeb: This looks more like a `code` than an `id`:

> id: a unique identifier for this particular occurrence of the problem.
> code: an application-specific error code, expressed as a string value.

bf4: Oh, looks like I somehow mixed up code and id here and in the error codes section as well

FWIW, I got it from https://github.com/emberjs/data/blob/d842890357840700f0ee17f152415be0cba16030/packages/ember-data/lib/adapters/errors.js#L82

``` es6
    Ember.deprecate('`InvalidError` expects json-api formatted errors.', false, { id: 'ds.errors.invalid-error-expects-json-api-format', until: '2.0.0' });
```

which I misread as an errors object

bf4: Updated

examples/index.md
```diff
@@ -162,3 +162,209 @@ Content-Type: application/vnd.api+json
...
+### Responses to failed resource request:
+
+Basic request:
+
+```http
+GET /posts/1,2,3,4
+```
+
+```http
+HTTP/1.1 404 Not Found
```
dgeb: Should be `400`, since errors include `404` and `403` statuses.

examples/index.md
```diff
@@ -162,3 +162,209 @@ Content-Type: application/vnd.api+json
...
+Multiple errors on `password` attribute, with error "id":
+
+```http
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "id":   "123",
```
dgeb: These really are codes as well, if they match up with the table above. The point is that `id` represents a specific instance of an error, that may be trackable in a database for instance. If the same `id` could be reused across requests, it's really a `code`.

bf4: Updated. I'll rebase if it looks good. 

So, 'id' could be a uuid generated at error time and logged as part of that request/response. Gotcha.

examples/index.md
```diff
@@ -162,3 +162,209 @@ Content-Type: application/vnd.api+json
...
+    }
+  ]
+}
+```
+
+Response with status:
+
+```http
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
```
bf4, line 224: 400 Bad Request?

dgeb, line 224: > 400 Bad Request?

:+1:

examples/index.md
```diff
@@ -162,3 +162,226 @@ Content-Type: application/vnd.api+json
...
+
+{
+  "errors": [
+    {
+      "title":  "Missing Association",
+      "source": { "pointer": "data/relationships/author" },
+      "detail": "Post author cannot be blank."
+    }
+  ]
```
bf4: refs in the wild: 
- https://github.com/cerebris/jsonapi-resources/blob/28ac3971f8b9e73073575d41439ffe263d7e56f5/test/controllers/controller_test.rb#L452-L462

examples/index.md
```diff
@@ -162,3 +162,226 @@ Content-Type: application/vnd.api+json
...
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "title":  "Invalid Attribute",
+      "source": { "pointer": "data/attributes/first-name" },
+      "detail": "First name must contain at least three characters."
+    }
```
bf4, line 196: refs in the wild: 
- https://github.com/cerebris/jsonapi-resources/blob/28ac3971f8b9e73073575d41439ffe263d7e56f5/test/controllers/controller_test.rb#L2412-L2429
- https://github.com/chingor13/json_api_client/blob/cf79aba96cbc9ff66a793d07f43b025fd8f70b38/lib/json_api_client/error_collector.rb#L34-L48
- https://github.com/manyminds/api2go/blob/3023e50d563e6ec642b46a4f044812cb232661a1/error.go
- https://github.com/emberjs/data/blob/1f866da3b4a4edcf70fdf346cd6e6be6cfaecfab/packages/ember-data/lib/adapters/errors.js#L3-L154

examples/index.md
```diff
@@ -162,3 +162,249 @@ Content-Type: application/vnd.api+json
...
+```http
+HTTP/1.1 400 Bad Request
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "title":  "Missing Query Parameter",
+      "source": { "parameter": "data" },
+      "detail": "The required parameter, data, is missing."
```
bf4: I've found examples of these in the wild, but they don't appear to adhere to the spec.  I chose this example because I would have thought a POST with the data member mispelled, and thus missing, wouldn't count as a missing query parameter http://jsonapi.org/format/#query-parameters

This could also have been `"source": { "pointer": "/data" }`, right?

bf4:
@dgeb updated and rebased and added a few more source examples.   What would a meta example be in an error object (not top-level member meta)?

examples/index.md
ethanresnick: As I understand it, "query parameter" in the spec is referring strictly to query parameters in the request uri (like `sort`, `fields`, etc.), not to parameters in a broader sense (e.g. payload data).So a GET request that included a `fields[TYPE]` parameter whose value specified a field that doesn't exist on resources of that type would be an example of when to use the `parameter` member in the error response.

bf4: @ethanresnick Thanks for confirming my suspicions.  I glanced through some docs on the web and didn't immediately find anything that distinguished url params from body params when discussing query params.

And I suppose it would be up the implementation to decide whether to ignore an invalid or missing parameter or to return an error.

I think I can keep this example as a POST with body `{ "datum": { } }` to return an error object with `"source": { "pointer": "/data" }`  and add a new one for parameter, but I can't think of an  example where it wouldn't make sense for the client to ignore an invalid or missing parameter.

`GET http://www.example.com/api/users/1/secrets.json`  could return a `missing auth_key parameter`?

ethanresnick: > GET http://www.example.com/api/users/1/secrets.json could return a missing auth_key parameter?

That would be fine. I think returning an error (rather than ignoring) an invalid `fields` or `include` parameter would be fine too, but perhaps your example is safer.

bf4: I was thinking an implementor might decide to ignore invalid params that aren't required, so it wouldn't be as strong an example.  Does that jive with what you're thinking?

ethanresnick: Yup

bf4:
New and improved :sparkles:

examples/index.md
```diff
@@ -162,3 +162,292 @@ Content-Type: application/vnd.api+json
...
+```http
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "title": "Adapter Error",
+      "code": "ds.errors.invalid-error-expects-json-api-format",
+      "detail": "The adapter rejected the commit because it was invalid"
```
ethanresnick: Can you make the detail message here a bit more distinct than the `title` than it is now?

Also, can you add `status` to one of the earlier examples, in addition to having it on the one below? I don't want to imply through the examples that `status` is only applicable in cases where there are multiple errors returned at once. It can be useful too when the code processing the JSON isn't directly aware of the HTTP response (and could be useful also if JSON API is eventually used over non-HTTP protocols).

examples/index.md
```diff
@@ -162,3 +162,292 @@ Content-Type: application/vnd.api+json
...
+
+{
+  "errors": [
+    {
+      "status": "403",
+      "source": { "pointer": "data/attributes/secret-powers" },
+      "detail": "Access to secret powers not authorized on Sundays."
+    },
+    {
+      "status": "404",
```
ethanresnick: The `status` for this error object should be `400`, I think. Because 404 only applies if the resource isn't found, but the `vorpal-blade` attribute isn't it's own resource; it's just (an invalid) part of a bigger resource that _was_ found.

examples/index.md
```diff
@@ -162,3 +162,292 @@ Content-Type: application/vnd.api+json
...
+```
+
+The only uniqueness constraint on error objects in a response is the `id` field.
+Multiple errors on the same attribute may be described by an error object
+for each error.
+
+Multiple errors on `first-name` attribute:
+
+```http
+HTTP/1.1 422 Unprocessable Entity
```
ethanresnick, line 253: 422 works here, but 400 is [also valid](http://stackoverflow.com/a/20215807/1261879), and some people still don't like using status codes from WebDAV for non-webDAV apps.

Therefore, maybe you could add a line at the bottom of this example saying that a 400 response would also be ok, and linking to that SO answer? I don't want JSON API to appear to be taking a position on 400 vs. 422 by only using one in our examples.

examples/index.md
```diff
@@ -162,3 +162,292 @@ Content-Type: application/vnd.api+json
...
+top-level members, but not the "data" member.
+
+```http
+PATCH /posts/1
+
+{ "datum": { } }
+```
+
+```http
+HTTP/1.1 422 Bad Request
```
ethanresnick: Typo; this should be 422 Unprocessable Entity

examples/index.md
```diff
@@ -162,3 +162,292 @@ Content-Type: application/vnd.api+json
...
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "title":  "Missing Data Member",
+      "source": { "source": "/data" },
+      "detail": "The required parameter, data, is missing."
```
ethanresnick: Can you get rid of `detail` here? It basically restates the title, so it isn't needed. More importantly though, I don't want to call `data` a "parameter" since that further blurs the distinction between `"source"` and `"parameter"`; thus far, the only thing we've called "parameters" are the URI query params.

examples/index.md
```diff
@@ -162,3 +162,292 @@ Content-Type: application/vnd.api+json
...
+Notice that the "source" indicates the error was causes not by a missing attribute, but
+by missing "data".
+
+
+### Responses to failed resource request:
+
+Basic request:
+
+```http
+GET /posts/1,2,3,4
```
ethanresnick: `/posts/1,2,3,4` is not a URI format that we recommend any more because (esp. with UUIDs) the URI length can somewhat quickly exceed the maximum length that many browsers/servers support. Personally, I think we should still be recommending this format for many cases (because it works, and offers a performance boost, 80% of the time). But, unless we add it back as a recommendation, I don't think we should have an example that uses it. 

Also, since `parameter` just refers to URI query parameters in the RFC 3986 sense, treating `/1,2,3,4` as a source of `"parameter"` errors is invalid.

There's probably a way to rework this example with a new URI to make it compliant, but maybe we should just get rid of it? After all, you already have a very good description of `"parameter"` in the next example.

examples/index.md
```diff
@@ -162,3 +162,292 @@ Content-Type: application/vnd.api+json
...
+Notice that the "source" indicates the query parameter which caused the error.
+This example with a missing "auth_key" parameter was chosen over an example with an invalid
+query parameter, since it may make sense to ignore invalid parameters in some cases, rather
+than return an error.  The "error" and "data" keys may not co-exist in the same response.
+Example of invalid paramters may be "?page[prev]=" (parameter present, but has no value),
+"?include=auther" (invalid parameter value, perhaps a typo), "?felds[author]=",
+(invalid paramter key, perhaps a typo), "?redirect_to=http%3A%2F%2Fwww.owasp.com" (invalid
+parameter, in this case, a phishing attack), etc.
+
+Response with one error object, with parameter error:
```
ethanresnick: Dangling line.

ethanresnick:
Hey @bf4,

Thanks again for this! I just had a chance to review the whole PR from my computer, and I've added inline comments for some final things that I think need tweaking. Once those are fixed, I think this'll be good to go!

examples/index.md
bf4: hangs head in shame :frowning:

examples/index.md
bf4, line 253: Thanks. I admit I was unsure of which to choose in some of these examples.

examples/index.md
bf4: Hmm, I wanted a 404 example.  I'll update this. I intended this to reference a link, not an attribute.

examples/index.md
bf4: Sure.  FWIW, I got that example title/code/detail off of the ember-data project :).

examples/index.md
bf4: Updated to 

``` json
{
      "status": "404",
      "links": { "about": {
        "href": "http://example.com/weapons/vorpal-blade",
        },
      },
      "title": "Weapon not found."
}
```

I'm a little unclear on how the error links object should work.

bf4:
@ethanresnick Updated, though I admit I'm getting a little code blindness.  Let me know if I should move the extension examples and the 'links: about' example into another pr if the rest could be merged sooner.

ethanresnick:
@bf4 Awesome, this is getting close now. Can you squash these commits down to one? Then, do you mind if I send a PR to your branch with a couple small, outstanding fixes—I think that's easier than asking you to do another round of revisions—and then we'll merge this?

bf4:
Great, will do

examples/index.md
```diff
@@ -162,3 +162,354 @@ Content-Type: application/vnd.api+json
...
+
+```http
+HTTP/1.1 422 Unprocesssable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "title":  "Missing Data Member",
+      "source": { "source": "/data" }
```
swalkinshaw: Is the 2nd `source` a mistake? Spec only mentions `pointer` and `parameter`. There's another occurrence of this in the very last example if so.

bf4: Good catch! It should be pointer :)

bf4:
@ethanresnick Rebased and updated with @swalkinshaw's catch where I had `source: { source: "" }` instead of `{source: {pointer: ""}`. I also changed all the the words I had in quotes to backticks to be more consistent and because they show up better.  Thanks to all!

Other thoughts:
- I don't really like how I've separated each example, but didn't prefer h4 nor hr nor bold, so left it for now.
- I think the two extensions examples should go on their respective pages, and be referenced here, rather than inline.
- It could probably do examples with an "id", and for DELETE
- It doesn't really address included relations or nested relations

examples/index.md
bf4: I think this could also have been, instead of links,  a `"source": { "pointer": "data/relationships/weapons"}`

examples/index.md
```diff
@@ -162,3 +162,241 @@ Content-Type: application/vnd.api+json
...
+      "title": "The backend responded with an error",
+      "detail": "Reputation service not responding after three requests."
+    }
+  ]
+}
+```
+
+The only uniqueness constraint on error objects is the `id` field. Thus,
+multiple errors on the same attribute can each be given their own error
+object. The example below shows multiple errors on the `first-name` attribute:
```
bf4: I believe we're formatting attribute names as `"first-name"`, so, note to self :)

examples/index.md
```diff
@@ -162,3 +162,241 @@ Content-Type: application/vnd.api+json
...
+      "title": "Invalid Attribute",
+      "detail": "First name must contain an emoji."
+    }
+  ]
+}
+```
+
+> Note: in the responses above with a 422 status code, `400 Bad Request` would
+also be acceptable. ([http://stackoverflow.com/a/20215807/1261879](More details.))
+JSON API doesn't take a position on 400 vs. 422.
```
bf4, line 274: extraneous `>`

examples/index.md
```diff
@@ -162,3 +162,241 @@ Content-Type: application/vnd.api+json
...
+For the example below, imagine the API docs specifed the following mapping:
+
+> | Code | Problem                                                   |
+> |------|-----------------------------------------------------------|
+> |  123 | Value too short                                           |
+> |  225 | Password lacks a letter, number, or punctuation character |
+> |  226 | Passwords do not match                                    |
+> |  227 | Password cannot be one of last five passwords             |
+
+Multiple errors on `password` attribute, with error `code`:
```
bf4: See above, should be `"password"` to be consistent with attribute quoting (note to self)

examples/index.md
ethanresnick, line 274: This is there to mark the section as an indented note. Cf the note above that says "The above example URI shows unencoded [ and ]..."

examples/index.md
ethanresnick: Yes, quoting the attribute values would be more consistent with the base spec. There, I believe we actually quote _all_ member values. So, probably, we should be quoting `code`, `detail`, `title`, etc. Honestly, though, it's not a big deal (to me anyway) one way or the other.

examples/index.md
```diff
@@ -162,3 +162,241 @@ Content-Type: application/vnd.api+json
...
+    }
+  ]
+}
+```
+
+In most cases, JSON API requires the server to return an error when it encounters
+an invalid value for a JSON API–defined query parameter. However, for API-specific
+query parameters (i.e. those not defined by JSON API), a server may choose to
+ignore an invalid parameter and have the request succeed, rather than respond with
+an error. API-specific query paramters must contain one non a-z character.
```
bf4: typo `parameter`, also, on first read, it's unclear to me what it means for API-specific query parameters to contain a non a-z character and I haven't yet found it in the `schema` or the format page.

? http://jsonapi.org/format/#document-member-names

> To enable an easy mapping of member names to URLs, it is RECOMMENDED that member names use only non-reserved, URL safe characters specified in RFC 3986.

ethanresnick: Thanks for catching the typo! Can you fix that, so I don't need to make a new PR?

Re api specific query parameters, here's the reference:

> "Implementation specific query parameters MUST adhere to the same constraints as member names with the additional requirement that they MUST contain at least one non a-z character (U+0061 to U+007A)."

Then, later:

> "This is to preserve the ability of JSON API to make additive additions to standard query parameters without conflicting with existing implementations."

examples/index.md
bf4, line 274: yeah.. noticed that.. 'noted'

examples/index.md
bf4: Ok, thanks, found it: http://jsonapi.org/format/#query-parameters  Don't know why I didn't see it.

What would be an example of this?  `auth_key=` is a valid api-specific query param but  `authkey=` is not?

ethanresnick: > What would be an example of this? auth_key= is a valid api-specific query param but authkey= is not?

Correct!

bf4:
@ethanresnick I amended the merge commit with those small changes.

Good for a final review by dgeb steveklabnik etc?

I'm bringing down some notes about future work from above:

> I think the two extensions examples should go on their respective pages, and be referenced here, rather than inline.

That'll be a new PR I'll make based on the excised material

> It could probably do examples with an "id", and for DELETE

I'll make another PR if there's interest

> It doesn't really address included relations or nested relations

I'm not sure what that would look like offhand. 
- I'd also like to add a proper `links` attribute as discussed in https://github.com/bf4/json-api/pull/1#discussion_r37645336
- Maybe add a note about handling the top level member `json-api` (which should be `jsonapi`) per https://github.com/bf4/json-api/pull/1#discussion_r37646127

examples/index.md
```diff
@@ -162,3 +162,242 @@ Content-Type: application/vnd.api+json
...
+  ]
+}
+```
+
+In most cases, JSON API requires the server to return an error when it encounters
+an invalid value for a JSON API–defined query parameter. However, for API-specific
+query parameters (i.e. those not defined by JSON API), a server may choose to
+ignore an invalid parameter and have the request succeed, rather than respond with
+an error. [API-specific query parameters must contain one non a-z
+character.](http://jsonapi.org/format/#query-parameters)
```
bf4, line 410: I added this link to the merge commit

bf4:
btw, I'm on the [#jsonapi channel](irc://irc.freenode.net/#jsonapi)

examples/index.md
```diff
@@ -162,3 +162,242 @@ Content-Type: application/vnd.api+json
...
+
+```http
+HTTP/1.1 400 Bad Request
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "status": "403",
+      "source": { "pointer": "data/attributes/secret-powers" },
```
ethanresnick: Pointer should start with a `/`

examples/index.md
```diff
@@ -162,3 +162,242 @@ Content-Type: application/vnd.api+json
...
+      "detail": "Editing secret powers is not authorized on Sundays."
+    },
+    {
+      "status": "422",
+      "source": { "pointer": "data/attributes/volume" },
```
ethanresnick: Ditto

examples/index.md
```diff
@@ -162,3 +162,242 @@ Content-Type: application/vnd.api+json
...
+      "detail": "Volume does not, in fact, go to 11."
+    },
+    {
+      "status": "500",
+      "source": { "pointer": "data/attributes/reputation" },
```
ethanresnick: Ditto

examples/index.md
```diff
@@ -162,3 +162,242 @@ Content-Type: application/vnd.api+json
...
+object. The example below shows multiple errors on the `"first-name"` attribute:
+
+```http
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "source": { "pointer": "data/attributes/first-name" },
```
ethanresnick: Ditto

examples/index.md
```diff
@@ -162,3 +162,242 @@ Content-Type: application/vnd.api+json
...
+      "title": "Invalid Attribute",
+      "detail": "First name must contain at least three characters."
+    },
+    {
+      "source": { "pointer": "data/attributes/first-name" },
```
ethanresnick: Ditto

examples/index.md
```diff
@@ -162,3 +162,242 @@ Content-Type: application/vnd.api+json
...
+```http
+HTTP/1.1 422 Unprocessable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "jsonapi": { "version": "1.0" },
+  "errors": [
+    {
+      "code":   "123",
+      "source": { "pointer": "data/attributes/first-name" },
```
ethanresnick: Ditto

examples/index.md
```diff
@@ -162,3 +162,242 @@ Content-Type: application/vnd.api+json
...
+      "title":  "Value is too short",
+      "detail": "First name must contain at least three characters."
+    },
+    {
+      "code":   "225",
+      "source": { "pointer": "data/attributes/password" },
```
ethanresnick: Ditto

examples/index.md
```diff
@@ -162,3 +162,242 @@ Content-Type: application/vnd.api+json
...
+      "title": "Passwords must contain a letter, number, and punctuation character.",
+      "detail": "The password provided is missing a punctuation character."
+    },
+    {
+      "code":   "226",
+      "source": { "pointer": "data/attributes/password" },
```
ethanresnick: Ditto

examples/index.md
bf4: oy

examples/index.md
```diff
@@ -162,3 +162,242 @@ Content-Type: application/vnd.api+json
...
+Therefore, the server responds:
+
+```http
+HTTP/1.1 422 Unprocesssable Entity
+Content-Type: application/vnd.api+json
+
+{
+  "errors": [
+    {
+      "source": { "pointer": "/" },
```
ethanresnick: My bad. This pointer should actually be the empty string `""`. Using `"/"` refers to a property `""` at the top level, i.e.

``` js
{
  "": "some value"    // The json pointer "/" would reference the string "some value"
}
```

examples/index.md
ethanresnick, line 410: :+1:

examples/index.md
bf4: It would be really great were there an easy way to validate example code against the schema.  That easiest right now would be a task that parses fenced code blocks, identifies the document, and then can validate that..

ethanresnick:
@bf4 Thanks for all your hard work! This looks great to me now.

Please fix the small errors with the JSON Pointers, and then I think this is good to merge, unless @dgeb or @tkellen want to review it first.

bf4:
Merge commit amended and force-pushed.

examples/index.md
```diff
@@ -162,3 +162,242 @@ Content-Type: application/vnd.api+json
...
+  "errors": [
+    {
+      "source": { "pointer": "" }, // The json pointer "/" would reference the string "some value"
+      "detail":  "Missing `data` Member at document's top level."
+    }
+  ]
+}
+```
+
+It uses `source` to point to the top-level of the document (`"/"`).
```
ethanresnick: This line needs to say `""` instead of the old `"/"`. 

Also, I think my comment about why `"/"` is inappropriate got accidentally merged into the response. It doesn't make sense there, though, because my example document with `{"": "some value"}` is missing, and because a server isn't allowed to respond with comments in its JSON. If you want to explain why `""` is used rather than `"/"`, please do it in this paragraph instead.

ethanresnick:
:+1:

examples/index.md
bf4: you a copy editor or I'm just tired of re-reading this ? :)

ethanresnick: Maybe both :)

bf4: Added an incremental commit to address this to aid in future reviewing. I'll rebase into the merge commit once we're happy.

bf4:
fixing merge conflict... hold on

bf4:
rebased

bf4:
@ethanresnick LGTM?

ethanresnick:
LGTM. Merged.

bf4:
omg omg :fireworks: :rainbow: :exclamation: :boom: :dancers: :heart: :heart_eyes: :smiley: :100:

joaomdmoura:
:tada: :tada: :tada: :tada: :tada:

dgeb:
Really pleased to see this merged. We've needed some error examples for months.

@bf4 thanks for all your work here! :100: 

@ethanresnick thanks for finishing the review! :tada:

Issue #829, opened 2015-08-06
Schema causes RegexpError: empty range in char class


kjwierenga:
I'm using the Ruby JSON Schema Validator (https://github.com/ruby-json-schema/json-schema) and an accompanying matcher to match API responses to JSON schema definitions.

When I match against the JSONAPI schema found at http://jsonapi.org/schema I get the following error:

```
 Failure/Error: response.body.should match_json_schema('json-api')
 RegexpError:
   empty range in char class: /^(?!relationships$|links$)\w[\w-_]*$/
```

The `[\w-_]` char class seems to be a problem.
See also: http://stackoverflow.com/questions/13161903/regular-expression-empty-range-in-char-class-error

The solution is to move the dash to the front of the char class like so: `[-\w_]`. I'll supply a PR shortly.

bf4:
Closed by https://github.com/json-api/json-api/pull/830 :)

PR #830, opened 2015-08-06
Fix invalid char class by moving the dash to the front of the char class.

kjwierenga:
See http://stackoverflow.com/questions/13161903/regular-expression-empty-range-in-char-class-error

Moving the dash to the front of the char class fixes issues  #829


tkellen:
Thanks!

PR #832, opened 2015-08-07
Pagination links are allowed to have `null` value.

kjwierenga:
See: http://jsonapi.org/format/#fetching-pagination

Keys MUST either be omitted or have a null value to indicate that a particular link is unavailable.


tkellen:
:+1: thanks!

kjwierenga:
Thanks for merging!

PR #834, opened 2015-08-08
add example of pagination links

bacarini:
We are adding a new feature on [AMS](https://github.com/rails-api/active_model_serializers) and to make sure we're getting it right, I think it is important to validate and add examples on `json-api`.

That is the [PR to support pagination links](https://github.com/rails-api/active_model_serializers/pull/1041) on AMS

What do you thing about that?

### DIFF
--- a/examples/index.md
+++ b/examples/index.md
@@ -154,6 +154,7 @@ Content-Type: application/vnd.api+json
     }
   ],
   "links": {
+    "self": "http://example.com/articles?page[number]=3&page[size]=1",
     "first": "http://example.com/articles?page[number]=1&page[size]=1",
     "prev": "http://example.com/articles?page[number]=2&page[size]=1",
     "next": "http://example.com/articles?page[number]=4&page[size]=1",
--- a/examples/index.md
+++ b/examples/index.md
@@ -126,3 +126,38 @@ Content-Type: application/vnd.api+json
 for readability. In practice, these characters must be percent-encoded, as
 noted in the base specification.
 
+## Pagination Links
+
+Example of how to add [pagination links](http://jsonapi.org/format/#fetching-pagination).
+
+Basic request:
+
+```http
+GET /articles?page=3&per_page=1
+```
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/vnd.api+json
+
+{
+  "data": [
+    {
+      "type": "articles",
+      "id": "1",
+      "attributes": {
+        "title": "JSON API paints my bikeshed!",
+        "body": "The shortest article. Ever.",
+        "created": "2015-05-22T14:56:29.000Z",
+        "updated": "2015-05-22T14:56:28.000Z"
+      }
+    }
+  ],
+  "links": {
+    "first": "?page=1&per_page=1",
+    "prev": "?page=2&per_page=1",
+    "next": "?page=4&per_page=1",
+    "last": "?page=13&per_page=1"
+  }
+}
+```
--- a/examples/index.md
+++ b/examples/index.md
@@ -154,10 +154,10 @@ Content-Type: application/vnd.api+json
     }
   ],
   "links": {
-    "first": "?page=1&per_page=1",
-    "prev": "?page=2&per_page=1",
-    "next": "?page=4&per_page=1",
-    "last": "?page=13&per_page=1"
+    "first": "http://example.com/articles?page=1&per_page=1",
+    "prev": "http://example.com/articles?page=2&per_page=1",
+    "next": "http://example.com/articles?page=4&per_page=1",
+    "last": "http://example.com/articles?page=13&per_page=1"
   }
 }
 ```
--- a/examples/index.md
+++ b/examples/index.md
@@ -144,7 +144,7 @@ Content-Type: application/vnd.api+json
   "data": [
     {
       "type": "articles",
-      "id": "1",
+      "id": "3",
       "attributes": {
         "title": "JSON API paints my bikeshed!",
         "body": "The shortest article. Ever.",
--- a/examples/index.md
+++ b/examples/index.md
@@ -128,12 +128,12 @@ noted in the base specification.
 
 ## Pagination Links
 
-Example of how to add [pagination links](http://jsonapi.org/format/#fetching-pagination).
+Example of a page-based strategy on how to add [pagination links](http://jsonapi.org/format/#fetching-pagination).
 
 Basic request:
 
 ```http
-GET /articles?page=3&per_page=1
+GET /articles?page[number]=3&page[size]=1
 ```
 
 ```http
@@ -154,10 +154,10 @@ Content-Type: application/vnd.api+json
     }
   ],
   "links": {
-    "first": "http://example.com/articles?page=1&per_page=1",
-    "prev": "http://example.com/articles?page=2&per_page=1",
-    "next": "http://example.com/articles?page=4&per_page=1",
-    "last": "http://example.com/articles?page=13&per_page=1"
+    "first": "http://example.com/articles?page[number]=1&page[size]=1",
+    "prev": "http://example.com/articles?page[number]=2&page[size]=1",
+    "next": "http://example.com/articles?page[number]=4&page[size]=1",
+    "last": "http://example.com/articles?page[number]=13&page[size]=1"
   }
 }
 ```


examples/index.md
```diff
@@ -126,3 +126,38 @@ Content-Type: application/vnd.api+json
...
+      "attributes": {
+        "title": "JSON API paints my bikeshed!",
+        "body": "The shortest article. Ever.",
+        "created": "2015-05-22T14:56:29.000Z",
+        "updated": "2015-05-22T14:56:28.000Z"
+      }
+    }
+  ],
+  "links": {
+    "first": "?page=1&per_page=1",
```
ethanresnick: I'm pretty sure these need to be complete URIs, not just query strings, since a JSON API document doesn't have a base uri. The spec could specify in the future that the `self` link is the base uri, but right now it doesn't (and the `self` link isn't even mandatory).

bacarini:
@ethanresnick :+1:

gr0uch:
> I'm pretty sure these need to be complete URIs

@ethanresnick Just to clarify, an absolute path without host is still a "complete URI" correct? i.e. `/collection?query`

bacarini:
specified which strategy on this exmaple as suggested in this note http://jsonapi.org/format/#fetching-pagination:

> Note: JSON API is agnostic about the pagination strategy used by a server.
> Effective pagination strategies include (but are not limited to):
> page-based, offset-based, and cursor-based. The `page` query parameter can
> be used as a basis for any of these strategies. For example, a page-based
> strategy might use query parameters such as `page[number]` and `page[size]`,
> an offset-based strategy might use `page[offset]` and `page[limit]`, while a
> cursor-based strategy might use `page[cursor]`.

examples/index.md
```diff
@@ -126,3 +126,38 @@ Content-Type: application/vnd.api+json
...
+      "attributes": {
+        "title": "JSON API paints my bikeshed!",
+        "body": "The shortest article. Ever.",
+        "created": "2015-05-22T14:56:29.000Z",
+        "updated": "2015-05-22T14:56:28.000Z"
+      }
+    }
+  ],
+  "links": {
+    "first": "http://example.com/articles?page[number]=1&page[size]=1",
```
bf4, line 158: maybe add a 'self' link to show we're current on page 3?

bacarini, line 158: @bf4  :+1:

bf4:
@dgeb @steveklabnik Thoughts on this example of pagination links?

examples/index.md
```diff
@@ -126,3 +126,39 @@ Content-Type: application/vnd.api+json
...
 noted in the base specification.
 
+## Pagination Links
+
+Example of a page-based strategy on how to add [pagination links](http://jsonapi.org/format/#fetching-pagination).
+
+Basic request:
+
+```http
+GET /articles?page[number]=3&page[size]=1
```
steveklabnik, line 136: Isn't this `%5B` and `%5D` rather than `[` and `]`?

examples/index.md
```diff
@@ -126,3 +126,39 @@ Content-Type: application/vnd.api+json
...
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/vnd.api+json
+
+{
+  "data": [
+    {
+      "type": "articles",
+      "id": "1",
```
steveklabnik: if we're on page 3, this should maybe not be ID 1

steveklabnik:
two small nits, but looks great!

bacarini:
Hi @steveklabnik, before I commit the changes is that okay? Isn't a nicer presentation with `[`and `]`?

``` http
GET /articles?page%5Bnumber%5D=3&page%5Bsize%5D=1
```

``` http
HTTP/1.1 200 OK

{
  "data": [
    {
      "type": "articles",
      "id": "3",
      "attributes": {
        "body": "The shortest article. Ever.",
        "created": "2015-05-22T14:56:29.000Z",
        "updated": "2015-05-22T14:56:28.000Z"
      }
    }
  ],
  "links": {
    "self": "http://example.com/articles?page%5Bnumber%5D=3&page%5Bsize%5D=1",
    "first": "http://example.com/articles?page%5Bnumber%5D=1&page%5Bsize%5D=1",
    "prev": "http://example.com/articles?page%5Bnumber%5D=2&page%5Bsize%5D=1",
    "next": "http://example.com/articles?page%5Bnumber%5D=4&page%5Bsize%5D=1",
    "last": "http://example.com/articles?page%5Bnumber%5D=13&page%5Bsize%5D=1"
  }
}
```

steveklabnik:
I do agree it looks nicer, it's just an invalid URL. :/ What do you think, @dgeb ?

dgeb:
In other examples, we've gone with the unencoded characters for clarity (see the home page example for instance), so @bacarini's being consistent here.

If we want to make a change to encoding all chars, we should do so throughout.

steveklabnik:
Okay, then I'm fine with not encoding them here.

bacarini:
:+1: all good to merge it?

dgeb:
LGTM - thanks @bacarini :+1:

bacarini:
tks guys!

bf4:
:100:

ethanresnick:
> @ethanresnick Just to clarify, an absolute path without host is still a "complete URI" correct? i.e. /collection?query

@0x8890 Sorry I'm just responding to this. Revising my original position a bit... I think that providing a partial URI, or, in RFC 3968 parlance, providing a ["relative reference"](https://tools.ietf.org/html/rfc3986#section-4.2) would be allowed but not ideal. That's true whether the relative reference uses an absolute path (as in your example), a relative path, or (as in this PR originally) just a query string. Any of those forms should be technically allowed because RFC 3968 provides [a mechanism](https://tools.ietf.org/html/rfc3986#section-5.1.3) for clients to determine the full URI in these cases. Namely, the client is supposed to use the URI at which the resource was requested as the base URI.

Still, I think a full URI is preferable, in case the document needs to be parsed without knowledge of how it was originally requested; this is the same reason a server might provide a `self` link. Moreover, unless/until the JSON API spec defines how the `self` link affects the way a base url is calculated, relying on a relative reference could lead compliant clients to construct an incorrect URI. Consider the following:

```
GET /articles/1

{
  "data": {
    //...
    "relationships": {
       "comments": {
         "links": {
           "data": [// first 10 resource identifier objects],
           "next": "?page[number]=2&include=comments"
         }
       }
    }
  }
}
```

Above, the `"next"` link would be interpreted, per RFC 3986, as having the target URI: `http://example.com/articles/1?page[number]=2&include=comments` when it "should" be `http://example.com/articles/1/relationships/comments?page[number]=2&include=comments`. The second interpretation doesn't apply because JSON API doesn't specifically describe the `self` link in the relationship object as constituting a new, embedded base URI (as would be required, I think, under [RFC 3986, section 5.1.1](https://tools.ietf.org/html/rfc3986#section-5.1.1)).

So, again, I'd suggest using full URIs whenever possible. However, I think that relative references should be allowed, strictly speaking—though @dgeb, @steveklabnik, or @tkellen may have a different interpretation. Moreover, your comment suggests that JSON API would benefit from a more precise definition of the interplay between `self` links and RFC 3986 base uris. Would you be open to working up a PR for that?

ethanresnick:
Changing my position on this again. From the RFC: "A URI is an identifier consisting of a sequence of characters matching the syntax rule named <URI> in [Section 3](https://tools.ietf.org/html/rfc3986#section-3)." Therefore, a relative reference is a string that can be converted to a URI, but it's not a URI itself. And, since JSON API says the `"self"` key et al. must hold a URI, it follows that relative references are simply not allowed. And that's probably a good thing, since extending the spec to allow them would probably break existing clients.

gr0uch:
@ethanresnick does that mean that only paths are not allowed anywhere?

ethanresnick:
Yeah, that's what it would mean. Per my comment in the other issue, though, I'm open to other solutions if you have ideas. I just think that many clients aren't currently expecting relative references (we don't use them in any examples) and would handle them wrong, and I don't think we can blame clients for that, since a relative reference isn't technically compliant, per my logic above

tkellen:
I discussed this with @dgeb way back in the day and we had landed on relative URIs being allowable. We do in fact have examples showing this. There is one shown in the [Resource Object](http://jsonapi.org/format/#document-resource-objects) section. That said, I'm on the fence about the value of allowing them at this point. Whatever we decide to do, it would be nice to make a non-normative clarification one way or the other.

ethanresnick:
> I discussed this with @dgeb way back in the day and we had landed on relative URIs being allowable. We do in fact have examples showing this. 

Ahh, shoot. I totally missed those examples. That complicates things a bit (though the examples aren't normative at least, so taking them out now probably wouldn't be a bc break). Curious what @dgeb thinks.

dgeb:
> Whatever we decide to do, it would be nice to make a non-normative clarification one way or the other.

Agreed.

I think that there is some confusion introduced because in many historical documents, the general term for an absolute URI or relative reference that can be converted into an absolute URI, is simply "URI". Take for example [RFC 1866](https://tools.ietf.org/html/rfc1866#page-34) which defines the `HREF` attribute of an `A` tag as: "gives the URI of the head anchor of a hyperlink." It goes on to provide examples which clearly indicate relative references.

It seems that in [RFC 3986 section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1) this concept of a "URI / relative reference" is defined as a "URI-reference".

As @tkellen said, he and I have always intended for relative references to be allowed but not recommended. Given the rather loose ways in which "URI" is used in other RFCs, I think we have some wiggle room to allow for its usage to mean "URI-reference". I am curious to know if there is further evidence of the usage of URI-reference vs. URI in RFCs more recent than 3986.

ethanresnick:
> Given the rather loose ways in which "URI" is used in other RFCs, I think we have some wiggle room to allow for its usage to mean "URI-reference".

I can see that. In fact, even the latest HTML spec seems to use the term "URL" to include what RFC 3986 calls relative references. 

If we do (continue to) allow relative references, though, that means we have to reopen #840 and define a procedure for resolving the base URI (which, in an ideal world, we probably don't want to be the same at the document, resource object, and relationship object levels). Who knows how clients have been resolving this so far, so we'd probably need a real-world audit too...

tkellen:
I think it's safe to say there aren't that many implementations in the wild using cross-domain links yet. For example, ember-data works just fine with domain relative "absolute" URIs.

ethanresnick:
> I think it's safe to say there aren't that many implementations in the wild using cross-domain links yet.

Right, I wasn't talking about cross-domain links so much as what path on the current domain to use. E.g.

``` js
GET /articles
Host: example.com

{
  "links": {
    // are clients currently resolving this as:
    // http://example.com/?page=2 or
    // http://example.com/articles?page=2
    // hopefully (probably?), it's the second.
    "next": "?page=2"
  }
  "data": [ /* ... */ ]
}
```

And, what I was saying about "in an ideal world, we probably don't want [the base url] be the same at the document, resource object, and relationship object levels" was:

``` js
GET /articles

{ 
  "data": [{
    "type": "articles",
    "id": "1",
    "links": {
    },
    "relationships": {
      "author": {
        "links": {

          // I would want, in an ideal world, for this to resolve to 
          // http://example.com/articles/1/author (that is, the self uri
          // in the same link object should form the base uri), but 
          // i almost guarantee clients aren't resolving it that way.
          "related": "../author"
        }
      }
    }
  }]
}
```

PR #836, opened 2015-08-12
First PHP client implementation: art4/json-api-client

Art4:
I would like to introduce a new client implementation in PHP: [art4/json-api-client](https://github.com/Art4/json-api-client)
Because there were no other client library in PHP I wrote it myself. It can be use as a client and as a validator.

The library is fully testet and should fit the json api specification nearly 100%. If you find an error I would be very happy about an issue or PR.


steveklabnik:
Awesome! Thanks so much. Glad to have some PHP on board!

PR #844, opened 2015-08-17
Update link to JSON API serializer

gr0uch:
I've moved all included serializers out of Fortune.js and into separate modules. Now there are links to both Fortune.js and the JSON API serializer.


ethanresnick:
Awesome!

PR #846, opened 2015-08-19
Added another client library and changed server library link

alex94cp:
I'm the author of both.


steveklabnik:
Thank you! Yeah, nice to have a real link rather than the shortened one :+1:

PR #848, opened 2015-08-20
Better explain why we don't use PUT

ethanresnick:
This is something that people are still [asking about](https://github.com/json-api/json-api/issues/419#issuecomment-132851928) in the issues and at least one [post](http://phalt.co/api-response-formats/) online has also said our current answer is no good.

In addition to better explaining the rationale for PATCH, this PR also removes the note about users being able to define their own PUT semantics since, until we have an extension mechanism, that may not be
a good idea.

### DIFF
--- a/faq/index.md
+++ b/faq/index.md
@@ -56,13 +56,26 @@ and capabilities and use the errors response to let users know. This error featu
 is still pending to be included in the standard since is still in
 [discussion](https://github.com/json-api/json-api/issues/7).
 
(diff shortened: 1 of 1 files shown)


steveklabnik:
:+1:

hubert:
this is a nice, concise explanation.

tkellen:
:+1: good to merge if you agree @dgeb!

dgeb:
Thanks for writing this up, @ethanresnick :+1:

benoittgt:
Thanks for this PR. Very appreciated.

PR #849, opened 2015-08-20
Update index.md

afilina:
Explicit encoding mention for pagination example.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -1095,6 +1095,10 @@ strategy might use query parameters such as `page[number]` and `page[size]`,
 an offset-based strategy might use `page[offset]` and `page[limit]`, while a
 cursor-based strategy might use `page[cursor]`.
 
+> Note: The above example URI shows unencoded `[` and `]` characters simply
+for readability. In practice, these characters must be percent-encoded, as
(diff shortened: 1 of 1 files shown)


format/index.md
```diff
@@ -1095,6 +1095,10 @@ strategy might use query parameters such as `page[number]` and `page[size]`,
 an offset-based strategy might use `page[offset]` and `page[limit]`, while a
 cursor-based strategy might use `page[cursor]`.
 
+> Note: The above example URI shows unencoded `[` and `]` characters simply
```
ethanresnick, line 1098: Thanks again for this!

Two small things: since we don't actually have a URI here (but query parameters), I think this should say "The above example query parameters use unencoded `[` and `]` characters simply for readability." Then, since this note is already in the base spec, it should instead end: "In practice, these characters must be percent-encoded, per the requirements in [RFC 3986](http://tools.ietf.org/html/rfc3986#section-3.4)."

bf4, line 1098: :+1:

ethanresnick:
@afilina Merged as part of #861. Thanks again!

Issue #855, opened 2015-08-23
the format page can be difficult to navigate


fivetanley:
Hi JSONAPI team!

I'm reading throughs some of the spec for some Ember Data implementation details. I'm using Chrome, and the menu can be kind of difficult to navigate because I have to scroll down to the bottom of the page in order to see the last few menu items, instead of the menu being scrolled separately.

Here's a video demonstrating the issue: ![video demonstrating issue](https://dl.dropboxusercontent.com/s/0h29u2bq2vlk5el/2015-08-23%20at%2010.58%20AM.gif)

(sorry for the quality, the tool i used wasn't that good)

And a screenshot of what happens when you do visit the bottom of the page. ![screenshot demonstrating scroll issue](https://dl.dropboxusercontent.com/s/jrfkfe8fleq5lq4/2015-08-23%20at%2010.56%20AM.png)

ethanresnick:
Thanks for submitting! We already have a couple issues for this, though your report inspired me to take another crack at a solution, so I'm closing this in favor of those tickets. See https://github.com/json-api/json-api/pull/856 and #831

PR #856, opened 2015-08-23
Make the sidebar scroll independently on too-small viewports

### LINKED ISSUE #831: /format sidebar overflow is not accessible
http://jsonapi.org/format/ accessed by Chrome Version 44.0.2403.130 (64-bit)

the information in the sidebar on the Format page extends below the window and is not accessbile via scroll. it seems like a result of fixed positioning.

might investigate later and make a PR but s'cool if someone wants to take over before i get to it.

ethanresnick:
This should fix #831/#855. It's an alternative to the fix in #785. @tkellen, @dgeb, @steveklabnik please merge whichever version you prefer. @wvteijlingen is welcome to chime in too!

This PR makes the sidebar scroll independently when it overflows (as requested in #855), while the other PR makes it not `position: fixed` if it can't fit.

### DIFF
--- a/javascripts/all.js
+++ b/javascripts/all.js
@@ -23,11 +23,21 @@ function fixElement(element, offset) {
         var scrollPosition = $(window).scrollTop();
 
         if(scrollPosition >= affixWaypoint) {
-            element.css('position', 'fixed');
-            element.css('top', offset + 'px');
+            element.css({
+                position: 'fixed',
+                top: offset + 'px',
+                bottom: '0',
+                overflow: 'scroll',
+                paddingBottom: offset + 'px'
+            });
         } else {
-            element.css('position', 'relative');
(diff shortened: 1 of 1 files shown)


ethanresnick:
:+1:

tkellen:
Thanks @ethanresnick !

wvteijlingen:
I had this solution in mind first, but unfortunately this doesn't work reliably in Safari. Often the sidebar is not scrollable and the main page scrolls instead. If you don't mind that, then this is ok of course. :)

ethanresnick:
@wvteijlingen Hmm... I see the behavior you're talking about, but have no idea why it's happening. Seems like a Safari bug. Do you know if this is documented anywhere or if there's an easy fix?

_EDIT_: I think I found a fix for Safari. Will post soon.

PR #857, opened 2015-08-23
Amendment to #856

ethanresnick:
Some small changes to make the scrolling effect a bit smoother. Now, only the list items will scroll, while the sidebar's heading "Format" stays put. This makes the scroll look far less awkward.

/cc @tkellen


tkellen:
:+1: that is a lot nicer

Issue #858, opened 2015-08-24
formatting is really messed up on Firefox/Linux


steveklabnik:
![2015-08-23-200848_604x511_scrot](https://cloud.githubusercontent.com/assets/27786/9430913/c3dbb638-49d2-11e5-9498-beceda08c50b.png)

O_O

jamesplease:
![](http://www.countryhealthsa.sa.gov.au/Portals/0/Graphics/IE%20Best%20viewed-paint7.png)

ethanresnick:
Should be fixed by #860.

PR #860, opened 2015-08-24
Sidebar fixes

ethanresnick:


ethanresnick:
Ok, this PR's ready for review.

Basically, it does two things:
1. It offers a partial fix to @steveklabnik's #858 by not showing a horizontal scrollbar, and only showing the vertical scrollbar when there's actually something to scroll. It also adds a bit more space between the sidebar and the spec text. The vertical scrollbar, when present, remains kinda obtrusive on non-OSX platforms, but I don't think there's much we can do about that; it's just those platforms' scrollbar. All these changes are the first commit.
2. The second commit has the whole list slide up as the footer comes on screen, to prevent the sidebar from overlapping with the footer, which has always been a problem. I also used this as an opportunity to refactor the JS a bit, so reviewing the new source will probably be easier than trying to follow the diff.

As @wvteijlingen pointed out in another issues, making the sidebar a scrollable column on small screens doesn't work reliably in Safari. That is, in Safari, when the user's cursor is over the sidebar and they scroll, about half the time it's the main section of the document that scrolls, rather than the sidebar scrolling. I have no idea why this is, and I tried a couple things to fix it, but neither ultimately worked. These were: using `z-index` to try to move the sidebar forward, to hint to Safari that it's what the user is focusing on, and making the links in the sidebar `display: block`, again to try to increase the odds that Safari thinks the user is interacting with the sidebar when their cursor is over the sidebar.

The alternative to this Safari unreliability (if we can't find a fix) is to make the sidebar `position: static` when it can't fit, rather than making it a fixed-but-scrollable region. I'm not sure if that's worth it, since it means that the nav items will be offscreen for users on small viewports once they've scrolled down a bit into the spec text (which defeats the point a bit). But it's an option, and Ward's PR in #785 gets us most of the way there (I just don't think it accounts for the footer yet in computing whether there's space for the sidebar to fit with `position: fixed`).

ethanresnick:
Updated this to also narrow the sidebar. See commit message for the details.

dgeb:
Definitely an improvement. Thanks @ethanresnick :shipit:

PR #865, opened 2015-08-26
Site simplifications

ethanresnick:
A number of simplifications. See commit messages for details.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -19,7 +19,7 @@ port: 9876
 navigation:
 - title: JSON API
   url: /
-- title: Format
+- title: Specification
   url: /format/
 - title: Extensions
   url: /extensions/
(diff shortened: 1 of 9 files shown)


tkellen:
I'm :+1: to all of this except hiding the sidebar. I think it should just move to the top of the page. That being said, it disappears on the current site, so this isn't making things any worse. @dgeb?

ethanresnick:
> I think [the sidebar] should just move to the top of the page.

Totally agree that we should do our best to always keep the sidebar's key info visible, and I can make more tweaks later with that aim (like moving it to the top or similar). For now, this is just meant to be a first step in that direction by hiding the sidebar in fewer cases than we do currently.

ethanresnick:
@dgeb Are you ok with me merging this?

dgeb:
Sure @ethanresnick - thanks for doing this cleanup :+1:

ethanresnick:
Cool :+1:

PR #866, opened 2015-08-27
Added ripozo to python implementations

timmartin19:
NOTE: I developed ripozo.


ethanresnick:
Hi @timmartin19, 

We're working on an objective way to order the implementations in each section (see #733), but, until that's ready, chronological ordering is what we've been going with. So, please update this Pr to put riposo at the end of the list, rather than at the top, and then I'll give this a merge.

timmartin19:
That absolutely makes sense. I bumped it to the bottom.

steveklabnik:
Thanks!

PR #869, opened 2015-08-31
Fix wrong title in recommendation's page

benoittgt:
The `=======` doesn't seems to be in the good place.

image


ethanresnick:
Good catch; thanks!

PR #871, opened 2015-09-06
"id" fields must be strings

erickt:


steveklabnik:
Nice catch, thanks so much!

Issue #873, opened 2015-09-08
archive of old release


equivalent:
Hi, is there a url somewhere to the older page version (before 1.0, from around dec 2014 ? ) 

the reason is that I've used that version to create application, and I've need to stick url to the app Documentation for the reference.

tkellen:
We're going to maintain a history of versions for everything post 1.0 on the website. If you want to link to something leading up to 1.0, you'll (unfortunately) need to mine through the git history and link directly to the markdown file.

equivalent:
ok there is one now http://jsonapi-rc3.herokuapp.com/

PR #876, opened 2015-09-10
include list of archived versions for further reference

equivalent:
had a situation where I have to reference the old documentation (yes it was just RC3 but yet it was implemented  in production on our app => I needed to give reference link to client and there was no resource)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,3 +34,10 @@ This standard is currently under development. To assist:
 
 The website is automatically built by [GitHub Pages](http://pages.github.com)
 when changes are pushed to the `gh-pages` branch.
+
+Archive
+-------
+
+older versions of the JSON-API documentations
+
+* RC3 - http://jsonapi-rc3.herokuapp.com/


equivalent:
https://github.com/json-api/json-api/issues/873

ethanresnick:
Thanks for submitting this @equivalent, and sorry for the delayed response. 

I don't think we want to link to anything off-site that we don't control, which is why this hasn't been merged in.

If none of the other editors mind, though, I'd be happy to tag the points in the git history for the old RCs, and possibly link to the Tags page from the jsonapi.org homepage.

equivalent:
Hi Ethan, no problem

If you give me your Heroku details I can transfer the ownership of the Heroku app & the Github  to you, it's up to you if you keep me in the project (I can at least do some maintenance  when Heroku deprecates cluster again or something)

I'm not planing to update any of the content inside the app, I was purely aiming to have some old RC reference for quick access (both me and anyone else on the internet)

tkellen:
@ethanresnick I am going to land this. I don't see any harm in having the content managed by a third-party. The worst case scenario is that we remove the link because the page is gone or is no longer accurate. Thank you @equivalent!

tkellen:
@equivalent If you'd be willing to open a PR to put rc3 in the [formats folder](https://github.com/json-api/json-api/tree/gh-pages/_format) we can get this on the main site.

PR #877, opened 2015-09-10
Show relationships in included resource objects

ethanresnick:
Even though the spec text says that the objects in `included` are
resource objects (and so can have relationships), I’ve had a couple
people tell me they weren’t sure whether this was allowed because we
didn’t have an example showing it. This aims to rectify that.

### DIFF
--- a/format/index.md
+++ b/format/index.md
@@ -437,6 +437,11 @@ A complete example document with multiple included relationships:
     "attributes": {
       "body": "First!"
     },
+    "relationships": {
+      "author": {
+        "data": { "type": "people", "id": "2" }
+      }
+    },
     "links": {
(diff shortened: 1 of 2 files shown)


dgeb:
Thanks, @ethanresnick. I've had this same question come up, so this is a great idea. Would you mind also updating the home page example with the same changes (since it's the same example)?

ethanresnick:
@dgeb Done! I also changed the `type` on the homepage example to match the one on the specification page.

ethanresnick:
@dgeb @tkellen this good to merge now?

tkellen:
:+1:

ethanresnick:
cool beans

PR #878, opened 2015-09-12
Versioning Infrastructure

ethanresnick:
This PR adds the structure necessary for us to host multiple versions of the spec simultaneously. It gives the user option to switch between versions easily and describes the status of each version. It's also meant to make development easy by having Jekyll dynamically output the text describing each version's status. Details on each change are in the commit messages, but here's a slightly higher level summary:
- There's a `latest_version` config option that we set in `_config.yml` and there are files for each version of the spec (`_format/1.0/index.md`, `_format/1.1/index.md` etc). All we have to do is add text to the file for the upcoming version (which we can do continuously) and then, when the new version is ready to go, up the `latest_version` setting.
- From the above, Jekyll generates the following routes:
  - `/format`. This always shows the latest published version of the spec.
  - `/format/upcoming`. This shows the latest draft of the next version of the spec. I think this is useful to have on the site so that potential contributors can review the upcoming changes more easily than if they were tucked away in a branch or PR somewhere.
  -  `/format/X.Y`, where `X.Y` is the version number (e.g. /format/1.0). This holds a copy of version X.Y of the spec. This is an archived, unchanging copy (labeled as such) once that version of the spec has been finalized and a changing draft of version X.Y until that version is finalized (also labeled as such).
- On each page that hold some version of the specification, there's a dynamically generated `<title>` and `<h1>` that alludes to the status of the page the user's looking at. This status is then explained more fully in the `Status` section at the top of the page, which is also customized for each version.
- There's a version switcher `<select>` at the top the sidebar on each page that has some version of the spec text.

Try it out and lmk what you think!

@tkellen @dgeb @steveklabnik I realize we haven't talked yet about how to do this, but I had some time today and just wanted to bang something out. If you guys had different ideas for how this should work, lmk and I'll be happy to edit!


steveklabnik:
:clap: :clap: :clap: 

can you host this somewhere so I can check it out in my browser?

ethanresnick:
Sure! Check it out [here](http://jsonapi.mymorguefile.com/)

ethanresnick:
P.S. I just realized that the version switcher disappears on mobile (since it's part of the sidebar, which also disappears). That's probably a problem. I'll fix it tomorrow :)

steveklabnik:
This looks great!

ethanresnick:
Awesome! I just pushed a fix for showing the version switcher on mobile too. I think this is all ready to go now, but let's see if @dgeb and @tkellen have any thoughts.

tkellen:
Until we add the required html to our format document to extract the `normative-statements.json`, we need to keep that sibling to the `format.md` file. Other than that, this looks great!

ethanresnick:
@tkellen Great catch! I've added a `normative-statements.json` file as a sibling of each format, so the normative statements will be available at `/format/x.y/normative-statements.json`. I also added an `{% include %}` that always makes the latest format's normative statements available at `/format/normative-statements.json`. 

The only URI missing now is `/format/upcoming/normative-statements.json`. I tried to create the `normative-statements.json` file under `/format/upcoming` and have it `{% include_relative %}` the `normative-statements.json` for whatever the next version is, but that didn't work because `{% include_relative %}` can't look in the parent directory (to get from into `../1.1`). So, we could add a non-symmetrical URI, like /format/upcoming-normative-statements.json` that would always have the right version, or we could just skip supporting that URI, or maybe there's some other option. Thoughts?

tkellen:
Could we just hard-code what the upcoming version is?

ethanresnick:
@tkellen I don't think that would help... the template can pretty easily figure out what the next version is and where that normative statements file is. The problem is that any file in the `upcoming` folder (i.e. any file that could have end up with an `/upcoming/*` url) literally can't include the relevant normative statements file because it lives in a different directory that jekyll's `{% include_relative %}` command (I guess for security reasons?) doesn't give it access to.

~~Thinking about this more, though, I have an idea for a fix that seems elegant enough. I'll commit it in a sec.~~ Nevermind. The two options that I thought of (moving the normative statements to `_includes` or creating a parallel `_normative-statements` collection for them) both separate the json files from the markdown files they correspond too, which seems worse than the problem we have now.

~~I think the best option is to just have a slightly asymmetric URI for this case. So we'd have `/format/upcoming-normative-statements.json` instead of `/format/upcoming/normative-statements.json`.~~ Jk, jk. I should think more before commenting. New idea to try and then I'll let you know.

ethanresnick:
Ok @tkellen, this all works now! The solution was to add empty front matter to normative statement json files, so that jekyll would treat them as pages of the formats collection rather than static files, which allowed me to access them by looping over the collection's pages. So we're all good now!

tkellen:
Nice!

hhware:
It's great that multiple branches can now be accessed via UI! However, IMHO, it would make sense to implement this via git submodules. Files containing each version of the spec could reside in separate branches, all these branches would be included into `gh-pages` via git submodules. `gh-pages` would then host only formatting and other general stuff, but not actual spec documents. Not sure whether submodule approach provides any tangible benefit over the current approach, but it is much like SVN branches vs git branches: the latter are arguably much more elegant, flexible and useful.

ethanresnick:
@tkellen This good to merge now? (It should be independent of the sidebar on mobile stuff, which we can keep working on in #879.)

@hhware Since, as you say, the submodule approach might not offer any tangible benefit, but it does come with extra complexity, I think we should stay away from it until/unless it proves necessary. (Extra complexity comes in the form of distinct commit histories for each submodule and `gh-pages`, even though, say, the homepage example in `gh-pages` does depend on the content of the latest spec doc. More importantly, it would make the workflow more complicated for contributors, since submodules aren't autodownloaded on clone. Etc.)

hhware:
> I think we should stay away from it until/unless it proves necessary

@ethanresnick, sure, it was just an unassuming suggestion, especially because submodules are not universally liked.

> Extra complexity comes in the form of distinct commit histories for each submodule

I actually view it as an advantage. I guess most contributors have developer-oriented mindset, so it should not be extraneous to anyone. Separate directories is certainly a good structure, but its level of abstraction is lower, IMHO.

> since submodules aren't autodownloaded on clone

Not by default, but `git clone --recursive` does it, and `git push --recurse-submodules=on-demand` pushes to all submodules at once.

ethanresnick:
Ok, I'm gonna give this a merge, as the issue @tkellen pointed out has been fixed. That way, we can move onto other changes that depend on having this in place. If I'm overlooking something, feel free to revert or add a fix, but I'm pretty sure this is all good.

tkellen:
:+1:

PR #879, opened 2015-09-13
Site tweaks

ethanresnick:
These are some small tweaks that were should have been included with, but were missing from, #865.

First, this redirects users who land on the old status page (which was obsoleted by #865 and is no longer linked to from anywhere) to the format page.

Second, it keeps the sidebar active on yet narrower screens still, per the discussion with @tkellen in #865.

Third, it replaces the old Google+ footer icon/link with one for the Discourse group, and cleans up the icons' css a bit.

(I had to update some gem versions to do the status page redirect, since jekyll-redirect-from doesn’t work with jekyll 1.0. The changes to the pygments setting in _config.yml and the removal of the --watch flag in the Rakefile reflect changes to Jekyll in 2.4.)


tkellen:
Taking a look at this locally now.

tkellen:
:+1: from me. I continue to be confused why we can't just put the sidebar above the content at the smallest sizes, but I honestly haven't looked closely (nor do I have the bandwidth to :(). Nice work @ethanresnick! I'd like @dgeb or @steveklabnik to +1 before we land since this affects local dev, though.

ethanresnick:
> I continue to be confused why we can't just put the sidebar above the content at the smallest sizes

Ahh, I see what you mean now. Wouldn't that push the content down really far, though (esp. on the spec page)? Maybe some sort of toggle to reveal the sidebar content would make more sense? (Or starting it partially shown but then having a way to expand it?) Idk, but I'm happy to think about it more and whip something up when I have time.

> I'd like @dgeb or @steveklabnik to +1 before we land since this affects local dev, though.

Sounds good.

tkellen:
Yes, it would push the content down if you resized the page, but on mobile devices the content wouldn't be pushed anywhere, it would just be below the menu.

ethanresnick:
Right, I meant pushed in the sense that the user would have to scroll a ways to see it...but maybe that's not a problem.

ethanresnick:
I'm gonna go ahead and merge this, since it's been sitting for quite a while and it has some good short term fixes (like unlinking the Google Group and linking Discuss). We can take the sidebar stuff further in a future PR and the changes to local dev aren't an issue, since they all already happened when we added the versioning infrastructure.

dgeb:
@ethanresnick sounds good! Sorry for letting this sit.

PR #881, opened 2015-09-14
Add marshmallow-jsonapi to Python implementations

sloria:


steveklabnik:
We tend to add things at the bottom of the list, rather than the top, so as not to play favorites. Would you mind moving it down?

sloria:
Sounds good. I've changed it.

ethanresnick:
Thanks! Merged.

PR #888, opened 2015-09-18
Fix headings going off-screen after following an anchor link

ethanresnick:
When the user loaded the page with an anchor already selected (from the URI hash), it was the case that the browser often jumped the user to below the header, causing the section heading to be off-screen. This fixes that, by putting all the `<a>`s before the header text, and updating the CSS appropriately.


tkellen:
<3 i've been meaning to do this forever

PR #893, opened 2015-09-23
Allow meta in linkage

elliotttf:
The [docs](http://jsonapi.org/format/#document-resource-identifier-objects) indicate that it is valid for linked resources to contain a meta object; however, this is missing from the schema.


ethanresnick:
Thanks!

PR #895, opened 2015-10-01
Improve wording for api2go

sharpner:
Since our library has come pretty far it would be awesome if we could promote it better :-)


steveklabnik:
Sure thing! :)

PR #907, opened 2015-10-27
Replace "Github" with "GitHub"

acme:
Just a quick fix, replacing "Github" with the more correct "GitHub".


ethanresnick:
Sure, why not. Thanks!

PR #908, opened 2015-10-27
Replace "it's" with "its" in two spots

acme:
Just a quick fix to replace "it's" with "its" in two spots.

(I know I've submitted two minor fixes, but that's because the rest is great!)


steveklabnik:
Minor fixes are still useful! Thanks so much.

PR #914, opened 2015-11-06
Fix actions discovery FAQ

ethanresnick:
The current answer made no sense since, of course, JSON API doesn't
(yet) define the format of an OPTIONS response and, even if it did, the
allowed methods would probably be in the `Allow` header and not the
response body.

This also removes the talk of profiles in favor of a link to a more
recent issue where these questions are being discussed.


ethanresnick:
@tkellen @steveklabnik @dgeb Any objections to me merging this?

tkellen:
:+1:

steveklabnik:
:+1:

PR #917, opened 2015-11-15
Design tweaks

ethanresnick:
Some subtle changes that I think make the site look a little crisper. Replaces gray backgrounds in a couple places with white ones to add contrast, and introduces just a hint of color (a quiet green) to liven things up.


steveklabnik:
any chance you could host this somewhere real quick so i could check it out?

ethanresnick:
Yup! http://jsonapi.mymorguefile.com/

tkellen:
:+1: by me

dgeb:
Some random thoughts:

The splashes of green are nice.

The home page looks odd to me now that the banner isn't full width. Might be because there's no border on sides (see red arrows below) like there is on the top and bottom. As is, I think I prefer the current full width navbar and home page banner, but it's not a huge deal.

screen shot 2015-11-19 at 11 02 51 pm

I like the use of green in the H1:

screen shot 2015-11-19 at 11 18 48 pm

But I prefer the full caps font we're using now. It matches the font used on the home page banner.

screen shot 2015-11-19 at 11 19 33 pm

ethanresnick:
@dgeb 

> The home page looks odd to me now that the banner isn't full width.

Yup. I had tried a full-width version, and was on the fence about it, but looking at it again I think you're right. PR updated!

> But I prefer the full caps font we're using now

I _think_ what you're reacting to here is how much heavier the new headings are, rather than the title case vs. all caps issue per se. I switched away from all caps for better readability, and to have the headings' style echo that of the selected item in the nav bar. But when I moved to title case, I had to make the heading bolder to keep the hierarchy intact, and I think I went a bit too far. So I've updated the PR to use a somewhat lighter font weight. Check it out and let me know if it feels better.

dgeb:
@ethanresnick thanks for making those changes.

My final nit is that the line under the top navbar should probably be full width on all pages, to match both the home page banner and the green line at the very top.

Other than that, I'm :+1:

ethanresnick:
> My final nit is that the line under the top navbar should probably be full width on all pages, to match both the home page banner and the green line at the very top.

I mocked this up [here](http://opt-a.mymorguefile.com/format/), but I'm not sure it's an improvement. 

To explain my thinking, let me back up a bit... It seems like there are two ways the layout can be made to "read" visually: 
1. The navbar can appear as a distinct, full-width element with some centered, fixed-width content below it. This is how the current design on jsonapi.org reads:
   
   current
2. The nav links can "read" as being part of the body content, with the whole page (including the nav links) "hanging" off the full-width line at the top. This is how the inner pages read in the design I'm proposing (and how the homepage did originally, with the fixed-width banner):
   
   proposed 

I'm ok with either structure, though I slightly prefer the first. Nevertheless, I switched to the second one in this PR (holding aside the homepage for a sec) because...

In both designs shown above, the nav links are aligned with the body content (not spread across the full width of the screen, for example), and this acts to naturally suggest the second "read". But, in the design currently up on jsonapi.org, the full-width background color behind the nav links and the (very dark) line below them serve to overcome the alignment and convincingly suggest a distinct, full-width navbar.

However, once I removed the gray background from behind the navbar and lightened the line below it as part of this PR, using just a full-width line below the nav links to suggest the first "read" felt insufficient and a little dissonant. Looking at [that design](http://opt-a.mymorguefile.com/format/), my brain still ultimately settles on the first interpretation, but there's a bit of tension because there are fewer cues to counterbalance the alignment and strongly suggest that the navbar is a distinct, full-width element. Meanwhile, narrowing the line to match the width of the main content removes this tension by cleanly suggesting the second interpretation instead. 

So that's what I went with in this PR, and it's why I'd prefer this PR as-is to making the line below the nav links full-width. But there are also couple other options that I want to throw out there, that might address the concern behind Dan's suggestion:
1. It's true that the homepage and the interior pages are inconsistent in this PR, because the homepage banner breaks the grid and goes full width. If this inconsistency is what's motivating Dan's suggestion, we could remove the inconsistency by going back to the fixed-width homepage banner, but [with the side borders](http://opt-c.mymorguefile.com/) that Dan proposed earlier. Personally, this inconsistency, which I tried to avoid at first, doesn't bother me anymore, because I think the full-width banner's stronger impact justifies it. But I could go either way.
2. If Dan's suggestion is motivated less by the inconsistency and more by a preference for designs with a distinct, full-width navbar, which I can understand, we could add back a background color to the navbar and make it full-width, so that it would function just as it does on the current site. However, my suggestion here would be to make the background a quiet green, rather than a gray (as my primary goal with this PR was to make the site less gray). That would mean substantially more color than this PR contains right now, though, which is why I was hesitant to suggest it at first. Still, I think it's a strong option, so I mocked it up [here](http://opt-d.mymorguefile.com/). Live with it for a bit and let me know what you think.

dgeb:
Wow, @ethanresnick - thanks for doing the mockups. It's difficult to discuss these things without seeing them.

I personally favor option a - http://opt-a.mymorguefile.com/format/ - but would prefer some very slight background color to the navbar, at least on pages other than the home page (similar to what we're doing now).

I think the dark green in option d is overkill, and I find the truncated horizontal line under the navbar in b and c to be abrupt.

Of course, that's just my opinion :)

ethanresnick:
> I personally favor option a - http://opt-a.mymorguefile.com/format/ - but would prefer some very slight background color to the navbar... I think the dark green in option d is overkill

@dgeb Cool! I'd also be happy with a full-width navbar, like in option a, once we put some background color behind it. So it sounds like the question is just picking that color. Would you be up to put together a version with a background color that you like? In the mean time, I'll also play with some versions that have subtler colors.

About the green used in option d, though: I do wonder if it feels like too much color just because we're so used to the current, gray design. As I've lived with option d for a bit (leaving it open, browsing around, and switching between it and other sites), the green bar has grown on me. So, if you haven't already, I'd encourage you to sit with it for a bit too and see if your feelings about it change.

dgeb:
> So it sounds like the question is just picking that color.

I tend to favor leaving the navbar background color the same minimalist gray for now: #fafafa. This one minor color change could also be revisited separately.

> About the green used in option d, though: I do wonder if it feels like too much color just because we're so used to the current, gray design.

I'm willing to be overruled here. @steveklabnik @tkellen @wycats what do you guys think about this green background to the navbar that Ethan favors?

http://opt-d.mymorguefile.com/format/

dgeb:
I'm not sure this comes across in my comments, but I do agree that the site could use a little more style / color.

One thought is to incorporate the logo on all pages in the nav bar, like we're doing over in http://discuss.jsonapi.org/

tkellen:
I like the green!

If ya'll want, I might be able to get some help from the Bocoup design team sometime in the coming months if we want a wholesale facelift.

dgeb:
> I like the green!

Well, if both you guys are onboard, I'm fine with it too.

> If ya'll want, I might be able to get some help from the Bocoup design team sometime in the coming months if we want a wholesale facelift.

Thanks - I think that would be awesome :100:

tkellen:
I'll open an issue on our trello board and see what kind of response I get. I'm not sure what all we have in the pipeline right now.

steveklabnik:
I feel neutral to vaguely positive about the green :+1:

dgeb:
I think I'd like the green better with a little less vertical padding (say 5px instead of 10px). That removes the weight imbalance that was my main objection.

P.S. how ironic is it to paint this particular bikeshed in public? :stuck_out_tongue_closed_eyes:

ethanresnick:
> P.S. how ironic is it to paint this particular bikeshed in public?

Yeah, it's a delicious irony :D

> I think I'd like the green better with a little less vertical padding (say 5px instead of 10px). That removes the weight imbalance that was my main objection.

I definitely see what you're talking about re the weight. I've shrunk the vertical padding somewhat (from 10px to 8px) and also desaturated the green a tiny bit so it's less heavy. I tried shrinking the vertical padding further, but it made the navigation feel a bit too narrow, and less inviting as a click target. 

I've updated this PR to reflect the newest version, which is also visible at the old Option D URI.

> If ya'll want, I might be able to get some help from the Bocoup design team sometime in the coming months if we want a wholesale facelift.

Sure! Thanks Tyler :)

dgeb:
@ethanresnick thanks - looks good to me! Was going to merge but there are conflicts - feel free to merge after resolving.

ethanresnick:
Awesome! Conflicts resolved & merged.

Thanks for the great feedback guys! The design we ended on is definitely better than the one I originally opened this PR with :)

dgeb:
Looking good! Thanks for seeing this through @ethanresnick :+1:

PR #921, opened 2015-11-21
Clarify sparse fieldset restrictions

ethanresnick:
The requirement about not including extra fields is _on a per type basis_.

This clarification inspired by http://discuss.jsonapi.org/t/clarification-on-the-sparse-fieldset-response/221


ethanresnick:
Any objections to my merging this?

dgeb:
I agree that this is necessary and is not breaking, but a clarification. :+1:

PR #922, opened 2015-11-25
Clarify relationship self link again

bintoro:
Makes the definition of the self link more explicit.


ethanresnick:
I'm :+1: on this as I said in the other issue. 

Any objections @dgeb @tkellen?

tkellen:
:+1: here too. Thanks @bintoro!

dgeb:
Yeah, that's a definite improvement. Thanks @bintoro - and welcome back :)

Issue #913, opened 2015-11-05
Unifying Resource Identifier and Link Objects


ethanresnick:
Conceptually, a link object and a resource identifier both reference some external data. And, if you think in web architecture terms, a relationship name in JSON API is just shorthand for what RFC 5988 would call an "extension relation name", so a resource identifier objects can be seen as the target of an extended relation in the same way a link object represents the target of a registered relation.

There are some differences between link and resource identifier objects, of course. But I think those differences are bridgeable (see below). So I'm tentatively thinking we should unify link and resource identifier objects into one concept, called a link object, with keys:
- `"href"` — This already exists on today's link objects and [has been requested](https://github.com/json-api/json-api/issues/477) for resource identifier objects
- `"content-type"` — The expected mime-type of the target href; this corresponds to the `"type"` [target attribute](https://tools.ietf.org/html/rfc5988#section-5.4) in RFC 5988
- `"type"` and `"id"` — The JSON API type/id of the target resource. If these keys are present, it implies that the `"content-type"` is `"application/vnd.api+json"` unless it's otherwise set.
- `"meta"` — Same as today.

Since we didn't make hypermedia more of a requirement in 1.0, the spec text would still need to say (for backwards compatibility) that these new link objects, when present in relationship `"data"` must include type/id, whereas the ones present in `"links"` must include href. Because of that caveat, this unification in the base spec might be a wash for users who just work with the base spec: it would remove some duplication (e.g. if we add `"href"` to resource identifier objects, or `"title"` to both links and resource identifier objects, we wouldn't need to specify these things in two places) and lessens the number of distinct concepts, but it does obfuscate the distinction in the required fields.

Still, what tips the scales for me in favor of unification is the groundwork that this would lay for the long-discussed "hypermedia required" extension, which would, hopefully, over time, become the standard way of using JSON API. In that "hypermedia required" world, in which each resource object has a `"self"` link, compound document linkage can be done by `"href"`, and `"type"`/`"id"` can be downgraded to just a target hint, like content-type. In fact, in such a world, you could get rid of relationship `"data"` all together (which is already optional when links are present) and just do the following:

``` js
{
  // my resource object...
  "relationships": {
    "my-relationship": {
      "links": {
        "self": "http://example.com/myresource/relationships/my-relationship",
        "related": "http://example.com/myresource/my-relationship",
        "item": [{"href": "http://example.com/articles/4", "type": "articles", "id": "4"}]
      }
    } 
  }
}
```

[...]

Issue #477, opened 2015-03-16
Suggestion: "target", a reserved member of relationship linkage 


hhware:
IMHO, the terminology of relationship specification is a bit confusing. URLs are provided for the relationship itself and for the resources referred to via the relationship, but there is no way to provide a direct URL of the resource being referred to (i.e., its `self`), the user _has to_ construct it from the `type`/`id` or make a call to `related` and get `self` from it. IMHO, it is an unnecessary limitation, especially for APIs which do not support compound documents.

I know that similar things were discussed (e.g., in #312 & #392), so I apologize if I am reintroducing something which has already been decided upon, but I do not think this particular suggestion has been discussed. 

I think it would be good to straighten up terminology and provide way to specify URLs for all 3 types of references. So we would have
- "self": relationship itself,
- "related": reference to the target object of the relationship through the relationship,
- "target": direct reference to the target object.

**Proposal**: each individual linkage item, for both to-one and to-many relationships, must contain at least on of the following:
- a combination of `type` and `id` members,
- a direct URL of the resource being linked as a member `target`.

An example to illustrate all possible combinations (obviously, any given API is not expected to mix them all, at least not in one relationship):

``` json
{
  "type": "articles",
  "id": "1",
  "title": "Rails is Omakase",
  "links": {
    "self": "http://example.com/articles/1",
    "author": {
      "self": "http://example.com/articles/1/links/author",
      "related": "http://example.com/articles/1/author",
      "linkage": { "type": "people", "id": "9" }
    },
    "editor": {
      "self": "http://example.com/articles/1/links/editor",
      "related": "http://example.com/articles/1/editor",
      "linkage": { "target": "http://example.com/people/10" }
    },
    "proofreader": {
      "self": "http://example.com/articles/1/links/proofreader",
      "related": "http://example.com/articles/1/proofreader",
      "linkage": { "type": "people", "id": "11", "target": "http://example.com/people/11" }
    },
    "illustrators": {
      "self": "http://example.com/articles/1/links/illustrators",
      "related": "http://example.com/articles/1/illustrators",
      "linkage": [
        { "type": "people", "id": "12", "target": "http://example.com/people/12" },
        { "target": "http://example.com/people/13" },
        { "type": "people", "id": "14" }
      ]
    }
  }
}
```

bintoro:
@hhware, the `self` URL (a.k.a. the "static" or "canonical" URL) of the resource is something I very much want to have in link objects, too.

[...]

PR #926, opened 2015-11-29
Support `"href"` in resource identifier objects

ethanresnick:
It is sometimes useful for a resource identifier object that points to a resource that isn't included in the document to contain the URI of that resource. That way, the client can request it without having to ask for the whole `"related"` collection, and without the server needing to support the related endpoint.

This need has come up repeatedly on the discourse forum, and is also the issue in #477.

Here, I'm suggesting we address this use case by adding an `"href"` key to resource identifier objects. I prefer this approach over, say, adding a `"links"` key to resource identifier objects, because it stresses the conceptual similarity between link and resource identifier objects (without going as far as I propose in #913). By contrast, adding a `"links"` key would suggest a similarity between resource objects and resource identifier objects, which are conceptually very distinct and are (if anything) already too easy to confuse.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -362,8 +362,14 @@ resource.
 
 A "resource identifier object" **MUST** contain `type` and `id` members.
 
-A "resource identifier object" **MAY** also include a `meta` member, whose value is a [meta] object that
-contains non-standard meta-information.
+A "resource identifier object" **MAY** also include the following members:
+
+* `href`: a URI that, when derefenced, retuns a JSON API resource object with
+  the same `type` and `id` as in the resource identifier object. If the
+  returned resource object contains a `"self"` link, that link's URI **SHOULD**
+  match the URI provided in this key.
+
+* `meta`: a [meta] object that contains non-standard meta-information.
 
 ### <a href="#document-compound-documents" id="document-compound-documents" class="headerlink"></a> Compound Documents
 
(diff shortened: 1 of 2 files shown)


ethanresnick:
Actually, let's hold off on merging this... I want to think more about whether it's appropriate. Because, while it's true that resource identifier objects point to a target of a particular link relation (and in that sense are very much like link objects), they also have a dual nature of being primary data (and so being paginatable etc.). Therefore, maybe a `"links"` key in them (with a `"related"` link??) is appropriate? Alternatively, we could still see resource identifier objects as conceptually a type of link object, if we just say that link objects can occur as primary data (in relationship endpoints...).

Thoughts?

dgeb:
Hmmm ... I agree that we need to pause and carefully consider whether `links` or `href` is more appropriate for resource identifier objects. Everywhere that we allow `links` we allow a `self` member, and yet the spec doesn't have the concept of allowing direct access to an identifier object (nor do I see value in allowing that). I suppose we could allow `links` and not `self`. As for `href` - I find it to be a bit inconsistent with the rest of the spec.

jelhan:
> By contrast, adding a `"links"` key would suggest a similarity between resource objects and resource identifier objects, which are conceptually very distinct and are (if anything) already too easy to confuse.

Why do you see them as conceptually very different. I thought of resource identifier objects as a subset of resource objects. To even stress it more: I don't see how a resource object without any field and without `meta` is distinct from a resource identifier object. At the end of the day it would be the same JSON.

Treating resource identifier objects as a subset of resource objects would simplify this question a lot. The `links.self` in a resource object could be the same as for the corresponding resource identifier object.

As dgeb pointed out, having dedicated self link for a resource identifier object does not seem to provide any value. But even if we want that, we could add a dedicated member to `links` of a resource object _and_ resource identifier object for it. We could name it `links.identifier` for example.

jelhan:
I'm going to close this. The PR is stale for a couple of years now. @ethanresnick and @dgeb shared in their last comments that they aren't convinced that it is the right direction.

PR #928, opened 2015-11-30
Standardize code block syntax highlighting

nonsensery:
#### What's this PR do?

Switches code blocks on the Specification page from `json` to `javascript` to improve the syntax highlighting.
#### Background context

This is a follow-up to PR #702 and should replace it. The discussion in that PR ended with the consensus to switch the code blocks to `javascript`, rather than modifying the content to appease the JSON syntax highlighter.

Many of the examples are partial JSON data structures or contain JS-style comments. The JSON syntax highlighter does not handle them very well, often not applying any highlighting at all. The JS highlighter does a better job with these code blocks.

Not all of the code blocks need to be marked as JS; some of them are perfectly valid JSON. It's my opinion that it's best to be consistent so that future contributors don't need to guess at which syntax highlighter to use, so I've marked them all as JS.

### DIFF
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -337,7 +337,7 @@ related to the resource.
 If present, this links object **MAY** contain a `self` [link][links] that
 identifies the resource represented by the resource object.
 
-```json
+```javascript
 // ...
 {
   "type": "articles",
@@ -388,7 +388,7 @@ objects][resource identifier object]) or to each other.
 
 A complete example document with multiple included relationships:
 
-```json
+```javascript
 {
   "data": [{
     "type": "articles",
@@ -516,7 +516,7 @@ either:
 
 The following `self` link is simply a URL:
 
-```json
+```javascript
 "links": {
   "self": "http://example.com/posts",
 }
@@ -525,7 +525,7 @@ The following `self` link is simply a URL:
 The following `related` link includes a URL as well as meta-information
 about a related resource collection:
 
-```json
+```javascript
 "links": {
   "related": {
     "href": "http://example.com/articles/1/comments",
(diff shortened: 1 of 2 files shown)


nonsensery:
@tkellen You requested a FAQ entry for this issue. I haven't added one yet because it seemed a little out of place. The existing entries seem to answer questions about the spec itself, rather than the code used in the website. What do you think?

tkellen:
Haha! I was kidding, I should've left a note saying I was being sarcastic :p

nonsensery:
Where's the sarcmark when you need it? :-)

ethanresnick:
So, I looked into this more today, and marking everything as JS, which I realize I suggested, actually brings its own troubles. The biggest issue is that, when js highlighting is used, the keys and the values in object literals seem to get classified the same way, so there's no way to distinguish them visually (as in the JSON highlighting). That's a big loss imo. Then there were other issues, like: even if we tag all the blocks as JS, the json that's inside http blocks still gets tagged (automatically on the client) as json and so ends up looking inconsistent.

After playing around with this, the only solution that I could get to work was:
1. Mark all the blocks (even those with comments) as json in the markdown
2. That'll force highlight.js (which runs on the client and actually does most of the highlighting work, confusingly enough) to treat every block as json, giving us nice key/value distinctions and http block consistency. (We're using highlight.js, I think, since it seems to apply much more informative class names.)
3. Then, if highlight.js catches a syntax error, assume that that "error" is a comment in json, and style it like a comment.

In the process of playing around with this, I ended up making all of the changes myself in the uniform-syntax-highlighting branch. So this PR isn't really needed anymore (sorry; I know that's annoying), but I'll make a new one for my changes and your feedback/further refinements there would be awesome.

PR #930, opened 2015-12-01
Uniform syntax highlighting

ethanresnick:
Following up on #702 and #928, I think this finally gives us consistent and usable syntax highlighting (i.e. with comments intact and different colors for object keys and string values).

Note: I suspect that there are still some stray code blocks that look a bit weird, and that we can clean up over time, but I think this PR gets the basic setup for consistent highlighting right (i.e. tag as json, use hightlight.js, and treat syntax errors (in json) as json comments.)


nonsensery:
Hey, this looks great! Nice work.

stylesheets/highlight.scss
```diff
@@ -12,9 +12,7 @@ pre code {
 }
 
 pre .comment,
-pre .template_comment,
-pre .diff .header,
-pre .javadoc {
+pre .err {
```
nonsensery: What do you think about scoping this rule so that it only affects JSON code blocks? Like `pre .language-json .err`

ethanresnick: Good idea! I'll update this now. I'm also gonna merge the syntax highlighting rules into the main compiled css, since there's no point in paying an extra HTTP request for a few rules.

ethanresnick:
Ok, I'm gonna merge this, since it's definitely an improvement over the status quo.

The only thing about it that I'm not crazy about is that it required me to add "HTTP/1.1" at the end of some request lines to get the highlighting working. Of course, the `HTTP/1.1` "should" be there, but it's just noise from a practical point of view. Maybe in the future we can figure out how to remove that and still get the highlighter to behave. (@nonsensery you're welcome to take a crack at that!)

bintoro:
Oh yeah, this is great. I looked into it sometime pre-1.0 and reached the same conclusion that designating as `json` and suppressing the syntax errors is seemingly the only way to get it right.

PR #934, opened 2015-12-03
Allow null data

elliotttf:
This was brought to my attention in elliotttf/jsonapi-validator#9. It appears that the schema is not allowing data to be `null`.

Per the spec:

> A server **MUST** respond to a successful request to fetch an individual resource with a resource object or `null` provided as the response document's primary data.
> 
> `null` is only an appropriate response when the requested URL is one that might correspond to a single resource, but doesn't currently.

It's a bit of a weird case since the validity of `null` is entirely conditional – which I don't think JSON schema really accounts for – but I thought I'd raise this anyway for discussion. I don't know if this will have further implications for elements that use `data` references.

### DIFF
--- a/schema
+++ b/schema
@@ -111,6 +111,10 @@
             "$ref": "#/definitions/resource"
           },
           "uniqueItems": true
+        },
+        {
+          "description": "null if the request is one that might correspond to a single resource, but doesn't currently.",
+          "type": "null"
         }
       ]
     },


cziegenberg:
A resource can be in pending creation so that no data can be returned yet:

> 202 Accepted
> 
> If a request to create a resource has been accepted for processing, but the processing has not been completed by the time the server responds, the server MUST return a 202 Accepted status code.

This is not an error, but a normal state. So you have to return data, which in this case is NULL (until processing has finished). Let's think of a resource that contains an image that needs to be processed, before it can be used and returned in requests. Of course you could return a type and id in the meantime, but this would be an incomplete data set. NULL is the better value to return here.

ethanresnick:
Allowing `null` doesn't really have to do with the 202 case (which should usually be handled as suggested [here](http://jsonapi.org/recommendations/#asynchronous-processing)).

Instead, `null` is used when you request the relationship URI of a to-one relationship that is currently empty. E.g. `/articles/1/relationships/primary-category` would return `{"data": null}` if the post is uncategorized. This is necessary for consistency with the `"relationships"` member's data in the article 1 resource object (if it's requested directly) and it makes sense conceptually. This is the case that's meant by:

> null is only an appropriate response when the requested URL is one that might correspond to a single resource, but doesn't currently.

If that's not clear, which I agree it might not be, let's open an issue/write up a PR to clarify it.

In the mean time, I'm inclined to merge this unless someone with more JSON Schema experience sees a reason to express this case differently. (If that's you, please chime in.)

cziegenberg:
@ethanresnick Thanks for the tip for handling 202s (although the recommendation requires to keep all queued jobs forever to 303 to the delayed recource - but that's another topic).

bintoro:
> Instead, `null` is used when you request the relationship URI of a to-one relationship that is currently empty. E.g. `/articles/1/relationships/primary-category` would return `{"data": null}` if the post is uncategorized. 

And the same applies to the related resource URL, i.e. `/articles/1/primary-category`. In fact, this is the case the quoted part refers to.

ethanresnick:
Ok, hearing no objections on JSON Schema formatting grounds, I'm going to merge this.

Thanks @elliotttf!

PR #936, opened 2015-12-04
Adding a Go API

derekdowling:
Hey! I previously added jsh to this list, as you can see from the modifications, which is a good stand-alone solution for someone with an existing API. On top of it, I built an API which can also be used to perform most of the heavy lifting and is perfect for micro-services.


tkellen:
Very cool! Thanks @derekdowling :D

PR #942, opened 2015-12-09
[Examples] - Fix typo

MQuy:
Correct the typo for `fields`


ethanresnick:
Hi @MQuy. Thanks for submitting this, but the typo is actually intentional here. (That's why it triggers an error!)

PR #944, opened 2015-12-09
Add katharsis-spring adapter

masterspambot:
Another adapter to the list of Java implementations.


ethanresnick:
Thanks @masterspambot! Do all these adapters need their own bullet points, though? It seems like it might be more appropriate to reference and link them as part of the main katharsis bullet point.

ethanresnick:
Yeah, I think that's fine. Merged!

PR #946, opened 2015-12-15
Associate "relationship object" with everything else

bintoro:
- Formally connects _relationship objects_ to relationships and, by extension, to relationships objects.
  
  The current language effectively doesn't specify what can go inside relationships objects.
  
  Also, clearly stating the parity between "each relationship" and "an object" hopefully helps emphasize the fact that the section is discussing two different concepts. The distinction between relationships objects and relationship objects is too easy to overlook at first.
- Edits the sentence that defines members of a relationships object as "relationships". Rather, the members only _represent_ relationships.
- Fixes a broken link nearby.

### DIFF
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -240,13 +240,14 @@ alongside other information to be represented in a resource object, these keys
 #### <a href="#document-resource-object-relationships" id="document-resource-object-relationships" class="headerlink"></a> Relationships
 
 The value of the `relationships` key **MUST** be an object (a "relationships
-object"). Members of the relationships object ("relationships") represent
-references from the [resource object][resource objects] in which it's defined to other resource
-objects.
+object"). Members of the relationships object represent "relationships", i.e.,
+references from the [resource object][resource objects] in which they are defined
+to other resource objects.
 
 Relationships may be to-one or to-many.
 
-A "relationship object" **MUST** contain at least one of the following:
+Each relationship **MUST** be represented by an object (a "relationship object")
+containing at least one of the following:
 
 * `links`: a [links object][links] containing at least one of the following:
   * `self`: a link for the relationship itself (a "relationship link"). This
@@ -268,7 +269,7 @@ A relationship object that represents a to-many relationship **MAY** also contai
 
 #### <a href="#document-resource-object-related-resource-links" id="document-resource-object-related-resource-links" class="headerlink"></a> Related Resource Links
 
-A "related resource link" provides access to [resource objects] [linked][links]
+A "related resource link" provides access to [resource objects][] [linked][links]
 in a [relationship][relationships]. When fetched, the related resource object(s)
 are returned as the response's primary data.
 
--- a/_format/1.0/normative-statements.json
+++ b/_format/1.0/normative-statements.json
@@ -643,7 +643,7 @@ layout: null
       "type": "normative-statements",
       "attributes": {
         "level": "MUST",
-        "description": "A \"relationship object\" **MUST** contain at least one of the following:]\\n\\n- `links`: a links object containing at least one of the following:\\n  - `self`: a link for the relationship itself (a \"relationship link\"). This link allows the client to directly manipulate the relationship. For example, it would allow a client to remove an author from an article without deleting the people resource itself.\\n  - `related`: a related resource link\\n- `data`: resource linkage\\n- `meta`: a meta object that contains non-standard meta-information about the relationship."
+        "description": "Each relationship **MUST** be represented by an object (a \"relationship object\") containing at least one of the following:\\n\\n- `links`: a links object containing at least one of the following:\\n  - `self`: a link for the relationship itself (a \"relationship link\"). This link allows the client to directly manipulate the relationship. For example, it would allow a client to remove an author from an article without deleting the people resource itself.\\n  - `related`: a related resource link\\n- `data`: resource linkage\\n- `meta`: a meta object that contains non-standard meta-information about the relationship."
       },
       "relationships": {
         "section": {
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -240,13 +240,14 @@ alongside other information to be represented in a resource object, these keys
 #### <a href="#document-resource-object-relationships" id="document-resource-object-relationships" class="headerlink"></a> Relationships
 
 The value of the `relationships` key **MUST** be an object (a "relationships
-object"). Members of the relationships object ("relationships") represent
-references from the [resource object][resource objects] in which it's defined to other resource
-objects.
+object"). Members of the relationships object represent "relationships", i.e.,
+references from the [resource object][resource objects] in which they are defined
+to other resource objects.
 
 Relationships may be to-one or to-many.
 
-A "relationship object" **MUST** contain at least one of the following:
+Each relationship **MUST** be represented by an object (a "relationship object")
+containing at least one of the following:
 
 * `links`: a [links object][links] containing at least one of the following:
   * `self`: a link for the relationship itself (a "relationship link"). This
@@ -268,7 +269,7 @@ A relationship object that represents a to-many relationship **MAY** also contai
 
 #### <a href="#document-resource-object-related-resource-links" id="document-resource-object-related-resource-links" class="headerlink"></a> Related Resource Links
 
-A "related resource link" provides access to [resource objects] [linked][links]
+A "related resource link" provides access to [resource objects][] [linked][links]
 in a [relationship][relationships]. When fetched, the related resource object(s)
 are returned as the response's primary data.
 
--- a/_format/1.1/normative-statements.json
+++ b/_format/1.1/normative-statements.json
@@ -643,7 +643,7 @@ layout: null
       "type": "normative-statements",
       "attributes": {
         "level": "MUST",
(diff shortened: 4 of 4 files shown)


_format/1.0/index.md
```diff
@@ -240,13 +240,14 @@ alongside other information to be represented in a resource object, these keys
 #### <a href="#document-resource-object-relationships" id="document-resource-object-relationships" class="headerlink"></a> Relationships
 
 The value of the `relationships` key **MUST** be an object (a "relationships
-object"). Members of the relationships object ("relationships") represent
-references from the [resource object][resource objects] in which it's defined to other resource
-objects.
+object"). Members of the relationships object represent "relationships", i.e.,
```
tkellen, line 243: While we're here:
...references from the resource objects in which **they are** defined...

Also, this needs to be updated here too:
https://github.com/json-api/json-api/blob/gh-pages/_format/1.1/index.md

_format/1.0/index.md
```diff
@@ -240,13 +240,14 @@ alongside other information to be represented in a resource object, these keys
...
+references from the [resource object][resource objects] in which it's defined
+to other resource objects.
 
 Relationships may be to-one or to-many.
 
-A "relationship object" **MUST** contain at least one of the following:
```
tkellen, line 249: This needs to be updated in a few more places:
https://github.com/json-api/json-api/blob/gh-pages/_format/1.0/normative-statements.json:
https://github.com/json-api/json-api/blob/gh-pages/_format/1.1/index.md
https://github.com/json-api/json-api/blob/gh-pages/_format/1.1/normative-statements.json:

tkellen:
:+1: from me with those minor nits

_format/1.0/index.md
bintoro, line 243: I wondered about that too. The "it" refers to the relationships object, which is singular. But yes, I agree that it would sound more correct if we referred to the relationships themselves.

bintoro:
Looks like the author of these rules in `normative-statements.json` made just the mixup this PR aims to prevent:

```
    "id": "resource-relationships-object",
    "type": "normative-statements",
    "attributes": {
      "level": "MUST",
      "description": "A \"relationship object\" **MUST** contain [...]
    },
```

`id` says "relationships object" but `description` says "relationship object".

This is my first time dealing with the normative statements file. May an existing `id` string be edited or will that screw something up?

tkellen:
You can edit the id strings if you like--to my knowledge nobody is using them (yet) :(.

Ideally, we'd mark up our format.md files such that we can generate the normative statement files, but we haven't done so yet. This is to assist with automated testing for the specification. If you have the bandwidth to make this happen, that would be **amazing**.

bintoro:
Commit updated. I ended up leaving the `id` alone for now, since `relationships-object` can also be read as "item `object` in section `relationships`", which leaves the hierarchy intact.

_format/1.0/index.md
```diff
@@ -240,13 +240,14 @@ alongside other information to be represented in a resource object, these keys
...
-references from the [resource object][resource objects] in which it's defined to other resource
-objects.
+object"). Members of the relationships object represent "relationships", i.e.,
+references from the [resource object][resource objects] in which they are defined
+to other resource objects.
 
 Relationships may be to-one or to-many.
 
-A "relationship object" **MUST** contain at least one of the following:
+Each relationship **MUST** be represented by an object (a "relationship object")
```
ethanresnick, line 249: So, if we're being nitpicky/precise, my mental model has always been that each relationship has two parts: its name and its contents (the linkage, not the related). So maybe we should say "Each relationship's contents....", or use the prior wording, or make the name + contents division more explicit?

bintoro, line 249: Yeah, "each relationship's contents" would seem to make sense.

The prior wording didn't associate relationship objects with any of the other definitions, which prompted the patch in the first place.

ethanresnick, line 249: One potential problem with "each relationship's contents" is that the relationship object also includes links, and I'm not sure if those are part of the contents???

bintoro, line 249: One way to sidestep the entire issue is to specify that _each member of a relationships object must be a relationship object_.

It's already mentioned that members of a relationships object represent relationships, so a relationship object's purpose could be defined in terms of either.

bintoro:
@ethanresnick: Looking at this again, I don't think "contents" is a necessary addition, particularly if you feel it's not well defined what makes up the contents.

By necessity, objects consist of key-value pairs, so if a key-value pair represents a relationship, then saying that "a relationship must be represented by an object" makes it very clear that we're referring to the _value_.

In the _Links_ section there's an exactly similar case:

> Each member of a links object is a "link". A link **MUST** be represented as either: [...]

As for a relationship's contents, there's one existing mention:

> [...] a related resource link **MUST NOT** change because its relationship's content changes.

It does suggest that the content might refer to the _set of the referenced resources_, although it would be fairly easy (and probably a good idea) to come up with a dedicated term for this, such as a "relationship's target".

tkellen:
Just checking in here... is the current round of bike-shedding complete? Are we ready to re-review?

bintoro:
I'm not sure. The question raised by @ethanresnick was whether it's okay to say

> Each relationship MUST be represented by an object [...]

when we actually mean

> Each relationship MUST be represented by a member whose value is an object [...]

I'm curious as to what others think. To be sure, the latter is more accurate but perhaps a bit awkward. Another angle is to figure out a term for the value part of a relationship and say

> Each relationship's value MUST be represented by an object [...]

ethanresnick:
First off, I agree that this PR as written is better than the current text, and it may be good enough. If we want to be totally accurate, though, I think there's a way to separate the definitions of the relationship's name, contents (which isn't a term we need to use necessarily), and "relationship object" without being too awkward. 

Here's probably what I'd say (key changes in bold):

> The value of the `relationships` key **MUST** be an object (a "relationships object"). Each member in the relationships object represents a "relationship", which is a **named** set of references from the resource object in which it is defined to other resource objects.
> 
> Relationships may be to-one or to-many.
> 
> **A relationship's name is given by its key. The value at that key MUST be an object ("relationship object") that contains the set of referenced resources or information pertaining to that set.**
> 
> A "relationship object" **MUST** contain at least one of the following:
> 
> ....

Thoughts?

bintoro:
> Each member in the relationships object represents a "relationship", which is a named set of references [...]

But isn't this accurate only for to-many relationships? A to-one relationship doesn't feel like a "set of references".

> A relationship's name is given by its key.

+1. I think this alone is enough to supplant the "named set" addition.

> The value at that key MUST be an object ("relationship object") that **contains the set of referenced resources or information pertaining to that set.**
> 
> A "relationship object" MUST contain at least one of the following: ...

Saying that the object "contains the set of referenced resources" sounds wrong to me; it only contains the references, not the resources. Or is this in preparation for embedded resource objects?

Since the entire bolded segment is basically a summary of what is normatively stated right in the subsequent paragraph, I'd just drop that part entirely:

> The value at that key MUST be an object ("relationship object") containing at least one of the following: ...

ethanresnick:
> But isn't this accurate only for to-many relationships? A to-one relationship doesn't feel like a "set of references".

In my mind, it applies for to-one relationships too (since a set can be limited to one item, which is what I think the next paragraph does), but I see how it could be confusing, so I'm fine with changing it. Ideas?

> +1. I think this alone is enough to supplant the "named set" addition.

Ok, but would that mean going back to "[relationships are] references from the resource object in which they are defined to other resource objects". If so, that just sounds a bit weird to me... in my head it's the resource identifier object that's the reference, not the relationship, which is why I'd said the relationship is a set of references.

> Saying that the object "contains the set of referenced resources" sounds wrong to me; it only contains the references, not the resources.

Good catch. If we were going to keep it, it should say something like "contains the references that constitute the relationship".

> Since the entire bolded segment is basically a summary of what is normatively stated right in the subsequent paragraph, I'd just drop that part entirely:

Works for me. (Though aren't we just back to your original complaint, then, that the definition of a "relationship object" isn't really connected conceptually to what comes before?)

bintoro:
I love how deep this rabbit hole is going :smile:

> in my head it's the resource identifier object that's the reference, not the relationship, which is why I'd said the relationship is a set of references.

Aha, this explains a lot. I never considered "reference" with this kind of static meaning in mind.

Thus, I've thought it's perfectly okay to simultaneously say that
- a relationship is a reference from one resource (object) to another
- a resource identifier is a reference to a particular resource object

where the first statement implies nothing about the concrete representation or the cardinality of the target.

If we want to narrow the usage of "reference", the first paragraph could say:

> Members of the relationships object represent "relationships", i.e., associations between the enclosing resource object and other resource objects.

In fact, "enclosing" might be preferable to "in which they are defined" regardless of the decision w.r.t. "reference".

> Though aren't we just back to your original complaint, then, that the definition of a "relationship object" isn't really connected conceptually to what comes before?

No. In the published spec, the definition of relationship object just appears out of nowhere. In the current iteration, we have the statement:

> The value at that key MUST be an object ("relationship object") ...

This bridges the gap and is associated with the concept of relationships via the first paragraph that defines relationships object members as relationships.

ethanresnick:
> I love how deep this rabbit hole is going :smile:

Indeed. At least we seem to be getting somewhere :smile: I'll take a look at this with fresh eyes tomorrow.

ethanresnick:
Ok, so it sounds like the text you're proposing is:

> The value of the relationships key MUST be an object (a "relationships object"). Each member in the relationships object represents a "relationship", i.e., an association between the enclosing resource object and other resource objects.
> 
> Relationships may be to-one or to-many.
> 
> A relationship's name is given by its key. The value at that key MUST be an object ("relationship object").
> 
> A "relationship object" MUST contain at least one of the following:
> 
> ....

(Or the above, but replacing "Each member..." with "Members...".)

If that's right, then I'm fine with it! Update the PR and we'll get it merged :)

jelhan:
Thanks a lot for working on this one. I reviewed it looking at latest state of v1.1. I think that clarification is still needed. I created a follow-up merge request at #1608 to finally get it in. I hope that's okay with everyone who had worked on this before. If you have time, I would appreciate a review of that one to ensure that I understood the consensus reached here correctly.

Issue #945, opened 2015-12-15
Redefine 204 Status Code conditions for POST requests


cziegenberg:
For POST request the 204 status code depends on if a client-generated ID has been used or not, because in this case the ID isn't generated by the server and so - theoretically - the resulting resource is the same as the resource in the request.

**Problem 1:**
For PUTs and DELETEs, this also depends on added meta data. This can be also added to POST requests, but isn't handled there. So meta data seems to be treated as important data for the client in some cases, in others not.

**Problem 2:**
The server could use default values or calculate values (the derived fields topic) when creating resources, so the client doesn't have to (or is not allowed to) set all fields in the request. So when the client generated the ID and a 204 response was returned, the client would not have the up-to-date/complete resource data.

**Solution:**
To cover all cases, the text should be changed similar to the one in the update section:
"A server MUST return a 204 No Content status code if the creation is successful and the representation of the resource in the request matches the result."

tkellen:
Can you open this as a PR for discussion please? I think I disagree with your framing here but it would be easier to examine in the context of the document itself.

cziegenberg:
Done. I left the "include a Client-Generated ID" part - it's no more neccessary, but perhaps makes it clearer.

PR #947, opened 2015-12-15
 Redefine 204 Status Code conditions for POST requests

cziegenberg:
PR for  issue #945:

For POST request the 204 status code depends on if a client-generated ID has been used or not, because in this case the ID isn't generated by the server and so - theoretically - the resulting resource is the same as the resource in the request - BUT this isn't true in all cases and so a 204 status can be wrong sometimes.

Problem 1:
For PUTs and DELETEs, this also depends on added meta data. This can be also added to POST requests, but isn't handled there. So meta data seems to be treated as important data for the client in other cases, but not in case of POSTs.

Problem 2:
The server could use default values or calculate values (the derived fields topic) when creating resources, so the client doesn't have to (or is not allowed to) set all fields in the request. So when the client generated the ID and a 204 response was returned, the client would not have the up-to-date/complete resource data.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1211,8 +1211,9 @@ to create a resource with a client-generated ID.
 
 ##### <a href="#crud-creating-responses-201" id="crud-creating-responses-201" class="headerlink"></a> 201 Created
 
-If a `POST` request did not include a [Client-Generated
-ID](#crud-creating-client-ids) and the requested resource has been created
+If the result of a `POST` request *doesn't match* the representation of the resource in the request 
+(which is the case when additional fields are added by the server on resource creation, e.g. when 
+no [Client-Generated ID](#crud-creating-client-ids) is used) and the requested resource has been created
 successfully, the server **MUST** return a `201 Created` status code.
 
 The response **SHOULD** include a `Location` header identifying the location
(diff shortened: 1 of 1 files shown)


bintoro:
I like the idea, but the wording needs to be adjusted. Now it only says what must happen when the result is identical to the request and specifies nothing about the opposite case.

One possibility is to leave the existing paragraph alone and simply add a qualifying statement like

> However, a server **MUST NOT** return a `204 No Content` if the omitted representation would have been different from the one received with the request.

cziegenberg:
@bintoro You were right. I thought this was defined in the 201 Status section, but this excluded requests with client-generated IDs. I adjusted both definitions, so all cases should be defined now.

bintoro:
The language is getting very confusing, and I think the root cause is the partitioning of the section by status code. It doesn't fit the content in the best possible way.

The client-generated ID thing (more generally, the _representation changed_ thing) creates these two scenarios that have zero relevance for 201, which is always OK to use.

If I were to decide, I'd keep things simple by merging the 201 and 204 segments and stating it like this:

---

If the requested resource has been created successfully, the server **MUST** return a `201 Created` status code.

The response **MUST** also include a document that contains the primary resource created.

Notwithstanding the above, a server **MAY** choose to return a `204 No Content` status code without an accompanying response document if the omitted document would have been identical to the original request document sent by the client.

> Note: In particular, usage of client-generated IDs is a prerequisite for returning a `204` response, as a response document is necessary to communicate server-generated IDs back to the client.

---

While the 204 exception _could_ remain under its own 204 heading, I believe it would be best to keep it right next to the main rule.

I'm not sure about the strict "identical" requirement, but I'm certainly in favor of generalizing from the current "client-generated ID" requirement. The main point is that if there are any changes, they must be relatively insignificant so as not to invalidate the client's representation.

cziegenberg:
I tried to keep the changes of the existing text minimal, also because @tkellen was against this change (feedback welcome), but I agree that your version sounds much clearer.

I'll adjust the PR, but seperate the two sections, because that's the way this is done for the other request methods and status codes.

cziegenberg:
I also agree that "identical" isn't correct here. 

The object properties can be in a different order or you can use an string instead of an object for a link. So what's meant is "identical with regard to contents" or "identical after normalization".

ethanresnick:
@dgeb @jmeas I'd like to merge some version of this. Seems like an easy fix. Thoughts?

dgeb:
@ethanresnick I agree with the spirit of this PR but think the following sentence needs to be qualified:

> If the requested resource has been created successfully, the server **MUST** return a `201 Created` status code.

The exception to this case is not mentioned until the 204 section.

ethanresnick:
@dgeb Yeah, I agree. Do you have preferred wording? If so, one of us could take this branch over, update the wording, and merge it.

PR #931, opened 2015-12-01
Added favicon.ico

iamsebastian:
Hej guys,

Thanks for your wonderful work.

While I'm creating a new node/Express-API in combination with a Ember app, I always have some pinned tabs open, that points to the jsonapi.org.
Is it possible, that you could add a favicon (this, or another) to your jekyll gh-pages? This would help a lot to see, which tab is jsonapi.org and which tab is another one, without a favicon.

Thanks in advance.

PR #953, opened 2015-12-24
Add a favicon

ethanresnick:
This is the alternative I mentioned in https://github.com/json-api/json-api/pull/931#issuecomment-166286832. Take a look and lmk if there are any objections!


tkellen:
This looks good to me! I think we should update disqus to match.
screen shot 2015-12-28 at 7 52 28 am

PR #954, opened 2015-12-24
Fully partition unicode into allowed and reserved characters

ethanresnick:
On discuss, a user [pointed out](http://discuss.jsonapi.org/t/clarification-on-allowed-characters-in-member-names/274) that the existing wording around reserved characters might be a bit confusing and that, currently, the "Allowed Characters" and "Reserved Characters" sets don't fully partition Unicode. In particular, the C0 controls and the delete key aren't in either group. This small PR attempts to address those concerns.


tkellen:
I'm good with this. My only concern is how strict we want to be about the fact that this is not a backwards compatible change.

If @dgeb or @steveklabnik is :+1: as well, let's land it.

dgeb:
Adds clarity and seems very unlikely to conflict with any implementations :+1:

PR #961, opened 2016-01-06
Additional NodeJS implementations

theninj4:
We've been pushing hard to clean up documentation and fix up any+all discrepancies between our implementation and the v1 spec - we're in a position where we've got a bunch of cool modules worthy of inclusion in your list of implementations.


steveklabnik:
Thank you!

PR #964, opened 2016-01-12
Fix typo in Query Parameters section: "proccess"

valscion:
This JSON API kicks ass! I noticed a typo in the `Query Parameters` section in [format page](http://jsonapi.org/format/#query-parameters) so here's a PR to fix that :smile:


ethanresnick:
Thanks!

PR #973, opened 2016-01-29
added py-jsonapi

AdorablePotato:
Hi,

I've written a Python3 library for this specification. It can be extended easily to work with every framework and database driver. Currently, support for tornado, flask, mongoengine and sqlalchemy is included. Following the invitation on jsonapi.org, I'd like to leave a short note, that it exists :)


steveklabnik:
Thanks so much!

PR #980, opened 2016-02-09
Suggested structural change and fix to meet schema

MrTrick:
Moved "relationship" out of "relationships" into its own definition, added missing "type":"object" requirement.

**Previously:**
1. There was no (good) way to reference the "relationship object" definition.
2. The schema did not check that "relationship object" was actually an object.


MrTrick:
This PR is related to #946, but could be merged with/without it.

PR #985, opened 2016-02-19
Provide wikipedia link about ''bikeshedding'

ianks:
Since not all people know the 'bikeshedding' colloquialism, provide them with information to research the term.

### DIFF
--- a/index.md
+++ b/index.md
@@ -5,7 +5,7 @@ show_masthead: true
 ---
 
 If you've ever argued with your team about the way your JSON responses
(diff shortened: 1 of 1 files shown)


steveklabnik:
Could we maybe link to http://bikeshed.org/ instead? It's more of a primary resource than Wikipedia.

That said, :+1: to this overall.

ianks:
@steveklabnik Done.

steveklabnik:
Thanks!

PR #988, opened 2016-02-22
Upgrade deps

steveklabnik:
So, GitHub upgraded to Jekyll 3: https://github.com/blog/2100-github-pages-now-faster-and-simpler-with-jekyll-3-0

But we don't build properly with it. I've been getting emails but only now had the chance to look into it.

This doesn't yet _work_ but it at least gets us back with what GitHub should be actually using. I couldn't figure out why it doesn't work, so I filed https://github.com/jekyll/jekyll/issues/4564. That should block this PR from landing.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,12 +3,12 @@
 source 'http://rubygems.org'
 
 gem "rake", "~> 10.0"
-gem "jekyll", "~> 2.4.0"
+gem "jekyll", "~> 3.0.0"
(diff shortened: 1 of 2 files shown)


ethanresnick:
Thanks for handling this Steve! I'll look into the redcarpet->kramdown switch soon

steveklabnik:
Cool! I _think_ that it should Just Work, but haven't had the chance to deal with it yet.

PR #992, opened 2016-02-25
Adjust so that description of paths goes along with accompanying note

cheerfulstoic:
This was really confusing to me when I read it, but then I realized that the note seems to go with the description of formatting `include` parameters, not with the item about identifying a relationship path.

### DIFF
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -926,12 +926,12 @@ If an endpoint supports the `include` parameter and a client supplies it,
 the server **MUST NOT** include unrequested [resource objects] in the `included`
 section of the [compound document].
 
+If a server is unable to identify a relationship path or does not support
+inclusion of resources from a path, it **MUST** respond with 400 Bad Request.
+
 The value of the `include` parameter **MUST** be a comma-separated (U+002C
 COMMA, ",") list of relationship paths. A relationship path is a dot-separated
 (U+002E FULL-STOP, ".") list of [relationship][relationships] names.
-
-If a server is unable to identify a relationship path or does not support
-inclusion of resources from a path, it **MUST** respond with 400 Bad Request.
 
 > Note: For example, a relationship path could be `comments.author`, where
 `comments` is a relationship listed under a `articles` [resource object][resource objects], and
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -926,12 +926,12 @@ If an endpoint supports the `include` parameter and a client supplies it,
 the server **MUST NOT** include unrequested [resource objects] in the `included`
 section of the [compound document].
 
+If a server is unable to identify a relationship path or does not support
+inclusion of resources from a path, it **MUST** respond with 400 Bad Request.
+
 The value of the `include` parameter **MUST** be a comma-separated (U+002C
 COMMA, ",") list of relationship paths. A relationship path is a dot-separated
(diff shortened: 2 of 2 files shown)


cheerfulstoic:
While we're on the topic, the reason I ran into this was because I had a collection of resources which were different types.  I wanted to `include` on a relationship which one of those `type`s has.  Has there been a discussion on allowing this?  I'm not really sure how it could be supported because it might be possible that resources of the type for which the `include` applies to might not be returned in a particular request...

For now I've found what is probably a more elegant solution anyway.

bintoro:
`include[<type>]=<path>,...` would be the logical way to do it. I don't remember if this never came up or if there was a reason not to support this format.

_format/1.0/index.md
```diff
@@ -926,13 +926,13 @@ If an endpoint supports the `include` parameter and a client supplies it,
 the server **MUST NOT** include unrequested [resource objects] in the `included`
 section of the [compound document].
 
+If a server is unable to identify a relationship path or does not support
+inclusion of resources from a path, it **MUST** respond with 400 Bad Request.
+
```
bintoro, line 931: If the problem is with the note, then I think it's the note that should be moved up. This paragraph clearly belongs after the next one.

cheerfulstoic, line 931: That's actually what I did at first, but then the paragraph starting with "For instance" seems like it applies to the paragraph starting with "If a server is unable to identify a relationship path".  The "For instance" paragraph is the start of examples about `include`ing, so perhaps it should be another section?

bintoro, line 931: I don't think so. A note is not really a cutoff of any sort. Then again, I also fail to see any issue with the current organization.

It occurs to me, though, that the note is somewhat superfluous. The same path is used in the example request that follows soon after. Maybe the substance of the note would be better placed as an ordinary paragraph after the example:

> `GET /articles/1?include=comments.author HTTP/1.1`
> `Accept: application/vnd.api+json`
> 
> Here, `comments` is a relationship listed under an `articles` resource object, and `author` is a relationship listed under a `comments` resource object.

cheerfulstoic, line 931: Sorry, I've got a ton of things on my plate right now.  Feel free to close the PR and resolve however you think is best.  The order just confused me and my pair for a few minutes when we were trying to go over the spec ;)

(thanks for JSON API, BTW, I'm loving it!)

ethanresnick:
The reasoning behind the current order is that the paragraph that's currently on top is the one that defines what a "relationship path" is. So, if we reverse the order, it seems to me like we're using the term relationship path before defining it.

I'm going to close this for now, but I'd be open to a PR that addresses this issue in another way (maybe by moving the note or working its content in after the example).

> While we're on the topic, the reason I ran into this was because I had a collection of resources which were different types. I wanted to include on a relationship which one of those types has.

My interpretation has always been that this is/should be allowed. That conclusion, though, rests on what it means for a server to be able to "identify a relationship path". If endpoint A returns types B and C, and type B has a relationship `x`, then I think the server can "identify" `x` as a valid relationship path on a request to A. That said, I agree that the spec is vague, and this is actually one of two vagueries with `include`. There was an attempt to specify this stuff more clearly pre-1.0, but all the approaches we tried got very wording and confusing. Maybe this is worth revisiting now... I'll open another issue.

Issue #587, opened 2015-05-01
JSON API / JSON-LD integration


elplatt:
I've been looking over both specifications, and I can see two approaches for apps wishing to use both:
1. LD over API: Serve the JSON-LD document as the "data" field in the JSON API result.
2. API over LD: Augment the JSON API result with a JSON-LD context object defining the JSON API schema.
### LD over API

The advantage is that JSON API wraps the data and JSON-LD formats the data, which seems in line with the intentions of both specs. The challenge is that the "data" object is required to have an "id" and "type" attribute, which is not part of the JSON-LD spec. It's workable but very inconvenient to use both the JSON API type/id and the JSON-LD @type/@id. It would be preferable to change JSON API to use JSON-LD-style @type/@id names and syntax.
### API over LD

The advantage is that JSON-LD is general enough that JSON API is a valid subset (if the root object is augmented with an appropriate context/schema). The disadvantage is that because none of the JSON-API data includes @type or @id keys, they would need to be defined in the context. There would be a lot of duplication because each context would have to include definitions for the standard JSON API type/id/self fields.
### Conclusion

It seems to me that the LD over API approach would be preferable, but that would require changes to JSON API that aren't backwards compatible (renaming type/self/id fields). I'd be happy to help make these changes.

dgeb:
We have discussed using prefixes such as `@` many times previously and decided against it. I'm sure that we won't be making such a change to the base spec this close to 1.0 (May 21).

My suggestion is to explore hybrid approaches in (non-additive) extensions.

ethanresnick:
As I was thinking tonight about extensions, I wanted to mock up what a JSON-LD–annotated version of a JSON-API resource object would look like. Below is what I came up with. I'm leaving it here for reference, in case it helps other implementors and/or our efforts at interoperability. Others can comment here too if they figure out a simpler way to do it.

```
{
  "@context": {
    "links": {"@id": "_:links", "@container": "@set"},
    "attributes": {"@id": "_:attributes", "@container": "@set"},
    "relationships": {"@id": "_:relationships", "@container": "@set"}
  },

  "@id": "http://example.com/live-events/4",
  "@type": "http://schema.org/Event",

  "id": "4",
  "type": "live-events",
  "links": {
    "@context": {
      "self": "@id",
      "collection": {"@id": "http://www.iana.org/assignments/link-relations/collection"}
    },
    "self": "http://example.com/live-events/4",
[...]

PR #995, opened 2016-02-26
Allow JSON LD to be used with JSON API

ethanresnick:
This change was originally part of the profile extensions PR, because I've long seen allowing JSON LD data as a key requirement for the extension system's design. However, I'm breaking this change out into a separate PR, per @dgeb's suggestion, to make it more visible and invite any discussion on it particularly.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -631,6 +631,26 @@ The following characters **MUST NOT** be used in member names:
 - U+007F DELETE
 - U+0000 to U+001F (C0 Controls)
 
+#### <a href="#document-member-names-at-members" id="document-member-names-at-members" class="headerlink"></a> @-Members
+
+Member names **MAY** also begin with an at sign (U+0040 COMMERCIAL AT, "@"). 
+Members named this way are called "@-Members". @-Members **MAY** appear 
+anywhere in a JSON API document. 
+
+However, JSON API processors **MUST** completely ignore @-Members (i.e. not 
+treat them as JSON API data). 
+
+Moreover, the existence of @-Members **MUST** be ignored when interpreting all 
+JSON API definitions and processing instructions given outside of this 
+subsection. For example, an [attribute][attributes] is defined above as any 
+member of the attributes object. However, because @-Members must be ignored 
+when interpreting that definition, an @-Member that occurs in an attributes 
+object is not an attribute. 
+
+> Note: Among other things, "@" members can be used to add JSON-LD data to a 
+JSON API document. Such documents should be served with [an extra header](http://www.w3.org/TR/json-ld/#interpreting-json-as-json-ld) 
+to convey to JSON-LD clients that they contain JSON-LD data.
+
 ## <a href="#fetching" id="fetching" class="headerlink"></a> Fetching Data
 
 Data, including resources and relationships, can be fetched by sending a
--- a/_format/1.1/normative-statements.json
+++ b/_format/1.1/normative-statements.json
@@ -89,7 +89,11 @@ layout: null
             { "id": "member-name-globally-allowed", "type": "normative-statements" },
             { "id": "member-name-url-safe", "type": "normative-statements" },
             { "id": "member-name-allowed-characters", "type": "normative-statements" },
-            { "id": "member-name-reserved-characters", "type": "normative-statements" }
+            { "id": "member-name-reserved-characters", "type": "normative-statements" },
+            { "id": "member-name-at-members-definition", "type": "normative-statements" },
+            { "id": "member-name-at-members-where", "type": "normative-statements" },
+            { "id": "member-name-at-members-no-process", "type": "normative-statements" },
+            { "id": "member-name-at-members-ignore-for-definitions", "type": "normative-statements" }
           ]
         }
       }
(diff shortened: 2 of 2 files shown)


tkellen:
:+1:

_format/1.1/index.md
```diff
@@ -631,6 +631,13 @@ The following characters **MUST NOT** be used in member names:
 - U+007F DELETE
 - U+0000 to U+001F (C0 Controls)
 
+#### <a href="#document-member-names-jsonld" id="document-member-names-jsonld" class="headerlink"></a> JSON-LD Exception
+
+When a JSON API document is [served as JSON-LD](http://www.w3.org/TR/json-ld/#interpreting-json-as-json-ld),
+the document **MAY** include member names that begin with an at sign
+(U+0040 COMMERCIAL AT, "@") to convey JSON-LD information. JSON API processers
```
tkellen: > JSON API processers that don't understand JSON-LD **MUST** ignore these members.

This seems redundant?

tkellen: By that I mean to say yes, this allows JSON-LD to be used, but we don't need to be prescriptive if people want to use it for something else.

ethanresnick: Re the new MUST, yes, it is redundant, in the sense that it's the logical consequence of the existing requirement that "Client and server implementations MUST ignore members not recognized by this specification.")

Re not being prescriptive, I'm not sure I follow... are you saying people should be able to use @-prefixed members for non-JSON LD data?

bintoro: +1 to what @tkellen said. What are the downsides of simply adding "@" to the set of allowed characters?

ethanresnick: So, allowing "@" as a first character everywhere is "neater", in that it doesn't create a special case just for JSON LD. My thinking for writing it this way, though, was: 
1. It preserves the ability for us to use "@" in, e.g., a future query parameter's syntax.
2. Using an "@"-prefixed key for non-JSON LD data is a bad idea, because it means that if JSON LD adds a key of the same name in the future, your use may cause problems. This narrower allowance takes away that footgun.
3. It prevents "@" signs from creeping in to attribute, relationship, and other names, which we may not want for aesthetic reasons (seeing how much we've done to keep prefixes out), and which could be confusing to readers trying to distinguish JSON API data from JSON LD data.

dgeb:
Thanks for writing this up, @ethanresnick. 

Explicitly calling out JSON-LD has a couple effects:
A) it makes the exception for `@` usage as narrow as possible (as @tkellen mentioned inline)
B) it implicitly allows JSON-LD members _anywhere_ in the document

We should ensure that both A + B are desirable. And we should decide if we want B to be made more explicit, since it's a pretty big exception we'd be carving out for JSON-LD.

ethanresnick:
@dgeb See my other note about part A. I could be probably persuaded either way on that.

Re part B... I don't think JSON LD integration will be workable if we restrict where in the document the JSON LD data can occur.  JSON LD's syntax just isn't expressive enough to effectively annotate a JSON API document unless its able to put those annotations pretty close to the data its referring to. See [here](https://github.com/json-api/json-api/issues/587#issuecomment-103367380) to get a sense of what I'm talking about.

steveklabnik:
I am also _very_ hesitant to add JSON-LD specific features into the spec, but a generic thing that allowed it to work better with JSON-LD is okay.

dgeb:
I'm a bit nervous about the implications of allowing `@` members for anything and everything custom, wherever in a document. But I'm probably even less thrilled about an exception explicitly for JSON-LD.

All things considered, I think that I favor a more general exception.

ethanresnick:
@steveklabnik @dgeb Can you talk more about your hesitancy to make a JSON-LD specific exception? Is it that it seems too ugly? Or that it'll be confusing?

To me, a narrow exceptions seems like the more cautious approach, in that we can always loosen the restrictions later to allow `@` members everywhere, but we can't do the reverse.

steveklabnik:
I do not want the spec to look like it's recommending JSON-LD, because I don't think that JSON-LD is a good way to build APIs. This is entirely a personal opinion, and I don't want to block people who _do_ enjoy it to use it. But I am afraid that a "And do this to work with JSON-LD" in the spec is an endorsement that may lead people astray.

ethanresnick:
> But I am afraid that a "And do this to work with JSON-LD" in the spec is an endorsement that may lead people astray.

Is there a way to both 1) have narrow-tailoring of the exception and 2) make it clear that we're not making an endorsement? Maybe by changing how we word the rule or the text around it?

ethanresnick:
Here's another possibility for the wording. It aims to narrowly-tailor the exception without mentioning JSON-LD specifically in the normative text:

> #### "@" Members
> 
> Member names **MAY** also begin with an at sign (U+0040 COMMERCIAL AT, "@"). However, these members **MUST** be completely ignored by JSON API parsers (i.e., not treated as JSON API data). Accordingly, such member names **MUST NOT** be given to [fields] or to any other members that JSON API consumers are expected to process.
> 
> > Note: Among other things, "@" members can be used to add JSON-LD data to a JSON API document. Such documents should be served with an [extra header](http://www.w3.org/TR/json-ld/#interpreting-json-as-json-ld) to convey to JSON-LD clients that they contain JSON-LD data.

Thoughts?

ethanresnick:
@steveklabnik @tkellen @dgeb Have you had a chance to look at my revised proposal for the wording above? I think it'd be awesome if we could discuss that today and tomorrow and have this issue settled before Wednesday's meeting, so we can move on to other things.

@bintoro Feel free to chime in to if you have any ideas.

steveklabnik:
I have not, but will be checking it out in a day or two, sorry :(

steveklabnik:
I think it's better, for sure.

ethanresnick:
@steveklabnik Anything you'd change?

dgeb:
@ethanresnick I think your [revision above](https://github.com/json-api/json-api/pull/995#issuecomment-194506665) strikes a better balance by mentioning JSON-LD non-normatively. 

My main concern is that the "member names" section describes members that are in user-space, like fields. And this is `@` exception allows defining members in the inverse space (i.e. where users can NOT define members otherwise). This distinction is important and I'm not sure it's covered completely by:

> Accordingly, such member names MUST NOT be given to [fields] or to any other members that JSON API consumers are expected to process.

How about: 

> Member names that begin with an at sign (U+0040 COMMERCIAL AT, "@") MAY appear anywhere in a JSON API document except as [fields] or other members that JSON API consumers are expected to process. These `"@"` members MUST be ignored by JSON API parsers (i.e. not treated as JSON API data).

steveklabnik:
@ethanresnick not at the moment, but I'm just trying to be cautious with my wording because I am extremely jetlagged. It reads just fine to me now, but what about once my brain is out of this fog?

ethanresnick:
@dgeb When you say:

>  this @ exception allows defining members in the inverse space (i.e. where users can NOT define members otherwise).

are you imagining that this exception would allow users to define @ members _only_ in the spaces where they can't define members otherwise? Or were you imagining that @ members could go both in that space and in the space that users can already add (non @)-members to?

I think I was imagining the latter, for the reason I alluded to [earlier](https://github.com/json-api/json-api/pull/995#issuecomment-189480062), namely: that JSON LD integration may not be possible if users are limited in where they can put their @ members.

I see problems (contradictions and potential ambiguities) with my proposed wording either way, but before resolving those I want to make sure we're on the same page conceptually about the exception's scope.

@steveklabnik Fair enough!

tkellen:
I'm strongly in favor of allowing `@` prefixed members anywhere. If we aren't winning JSON-LD support with this addition, I have a hard time seeing the value in it at all.

dgeb:
Ok, I can accept allowing `@` prefixed members everywhere, since technically they should already be ignored by 1.0 processors. It's not clear (to me) that [the version](https://github.com/json-api/json-api/pull/995#issuecomment-194506665) that @ethanresnick wrote above applies to members anywhere in the document (and not just in places where user-defined members are explicitly allowed like `attributes`). Just a minor nit, so let's move this forward.

ethanresnick:
> Ok, I can accept allowing `@` prefixed members everywhere, since technically they should already be ignored by 1.0 processors. 

:+1: 

> It's not clear (to me) that the version that @ethanresnick wrote above applies to members anywhere in the document (and not just in places where user-defined members are explicitly allowed like attributes). 

Agreed! I'm going to take a stab at rewriting it this weekend.

ethanresnick:
@tkellen @dgeb @steveklabnik Can ya'll take a look at this before our meeting so we can try to get this merged today?

_format/1.1/index.md
```diff
@@ -631,6 +631,26 @@ The following characters **MUST NOT** be used in member names:
...
+
+Member names **MAY** also begin with an at sign (U+0040 COMMERCIAL AT, "@"). 
+Members named this way are called "@-Members". @-Members **MAY** appear 
+anywhere in a JSON API document. 
+
+However, JSON API processors **MUST** completely ignore @-Members (i.e. not 
+treat them as JSON API data). 
+
+Moreover, the existence of @-Members **MUST** be ignored when interpreting all 
+JSON:API definitions and processing instructions given outside of this
```
dgeb: "JSON:API" => "JSON API" (until we perform a global swap of the terms).

dgeb:
:+1: LGTM except that minor nit. Thanks for your patience here @ethanresnick

steveklabnik:
:+1:

ethanresnick:
Awesome! Updated with that change, and to add the new content to the normative statements file.

dgeb:
Great! Thanks for seeing this through @ethanresnick :shipit:

PR #1000, opened 2016-03-01
Clarify the Location requirement (fixes #999)

### LINKED ISSUE #999: Make 201s more RFC 7231 compliant
The “Creating Resources” section [says](http://jsonapi.org/format/upcoming/#crud-creating-responses-201) that a `Location` header **SHOULD** be sent on a 201 response.

But [RFC 7231 § 6.3.2](http://tools.ietf.org/html/rfc7231#section-6.3.2) says:

> The primary resource created by the request is identified by either a Location header field in the response or, if no Location field is received, by the effective request URI.

A (hypothetical) implementation of this RFC 7231 text would infer the wrong URI from JSON API’s 201 response without `Location`.

To avoid confusing implementors, I suggest elevating the requirement for `Location` to a **MUST**, if that’s OK by your compatibility policy. If not, then maybe just include a note along the lines of:

> If you choose not to send `Location` with a 201 response, you will violate RFC 7231 § 6.3.2.

vfaronov:
I didn’t touch `normative-statements.json` because I don’t know what’s that for and the change is non-normative anyway.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1220,7 +1220,8 @@ ID](#crud-creating-client-ids) and the requested resource has been created
 successfully, the server **MUST** return a `201 Created` status code.
 
 The response **SHOULD** include a `Location` header identifying the location
-of the newly created resource.
+of the newly created resource, in order to comply with [RFC
+7231](http://tools.ietf.org/html/rfc7231#section-6.3.2).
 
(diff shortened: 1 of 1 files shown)


dgeb:
:+1:

ethanresnick:
LGTM! Merged.

PR #1002, opened 2016-03-05
Added jsonapi-converter link to the list of client libraries

jasminb:
This PR adds link to the Java client library that I created recently.

Following is the list of futures:
- Serialisation/deserialisation using annotated Java classes
- Retrofit plugin
- Relationship resolution using pluggable resolves (global or per type)

Library uses jackson for JSON handling.

More details: https://github.com/jasminb/jsonapi-converter.


ethanresnick:
Thanks!

PR #1003, opened 2016-03-06
removes trailing comma from JSON object

jfinkels:


steveklabnik:
Nice catch, thank you!

PR #1012, opened 2016-03-20
`oh-my-jsonapi` is now `jsonapi-mapper` - renaming and updating link

jamesdixon:


dgeb:
Thanks for the update!

PR #1015, opened 2016-03-26
Fix footer Social & License links

A-Helberg:
The clickable area for the links in the footer is slightly to small and does not extend the entire child image/icon.

This PR aims to fix that


dgeb:
That is an improvement. Thanks for cleaning up these links @A-Helberg, and sorry for the delay reviewing!

PR #1018, opened 2016-04-03
Remove jsonapify from server implementations

alex94cp:
Decided to remove jsonapify from the list while I work on [node-apioid](https://github.com/alex94puchades/node-apioid): an ES6-ready, more powerful and easy way to deliver jsonapi-compatible APIs.


ethanresnick:
Ok!

PR #1026, opened 2016-04-13
Deprecate existing extensions

ethanresnick:
This follows a conversation with @dgeb today. The thinking is that keeping the old extensions, and the old negotiation rules, online does more harm than good, as we don’t want people to use them anymore nor do we want to support them going forward.


dgeb:
LGTM :+1:

galiat:
@ethanresnick - I'm glad you updated the docs since I was about to start implementing the now-deprecate extensions. I'm curious what the plan moving forward is. Is there some place I could follow along with the discussion?

ethanresnick:
Yes, in #957

ethanresnick:
Yes, in #957

ethanresnick:
Yes, in #957

PR #1029, opened 2016-04-17
adding 404 response for deleteing non-existent resource

### LINKED ISSUE #1023: Add response for DELETE on resource that does not exist
So as far as I understand when I run `DELETE /resource/1` and this resource does not exists, the correct response is a `204` because the result is the expected one and the same as if the resource would have existed. Is this correct? 

If so, I would suggest to add this to the specs to make it explicit.

lukasoppermann:
This PR fixes #1023 which is about a not specified response for a `DELETE` to a resource that does not exist.

### DIFF
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -1719,6 +1719,11 @@ request is successful and no content is returned.
 A server **MUST** return a `200 OK` status code if a deletion request is
 successful and the server responds with only top-level [meta] data.
 
+##### <a href="#crud-deleting-responses-404" id="crud-deleting-responses-404" class="headerlink"></a> 404 NOT FOUND
+
+A server **SHOULD** return a `404 Not Found` status code if a deletion request fails
+due to the resource not existing.
+
 ##### <a href="#crud-deleting-responses-other" id="crud-deleting-responses-other" class="headerlink"></a> Other Responses
 
 A server **MAY** respond with other HTTP status codes.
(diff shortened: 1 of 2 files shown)


ethanresnick:
This looks good to me. Any objections @json-api/owners?

One question is whether we should say a 404 MAY be returned or SHOULD be returned. I'm tempted to say SHOULD, because HTTP's most recent specification pretty clearly implies that DELETE being idempotent doesn't mean you have to return the same status code each time (which was the argument for MAY in #1023):

> Idempotent methods are distinguished because the request can be repeated automatically if a communication failure occurs before the client is able to read the server's response... It knows that repeating the request will have the same intended effect, even if the original request succeeded, **though the response might differ**. (Emphasis added.)

But, if there's disagreement on a SHOULD, I don't think that technical debate should hold this up, as we can always strengthen a MAY to a SHOULD later.

lukasoppermann:
Even though I am not part of the owners, I would be in favour of SHOULD.

dgeb:
I am fine strengthening to a SHOULD here as well. I think a MAY is almost meaningless compared with the general catchall response section.

dgeb:
@lukasoppermann please update to SHOULD and then we'll get this merged.

lukasoppermann:
@dgeb updated.

dgeb:
Thanks @lukasoppermann! :shipit:

PR #1034, opened 2016-04-26
Fix miss link

GiaoGiaoCat:


ethanresnick:
Thank you for this! My guess is that the link isn't showing up currently because the parser's getting confused by having three square-bracketed segments right next to each other (only separated by spaces). So this should fix it.

PR #1035, opened 2016-04-27
Clarify full linkage.

beauby:
The current definition of _full linkage_ formally allows full linkage to happen only between resources within `included`, whereas the intention is (I believe) to ensure every resource object within `included` is identified by at least one resource identifier object within the transitive closure of primary data by its relationships. Any input on better wording warmly welcome. (:

TL;DR: Currently the spec could be interpreted to allow the following payload:

``` json
{
  "data": null,
  "included": [
    {
      "type": "people", 
      "id": "1",
      "relationships": {
        "best-friend": {
          "data": { "type": "people", "id": "2" }
        }
      }
    },
    {
      "type": "people", 
      "id": "2",
      "relationships": {
        "best-friend": {
          "data": { "type": "people", "id": "1" }
        }
      }
    }
  ]
}
```

### DIFF
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -375,17 +375,19 @@ In a compound document, all included resources **MUST** be represented as an
 array of [resource objects] in a top-level `included` member.
 
 Compound documents require "full linkage", meaning that every included
-resource **MUST** be identified by at least one [resource identifier object]
-in the same document. These resource identifier objects could either be
-primary data or represent resource linkage contained within primary or
-included resources.
+resource **MUST** be identified by at least one [resource identifier object] in
+the same document, representing resource linkage contained within an
+*accessible* resource object. A resource object is *accessible* if it is either
+primary data or it is identified by at least one [resource identifier object] in
+the same document, representing resource linkage contained within an
+*accessible* resource object.
 
 The only exception to the full linkage requirement is when relationship fields
 that would otherwise contain linkage data are excluded via [sparse fieldsets](#fetching-sparse-fieldsets).
 
-> Note: Full linkage ensures that included resources are related to either
-the primary data (which could be [resource objects] or [resource identifier
-objects][resource identifier object]) or to each other.
+> Note: Full linkage ensures that included resources are related (directly or
+indirectly) to the primary data (which could be [resource objects] or [resource identifier
+objects][resource identifier object]).
 
 A complete example document with multiple included relationships:
 


bintoro:
Good catch. The present wording does seem to contain a loophole.

The proposed normative text has a bug: the origin reference need not be a resource object (think primary data from a relationship URL).

The simple change you made to the trailing note was a really good one. It now explains the whole thing so well that if the normative part can't achieve the same level of clarity, I would just promote the note to normative and build on that.

Suggestion:

Compound documents require "full linkage", meaning that every included resource **MUST** be referenced from the primary data either directly or via intermediate included resources.

beauby:
You are right, I didn't take into consideration the case of fetching relationships directly and including its data in `included` (though I'm not sure I can think of a useful use-case – maybe polymorphic has-many?).

I'm not satisfied with my current proposed wording because it feels quite redundant (this is because I chose to take the transitive closure of _neighbors_ of primary resource identifiers, and because we somehow need to identify resource objects and resource identifier objects).

How about the following:

> Compound documents require "full linkage", meaning that every included
> resource **MUST** be identified by at least one _accessible_ 
> [resource identifier object] in the same document.
> A [resource object] or [resource identifier object] is _accessible_ if it is either primary data,
> or it represents resource linkage contained within an _accessible_ resource object.

Edit: concerning your suggestion, I find it great, as it conveys the intuition of what's meant to someone familiar with the spec, but I'm afraid it may be too vague for newcomers (though I'm aware being too formal won't help much either).

beauby:
Formally, what we are trying to say is:

> A JSON API document represents an underlying directed multigraph where resources are vertices and relationships represent arcs. Formally, for every pair of resources `A`, `B`, there is an arc from `A` to `B` labelled `R` if `B` is represented by a resource identifier object within a relationship `R` in `A`.
> In this context, full linkage expresses that all resources in `included` must belong to the transitive closure of `data`.

beauby:
Any wording suggestion on this one?

wimleers:
I'd love to see this clarified too, I know that "full linkage" has been a point of discussion multiple times while working on https://www.drupal.org/project/jsonapi.

jelhan:
This clarification seems to be highly needed. I think the proposed change points into the correct direction. But as pointed out by @bintoro it does not work if the primary data is resource identifier objects. I tried to improve the proposal to also consider that case.

> Compound documents require "full linkage", meaning that every included
> resource **MUST** be identified by at least one [resource identifier object]
> in the same document. These resource identifier object can be primary
> data or represent resource linkage contained within an accessible resource.
> A resource is accessible if it is included in the same document as primary
> data or if it is identified by a resource identifier object in primary data
> or represents resource linkage contained within an accessible resource.

jelhan:
@beauby Are you still motivated to push this forward? If not, I can fully understand after more than four years. I could take over in that case and provide a new pull request with my proposal based on your original one.

beauby:
Hey @jelhan, feel free to take over and I'll close this PR once yours is up!

Issue #1033, opened 2016-04-25
How to deal with requests to create/update relationships while the related resources do not exist?


RussellLuo:
According to the [spec](http://jsonapi.org/format/), there may be a request to create an article with an author relationship:

```
POST /articles HTTP/1.1
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json

{
  "data": {
    "type": "articles",
    "attributes": {
      "title": "JSON API paints my bikeshed!"
    },
    "relationships": {
      "author": {
        "data": { "type": "people", "id": "9" }
      }
    }
  }
}
```

Or a request to update the existing author relationship:

```
PATCH /articles/1/relationships/author HTTP/1.1
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json

{
  "data": { "type": "people", "id": "12" }
}
```

If the related author (`/people/9` or `/people/12`) does not exist, how to deal with the above requests:
1. Respond errors for the missing author?
2. Always create/update the relationship regardless of the missing author?

dgeb:
You should return a 404 here. See http://jsonapi.org/format/#crud-updating-responses-404

> A server MUST return 404 Not Found when processing a request that references a related resource that does not exist.

PR #1036, opened 2016-04-27
Add response status for POST with inexistent related resource.

beauby:
related: #1033

### DIFF
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -1271,6 +1271,11 @@ had returned it back in a `201` response.
 A server **MAY** return `403 Forbidden` in response to an unsupported request
 to create a resource.
 
+##### <a href="#crud-creating-responses-404" id="crud-creating-responses-404" class="headerlink"></a> 404 Not Found
+
+A server **MUST** return `404 Not Found` when processing a request that
+references a related resource that does not exist.
+
 ##### <a href="#crud-creating-responses-409" id="crud-creating-responses-409" class="headerlink"></a> 409 Conflict
 
 A server **MUST** return `409 Conflict` when processing a `POST` request to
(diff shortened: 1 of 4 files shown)


dgeb:
I agree that this should be spelled out under creating resources, as it is under updating resources (see http://jsonapi.org/format/#crud-updating-responses-404).

We need to decide if this should land in 1.0, 1.1, or both.

beauby:
Anything more I can do to help on this one?

dgeb:
@beauby I've labeled this as needing further discussion because of the MUST.

beauby:
Fair enough – my understanding was that this was highly implied by the spirit of the spec, but forgotten to be clearly spelled out.

dgeb:
I agree. I think this change is entirely consistent with the "updating resources" section, and I am not really in favor of downgrading the MUST because its omission was just an oversight.

ethanresnick:
I  trust @dgeb's judgement re adding the MUST.

Re landing in 1.0, 1.1 or both...to date, we've been pretty loose about backporting things to 1.0, and I don't think that's been a problem. If we want to get stricter at some point about the published versions being immutable artifacts, we can, but I'm not sure what the value of that would be now.

Regardless, @beauby please update this pr so that it changes the normative-statements.json files as well. (See e.g. #1102.)

beauby:
@ethanresnick Done.

dgeb:
@beauby please update _format/1.1/index.md as well

beauby:
@dgeb Done as well.

dgeb:
@beauby Thanks again. LGTM.

@ethanresnick Please review.

ethanresnick:
LGTM too. Merged! Thanks @beauby

PR #1038, opened 2016-04-30
Removed GOintegro / HATEOAS

LavaToaster:
It hasn't been updated in over a year, and is not compliant with version 1.0 of the spec.


steveklabnik:
Thanks!

Issue #990, opened 2016-02-23
Redcarpet -> Kramdown


steveklabnik:
We need to move to Kramdown by May 5. We're currently using redcarpet's tables extension, hopefully kramdown supports it.

ethanresnick:
I'm working on this in the gh-pages-compliance branch.

Tables haven't been a problem (kramdown actually has support built-in) and, with a bit of configuration, I've was able to make most of the other behavior match redcarpets.

The big exception to the above is syntax highlighting, where we were relying on redcarpets to pass the language of each code block through to HTML as a class attribute, which kramdown doesn't seem to do.

steveklabnik:
:+1:

ethanresnick:
It may be a rouge issue too, rather than a kramdown one. Ref https://github.com/jneen/rouge/issues/361 https://github.com/jneen/rouge/issues/379

steveklabnik:
Rouge is very much not-maintained (for good reason, I might add (at leas this is the impression I got from the author on twitter, it looks like it does get a few commits every few months)), so hopefully if it is we can work around it.

jamesplease:
Is this issue closeable since #1039 was merged? :)

ethanresnick:
Yup, thanks!

PR #1039, opened 2016-05-01
Finish switch to kramdown

ethanresnick:
This is a pr for #990.

The diff's pretty big, but almost all the new files are just me storing the highlight.js source in our repo, in case we need to rebuild from it. We're intentionally not using the latest highlight.js version, and the language auto detect will only work right if our build contains only a small number of languages, and I modified one of the language files, so I wanted to preserve all that in our repo.

For the background behind the highlight.js stuff, and the general fix being used here to our syntax highlighting woes, see [this commit](https://github.com/json-api/json-api/commit/de80a0408f32d7e691678be95ef75f75b62d5b75)'s message.


steveklabnik:
Nice!

ethanresnick:
Ok, hearing no objections, I'm going to merge this.

Note: a different approach here would have been to remove highlight.js entirely and set our highlighting styles solely based on the rouge-outputted server side syntax classes. That would obviously be simpler. There are a few cases, though, where the highlighting that approach produces isn't quite as good. The biggest is in the Links section on the specification page, where we have some invalid json examples that trips up rouge, but that we're able to hack highlight.js into handling correctly. The other case is that Rouge doesn't apply a distinct class to the http status in the response line, so we can't call it out from the "HTTP/1.1" part. (Though we might be able to hack around these limitations with some clever CSS sibling selectors... I didn't try.) If we decide in the future that those are sacrifices worth making, we can ditch highlight.js then pretty easily. For now, though, I just wanted to reproduce the highlighting exactly as it was.

PR #1043, opened 2016-05-06
[README] – Adding Devour to the list of JavaScript client implementations

Emerson:
Devour is a new JSON API client implementation written in JavaScript that we are actively developing here at [The Working Group](http://twg.ca). We are using it in a couple of projects already and it's been working fairly well and we look forward to building it out even more.


dgeb:
Nice - thanks!

PR #1044, opened 2016-05-06
docs: add Lux to implementations

papa-zakari:
[Lux](https://github.com/postlight/lux) is a MVC style Node.js framework that supports the spec all the way down to the status codes and error handlers. [Lux](https://github.com/postlight/lux) is able to offer unique features because of the spec being integrated from the ground up such as, built-in filter, sort, pagination, and auto-optimized SQL queries based on `fields` and `include` params (with sanitization).
## 

Great work on the spec!


steveklabnik:
Thanks so much!

PR #1066, opened 2016-07-22
Update index.md - change from 'post' -> 'article'

wontheone1:
In the example response, the resource name is 'article' but below, the word 'post' is used instead. Corrected to 'article'.

### DIFF
--- a/index.md
+++ b/index.md
@@ -96,9 +96,9 @@ Here's an example response from a blog that implements JSON API:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


wontheone1:
It's a simple fix, almost like fixing typo.

steveklabnik:
@wontheone1 it's true, but we all have busy lives. Sometimes it takes some time to get to it. Thank you for the PR.

PR #1074, opened 2016-08-11
Updated links to katharsis

masterspambot:
Repository was integrated into one multi-module repository so that existing links should not redirect to old repos.


dgeb:
@masterspambot sorry for the delay!

PR #1077, opened 2016-08-19
Fix Fortune.js link on implementations page

cdl:
Noticed the link to [fortune.js](http://fortune.js.org/) was out of date on the [Implementations](http://jsonapi.org/implementations/#server-libraries-node-js) page - this PR updates that link.

(Let me know if the commit message format is alright! Didn't see anything about it in CONTRIBUTING.md, so I just went with something standard.)


dgeb:
Thanks!

PR #1087, opened 2016-09-12
Fix level for `error-object-key` in normative statements JSON

ethanresnick:


ethanresnick:
Very simple, typo change, so I'm just going to merge this.

Issue #1092, opened 2016-09-16
Valid blank relationship?


richmolj:
Per the docs:

> The value of the relationships key MUST be an object (a “relationships object”)
> A “relationship object” MUST contain at least one of the following:

And then specifies `data`, `links`, and `meta`.

However, both of the following seem incorrect according to that documentation, but pass schema validation:

`relationships: { author: {} }`
`relationships: { author: null }`

My question is: are either of these responses actually considered valid? Should the documentation or schema be improved to reflect this?

dgeb:
@richmolj Good catch! Neither value is valid and the schema should be updated to reflect this.

richmolj:
@dgeb thanks for the clarification!

As a quick follow-up, we are discussing this over at AMS - let's say I'm requesting `/posts`. And we _could_ sideload comments, but for this request we're not. And no `link` is available for comments.

Ideally I'd like to tell the consumer "the 'comments' relationship exists, but this response doesn't contain any information about it". I am thinking the best way to do this would be to respond with:

`relationships: { comments: { meta: {} } }`

Sound about right?

ethanresnick:
If no link is available, then `relationships: { comments: { meta: {} } }` seems reasonable. If a link is available, I think `relationships: { comments: { links: {/*self or related*/} } }` would be preferable, of course.

NullVoxPopuli:
@ethanresnick what should be the default in this scenario? should 

`relationships: { comments: value }`
where value could be `{ data: {} }`, `{ meta: {} }`, or `{ links: {} }` ?

which one means what?

richmolj:
PR to fix the schema: https://github.com/json-api/json-api/pull/1094

ethanresnick:
@NullVoxPopuli 

>  what should be the default in this scenario? should
> 
> relationships: { comments: value }
> where value could be { data: {} }, { meta: {} }, or { links: {} } ?

`data` and `links` are generally more useful than `meta`, because they tell the client either what resources are in the relationship, or where to go to look that up.  So I'd prefer those keys. 

That said, `{data: {}}` is not a valid value; if you're trying to express an empty relationship `data: null` is the way to do that. (The `data` key in a relationship object holds [resource linkage](http://jsonapi.org/format/#document-resource-object-linkage).)

If you don't want to provide info in `data` (e.g. because it's too much work to fetch it, and clients are unlikely to need it), then provide links, so clients can fetch it manually if needed. 

[...]

PR #1094, opened 2016-09-19
Update schema to require relationship keys.

richmolj:
According to the documentation, both

```
relationships: { author: {} }
relationships: { author: null }
```

Are invalid. However, they validate against the current schema. This
ensures a relationship object is only valid when one of meta/links/data
are present.

See: https://github.com/json-api/json-api/issues/1092

### DIFF
--- a/schema
+++ b/schema
@@ -225,6 +225,11 @@
               "$ref": "#/definitions/meta"
             }
           },
+          "anyOf": [
+            {"required": ["data"]},
+            {"required": ["meta"]},
+            {"required": ["links"]}
+          ],
           "additionalProperties": false
         }
       },


NullVoxPopuli:
if a relationship's meta/links/data object is nil, does each of those have a particular meaning? and is that documented somewhere?

richmolj:
@NullVoxPopuli 
- [links cannot be nil](http://jsonapi.org/format/#document-links)
- [meta cannot be nil](http://jsonapi.org/format/#document-meta) (but can be empty)
- `data` I think is self-explanatory, but just to verify [primary data can be nil](http://jsonapi.org/format/#document-top-level)

NullVoxPopuli:
links can be empty according to the schema. 

valid with the http://www.jsonschemavalidator.net/

```
{
  "data": {
    "id": "1",
    "type": "my-type",
    "attributes": {
      "name": "my name"
    },
    "relationships": {
      "posts": {
         "links": {} 
      }
    }
  }
}
```

richmolj:
Then I would suggest a separate issue/PR :)

NullVoxPopuli:
I think links being empty is totally valid
"You cannot access this relationship independently of this resource"

richmolj:
@NullVoxPopuli not saying your thought is invalid, just that this is the wrong place to discuss it.

NullVoxPopuli:
@richmolj I was just clarifying what you wrote:

links are !nil || empty
meta is !nil || empty
data is nil || !empty

richmolj:
You are right, I should have suggested a separate issue in the first place instead of responding :)

To summarize: we know of one `meta`, `links`, or `data` is required for `relationships`. This PR is to make the schema update to reflect that.

It looks like `links` may have a similar issue, where the documentation does not match the schema. So we should clarify with the maintainers and submit PR if needed. But that should not prevent this PR from getting merged, right?

schema
```diff
@@ -225,6 +225,11 @@
               "$ref": "#/definitions/meta"
             }
           },
+          "anyOf": [
+            {"required": ["data"]},
+            {"required": ["meta"]},
+            {"required": ["links"]},
```
ethanresnick: Please remove this trailing comma; it's [not valid JSON](http://www.2ality.com/2013/07/trailing-commas.html).

ethanresnick (changes requested):
Minus the small nit in the comment, this LGTM!

Thanks @richmolj.

cc @dgeb

schema
richmolj: Whoops, fixed

dgeb:
Thanks @richmolj!

PR #1110, opened 2016-10-15
add typescript section; add ts-jsonapi

mohuk:
ts-jsonapi is a port of jsonapi-serializer written purely in Typescript with a minor additions


steveklabnik:
Nice! Could you switch the order of the two though please? We keep the list in chronological order.

mohuk:
done

abdulhaq-e:
Could you also please move `ngrx-json-api`.

remmeier:
you may also add katharsis as java client library, most recent releases support both client and server use

mohuk:
@remmeier there should be another PR/Issue for the Java libs I believe :)

dgeb:
@steveklabnik this LGTM now. What do you think?

dgeb:
Sorry to take so long to merge this. And thanks again for the contribution!

PR #1115, opened 2016-11-12
Fix duplicate range in character class (in schema)

omghax:
This removes redundancy in a couple character classes within regular expressions in the JSON schema. Specifically, there were 2 places where a character class like `[\w_]` was being used, which is redundant because `\w` already includes `_`.

I happened to notice this because I was using the `json-schema-rspec` library in a Ruby project (with warnings turned on), and saw the following printed to the console:

```
  warning: character class has duplicated range: /^(?!relationships$|links$)\w[-\w_]*$/
```

### DIFF
--- a/schema
+++ b/schema
@@ -194,7 +194,7 @@
       "description": "Members of the attributes object (\"attributes\") represent information about the resource object in which it's defined.",
       "type": "object",
       "patternProperties": {
-        "^(?!relationships$|links$)\\w[-\\w_]*$": {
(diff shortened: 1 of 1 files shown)


Relequestual:
I don't see this regular expression in the current JSON Schema. I suggest this PR be closed.

jelhan:
I'm sorry that this hasn't received earlier feedback. As @Relequestual already mentioned, it has been fixed in the meantime. Closing this one as obsolete.

Please speak up if I missed something.

PR #1121, opened 2016-11-22
Switch fork of Munson to an active fork

coryodaniel:
Gem releases have always come from my fork. I no longer have access to the StackSocial fork...


dgeb:
@coryodaniel thanks for the correction. I can see that you're the definitive owner of the gem 👍

PR #1126, opened 2016-12-11
Added "allow" attribute for links to restrict access

aimeos:
This PR relates to the discussion in https://github.com/json-api/json-api/issues/745

If access to a resource is limited by a permission system, the HTTP header that can be sent along with the resource is "Allow: GET, PATCH, DELETE". For included resources, the link attribute should be able to list the allowed methods the same way:
```
"links": {
  "related": {
    "href": "http://example.com/articles/1/comments",
    "allow": ["GET", "PATCH", "DELETE"]
  }
}
```
to be as close as possible to the HTTP header. Only "Allow" should lower-case to be in line with all other keys in the JSON payload.

### DIFF
--- a/_format/1.1/normative-statements.json
+++ b/_format/1.1/normative-statements.json
@@ -857,7 +857,7 @@ layout: null
       "type": "normative-statements",
       "attributes": {
         "level": "MUST",
-        "description": "Each member of a links object is a \"link\". A link **MUST** be represented as either:\\n\\n- a string containing the link's URL.\\n- an object (\"link object\") which can contain the following members:\\n  - `href`: a string containing the link's URL.\\n  - `meta`: a meta object containing non-standard meta-information about the link."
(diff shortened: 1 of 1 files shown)


jelhan:
Thanks a lot for your proposal.

Looking at v1.1 this could be addressed using an extension as well. Doing so would be better in my perspective:

1. Extensions allow us to experiment outside of the main spec. An extension dies not have the same backward compatibility guarantee as the main spec. This allows it to move much faster.
2. Discovering available methods for an API endpoint programmatically is a requirement only some APIs have. I think it would be better to not force the complexity of adding a feature to all others as well.

bednic:
HTTP methods discovery should be done by HEAD or OPTIONS methods

aimeos:
@bednic That's not practical for all the links in the sent JSON response because they can have different allowed methods, e.g. GET for all resources, POST/DELETE only for adding/removing products into/from the basket. Doing OPTIONS requests for each link would generate a lot of request which should be avoided.

Issue #867, opened 2015-08-27
Creating new resources and validating with JSON Schema


mvdstam:
The official JSON API Schema (as described at http://jsonapi.org/schema) explicitly expects the ID field to be sent by clients when creating new resources. This conflicts with the following example (pulled from the official documentation):

``` json
POST /photos HTTP/1.1
Content-Type: application/vnd.api+json
Accept: application/vnd.api+json

{
  "data": {
    "type": "photos",
    "attributes": {
      "title": "Ember Hamster",
      "src": "http://example.com/images/productivity.png"
    },
    "relationships": {
      "photographer": {
        "data": { "type": "people", "id": "9" }
      }
    }
  }
}
```

When I validate the above to the official JSON Schema document, validation fails. Perhaps a seperate JSON Schema document can be provided solely for creating new resources, where the ID field is optional?

ethanresnick:
See #851, which is another reason not to use the published schema to validate incoming requests. Basically, that schema is only for validating output/response documents, and we're already thinking about whether to create a separate one for incoming requests (or just to keep a single one, but make it more lenient).

sebilasse:
Citing the official documentation

> Exception: The id member is not required when the resource object originates at the client and represents a new resource to be created on the server.

ethanresnick:
@dgeb @tkellen @steveklabnik Do you guys have an opinion here on whether it's worth it to maintain multiple, stricter schemas or just go with one looser one? I was leaning toward the looser schema approach and started mocking it up, as you can see from the links. But, on reflection, ditching the `id` requirement for all resource objects seems like a big loss, so now I'm not sure.

mvdstam:
If I may offer my 2 cents (as someone who has used JSON-API in a production environment and had the chance to look at it quite thoroughly). One of the most nice aspects about JSON-API is that once a request goes through a JSON-Schema, you can trust it's okay because it's so strict, and you _know_ that certain data is at your disposal.
I'd go for multiple stricter schemas instead of going with 1 looser schema, because going down that road means potentially making the schema even more looser when more use-cases show up thus making the standard less, well, standard. The `id` field thing is a major dealbreaker IMHO and should be required for all requests aside from updating existing resources.

sebilasse:
@ethanresnick : same like @mvdstam 
[...]

PR #1127, opened 2016-12-12
Separate out JSON Schema into schema, schema-update, schema-create

bf4:
Work by @eneuhauser from discussion in #867 (esp see https://github.com/json-api/json-api/issues/867#issuecomment-142693593 and https://github.com/json-api/json-api/issues/867#issuecomment-140128809 )just to maybe spur some work.  (I'm still not confident with composing schema haven't built many myself.) cc @sebilasse


n2ygk:
@bf4 FYI I've recently done something similar in which I used the jsonapi schema to create a [RAML 1.0 type library](https://github.com/n2ygk/raml-snippets) (JSON->YAML->RAML). I made a `resource_post` which allows the 'id' to be optional and then `resource` extends `resource_post` to make 'id' required. 

I also found I had to do a similar thing for my "user" data types that extend `resource`, making mytype_patch which allows all attributes to be optional (for PATCHing) which is then extended by mytype_post (for POSTing) which identifies the required attributes and then mytype (for GETting) which extends the post to require the 'id'. Still not quite correct and not completely DRY.

Oh, and from another thread, the `info` response type goes with the [DELETE 200 response](http://jsonapi.org/format/#crud-deleting):

> 200 OK
>
> A server MUST return a 200 OK status code if a deletion request is successful and the server responds with only top-level meta data.

I wish I had seen this work before reinventing it!  Oh well.

bf4:
Incidentally, I think the schema naming of success/failure/info is a little misleading. 

```json
  "oneOf": [
    {
      "$ref": "#/definitions/success"
    },
    {
      "$ref": "#/definitions/failure"
    },
    {
      "$ref": "#/definitions/info"
    }
  ],
```

- a `success` schema has `data` and optionally `meta`, `links`, `included`, `jsonapi`
- a `failure` schema has `errors` and optionally `meta`, `jsonapi`,(unclear if  `links` is allowed)
- an `info` schema has `meta` and optionally `links`, `jsonapi`

Why not just call them `data`, `errors`, `meta`?

webjunkie:
I would need a schema for "create". Too bad there hasn't happened anything here in a while.

Relequestual:
@dgeb I started to break them up, but my current requirements and timelines don't allow me to complete that right now.
Is this something you're interested in?
I also note, the schema only validates responses, and separate schemas would be required to validate requests.

jelhan:
I'm going to close this in favor of #1600. I hope that we can merge #1600 soon.

Thanks a lot @bf4 for starting the work!

PR #1129, opened 2016-12-13
New flask library to create rest api

akira-dev:
I have created a Flask extension that let you build Rest api according to JSONAPI specification. It is based on flask and marshmallow-jsonapi and SQLAlchemy or MongoDB. This library is very esay to use and offer a large amount of customization for real life projects.


steveklabnik:
Thank you!

PR #1135, opened 2016-12-22
Add json-api-document-viewer to implementations

tadast:
https://github.com/tadast/json-api-document-viewer is a React.js based tool that aims to visualise relationships in large json:api documents with a lot of "included" objects. It visually nests the relationships under their parents but does it one level at a time, thus avoiding infinite loops ... unless the user is infinitely patient in expanding them :)


steveklabnik:
Neat, thanks!

PR #1139, opened 2017-01-18
Fix display of Reverse Solidus and Vertical Line characters

chris-hamper:
The "Reverse Solidus" and "Vertical Line" characters as shown in the "Member Names" subsection do not display correctly on some browsers (latest Chrome on Mac OS X). Replace them with their HTML entities to ensure they appear correctly.


dgeb:
Thanks!

PR #1152, opened 2017-04-03
Add python JSON API client to implementations/index.md

tuomas2:
Add new python JSON API client implementation by Qvantel.


dgeb:
Thanks!

PR #1153, opened 2017-04-04
Add cloudcreativity libs to implementations page

dtr0yan:
This libs were used in production projects, have good quality of code and can be used by community


dgeb:
Thanks!

PR #1154, opened 2017-04-04
Remove JSONAPI.NET from implementations

csantero:
I am no longer actively maintaining this JSONAPI.NET, and it does not support .NET Core. New projects should use one of the other .NET server implementations.


dgeb:
Thanks for following up!

PR #1156, opened 2017-04-18
Remove uniqueItems restriction on relationships.

FlorentTorregrosa:
Discussion opened: http://discuss.jsonapi.org/t/impossible-to-have-a-relationship-on-another-object-twice/951
This patch is here to highlight the proposition.

### DIFF
--- a/schema
+++ b/schema
@@ -254,8 +254,7 @@
       "type": "array",
       "items": {
         "$ref": "#/definitions/linkage"
-      },
-      "uniqueItems": true
+      }
     },
     "empty": {
       "description": "Describes an empty to-one relationship.",


beauby:
I fail to see a case where this would be useful. Could you expand on why it is necessary for Drupal? (Or is it just a limitation of the current implementation?)

FlorentTorregrosa:
Hello,

In Drupal, you don't have limitation to reference multiple times the same tag or content or entity if you want to. I agree that the case must be very very rare to have to reference the same entity twice.

I found randomly the case when using a dummy content generation module that was generating content referencing multiple times another entity and then found that the JSON API endpoints on my Drupal site were no more working.

rpkilby:
> dummy content generation module that was generating content referencing multiple times another entity

Why can't just you remove the duplicate references from the generated content?

FlorentTorregrosa:
Hello,

@rpkilby: About the dummy content generation module, yes that's a solution, but the goal of a such module is to avoid having to create or edit content for the tests.

But that's not the goal of this dicussion. As said in my previous comment, you are not limited in Drupal to reference the same entity twice. So if not dummy content, it can for functional needs be the case. I agree that the case will be rare to reference the same entity twice (I never meet the case for the moment), but it can exist.

beauby:
@FlorentTorregrosa Note that you can currently convey that semantic within the spec by using a `meta` member on the resource identifiers within the linkage of a relationship. 

Example:

```json
{
  "data": {
    "type": "foo",
    "id": "bar",
    "relationships": {
      "baz": {
        "data": [
          { "type": "foobar", "id": "123", "meta": { "arity": 5 } },
          { "type": "foobar", "id": "456", "meta": { "arity": 2 } },
          { "type": "foobar", "id": "789", "meta": { "arity": 1 } }
        ]
      }
    }
  }, 
  "included": {...}
}
```

wimleers:
@beauby That's not linking to the same resource multiple times though, is it?

I can't believe the original Drupal issue where this was discovered hasn't been mentioned yet. You can find every discussed detail there: https://www.drupal.org/project/jsonapi/issues/2864680

FlorentTorregrosa:
Hello,

@wimleers Thanks for updating this issue.

The original Drupal issue has been mentioned but in the jsonapi forum (http://discuss.jsonapi.org/t/impossible-to-have-a-relationship-on-another-object-twice/951) with a link in my first comment on this conversation.

wimleers:
We've implemented @beauby's suggestion 4 months ago in https://www.drupal.org/project/jsonapi/issues/2864680. This can be closed.

wimleers:
@FlorentTorregrosa Mind closing this PR? :)

FlorentTorregrosa:
@wimleers Thanks for having pushed this forward :).

PR #1160, opened 2017-05-02
Allow 'links.self' as objects in the schema

### LINKED ISSUE #1159: `self` link inconsistency
The spec says that [a link can be either a string or an object](http://jsonapi.org/format/#document-links).

However the schema enforces that [`self` links must be strings](https://github.com/json-api/json-api/blob/509ee949c621cee5683c959d4268386e1a592b83/schema#L158-L162).

(It is also a bit odd that the properties and descriptions in `definitions/links` seems to be [relationship-specific](https://github.com/json-api/json-api/blob/509ee949c621cee5683c959d4268386e1a592b83/schema#L154-L168), but this definition is referenced from [resource](https://github.com/json-api/json-api/blob/509ee949c621cee5683c959d4268386e1a592b83/schema#L144-L146) and [top level](https://github.com/json-api/json-api/blob/509ee949c621cee5683c959d4268386e1a592b83/schema#L38-L43) link objects.)

I was wondering if the `self` links are an exception and must be strings intentionally, or is it a bug in the schema? If it's the latter I'd be happy to open a PR to fix the schema.

Sample use case: on a project I'm working on we prefer links as objects v.s. links. We have a convention to pass extra meta data for each link e.g. request header, http verb (GET/PUT/POST) and so on. However the schema validations fail if a resource `self`link is an object, even though there doesn't seem to be anything in the documentation that makes `self` links exceptional.

tadast:
See #1159 for context.

This change allows the self link to be an object as well as a string [as per spec](http://jsonapi.org/format/#document-links).

> A link MUST be represented as either:
> - a string containing the link’s URL.
> - an object (“link object”)

It also makes sure that each property in the `links` object is a `definitions/link`, as per spec.
> The value of each links member MUST be an object (a “links object”).

It also separates a `relationshipLinks` definition that seems to carry
some relationship-specific descriptions. Although it could easily use the modified `links` definition as well with the same effect. If we don't care about these descriptions I can modify this PR for the relationship.links to use the new `links` definition.

Fixes #1159


steveklabnik:
Thank you!

@dgeb what do you think?

jamesplease (approved):
Looks good to me. Good catch @tadast

dgeb:
@tadast thanks much! Apologies for taking so long to review.

PR #1161, opened 2017-05-03
Add JsonApiSerializer to .NET implementations.

alex-davies:
Adding another .NET Implmentation.


dgeb:
Thanks - and sorry for the delay!

PR #1168, opened 2017-05-18
1.1: JSON API Object: using 1.1 in the example

adam-lynch:
Because why not I guess :)


beauby:
This makes sense to me, and shouldn't be too long to review by one of the maintainers (:

dgeb:
Haha ... seems kind of important. Thanks!

Issue #898, opened 2015-10-07
RFC: URLs in links


bf4:
- https://github.com/json-api/json-api/issues/840#issuecomment-141631828
- https://github.com/json-api/json-api/pull/834#issuecomment-131192713
- https://github.com/json-api/json-api/pull/834#issuecomment-141631788

>  relative references are simply not allowed.

So, links should be `"links": { "self": "http://example.com/link_author/1"}` not `{"links": { "self": "//link_author/1"}`

How about a link such as `"links": { "self": "//link_author/1"}` discussed in [ActiveModel::Serializers](https://github.com/rails-api/active_model_serializers/pull/1246#discussion_r41348842)

bf4:
cc @beauby

masterspambot:
:-1:

steveklabnik:
This already has a meaning in URLs: it's a "network relative" URL, though it's often called "protocol relative." So this isn't going to work, directly.

patrykorwat:
When we take a look at other specs: 
- JSON-LD defines `@base` annotation to support relative IRI-s (http://www.w3.org/TR/json-ld/#base-iri)
- HAL, Collections+JSON uses URI term, which can be relative

JSON API uses URL (as @steveklabnik said), so just use the `full` version.

beauby:
Wait wait wait, I think there has been a huge misunderstanding here. I forgot to put the domain in `//example.com/link_author/1` in a test for a PR supporting resource-level links in AMS, which resulted in a debate which I _believed_ was about whether protocol-relative URIs were allowed in the spec (which to my knowledge are at least not explicitly forbidden).

masterspambot:
@beauby In most cases urls are used by other systems (not client apps) to crawl the resources, so that the need for using full urls.

patrykorwat:
Oh, ok. 
Personally, I don't use this feature and dislike, especially when JSON API is not used to provide static content ;).
I don't imagine a situation when a resource, that is accessible using HTTPS, should be accessible using HTTP.

bf4:
[...]

PR #1169, opened 2017-05-19
updates schema to allow URI references

jelhan:
As discussed in #898 JSON API specification does not make any assertions
about URI structure. Also specification contains examples using "relative
URIs".

RFC 3986, which is referenced by JSON Schema specification, distinguish
between an uri and an uri reference like "relative" or "partial URIs":

> 1.2.3. Hierarchical Identifiers
> [...]
> NOTE: Previous specifications used the terms "partial URI" and "relative
> URI" to denote a relative reference to a URI. As some readers misunderstood
> those terms to mean that relative URIs are a subset of URIs rather than
> a method of referencing URIs, this specification simply refers to them as
> relative references.

Therefore relative URIs would fail validation if `uri` format (8.3.6 in JSON
Schema draft) is used. `uri-reference` format (8.3.7) allow both.

### DIFF
--- a/schema
+++ b/schema
@@ -158,7 +158,7 @@
         "self": {
           "description": "A `self` member, whose value is a URL for the relationship itself (a \"relationship URL\"). This URL allows the client to directly manipulate the relationship. For example, it would allow a client to remove an `author` from an `article` without deleting the people resource itself.",
           "type": "string",
-          "format": "uri"
+          "format": "uri-reference"
         },
         "related": {
           "$ref": "#/definitions/link"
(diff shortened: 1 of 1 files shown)


bf4:
FWIW, linking to comment that the JSON SCHEMA is not authoritative https://github.com/json-api/json-api/pull/440#issuecomment-151013219

dgeb:
Thank you @jelhan!

@bf4 Yep, the spec itself (not the schema) should be considered authoritative.

bf4:
@dgeb perhaps we should add a comment somewhere to that effect?

PR #1171, opened 2017-05-24
Fix: a resource can't have an attribute or relationship named type or id

meyerbaptiste:
According with the latest specification (v1.0):

> Fields for a resource object MUST share a common namespace with each other and with type and id. In other words, **a resource** can not have an attribute and relationship with the same name, **nor can it have an attribute or relationship named type or id**.


dgeb:
Good catch - thanks!

PR #1182, opened 2017-06-13
Add typescript de(serializer) jsona to implementations

olosegres:
Hi, this adds link to (de)serializer lib, typified with TypeScript, merge please :)


dgeb:
Hmmm ... this lib is listed above under JS as well. I wonder if we should merge the JS + TS sections, and recommend that TS libs include a note about being written in TS.

PR #1183, opened 2017-06-13
Added a JSON API Validator to the list of Implementations/Related Tools

aren55555:
http://corroborate.arenpatel.com/ is a JSON API validator that I created; the source for it is available here: https://github.com/aren55555/corroborate


dgeb:
Thanks!

PR #1188, opened 2017-06-27
add PHP server implementation, oligus/jad

oligus:
Jad is a library that turns doctrine entities into json api resource, or collection of resources, automagically.


steveklabnik:
Thanks!

oligus:
Thank you too!

PR #1195, opened 2017-07-23
Profile Extensions, Take 3: Extension Rules

ethanresnick:
Hey all!

Here is an updated PR for profile extensions. I’d really like to get this merged soon, or else get a concrete list of blockers if it’s not mergeable as is. We’ve been sitting on this for two years, and the design has gone through quite a lot of vetting and revision, so I think it’s time. Besides, 1.1 is still a draft, so we can still change things if this design proves unworkable.

## History
For anyone reviewing this PR who hasn’t followed (or has forgotten) the prior discussions around JSON:API extensions, here are some links describing, broadly, why this proposal works the way it does:

1. Issue #614 (starting [here](https://github.com/json-api/json-api/issues/614#issuecomment-101459115)) lays out the broad design goals for JSON:API extensions and groups possible extensions into two different types. Specifically, it separates out “profile extensions”—which are limited in what they can do, in order to simplify processing, maximize interoperability, and compose seamlessly—from more powerful “comprehensive extensions”.

2. Issue #650 contains my first stab at specifying the extension system. This design tried to cover both profile extensions and comprehensive extensions, as well as an extension advertising/discovery mechanism.

3. After #650 proved to be too big to consider all at once, I decided to focus on profile extensions first, and I put forward a refined design for them in #915. The first post in #915 offers some background on why I chose to focus on profile extensions and also why I think extensions are so important generally. That post also offers some additional design goals unique to profile extensions.

4. The discussion in #915 was wide-ranging. It covered what restrictions should go on profile extensions, whether and how they should be registered, whether and how they should be versioned, and how extension-added data should be distinguished from data that might be defined by future versions of the base specification. We explored many different possibilities and considered models developed by, and lessons learned from, other standards efforts. I was particularly influenced by a W3C TAG–recommended model of compatibility (defined in detail [here](https://github.com/json-api/json-api/issues/915#issuecomment-165319520)), which undergirds the approach to extension versioning used in this PR (namely, extensions are unversioned and must change in fully-compatible ways over time).

5. Finally, #957 attempted to put the conclusions of #915 into specification-ready text. This text was reviewed at an editors meeting and tweaked—most notably, to require that extension-defined data live under the `meta` object. This PR is based on #957.

## New Changes
For easier review, this PR rebases the latest version of #957, and some changes from #957 have been split off into a separate PR. In particular, this PR defines profile extensions, while #1196 outlines the (optional but encouraged) registration process. Beyond splitting up #957, this latest revision contains some small changes:

1. I removed the requirement that an array of `profile` links be added in the `links` object to indicate the extensions in use. The extensions in use are still conveyed in the `aliases` object in the document body, and in the Content-Type header. 

2. In the error object, I introduced `type` as a top-level key, rather than as a link. (For more on the error object changes, see the commit messages; they provide some context on why these changes are needed and why they’re acceptable from a compatibility POV.)

3. I've decided to keep the `aliases` key at the top-level, rather than nest it under `jsonapi`, because it includes data that applies to the current document, not necessarily to the server as a whole (which is more what the `jsonapi` key is for iiuc).

## Next Steps

Again, I think this is either ready to merge or very close, especially since we'd only be merging it into a draft spec. Still, I'm open to any feedback, although I'd rather not get bogged down bikeshedding names. I know it's a lot to review, so I've tried to break it up into coherent commits with detailed commit messages.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1819,7 +1819,7 @@ to standard query parameters without conflicting with existing implementations.
 ## <a href="#extending" id="extending" class="headerlink"></a> Extending JSON API
 
 An API **MAY** support one or more "profile extensions" to supplement the
-capabalities of this specification.
+capabilities of this specification.
 
 A profile extension is a small, separate specification that defines a set of
 values that can be added to a JSON API document, and the meaning of those values
@@ -2066,7 +2066,7 @@ GET /articles/1 HTTP/1.1
     "templates": [{
       "name": "publish",
       "method": "PATCH",
-      "uri": "/articles/1"
+      "uri": "/articles/1",
       "fields": [{
         "name": "publish-date",
         "meta": {
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -5,7 +5,8 @@ version: 1.1
 ## <a href="#introduction" id="introduction" class="headerlink"></a> Introduction
 
 JSON API is a specification for how a client should request that resources be
-fetched or modified, and how a server should respond to those requests.
+fetched or modified, and how a server should respond to those requests. JSON API
+can also be easily [extended](#extending).
 
 JSON API is designed to minimize both the number of requests and the amount of
 data transmitted between clients and servers. This efficiency is achieved
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1987,12 +1987,14 @@ keyed by `errors` in the top level of a JSON API document.
 An error object **MAY** have the following members:
 
 * `id`: a unique identifier for this particular occurrence of the problem.
+* `type`: a URI that identifies the type of error that this particular error 
+  is an instance of. When dereferenced, this URI **SHOULD** provide a 
+  human-readable explanation of the general error.
 * `links`: a [links object][links] containing the following members:
   * `about`: a [link][links] that leads to further details about this
     particular occurrence of the problem.
 * `status`: the HTTP status code applicable to this problem, expressed as a
   string value.
-* `code`: an application-specific error code, expressed as a string value.
 * `title`: a short, human-readable summary of the problem that **SHOULD NOT**
   change from occurrence to occurrence of the problem, except for purposes of
   localization.
@@ -2005,10 +2007,33 @@ An error object **MAY** have the following members:
     for a primary data object, or `"/data/attributes/title"` for a specific
     attribute]. This **MUST** point to a value in the request document that
     exists; if it doesn't, the client **SHOULD** simply ignore the pointer.
-  * `parameter`: a string indicating which URI query parameter caused
-    the error.
-* `meta`: a [meta object][meta] containing non-standard meta-information about the
-  error.
+  * `parameter`: a string indicating which URI query parameter used in the
+    request caused the error.
+  * <a id="error-objects-source-missing"></a>`missing`: an object indicating
+    that the error was caused by a missing extension, key, or URI query parameter.
+    If present, this object **MUST** contain exactly one of the following keys:
+        * `extension`: the URI of a missing extension. This key **SHOULD** be
+          used with the `pointer` member to indicate which value in the request
+          document is missing the extension.
+        * `key`: the name of a missing key. This key **SHOULD** be used with the
+          `pointer` member to indicate which value in the request document is
+          missing the key.
+        * `parameter`: the name of a missing URI query parameter.
+
+    This object **MAY** also contain an `alternatives` key, whose value is an
+    array listing alternate extension URIs, key names, or parameter names
+    (depending on which type of value is missing) that can be added to resolve
+    this error.
+* `meta`: a [meta object][meta] containing additional information about the error.
+
+> Note: Previous versions of this specification defined a `code` member in
+  error objects. The role of that member is now played instead by the `type`.
+>
+> Old APIs that send the `code` member will continue to function with newer
+  clients, but these clients will likely ignore that member. Clients may choose
+  to continue to support `code` if they must interact with older APIs. However,
+  new APIs should not produce error objects with this member, even though doing
+  so is technically allowed.
 
 [resource objects]: #document-resource-objects
 [attributes]: #document-resource-object-attributes
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1830,6 +1830,71 @@ be a space-separated (U+0020 SPACE, " ") list of profile extension URIs. This
 list **MUST** be surrounded by quotation marks (U+0022 QUOTATION MARK, "\""), in
 accordance with the HTTP specification.
 
+### <a href="#extending-profile-extensions-applying" id="extending-profile-extensions-applying" class="headerlink"></a> Adding Extensions to a Document
+
+The client or the server **MAY** add profile extensions to a JSON API document,
+whether or not the other party [requests](#extending-profile-extensions-requesting)
+them. 
+
+When one or more profile extensions are used in a JSON API document, the document 
+**MUST** define at least one [alias][aliases] for each extension's URI.
+
+An alias for an extension's URI **MAY** be used as a key (an "extension-associated
+key") within a `meta` object. The value at any such key is interpreted according the 
+specification of the extension to which the key name maps.
+
+The same extension-associated key **MAY** appear more than once in a given
+document.
+
+An extension-associated key **MUST NOT** be added at any location where the
+extension's specification does not define how to interpret the key's value.
+
+> In practice, extension-associated keys are only allowed as children of a
+  `"meta"` object, and only when that `"meta"` object is not within the
+  `"jsonapi"` object. This is because extensions are also [limited](#extending-profile-extensions-authoring)
+  in where they can permit their data to appear.
+
+The following document demonstrates these rules by adding a hypothetical
+extension that has the URI `http://example.org/extensions/last-modified`:
+
+```json
+{
+  "aliases": {
+    "last-modified": "http://example.org/extensions/last-modified"
+  },
+  "data": {
+    "type": "people",
+    "id": "9",
+    "attributes": {
+      "first-name": "Dan",
+      "last-name": "Gebhardt"
+    },
+    "relationships": {
+      "father": {
+        "data": {
+          "type": "people", "id": "7"
+        },
+        "meta": {
+          "last-modified": "2013-09-24T00:00:00Z"
+        }
+      }
+    },
+    "meta": {
+      "last-modified": {
+        "date": "2015-01-01T00:00:00Z", "fields": ["first-name"]
+      }
+    }
+  }
+}
+```
+
+The document above defines `"last-modified"` as an alias for the exension's
+URI (though any legal [alias][aliases] name would do). Then, `"last-modified"`
+is used as an extension-associated key throughout the document. The 
+`"last-modified"` key holds values that have different formats and meanings 
+in each location that it appears, which is allowed assuming the extension's 
+specification describes those formats.
+
 ### <a href="#extending-profile-extensions-sending" id="extending-profile-extensions-sending" class="headerlink"></a> Sending Extended Documents
 
 When sending a JSON API document that uses profile extensions, whether from the
@@ -1949,6 +2014,59 @@ of explicitly allowed extensions that it maintains.
 
 Conversely, a profile extension **MAY** define a set of values that can be used
 in objects added to the document by other profile extensions.
+
+For example, imagine an extension that defines a template format that servers
+can insert into their responses to tell clients how to construct valid requests
+to trigger various "actions" associated with a resource. Imagine also that this 
+extension allows the server to use other JSON:API extensions to communicate 
+validation constraints for the request, rather than the extension definining its 
+own validation format. In that case, applying this hypothentical extension to a 
+document might look like:
+
+```http
+GET /articles/1 HTTP/1.1
+```
+
+```json
+{
+  "aliases": {
+    "templates": "http://example.org/extensions/request-template",
+    "common-validators": "http://example.org/extensions/validators"
+  },
+  "meta": {
+    "templates": [{
+      "name": "publish",
+      "method": "PATCH",
+      "uri": "/articles/1"
+      "fields": [{
+        "name": "publish-date",
+        "meta": {
+          "common-validators": { "type": "date" }
+        }
+      }]
+    }
+  },
+  "data": {
+    "type": "articles",
+    "id": "1",
+    "attributes": {
+      // ... this article's attributes
+    },
+    "relationships": {
+      // ... this article's relationships
+    }
+  }
+}
+```
+
+In the example above, the server has added the `http://example.org/extensions/request-template`
+extension to its response to communicate that the client can publish the 
+article by making a `PATCH` request to `/articles/1` that includes a 
+`"publish-date"` field. The request template extension has also speficied a 
+`meta` member in its field objects, allowing foreign extensions to add extra 
+information. The server has used this opportunity to communicate that the 
+`"publish-date"` field must contain a date, by applying the 
+`http://example.org/extensions/validators` extension.
 
 #### <a href="#extending-profile-extensions-authoring-updating" id="extending-profile-extensions-authoring-updating" class="headerlink"></a> Updating a Profile Extension
 
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1827,6 +1827,29 @@ use profile extensions in subsequent interactions with the same API.
 support JSON API at all or that the client has failed to provide a
 [required extension](#extending-profile-extensions-required).
 
+
+### <a href="#extending-profile-extensions-negotiation" id="extending-profile-extensions-negotiation" class="headerlink"></a> Negotiating Extensions
+
+#### <a href="#extending-profile-extensions-requesting" id="extending-profile-extensions-requesting" class="headerlink"></a> Requesting Extensions
+
+The client **MAY** request that certain profile extensions be applied to the
+server's response by listing those extensions in a `profile` parameter in one
+of its references to the JSON API media type in the `Accept` header.
+
+For example, in the following request the client, the client asks that the
+server apply the `http://jsonapi.org/extensions/last-modified` extension if
+it is able to.
+
+```http
+Accept: application/vnd.api+json;profile="http://jsonapi.org/extensions/last-modified", application/vnd.api+json
+```
+
+> Note: The second instance of the JSON API media type in the example above is
+  required under the [client's content negotiation responsibilities](#content-negotiation-clients).
+  It is used to support old servers that don't understand the profile parameter.
+
+The server **SHOULD** include as many of the requested extensions as it supports.
+
 ## <a href="#errors" id="errors" class="headerlink"></a> Errors
 
 ### <a href="#errors-processing" id="errors-processing" class="headerlink"></a> Processing Errors
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -96,6 +96,7 @@ A document **MAY** contain any of these top-level members:
 * `links`: a [links object][links] related to the primary data.
 * `included`: an array of [resource objects] that are related to the primary
   data and/or each other ("included resources").
+* `aliases`: an object defining string aliases for referring to (longer) URIs
 
 If a document does not contain a top-level `data` key, the `included` member
 **MUST NOT** be present either.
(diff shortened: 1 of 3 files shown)


_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
 > Note: This is to preserve the ability of JSON API to make additive additions
 to standard query parameters without conflicting with existing implementations.
 
+## <a href="#extending" id="extending" class="headerlink"></a> Extending JSON API
+
+An API **MAY** support one or more "profile extensions" to supplement the
+capabalities of this specification.
```
jamesplease: capabilities*

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+{
+  "aliases": {
+    "templates": "http://example.org/extensions/request-template",
+    "common-validators": "http://example.org/extensions/validators"
+  },
+  "meta": {
+    "templates": [{
+      "name": "publish",
+      "method": "PATCH",
+      "uri": "/articles/1"
```
jamesplease: missing comma here

jamesplease (approved):
I haven't read all of the issues that you've linked to from this issue, but your decision to focus on "profile extensions" seems good.

Others may be more familiar with all of the discussions around extensions, but from my outsider's perspective, this seems to be in a good place to merge and iterate on.

Thanks for doing this work, @ethanresnick !

ethanresnick:
Thanks for the review @jmeas, and the encouraging comments. Typos fixed!

dgeb:
Thanks for rebasing your work and reopening this discussion, Ethan.

I'd like to try to establish consensus on some fundamentals before engaging in a thorough detailed review.

You have defined profile extensions as follows:

> An API **MAY** support one or more "profile extensions" to supplement the capabilities of this specification.

> A profile extension is a small, separate specification that defines a set of values that can be added to a JSON API document, and the meaning of those values at the different places they can occur.

My preference would be to define profile extensions as separate specifications that define particular usage patterns that are wholly compatible with the base JSON API specification.

I believe that profile extensions can be much more useful if they are allowed to define usage expectations of any specific part of the spec including query params, attributes, relationships, meta, etc. The only requirement is that these extensions must fully conform with the base spec.

Here are a few of the use cases we have discussed in the past:

* particular pagination patterns through specifying usage of the `page` param

* particular filtering patterns through specifying usage of the `filter` param

* inclusion of set data in top-level `meta` (e.g. pagination statistics in a partial set)

* inclusion of timestamps as either `attributes` or `meta` info per resource

* expectations about the nature of IDs - i.e. client or server generated, UUIDs, etc.

I am concerned that this PR does not allow for some of these use cases. Do you agree that they should be in scope? Do we have a fundamental disagreement on the purpose of profile extensions?

ethanresnick:
Hey Dan!

Good to be chatting with you again. And I'm glad we're backing up to look at the fundamentals before we get lost in the weeds.

I want to focus on 3 of your use cases, which I agree aren't neatly addressed by the current design:

> - particular pagination patterns through specifying usage of the page param
>
> - particular filtering patterns through specifying usage of the filter param
>
> - expectations about the nature of IDs - i.e. client or server generated, UUIDs, etc.

I see the value in an API being able to communicate all of these things to clients. In my head, though, this has been a slightly different feature than profile extensions. In particular, most of these usage patterns apply on an API-wide level, so my assumption is/has been that they should ultimately be described in some sort of "home" document living at the root of the API. That document might reference some separate specifications, but these would be a different a type of extension than profile extensions.

A key distinction, I think, is whether the extension adds meaning to contents in the request/response document vs. whether it specifies things that are irrelevant to interpreting the current document, but might be useful in other ways (like in helping clients constructing future requests). To me, profile extensions are only about the former, whereas the extensions you're describing would also do the latter. 

I guess we could make one type of extension that does both, but there are some ways in which, to me, they really do feel like different concerns. For example, a client might want to request that the server apply a profile extension (in its current definition) to a document, in order to get extra info, but it really doesn't make sense to have a client requesting/negotiating based on extensions describing API usage patterns.

About your other use cases...

> inclusion of timestamps as either attributes or meta info per resource

I agree that extension being able to give meaning to an attribute would be nice. For example, a resource object like: 

```
{ 
  "type": "examples",
  "id": "1",
  "attributes": { 
    // let's imagine both of these attributes have extension-defined meanings.
    // maybe the color extension defines a standard set of colors.
    "last-modified": "2017-07-24T20:08:45.497Z",
    "color": "green"
  }
}
```

is obviously nicer JSON than

```
{ 
  "type": "examples",
  "id": "1",
  "attributes": {},
  "meta": {
    "last-modified": "2017-07-24T20:08:45.497Z",
    "color": "green"
  }
}
```

and being able to keep the data in the attributes object is a plus. So, I'm all for changing the current design to support assigning meaning to attributes. And I think doing that shouldn't be too hard (it seems to amount to, in one way or another, letting the aliases functionality reach into the attributes bag).

One thing I want to preserve, though, is the ability for someone reading a JSON:API document to know which extensions are adding meaning to which document members, just based on the contents of the document. In the current design, if I see `{ "meta": { "last-modified": "..." } }`, I know that the meaning of the value at the `"last-modified"` key is defined according to whatever extension is associated with the `"last-modified"` alias. I can then easily look up its meaning in that extension's spec. By contrast, if extensions aren't explicitly associated with the data they describe, I have to be familiar with all the extensions in use on the document to know which of them, if any, are defining a `"last-modified"` attribute. 

Also, if each attribute isn't explicitly associated with an extension, it's possible for two extensions to give the same attribute contradictory meanings—either when the extensions are first created or, importantly, as they're updated over time to add new features. Part of the design goals I set for profile extensions was to allow extensions to evolve independently and without versioning schemes, and this isn't possible if multiple extensions can be applying to the same piece of data.

>  define usage expectations of ... relationships

As far as extensions applying to `relationships`: I think this is almost certainly just asking for a way to map relationship names to what RFC5889 would call "extended relation names", which I totally agree we should have (and which we've discussed a bit going all the way back to #159), but I think that's also a separate feature from profile extensions.

dgeb:
@ethanresnick - Thanks as always for engaging so thoughtfully!

In order to establish the boundaries of the responsibilities of profile extensions, I would find it helpful to step back and try to carve up the responsibilities of the "neighbors" to profile extensions (i.e. other extensions, schemas, profiles, home documents, etc. that help to define and clarify an API). Otherwise, we risk introducing an unintentionally "gerrymandered" concept prematurely (sorry for the tortured analogy).

To be clear, I don't want to take away from your hard work here and all the discussions that have taken place. I believe much of this work is useful even if its boundaries shift a little bit after discussion.

I would find it useful to start a new discussion (very soon!) with a goal of providing a 10K foot view of the potential ways to define and standardize API structure and capabilities (i.e. profile extensions and all its neighbors). If we can agree on the high order bits, we can empower the community to explore some of these areas, and we can come back to review your PR here with confidence.

ethanresnick:
> In order to establish the boundaries of the responsibilities of profile extensions, I would find it helpful to step back and try to carve up the responsibilities of the "neighbors" to profile extensions (i.e. other extensions, schemas, profiles, home documents, etc. that help to define and clarify an API). Otherwise, we risk introducing an unintentionally "gerrymandered" concept prematurely (sorry for the tortured analogy).

Hey Dan! I also don't want us to ship something that's narrower than it needs to be and/or ultimately makes JSON:API's design unnecessarily fragmented. To avoid that, I'm open to having the high-level discussion you're describing, and will try to go into it not totally attached to the existing design. 

That said, I think there's also huge value in just shipping *something* soon, as long as it gets us, say, 80% of the way there. Right now, there are a number of use cases that JSON:API isn't addressing, that us editors haven't had the bandwidth to address, and we're starting to see non-compliant implementations and (I'd imagine) people just abandoning JSON:API as a result. While the current profile extension design is intentionally limited, I do believe it would immediately unblock the community to start addressing many of these issues in ways that we can learn from, without being bound to adopt their solutions in the base spec. I don't want the perfect to be the enemy of the good here. 

So, in the name of making sure we ship something, do you have ideas for how we can constrain this 10k foot conversation so it doesn't totally derail shipping something? I'd really like to figure out the parameters for such a conversation before we open what could be a giant can of worms.

dgeb:
@ethanresnick I agree that we've got to build some momentum. I'm willing to say that we should at the very least time box such a potentially sprawling open-ended conversation. I think we could hash some rough parameters out in a couple weeks. I'm willing to get the discussion started this week and try to build consensus by August 11.

I think it's critical that we proceed cautiously and deliberately, but that we do in fact proceed.

If there are other ways you'd like to constrain this conversation before we start, please let me know.

ethanresnick:
@dgeb Cool. Timeboxing this to August 11th sounds like a good solution to me. I think we should also define the scope and desired output a bit more, but I'll let you do that in your OP, since I want to make sure we're including all the things you think are important to include.

dgeb:
@ethanresnick Please see https://github.com/json-api/json-api/issues/1207 for my attempt at a high-level discussion about extending and describing the spec. Sorry that it took me longer than expected.

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+at the different places they can occur.
+
+A profile extension is identified by a URI [[RFC 3986](https://www.ietf.org/rfc/rfc3986.txt)].
+When dereferenced, this URI **SHOULD** return documentation on the extension.
+
+For requesting extensions and indicating those in use, the JSON API media type
+defines a `profile` parameter. If present, the value of this parameter, **MUST**
+be a space-separated (U+0020 SPACE, " ") list of profile extension URIs. This
+list **MUST** be surrounded by quotation marks (U+0022 QUOTATION MARK, "\""), in
+accordance with the HTTP specification.
```
gabesullice, line 1839: I think an example would be useful here.

_format/1.1/index.md
```diff
@@ -543,14 +550,46 @@ objects in the future. It is also possible that the allowed values of
 additional members will be expanded (e.g. a `collection` link may support an
 array of values, whereas a `self` link does not).
 
+### <a href="#document-aliases" id="document-aliases" class="headerlink"></a> URI Aliases
+
+The top level of a JSON API document **MAY** include an `aliases` member.
+If present, the value of this member **MUST** be an object (an "aliases object").
+
+Each key–value pair in the aliases object is an "alias definition". The key,
```
gabesullice, line 560: I think this is missing where in the document aliases should be actively "dereferenced". Do alias only apply to JSON keys or do they apply to values as well? Should the be dereferenced only within a certain member like `jsonapi` or `meta`, or could they also be dereferenced for keys under `data.attributes`? Perhaps it's globally?

I think that no matter the answer, it would be valuable to specify it :)

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+be a space-separated (U+0020 SPACE, " ") list of profile extension URIs. This
+list **MUST** be surrounded by quotation marks (U+0022 QUOTATION MARK, "\""), in
+accordance with the HTTP specification.
+
+### <a href="#extending-profile-extensions-processing" id="extending-profile-extensions-processing" class="headerlink"></a> Processing Extensions
+
+The recipient of a document **MUST** ignore any profile extensions in that
+document that it does not understand.
+
+Recipients **MAY** process in parallel the data from different profile extensions
```
gabesullice, line 1846: What is this permitting/restricting?

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+### <a href="#extending-profile-extensions-applying" id="extending-profile-extensions-applying" class="headerlink"></a> Adding Extensions to a Document
+
+The client or the server **MAY** add profile extensions to a JSON API document,
+whether or not the other party [requests](#extending-profile-extensions-requesting)
+them. 
+
+When one or more profile extensions are used in a JSON API document, the document 
+**MUST** define at least one [alias][aliases] for each extension's URI.
+
+An alias for an extension's URI **MAY** be used as a key (an "extension-associated
```
gabesullice, line 1858: Ah, this is where aliases apply. Maybe the "where" part of this could be moved to the aliases section while the "how" past of how aliases should be interpreted in the context of an extension could be left here.

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+When sending a JSON API document that uses profile extensions, whether from the
+server to the client or vice-versa, the `profile` parameter **MUST** be used in
+the `Content-Type` header to indicate which extensions are in use.
+
+For instance, the following header would be sent with a document that makes use
+of the `http://jsonapi.org/ext/example-1/` and `http://jsonapi.org/ext/example-2/`
+extensions:
+
+```http
+Content-Type: application/vnd.api+json; profile="http://jsonapi.org/ext/example-1/ http://jsonapi.org/ext/example-2/"
```
gabesullice, line 1925: Interesting! The URI is really powerful. Perhaps there should be an HTTP(S) caveat here?

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+`415 Unsupported Media Type` error.
+
+After attempting to rule out other possible causes of this error, a client that
+receives a `415 Unsupported Media Type` **SHOULD** remove the profile extensions
+it has applied to the document and retry its request without the `profile` media
+type parameter. If this resolves the error, the client **SHOULD NOT** attempt to
+use profile extensions in subsequent interactions with the same API.
+
+> The most likely other causes of a 415 error are that the server doesn't
+support JSON API at all or that the client has failed to provide a
```
gabesullice, line 1941: "The most common causes of a 415 error are that a server doesn't support JSON API or that the client"

I think this ^ might read more clearly

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+The client **MAY** request that certain profile extensions be applied to the
+server's response by listing those extensions in a `profile` parameter in one
+of its references to the JSON API media type in the `Accept` header.
+
+For example, in the following request the client, the client asks that the
+server apply the `http://jsonapi.org/extensions/last-modified` extension if
+it is able to.
+
+```http
+Accept: application/vnd.api+json;profile="http://jsonapi.org/extensions/last-modified", application/vnd.api+json
```
gabesullice, line 1958: Nit: missing a space after the semicolon `;` in order to be consistent with the example above.

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+```
+
+> Note: The second instance of the JSON API media type in the example above is
+  required under the [client's content negotiation responsibilities](#content-negotiation-clients).
+  It is used to support old servers that don't understand the profile parameter.
```
gabesullice, line 1963: d/old

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+
+The server **SHOULD** include as many of the requested extensions as it supports.
```
gabesullice, line 1965: "include every supported extension that was requested"

Perhaps that would read more clearly?

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+
+A server **SHOULD NOT** require that clients send documents containing certain
+extensions, as such requirements reduce interoperability.
+
+However, in the event that the server does require one or more extensions (e.g.
+for  authentication), and the client fails to use those extensions, the server
+**MUST** send a `415 Unsupported Media Type` error. In the error's response body,
+there **MUST** be an [error object][error objects] for each required extension
+that is missing, and this error object **MUST**:
+
```
gabesullice, line 1977:
```
**MUST** send a `415 Unsupported Media Type` error with a corresponding error object for each unsupported extension. Each error object **MUST**:
```

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+  - have a [`type`](#error-objects) who's URI is
```
gabesullice, line 1978:
```
have a `type` key with the value `http://jsonapi.org/errors/missing-profile-extension`.
```

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+    [`missing` object](#error-objects-source-missing).
+
+### <a href="#extending-profile-extensions-authoring" id="extending-profile-extensions-authoring" class="headerlink"></a> Authoring Profile Extensions
+
+As mentioned above, a profile extension defines a set of values that can be 
+added to a JSON API document, and the meaning of those values at the different 
+places they can occur.
+
+These values **MAY** be defined for use in any spec-defined [meta object][meta], 
+with the exception is that an extension **MUST NOT** define values for use in
```
gabesullice, line 1991: `s/is that/that`

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+**MUST NOT** define any values for use outside of a meta object.
+
+The meaning of an extension-defined value **MAY** vary based on where it occurs
+in the document but **MUST NOT** vary based on the presence or absence of other
+profile extensions.
+
+The keys in any profile-extension-defined objects **MUST** meet this specification's
+requirements for [member names].
+
+> If you author an extension, you are **strongly encouraged to [register](/extensions/#extension-creation-registration)
```
gabesullice, line 2002:
```
Extension authors **SHOULD** register their extension with the JSON API extension registry if the extension is intended for use in a public-facing implementation. This is intended to reduce duplication of efforts and reduce extension collisions.
```

_format/1.1/index.md
```diff
@@ -1777,6 +1816,301 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+{
+  "aliases": {
+    "templates": "http://example.org/extensions/request-template",
+    "common-validators": "http://example.org/extensions/validators"
+  },
+  "meta": {
+    "templates": [{
+      "name": "publish",
+      "method": "PATCH",
+      "uri": "/articles/1",
```
gabesullice, line 2073: Nit: this is not a URI, just a path

_format/1.1/index.md
```diff
@@ -1800,12 +2134,14 @@ keyed by `errors` in the top level of a JSON API document.
...
 * `id`: a unique identifier for this particular occurrence of the problem.
+* `type`: a URI that identifies the type of error that this particular error 
+  is an instance of. When dereferenced, this URI **SHOULD** provide a 
+  human-readable explanation of the general error.
 * `links`: a [links object][links] containing the following members:
   * `about`: a [link][links] that leads to further details about this
     particular occurrence of the problem.
 * `status`: the HTTP status code applicable to this problem, expressed as a
   string value.
-* `code`: an application-specific error code, expressed as a string value.
```
gabesullice, line 1812: I'm actually 100% with you that `code` is less useful than `type` ([we've had our struggles](https://www.drupal.org/project/jsonapi/issues/2934362)), but can this be removed during the 1.0 lifecycle?

Edit: after reading below, I see the answer to this. However, perhaps this should be kept here with an additional caveat like: "this member is no longer recommended for use, however some implementations may continue to send this key. Implementors **SHOULD** use the `type` key for new APIs.

gabesullice (commented):
Wow! @ethanresnick great job! I'm incredibly excited for something like this to take shape.

We've spent the last 3 years developing a JSON API module for Drupal which will soon be included in Drupal itself and will enable every Drupal installation to provide its content using JSON API out of the box.

There are a number of things that we've come up against that this proposal will immediately solve for us. 

Namely, communicating access/allowed operations and better error handling.

I'm really excited to see where this goes!

The _one_ thing which I think it left a little too unspecified is how profiles can interact with query parameters. I've seen your comment here: https://github.com/json-api/json-api/issues/1207#issuecomment-320179611

But I'm afraid the URL templating won't be sufficient to express very rich use cases. Specifically, [our filter expressions](https://www.drupal.org/docs/8/modules/json-api/filtering) could not be templated.

I wonder if we might be able to add a mechanism to support richer uses of the reserved query parameters in a more structured way.

gabesullice:
Just wanted to chime back in and say that I've since read your further comments in https://github.com/json-api/json-api/issues/1207#issuecomment-320864590 and am less convinced link templates could not handle our filters. Good explanations!

However, I'm still ever so slightly on the side of @dgeb that I would like profiles to have greater power to define the usage of reserved query parameters without adopting link templates.

PR #1196, opened 2017-07-23
Profile Extensions, Take 3: Registration

ethanresnick:
This PR is identical to #1195, which should be merged first, except that it has adds two commits with copy about how I'm imagining extension registration would work. I've separated this copy from #1195 so they can be discussed separately.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -93,20 +93,21 @@ The members `data` and `errors` **MUST NOT** coexist in the same document.
 
 A document **MAY** contain any of these top-level members:
 
-* `jsonapi`: an object describing the server's implementation
-* `links`: a [links object][links] related to the primary data.
 * `included`: an array of [resource objects] that are related to the primary
   data and/or each other ("included resources").
+* `links`: a [links object][links] related to the primary data.
 * `aliases`: an object defining string aliases for referring to (longer) URIs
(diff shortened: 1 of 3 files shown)


_format/1.1/index.md
```diff
@@ -543,14 +551,46 @@ objects in the future. It is also possible that the allowed values of
...
-API version supported. This object **MAY** also contain a `meta` member, whose
-value is a [meta] object that contains non-standard meta-information.
+under a top-level `jsonapi` member. If present, the value of the `jsonapi`
+member **MUST** be an object (a "jsonapi object").
+
+The jsonapi object **MAY** contain a `version` member whose value is a string
+indicating the highest JSON API version supported.
+
+This object **MAY** also contain a `meta` member, whose value is a [meta object][meta]
+that contains non-standard meta-information. The `meta` object in the jsonapi
```
approxit, line 592: The `non standard meta-information` phrase are replaced few times by `additional information`. Should `additional information` be used here too?

ethanresnick: No, this discrepancy is actually intentional, because extension data  (the only meta info that isn't "non-standard"), cannot appear in the jsonapi object in the current extension design

ethanresnick:
Superseded by #1311.

Issue #1089, opened 2016-09-13
Support Embedded Relationships


richmolj:
### Intent

A number of issues have tried to address this in the past (https://github.com/json-api/json-api/issues/202, https://github.com/json-api/json-api/issues/205, https://github.com/json-api/json-api/issues/753, https://github.com/json-api/json-api/issues/536). Most of those issues were closed in favor of https://github.com/json-api/json-api/issues/795, and most Discourse threads on this topic eventually point to #795.

However, #795's scope involves two main points:
- Creating/updating a single resource with its relations in one go
- Multiple operations in a single request (primarily to support transactions)

This issue's scope is **only the former**. The intent is to resolve a common issue without trying to solve every other related issue at the same time. In other words, let's support something like [accepts_nested_attributes_for](http://api.rubyonrails.org/classes/ActiveRecord/NestedAttributes/ClassMethods.html) in Rails.
### Scope

As compared to https://github.com/json-api/json-api/issues/795:

``` diff
- * It should allow mixed actions in a single request (e.g. creating + updating).
+ * It should allow creating/updating a resource and its relationships in a single request
* It should work for multiple types of resources (e.g. articles + tags) that would otherwise be accessed
at different endpoints.
* Every request should be transactional (i.e. all operations should succeed or fail together).
* Additive to the base spec (probably as an extension).
+ * Minimal footprint possible, favoring #795 for more advanced use cases
+ * Forwards-compatible with the larger scope of #795.

Here are a couple use-cases that test these requirements:

- * Create two articles, add one existing tag to both, create a new tag and add to both.
* Change the title of an article and update its tags: add one existing tag, create another,
and remove another (without replacing the whole tag set).
```
### Example

<img width="429" alt="screen shot 2016-09-01 at 9 29 31 am" src="https://cloud.githubusercontent.com/assets/55264/18484902/69880558-79b8-11e6-8a7f-111168783df0.png">

In this form, we need to:
- Add already-persisted tags (say, from an autocomplete)
- Add unpersisted tags (based on user input)
- Remove tags (when editing)
- Submit the post and its tags at the same time, so validations can run together.
### Proposal
- Embrace @beauby's [proposal for dealing with partial updates to collections](https://github.com/json-api/json-api/issues/1097). TL;DR: introduce `added-data`, `removed-data`, and `updated-data` keys for partial updates, keeping `data` for a full replacement of the collection.
[...]

Issue #1176, opened 2017-05-30
Extra Fields


richmolj:
The spec has great support for sparse fieldsets. I'd also like the opposite - rendering an attribute only when it is specifically requested.

Why might you want to do this? Well, some fields are computationally expensive, or only make sense in a specific context. For instance a `Person` might have a `net_worth` attribute that requires a number of calculations adding up assets, factoring in debt, etc.

The alternative is to make fields like this a relationship, and fetched via `include`. I believe this often adds additional awkwardness to the client and makes the API less expressive at best and misleading at worst.

I propose an `extra_fields` parameter with the same signature as `fields`, used to only render an attribute when it is specifically requested:

> /people?extra_fields[people]=net_worth

This is currently implemented in my [jsonapi_suite](https://jsonapi-suite.github.io/jsonapi_suite/#extra-fields) project.

beauby:
How about making use of the `fields` query parameter for this, by introducing some kind of "relative" prefix (namely `+`/`-`)?

Example:
> /people?fields[people]=+net_worth,-name

The idea is based on having a "default" fieldset, which can either be modified by `+`/`-`ing some fields, or fully specified with the current syntax.

richmolj:
@beauby either approach sounds good to me, as long as we can get it through.

dgeb:
The suggestion by @beauby is the one we considered initially. We pulled back on including this in 1.0 but deliberately left room for its inclusion later. 

One design constraint to consider is that the spec will probably never include multi-word params to allow for API-specific params and to avoid taking a stand on separator characters.

The obvious caveat to include with the usage of `+` and/or `-` prefixes is that they could not be combined with unprefixed field names for a given type.

ethanresnick:
+1 to @beauby's approach

richmolj:
👍 

So would you guys accept a PR with this @beauby's approach? Think we are all in agreement here.

ethanresnick:
Hey, sorry, I'm realizing I may have spoken too soon. 

I think that, if we do this, @beauby's approach is definitely the best. I also think we should definitely make explicit that, when no fields parameter is present, servers can return a default set of fields that omits any which are computationally expensive or rarely used. 

[...]

PR #1197, opened 2017-07-26
[WIP] Sideposting draft

beauby:
This PR is a work in progress, following discussions in #1089, #795, and #536.

Questions:

1. `POST`ing
    1. Expressing linkage between (fresh) resources in the request document
        - [x] Should we use ~~JSON pointers~~ or temp ids?
            - [x] Should temp ids be unique throughout a document or ~~just unique per type~~?
            - [x] What key should we use for temp ids? `tempid`? ~~`tid`? `tmpid`?~~
        - [ ] Should sideposted resources be within the `included` member or within some new ad-hoc member? (see #1215)

    2. Expressing linkage between (fresh) resources in the request document and persisted resources in the response document
        - [x] How do we tell the client which created resources correspond to which resources in the request? **Keep the temp ids in the response**
          - [ ] Should the temp id be in each resource object in the response, or should there be a single, document-level temp id => permanent id mapping object?
        - [ ] What should happen if a client uses `?include` on their POST and (for some reason) doesn't list in that param all the relationship paths for which they sideposted resources?
        - [ ] Should created sideposted resources in the response be within the `included` member or within some other ad-hoc member? (see #1215)

    3. Supported linkage graphs for sideposting
        - [ ] Should we put restrictions on the kinds of graphs allowed or should we leave it to the server implementations?
        - [ ] Should we put minimum requirements of the kind of graphs a server must support?
        - [ ] Should we offer further guidance on responding to unsupported sideposting requests? (Unsupported graph, etc.)

    4. Updating/deleting related resources
        - [x] ~~Should we allow updates of related resources while `POST`ing a new primary resource?~~ Not in 1.1, but maybe down the line
        - [x] ~~Should we allow deletion of related resources while `POST`ing a new primary resource?~~ Not in 1.1, but maybe down the line

2. `PATCH`ing
    - [ ] Should we allow updates of related resources while `PATCH`ing a resource endpoint?
    - [ ] Should we allow deletion of related resources while `PATCH`ing a resource endpoint?

3. Other
    - [ ] Should we allow bulk creation? (which is simply a matter of relaxing the constraint on the primary data being a single resource object).
    - [ ] If so, should we allow bulk updates? deletions?

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1330,6 +1330,14 @@ Content-Type: application/vnd.api+json
 }
 ```
 
+If a `POST` request *did* include some [Sideposted resources](#sideposting), the server
+**MAY** default to including the created related resources within the `included`
+member.
+
+If a sideposted resource is present in the response document, then it **MUST**
+have a `temp-id` member that identifies the corresponding resource from the request
+document.
+
 ##### <a href="#crud-creating-responses-202" id="crud-creating-responses-202" class="headerlink"></a> 202 Accepted
 
 If a request to create a resource has been accepted for processing, but the
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1235,13 +1235,12 @@ to create a resource with a client-generated ID.
 #### <a href="#crud-sideposting" id="crud-sideposting" class="headerlink"></a> Sideposting
 
 A server **MAY** accept a request to create related resources along with the primary
-resource, which is referred to as *sideposting*. The related resources **MUST**
-be placed within the `included` member. A sideposted resource **MUST** be referred to
-(from the linkage data of resource relationships) either via standard
-[resource identifier objects][resource identifier object] when using a
-client-generated ID, or as a [resource pointer object][resource pointer object]
-otherwise (that is an object with a single `pointer` member the value of which is a
-JSON Pointer [[RFC6901](https://tools.ietf.org/html/rfc6901)] to a resource object).
+resource, which is referred to as *sideposting*. A related resource **MUST**
+be placed within the `included` member, and have a `temp-id` member the value of which
+is an arbitrary string that uniquely identifies a related resource within the request
+document. The related resources **MUST** be referred to (from the linkage data of
+resource relationships) via a *temporary resource identifier object* (that is an
+object with a single `temp-id` member referring to a related resource).
 
 The document **MUST** still respect the full linkage requirement.
 
@@ -1262,20 +1261,22 @@ Accept: application/vnd.api+json
     "relationships": {
       "tags": {
         "data": [{ "type": "tags", "id": "9" },
-                 { "pointer": "/included/0" },
-                 { "pointer": "/included/1" }]
+                 { "temp-id": "1" },
+                 { "temp-id": "2" }]
       }
     }
   },
   "included": [
     {
       "type": "tags",
+      "temp-id": "1",
       "attributes": {
         "label": "JSON"
       }
     },
     {
       "type": "tags",
+      "temp-id": "2",
       "attributes": {
         "label": "REST"
       }
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1235,12 +1235,17 @@ to create a resource with a client-generated ID.
 #### <a href="#crud-sideposting" id="crud-sideposting" class="headerlink"></a> Sideposting
 
 A server **MAY** accept a request to create related resources along with the primary
-resource, which is referred to as *sideposting*. A related resource **MUST**
-be placed within the `included` member, and have a `temp-id` member the value of which
-is an arbitrary string that uniquely identifies a related resource within the request
-document. The related resources **MUST** be referred to (from the linkage data of
-resource relationships) via a *temporary resource identifier object* (that is an
-object with a single `temp-id` member referring to a related resource).
+resource, which is referred to as *sideposting*. Related resources **MUST**
+be placed within the `included` section, and have a `temp-id` member the value of
+which is an arbitrary string.
+Linkage with sideposted resources **MUST** be expressed via *temporary resource
+identifier objects*, that is [resource identifier objects][resource identifier object]
+where the `id` member is replaced with a `temp-id` member. A temporary resource
+identifier object **MUST** identify an individual resource within a request document.
+
+In case a sideposted resource refers to the primary resource, the primary resource
+**MUST** have a `temp-id` member, and linkage **MUST** be expressed using a
+temporary resource identifier object.
 
 The document **MUST** still respect the full linkage requirement.
 
@@ -1261,8 +1266,8 @@ Accept: application/vnd.api+json
     "relationships": {
       "tags": {
         "data": [{ "type": "tags", "id": "9" },
-                 { "temp-id": "1" },
-                 { "temp-id": "2" }]
+                 { "type": "tags", "temp-id": "1" },
+                 { "type": "tags", "temp-id": "2" }]
       }
     }
   },
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1232,6 +1232,24 @@ Accept: application/vnd.api+json
 A server **MUST** return `403 Forbidden` in response to an unsupported request
 to create a resource with a client-generated ID.
 
+#### <a href="#crud-sideposting" id="crud-sideposting" class="headerlink"></a> Sideposting
+
+A server **MAY** accept a request to create related resources along with the primary
+resource, which is referred to as *sideposting*. The related resources **MUST**
+be placed within the `included` member. A sideposted resource **MUST** be referred to
+(from the linkage data of resource relationships) either via standard
+[resource identifier objects][resource identifier object] when using a
+client-generated ID, or as a [resource pointer object][resource pointer object]
+otherwise (that is an object with a single `pointer` member the value of which is a
+JSON pointer to a resource object). The document **MUST** still respect the full
+linkage requirement.
+
+A server **MUST** handle sideposting requests transactionally, and either create all
+requested resources or none.
+
+A server **MUST** return `403 Forbidden` in response to an unsupported request to
+create a resource with sideposting.
+
 #### <a href="#crud-creating-responses" id="crud-creating-responses" class="headerlink"></a> Responses
 
 ##### <a href="#crud-creating-responses-201" id="crud-creating-responses-201" class="headerlink"></a> 201 Created
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1338,6 +1338,46 @@ If a sideposted resource is present in the response document, then it **MUST**
 have a `temp-id` member that identifies the corresponding resource from the request
 document.
 
+```http
+HTTP/1.1 201 Created
+Content-Type: application/vnd.api+json
+
+{
+  "data": {
+    "type": "articles",
+    "id": "123",
+    "attributes": {
+      "title": "Sideposting with json:api"
+    },
+    "relationships": {
+      "tags": {
+        "data": [{ "type": "tags", "id": "9" },
+                 { "type": "tags", "id": "14" },
+                 { "type": "tags", "id": "42" }]
+      }
+    }
+  },
+  "included": [
+    {
+      "type": "tags",
+      "id": "14",
+      "temp-id": "1",
+      "attributes": {
+        "label": "JSON"
+      }
+    },
+    {
+      "type": "tags",
+      "id": "42",
+      "temp-id": "2",
+      "attributes": {
+        "label": "REST"
+      }
+    }
+  ]
+}
+```
+
 ##### <a href="#crud-creating-responses-202" id="crud-creating-responses-202" class="headerlink"></a> 202 Accepted
 
 If a request to create a resource has been accepted for processing, but the
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1241,8 +1241,9 @@ be placed within the `included` member. A sideposted resource **MUST** be referr
 [resource identifier objects][resource identifier object] when using a
 client-generated ID, or as a [resource pointer object][resource pointer object]
 otherwise (that is an object with a single `pointer` member the value of which is a
-JSON Pointer [[RFC6901](https://tools.ietf.org/html/rfc6901)] to a resource object). The document **MUST** still respect the full
-linkage requirement.
+JSON Pointer [[RFC6901](https://tools.ietf.org/html/rfc6901)] to a resource object).
+
+The document **MUST** still respect the full linkage requirement.
 
 For instance, a new article might be created along with two tags with the following 
 request:
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1244,6 +1244,45 @@ otherwise (that is an object with a single `pointer` member the value of which i
 JSON pointer to a resource object). The document **MUST** still respect the full
 linkage requirement.
 
+For instance, a new article might be created along with two tags with the following 
+request:
+
+```http
+POST /articles HTTP/1.1
+Content-Type: application/vnd.api+json
+Accept: application/vnd.api+json
+
+{
+  "data": {
+    "type": "articles",
+    "attributes": {
+      "title": "Sideposting with json:api"
+    },
+    "relationships": {
+      "tags": {
+        "data": [{ "type": "tags", "id": "9" },
+                 { "pointer": "/included/0" },
+                 { "pointer": "/included/1" }]
+      }
+    }
+  },
+  "included": [
+    {
+      "type": "tags",
+      "attributes": {
+        "label": "JSON"
+      }
+    },
+    {
+      "type": "tags",
+      "attributes": {
+        "label": "REST"
+      }
+    }
+  ]
+}
+```
+
 A server **MUST** handle sideposting requests transactionally, and either create all
 requested resources or none.
 
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1241,7 +1241,7 @@ be placed within the `included` member. A sideposted resource **MUST** be referr
 [resource identifier objects][resource identifier object] when using a
 client-generated ID, or as a [resource pointer object][resource pointer object]
 otherwise (that is an object with a single `pointer` member the value of which is a
-JSON pointer to a resource object). The document **MUST** still respect the full
+JSON Pointer [[RFC6901](https://tools.ietf.org/html/rfc6901)] to a resource object). The document **MUST** still respect the full
 linkage requirement.
 
 For instance, a new article might be created along with two tags with the following 


brainwipe:
It's well worded and I found it easy to understand. 

Is this for creation of new items only? Does it cover update/delete scenarios?

ethanresnick:
Awesome start!

Some general thoughts:

1. I agree with @brainwipe that we should think about if/how this applies to PATCH on resource endpoints, and PATCH/POST on relationship endpoints. We don't have to include rules for all those endpoints in this PR, but we should make sure that whatever design we come up with could logically be extended to them. Maybe you can post some examples of how those cases might work in the thread?

2. Where do the created side-posted resources end up in the response? In the `included` array? If so, just consider that `?include` can be used on a POST request to ask for related resources to be returned with the newly-created resource. For instance, if the client POSTs an article with a sideposted author to `POST /articles?include=comments`, the current spec says they should get back the just-POSTed article, and its comments (from the linkage in the POST), but NOT the sideposted author...maybe that's ok, actually. Regardless, it'd be good to have some text about this, and an example response, in the PR.

3. I'm not sure how I feel about supporting pointers and temp ids...presumably we would only support one, right? If so, I think I'm leaning toward temporary ids, because they're probably easier for a client to generate than a pointer. Do you or anyone have strong feelings about this?

4. While we can require 403 forbidden from new servers that don't support sideposting, we should warn that 1.0 servers may return a different error. My guess is that most of them will return a 400, if we use an unrecognized pointer format (e.g., one missing an `id` key). However, if client side ids are in use and the pointer is just `{type: "...", id: "client-generated-id"}`, the server should technically return a 404, because it should ignore the presence of `included`, try to link a related resources based on the type/id, and not find the related resource. Some servers may instead create a link to a resource that doesn't exist (if their db doesn't do a foreign key check) and return a 200, without having created the sideposted resource at all. To avoid this, and the 400 vs. 404 vs. 403 discrepancy, it might be good to always use a pointer with a temporary id, even if client-side ids are in use. 

5. Finally, in the current wording, it's not clear to me if a resource in `included` can link to another sideposted resource in `included`. (Like, can you create a newsletter issue with sideposted articles,  with each of those articles linked to a sideposted tag?) Personally, I think JSON:API should allow clients to formulate very complex requests like this, but then allow servers to reject them with various errors. My comments on this from #795:

> the extension would be required to support graphs in which the only links between the resources in the request document are relationships from the primary data to embedded resources. (That is, the embedded resources don't link to one another, and the primary data's resource objects, if there's more than one, don't link to one another. But embedded resources can link to other, pre-existing resources.) This could be extended to say that any links are valid so long as the resulting graph is acyclic, to support the recursively embedded resources that @tkellen asked about on the original embedded resources proposal, in addition to interlinking between primary data resources, etc.
>
> As far as supporting other graphs goes, the extension would have feature flags (again, mechanism tbd) indicating the types of graphs it supports. The structure of those flags would be defined over time as we collect more real-world use cases. A request might have a way to specify which type of graph it's sending, in order for the server to use more efficient processing.

As things stand today, I think feature flags etc etc are way too complicated, but just delineating a few different graph types and creating a few different  graph type not supported errors might be good.

antonkomarev:
Just a tiny note that we shouldn't forget about validation messages. If something bad has happened, we should have a way to point to the attribute which was malformed.

beauby:
Thanks for providing constructive feedback so quickly! :+1:

1. > Is this for creation of new items only? Does it cover update/delete scenarios? @brainwipe 

    As it stands, this PR is only about resource creation, though the goal is to agree on a general mechanism to handle "compound requests". As @ethanresnick mentioned, we indeed have to make sure the retained mechanism generalizes to updates. [TODO: add example payloads for resource PATCH]
2. > Where do the created side-posted resources end up in the response? @ethanresnick 

    The created related resources would probably be sent back within `included`. I haven't written that part yet as the exact way those resources are matched with those from the request depends on whether we use JSON pointers or temp ids. This is indeed totally in the scope of this PR.
    >  If so, just consider that ?include can be used on a POST request to ask for related resources to be returned with the newly-created resource. @ethanresnick 

    Fair point about POSTs with `?include` query params – I guess this is what initially drove you towards using an `embedded` member instead? Not sure whether this is an issue or not.
3. > I'm not sure how I feel about supporting pointers and temp ids...presumably we would only support one, right? @ethanresnick 

    To clarify, I was merely mentioning that JSON pointers and temp ids are two possible solutions for linking resources while sideposting, not that we should support both (which would add unnecessary complexity IMO). JSON pointers seem semantically cleaner, at the expense of ease of implementation. Plus they would force us to introduce some kind of mechanism to link the newly created resources with the request payload. I personally do not feel strongly about this, and neither does @richmolj AFAIK.
4. > While we can require 403 forbidden from new servers that don't support sideposting, we should warn that 1.0 servers may return a different error. [...] To avoid this, and the 400 vs. 404 vs. 403 discrepancy, it might be good to always use a pointer with a temporary id, even if client-side ids are in use. @ethanresnick 

    Fair point – this would make server implementations easier as well.
5. > Finally, in the current wording, it's not clear to me if a resource in included can link to another sideposted resource in `included`. @ethanresnick 

    I intentionally didn't put any restriction on the kind of linking, leaving it to server implementations to decide what is a "supported sideposting request" (which resources are allowed, what kind of graph (possibly none if the server does not implement sideposting on that endpoint), etc.). I agree that it may be useful to offer guidance for responding to unsupported sideposting requests, though the current `errors` mechanism with JSON pointers should be sufficient. Thoughts?
    > Just a tiny note that we shouldn't forget about validation messages. If something bad has happened, we should have a way to point to the attribute which was malformed. @a-komarev 

    Again, the current `errors` mechanism with JSON pointers should be sufficient. Did you have a specific situation in mind?

antonkomarev:
> Again, the current errors mechanism with JSON pointers should be sufficient. Did you have a specific situation in mind? @beauby 

Should we point to the `data.relationships.{name}.0` or to `included.0`? I mean thats will be better to have in specification to prevent future questions about it.

beauby:
@a-komarev Let's say you have the following payload (where the first sideposted tag has an invalid `label` attribute):
```http
{
  "data": {
    "type": "articles",
    "attributes": {
      "title": "Sideposting with json:api"
    },
    "relationships": {
      "tags": {
        "data": [{ "type": "tags", "id": "9" },
                 { "pointer": "/included/0" },
                 { "pointer": "/included/1" }]
      }
    }
  },
  "included": [
    {
      "type": "tags",
      "attributes": { 
        "label": null // This tag will fail validation as the server does not allow nil labels for tags
      } 
    },
    {
      "type": "tags",
      "attributes": {
        "label": "REST"
      }
    }
  ]
}
```
Then the error response would include an error about that attribute, pointing to where it is defined in the request payload, i.e. `/included/0/attributes/label`.

_format/1.1/index.md
```diff
@@ -1232,6 +1232,65 @@ Accept: application/vnd.api+json
...
+{
+  "data": {
+    "type": "articles",
+    "attributes": {
+      "title": "Sideposting with json:api"
+    },
+    "relationships": {
+      "tags": {
+        "data": [{ "type": "tags", "id": "9" },
+                 { "temp-id": "1" },
```
antonkomarev: Shouldn't it be `{ "type": "tags",  "temp-id": "1" },` ? Because relationship name could differ from resource type.

beauby: The idea is that temp-ids identify a unique resource within the document, so the type is redundant. I agree that this introduces a difference with the "normal" resource identifiers, which only require the id to be unique *per type*. I do not feel strongly about it, so let's wait for others to chip in before making a call on that.

antonkomarev: I imagined the situation where we are accepting user generated IDs. Client generates unique id and send it as `temp-id`. But for example if client generates UUIDv5, this means that IDs could intersect between entities of different types. Possibility critically low, but it exists.

sandstrom:
Awesome! ⛵️ 

- I think temp-ids are better (but I'm not totally against pointers).
- As others have mentioned, it would be good to think of the update/delete scenarios too. I understand if you don't want to expand the scope too much here, but in many real-world uses all three would be needed.

Super-excited to see this PR! 🎉

beauby:
Cases not covered by the draft yet:

## `PATCH /articles/123` 
I would suggest the following rule:
> Any side-posted resource with a `temp-id` is considered a new resource, any other side-posted resource is considered an update. (Again, it would be the responsibility of the server to decide which related resources are allowed to be side-posted/side-updated.)

### Example: 
```ruby
PATCH /articles/123

{
  "data": { 
    "id": "123", 
    "type": "articles", 
    "attributes": { "title": "New Title" },
    "relationships": {
      "tags": {
        "data": [{ "type": "tags", "id": "14" }, { "type": "tags", "temp-id": "43" }]
      }
    }
  },
  "included": [
    {
      "type": "tags",
      "id": "14",
      "attributes": { "label": "New Label" }
    },
    {
      "type": "tags",
      "temp-id": "43",
      "attributes": { "label": "New Tag" }
    }
  ]
}
```

## `POST /articles/123/relationships/tags`

TODO

## `PATCH /articles/123/relationships/tags`

TODO

Finally, the one thing that hasn't been brushed by this draft is the deletion of related resources. Thoughts?

Edit: We could move from `included` to `embedded` and introduce a `deleted` top-level member.

_format/1.1/index.md
beauby: I updated the draft to make temp-ids type-relative. I don't believe the collision probability is an issue but it makes more sense (and simplifies wording) to keep the definition close to standard resource identifier objects.

richmolj:
@beauby if you're inferring anything sent with `id` should be updated, how are you accomodate client-generated IDs?

beauby:
@richmolj I think you misread, I was suggesting to infer that anything sent with a **`temp-id`** should be created (and the rest be updated), which solves the ambiguity that you're raising.

richmolj:
@beauby right, but I don't know if it should be created with the id matching the value of `temp-id`, or created with an auto-generated primary key.

beauby:
With an auto-generated id, unless there is a client-generated id.

richmolj:
@beauby how do I tell if `temp-id` is a client-generated id, or a random uuid? How do I know "I should persist temp-id" versus "I should use temp-id only for linkage"?

For example, let's say I want a new `Person` to be created. I would use `temp-id`, but the server would assign an auto-generated primary key underneath the hood.

Now let's say I have a client-generated ID. It would be the exact same payload. How would the server know to persist that client-generated id, if the payloads are the same?

Or similarly, if I were to use `id` instead of `temp-id`, how does the server know to create this versus attempt an update to it?

To solve this, I propose a `method` (name can change) key at the same level:

```ruby
# create with client-generated id
{ id: 's0m3uu1d', method: 'create', type: 'people' }

# create with auto-incremented primary key
{ temp-id: 's0m3uu1d', method: 'create', type: 'people' }
```

Not only does this solve the immediate issue, it matches a general RESTful pattern. We usually send our resource to the server with an HTTP verb. Now we are sending multiple resources to the server, and so we will send multiple verbs. This also allows for additional functionality:

```ruby
# destroy the object
{ id: 's0m3uuid', method: 'destroy', type: 'people' }

# nil the foreign key
{ id: 's0m3uuid', method: 'disassociate', type: 'people' }
```

You could make the case `method` could be a known whitelist, or allow additional values (so we could see edge-case implementations like "delete this object and all it's children" etc.)

beauby:
@richmolj To clarify, `temp-id`s are *never* to be persisted, and only exist to indicate linkage within a document. 

> For example, let's say I want a new Person to be created. I would use temp-id, but the server would assign an auto-generated primary key underneath the hood.

Yes.

> Now let's say I have a client-generated ID. It would be the exact same payload. How would the server know to persist that client-generated id, if the payloads are the same?

No it would not – because the payload would contain *both* `id` and `temp-id`.

> Or similarly, if I were to use id instead of temp-id, how does the server know to create this versus attempt an update to it?

The proposed rule is "If there is a temp-id, then it is a new resource, otherwise it is an existing resource, so it must be updated".

richmolj:
@beauby OK, making them *both* be sent for client-generated IDs now makes sense, thanks for clarifying 👍  I do disagree with this approach, though, as I find it more confusing and supports less use cases than something like the `method` approach above.

ethanresnick:
> No it would not – because the payload would contain both id and temp-id.

I haven't thought too much about the `method` approach, so I'm not comparing the two, but, just in general: I really like `temp-id` being present even when `id` is too. Yes, it makes for a tiny bit of repetition in the client-generated id case (`temp-id` and `id` might hold the same value), but it creates uniform processing rules for both the client- and server-assigned id cases by separating what really can be orthogonal concerns—declaring that two unpersisted resources should be linked together and assigning resources a permanent id.

## Checklist

@beauby I really like the checklist you made at the top of the issue. Going through it point by point:

> Should we use JSON pointers or temp ids?

It seems like most people in this issue marginally leaning toward temp-ids. I agree with that wholeheartedly. I think we should commit to that and move on. As far as the temp-id object format goes, I actually like your original approach of `{"temp-id": "...."}` rather than `{"temp-id": "....", "type": "...."}`. I agree with you that the `type` isn't necessary, and I think including it maybe even obfuscates what these objects' true purpose is, and invites confusion.

I understand what you're saying about it being a little harder to spec without the `type` key though... I just went through that exercise myself. Still, I'd bite the bullet. Here's how I think think I'd approach that: 

1. Create two subsections under the current "Resource Identifier Objects" section. One called "Permanent Resource Identifier Objects", which holds the content that's there today but clarifies that it's for identifying resources that have already been persisted, and one called "Temporary Resource Identifier Objects", which describes the `{temp-id: ""}` object format for linking resources that are just being created.

2. Just link to the definition or temporary resource identifier objects in the sideposting section.

> Should we put restrictions on the kinds of graphs allowed or should we leave it to the server implementations?

I don't think I want to put restrictions in the sense of saying that the server *is not allowed* to support certain graphs. My comment was actually trying to suggest the opposite: that your current wording might be imposing restrictions you didn't mean to impose. In particular, the wording that could be imposing restrictions now is the part that says "The document **MUST** still respect the full linkage requirement." I believe that would forbid a graph where there's just one resource in `included`, and it's not connected to the primary resource (i.e., it's really just specifying a second, independent creation).

That said, we should probably have a discussion, conceptually, about whether we want restrictions before we get lost in the details of how to make sure the copy is imposing/not imposing the restrictions we want or don't want.

> Should we put minimum requirements of the kind of graphs a server must support?

I edited your OP to add this to the checklist. I hope you don't mind. (I'd never edit a post to change/delete something you wrote, but I wanted to keep all the open questions in one checklist, so I thought just adding something would be ok.)

Basically, the idea would be that, even if we don't prevent servers from supporting any graphs, requiring them to support certain graphs might be useful. We already do this, e.g., by requiring conforming implementations to support sparse fieldsets, and the rationale is that, if all servers support a common baseline of functionality, it gives client libraries much more incentive to implement support for that functionality. I think the same argument would apply here. I also think there's a pretty obvious choice for what an appropriate baseline would be, given the actual, requested use cases that motivated this feature (create an invoice with its items, an author with their first post, etc). Quoting from my earlier comment, that baseline could be:

> the extension would be required to support graphs in which the only links between the resources in the request document are relationships from the primary data to embedded resources. (That is, the embedded resources don't link to one another, and the primary data's resource objects, if there's more than one, don't link to one another. But embedded resources can link to other, pre-existing resources.)

> Should we offer further guidance on responding to unsupported sideposting requests? (Unsupported graph, etc.)

Let's defer this until #1199 is solved.

> Should sideposted resources (as well as the created sideposted resources in the response) be within the included member or within a new embedded member?

Tbh, `embedded` never totally made sense. I'd say we should just reuse `included`.

> What should happen if a client uses ?include on their POST and (for some reason) doesn't list in that param all the relationship paths for which they sideposted resources?

<a id="include-param"></a>This is another item I added to the checklist, and it's the one issue to figure out when using `included`. Should the server omit the sideposted resources from its response (more consistent with current `?include` handling) or always include them (probably simpler to implement). Example:

```
POST /articles?include=comments

{
  "data": { 
    "type": "articles",
    "relationships": {
      "author": {
        "data": { "temp-id": "2" }
      }
      "comments": {
        "data": [
          { "type": "comments", "id": "1" },
          { "temp-id": "1" },
        ]
      }
    }
  },
  "included": [
    { /* comment with temp-id 1 */},
    { /* author with temp-id 2 */},
  ]
}
```

Possible responses:

1. No author, honoring include:

```
{
  "data": { 
    "type": "articles",
    "relationships": {
      "author": {
        "data": { "type": "people": "id": "1" }
      }
      "comments": {
        "data": [
          { "type": "comments", "id": "1" },
          { "type": "comments", "id": "2" },
        ]
      }
    }
  },
  "included": [
    { "type": "comments", "id": "1", /* .... */ },
    { "type": "comments", "id": "2", "temp-id": "1", /* ... */}
  ]
}
```

2. All sideposted resources are always in `included`; `?include` is still honored in the sense of providing including extra resources at the requested paths (here, the comment with server-id 1):

```
{
  "data": { 
    "type": "articles",
    "relationships": {
      "author": {
        "data": { "type": "people": "id": "1" }
      }
      "comments": {
        "data": [
          { "type": "comments", "id": "1" },
          { "type": "comments", "id": "2" },
        ]
      }
    }
  },
  "included": [
    { "type": "comments", "id": "1", /* .... */ },
    { "type": "comments", "id": "2", "temp-id": "1", /* ... */},
    { "type": "people", "id": "1", "temp-id": "2", /* .... */}
  ]
}
```

The tricky part with option 1 is that, if the no-longer included author had also linked to other sideposted resources, presumably those wouldn't be returned either — unless a different resource that still is part of `included` linked to those other sideposted resources too. Implementing that seems like a lot of extra complexity for little gain.

> Should we allow updates of [related] resources when POSTing a new primary resource?

I'm inclined to say no, but I'd be interested to hear the use cases. My gut reaction is that updating an existing, resource at the same time as creating a new one is rarely going to need to happen transactionally, and I think we should focus sideposting on those cases where things really must happen transactionally (e.g. to satisfy consistency/validation constraints: "a user always needs an address, an author always needs at least one post, an invoice always needs at least one item, so I need to create them together").

ethanresnick:
Btw, I realize we're discussing a lot of issues in parallel here. Maybe we should create separate threads/issues for the pieces of this that can be considered independently, so the posts here don't get super long and hard to follow. E.g., restrictions + requirements on sideposted graphs could probably be its own issue, as could the interactions with `?include`. Then, assuming we've settled on temp-ids (which I really hope), this issue could just focus on sanity checking this approach could extend well to all the other types of requests.

jgornick:
@beauby Nice work here! I really like the approach you're using.

A couple of comments though:

* Should `temp-id` be renamed to `tempId` or `tid` so that it's easier to access in languages like JavaScript? Just at first glance, `temp-id` looks out of place.
* Regarding the question if the temp IDs should be unique for the document or per type, and my vote is for unique per type.

ethanresnick:
> Should `temp-id` be renamed to `tempId` or `tid` so that it's easier to access in languages like JavaScript?

I don't want to get bogged down bikeshedding names, but this comment does bring up something important to flag: JSON:API wants to stay away from adding any multi-word keys, so we don't have to take a stand one way or the other on camel casing vs. dasherizing. (See @dgeb's comment [here](https://github.com/json-api/json-api/issues/1176#issuecomment-318174236) about that.) Therefore, something like `tempid` or `tid` would be preferable.

rpkilby:
Hi - just throwing in my two cents.

> JSON pointers seem semantically cleaner, at the expense of ease of implementation. 

Out of curiosity, what's the reasoning behind this? Given that we seem to be moving towards not including `type` in the linkage, I'm not sure what the practical differences are between `{"tempid": "1"}` and `{"pointer": "/included/1"}`.

I'd also like to point out that error responses would use JSON pointers to point to the included data. It would seem convenient that relationship linkage and error detail use the same pointer.

beauby:
@rpkilby What I meant was that a JSON pointer expresses "this is the chunk of data I'm talking about here", whereas a temp id expresses "there is an abstract concept of temporary resource that I know of and I am referring to one of those".  The practical difference though is that in the response document, it is quite clear that adding temp ids signifies "this is the actual resource that you referred to as X", whereas adding a pointer makes it slightly less clear because it refers to a part of the request document, rather than a defined concept.

ethanresnick:
Also, re temp ids being easier to implement, my reading is that mostly applies to clients constructing requests. Like, if the client is building up a composite request (say, as the user goes through multiple screens of a form in a CMS), having each part of the form just give its new resource objects an opaque id is easier than it having to know where that resource object will end up in the request document (to make the pointer).

brainwipe:
To add support for client creation of IDs over pointers: microservice/CQRS/eventing architectures _need_ the client to create an id because a service accepting a POST/PUT (by strict definition) does not return data. In those cases, the unique identifier that the database uses never leaves the confines of the microservice.

Rio517:
It seems like this discussions has stalled on picking a name for the temp ids. Can I advocate for `tid`?  I think it is more consistent other common id modifiers - `uuid` and `jid` coming to mind.

JohnBDonner:
Isn't it a little redundant to have the `type` twice on the request?
```
"relationships": {
  "tags": {
    "data": [
      { "type": "tags", "id": "9" },
      { "type": "tags", "temp-id": "1" },
      { "type": "tags", "temp-id": "2" }
    ]
  }
}
```

Here we could easily have the `data` for `tags` with only the `id` or `tid`, without repeating the `type` for every object, as it is already assigned two levels up in the body.

stephenprater:
I'd also like to advocate for `tid` (or even `tempid`) since `temp-id` is not a valid variable / method name in a fair number of languages due to the dash.

ethanresnick:
FYI, I've opened a couple separate issues to discuss the various sub-problems for sideposting, per my comment [here](https://github.com/json-api/json-api/pull/1197#issuecomment-318918876).

@beauby If you feel like I grouped the checklist items wrong, or left any out, in the splinter issues I opened, feel free to open new ones instead or comment on them and I'll gladly update the OPs.

scott-mcdonald:
I am coming to this pull request/discussion late but I wanted to share how we at movietickets.com handled "sideposting" with our hypermedia API (json:api as the media type).

_A quick background_, as director I pitched adopting json:api to the company and development team over 2 years ago and it is our official json based media type for our "Level 3 REST API". We currently have in the teens of 3rd party integrators with a good number of these being major companies with well known name recognition. So far the majority of these integrators have given us and the json;api specification "thumbs up" so that is a good thing. The one area where json:api was **not** pragmatic was _creating a single resource at a time constraint_ - in the real world this was not going to work. I felt like creating another "wrapper" or "aggregate" resource just for creating a resource and it's related resources was not a real solution - why when the json;api specification itself already defines a standard on how a document can be composed of a primary resource with optional related resources. So after brainstorming we did come up with a solution that we documented for 3rd party integrators but had to add the disclaimer this was non-standard json:api with our "rationale" for being non-standard. Hence why this topic is "interesting" to me at movietickets.com...

_The solution_ we are using for creating a resource and related resources is the following:

- We use a standard json:api 1.0 compliant document.
- All resources are created atomically, either they are all created or none are created.
- If the client wants to create related resources it puts the related resources in the `included` section and creates "temporary" json;api resource linkage just for the request itself.
- This leads to the convention for resource linkage interpretation:
  - If the referenced resource is in the document, the client intends to create a new resource,
  - If the referenced resource is not in the document, the client intends the referenced resource already exists in the system.

What we liked about this solution was the POST payload was a json:api 1.0 compliant document and felt straightforward to explain to 3rd party integrators as we did not introduce any new "keywords" to the specification.

An _example_ showing how a 3rd party integrator would create a payment for an order in our system:
```json
{
  "data": {
    "type": "order-payments",
    "relationships": {
      "payment-option": {
        "data": {
          "type": "payment-options",
          "id": "15-2544-CreditCard-Visa"
        }
      },
      "order-payment-items": {
        "data": [
          {
            "type": "order-payment-items",
            "id": "1"
          },
          {
            "type": "order-payment-items",
            "id": "2"
          },
          {
            "type": "order-payment-items",
            "id": "3"
          },
          {
            "type": "order-payment-items",
            "id": "4"
          },
          {
            "type": "order-payment-items",
            "id": "5"
          },
          {
            "type": "order-payment-items",
            "id": "6"
          }
        ]
      }
    }
  },
  "included": [
    {
      "type": "order-payment-items",
      "id": "1",
      "attributes": {
        "Value": "4111111111111111"
      },
      "relationships": {
        "payment-item": {
          "data": {
            "type": "payment-items",
            "id": "15-2544-CardNumber"
          }
        }
      }
    },
    {
      "type": "order-payment-items",
      "id": "2",
      "attributes": {
        "Value": "John"
      },
      "relationships": {
        "payment-item": {
          "data": {
            "type": "payment-items",
            "id": "15-2544-CardFirstName"
          }
        }
      }
    },
    {
      "type": "order-payment-items",
      "id": "3",
      "attributes": {
        "Value": "Doe"
      },
      "relationships": {
        "payment-item": {
          "data": {
            "type": "payment-items",
            "id": "15-2544-CardLastName"
          }
        }
      }
    },
    {
      "type": "order-payment-items",
      "id": "4",
      "attributes": {
        "Value": "05"
      },
      "relationships": {
        "payment-item": {
          "data": {
            "type": "payment-items",
            "id": "15-2544-CardExpirationMonth"
          }
        }
      }
    },
    {
      "type": "order-payment-items",
      "id": "5",
      "attributes": {
        "Value": "2020"
      },
      "relationships": {
        "payment-item": {
          "data": {
            "type": "payment-items",
            "id": "15-2544-CardExpirationYear"
          }
        }
      }
    },
    {
      "type": "order-payment-items",
      "id": "6",
      "attributes": {
        "Value": "33472"
      },
      "relationships": {
        "payment-item": {
          "data": {
            "type": "payment-items",
            "id": "15-2544-BillingPostalCode"
          }
        }
      }
    }
  ]
}
```
Now you can see why the current json;api constraint of creating 1 resource at a time was not pragmatic for us. BTW we also support this being a resource collection document for "bulk" creation purposes.

I don't know if this helps, I am not saying I am right, I just wanted to share what we are doing in the "real world" to meet our integration needs in the present...

beauby:
@scott-mcdonald I believe this is roughly where the current draft (which has not been updated to the latest consensus) is headed (modulo the temp ids being explicitly marked as such). Have I missed something?

scott-mcdonald:
@beauby No sir you have not missed anything as what I came up with a year ago is indeed almost identical to what your proposal is with the only difference being the simplification of not needing the whole "temp identifier" paradigm instead relying on _convention_. I just wanted to share to see if that different way of thinking about it adds any value for discussion in your proposal - eliminating tempid basically...

In the end I'll abide by the standard as I am also a .NET implementer of the specification with [JsonApiFramework](https://github.com/scott-mcdonald/JsonApiFramework) and have been looking forward to a 1.1 version of the specification for a **long** time. I am still waiting for my [revision added as a top-level member of resource](https://github.com/json-api/json-api/pull/824) pull request to be finalized since December of 2015 which is/was supposed to be a part of v1.1 of the specification.

vsoroka:
Just want to share how "sideposting" was implemented in Oro Platform: [Create and Update Related Resources Together with a Primary API Resource](https://www.orocrm.com/documentation/2.3/book/data-api#create-and-update-related-resources-together-with-a-primary-api-resource)

scott-mcdonald:
@vsoroka Your approach for "sideposting" is the exact same approach we came up with for our hypermedia API at our company. Only difference was for the update (PATCH) use case we did not require a `meta` attribute for the updating of related resources, instead used the _convention_ if the client "included" a resource then they meant for the included resource to be updated atomically with the primary resource.

I hope our approaches give the authors some insight to some prior art on this topic and it helps them in some way. I do look forward to a standard on this topic someday which would **greatly** improve json:api in my opinion.

richmolj:
@scott-mcdonald I appreciated reading your background on using the jsonapi spec and sideposting.   My solution in [jsonapi_suite](https://jsonapi-suite.github.io/jsonapi_suite/) is similar to yours and the current proposal. For a bit of *my* background, I wanted to create libraries that facilitated the development of common applications like the one from my [tutorial](https://jsonapi-suite.github.io/jsonapi_suite/tutorial):



This is why I ended up needing an explicit `temp-id`. The [code for that app](https://github.com/jsonapi-suite/employee-directory/tree/step_9_validations) is something like this:

```typescript
let department = new Department({ name: 'QA' });
let position = new Position({ title: 'Safety Inspector', department });
employee.positions.push(position);
// saves everything in one request
employee.save({ with: { positions: 'department' });
```

Your proposed spec could do this too. However, now I have this in-memory `position` object, and I need to assign it the server-minted `id`. In other words, I need the response to contain *both* the temporary identifier I sent to the server *and* the "real" id so I can update my in-memory objects and ensure my single-page-app has correct state.

This brings up what the current spec in this issue hasn't addressed yet - deletions and disassociations. Scenarios would be something like:

* Deletion: Save the employee while deleting a previous position and adding a new one.

* Disassociation: Given a position already has a department, remove the department from the position without destroying the underlying record (ie, make the foreign key `nil`). The most common use case for this is removing tags.

To solve these I added `method`:

```ruby
{
  relationships: {
    department: {
      data: {
        id: '1',
        type: 'departments',
        method: 'disassociate' # can be create/update/destroy/disassociate
      }
    }
  }
}
```

I like this because it's just REST - I'm sending a resource to the server with a corresponding verb.

Hope this proposed spec evolves to the point it can actually power my sample app!

scott-mcdonald:
@richmolj Your [jsonapi_suite](https://jsonapi-suite.github.io/jsonapi_suite/) website is **impressive** - excellent work! Thank you for the example demonstrating the reasoning behind "temporary identifiers". I understand the "temporary identifier" paradigm clearly now, you basically want to reconcile the client-side generated document passed in the POST call to the server-side generated document possibly returned in the POST call.

I think the JSON API specification is great and I want to see it succeed but after using it for real world purposes I believe it needs to evolve in some areas:

- Revision as a top-level resource member for optimistic concurrency control and caching purposes. Think of revision as the `etag` per resource.
- Sideposting per this proposal
- Batch operations meaning multiple create/update/delete of resources as an atomic transaction.
- Filtering standard

If the JSON API specification had these things, dare I say it would be **awesome**. It is a little disheartening that JSON API specification has not officially evolved at all since 1.0 release...

DLiblik:
In 2015 we implemented the following solution.  It has been in use successfully since then (millions of transactions processed), with the following features:

* Handled as a POST with a ```data``` holding an array, and a ```meta``` tag indicating it is a batch.
* Each object in ```data``` has a ```meta``` tag indicating what the operation is for that object
* For posted objects, each has a "_locId" - an ID that is local only to the message, so that other objects in the same message can reference them by id. These IDs are deliberately formatted to be non-overlapping with global IDs. They are also returned in the response to allow the client to know which local ID goes with which global ID (if of interest).
* In our case, the server handles the request in a single transaction (atomic), which can be general (do adds, then patches, then deletes) or specific (i.e. custom handler), but depending on the nature of the transaction this may not be semantically essential
* The response contains a data array with the results for each operation

We recently updated to use a (__EDIT__: 'MERGE') verb to respect the JSON API protocol (i.e. POSTing a data array was not valid), but that's pretty minor.  Don't ask me why we put underscores before some of the meta attributes. We probably felt we needed a bike-shed inside our bike-shed.

__EDIT__: The reason the object to DELETE is sent with attributes in this implementation is to ensure it matches the current one on the server - this is not necessary in every impl'n, but is supported on ours.

### Sample

This sample posts a set of objects having the following structure:
* survey-set
    * survey
        * survey-feature


__POST /survey-sets__ (updates a survey-set and its related survey and survey-feature objects)
```json
{
    "meta": {
        "is-batch": true,
        "batch-id": "1"
    },
    "data": [{
        "attributes": {},
        "relationships": {},
        "type": "survey-set",
        "id": "8",
        "meta": {
            "_op": "PATCH"
        }
    }, {
        "attributes": {
            "occupancy-type": "assembly,storage",
            "sprinkler-types": "dry"
        },
        "relationships": {},
        "type": "survey",
        "id": "8",
        "meta": {
            "_op": "PATCH"
        }
    }, {
        "attributes": {
            "type": "secondaryEntrance",
            "sub-type-ordinal": 4
        },
        "relationships": {
            "survey": {
                "data": {
                    "type": "survey",
                    "id": "8"
                }
            }
        },
        "type": "survey-feature",
        "id": "688",
        "meta": {
            "_op": "DELETE"
        }
    }, {
        "attributes": {
            "center-lat": 11.111,
            "center-lng": -11.111
        },
        "relationships": {},
        "type": "survey-feature",
        "id": "690",
        "meta": {
            "_op": "PATCH"
        }
    }, {
        "attributes": {
            "sub-type-ordinal": 4
        },
        "relationships": {},
        "type": "survey-feature",
        "id": "2483",
        "meta": {
            "_op": "PATCH"
        }
    }, {
        "attributes": {
            "type": "fireServiceConnection",
            "sub-type-ordinal": 1,
            "center-lat": 11.222,
            "center-lng": -11.222
        },
        "relationships": {
            "survey": {
                "data": {
                    "type": "survey",
                    "id": "8"
                }
            }
        },
        "type": "survey-feature",
        "meta": {
            "_op": "POST",
            "_locId": "_1"
        }
    }]
}
```

__RESPONSE (201)__
```json
{
  "data": [
    {
// redacted -- all PATCHed and POSTed objects
    }
  ],
  "meta": {
    "is-batch": true,
    "batch-id": "1",
    "results": [
      {
        "_op": "PATCH",
        "type": "survey-set",
        "id": "8",
        "status": 200
      },
      {
        "_op": "PATCH",
        "type": "survey",
        "id": "8",
        "status": 200
      },
      {
        "_op": "DELETE",
        "type": "survey-feature",
        "id": "688",
        "status": 204
      },
      {
        "_op": "PATCH",
        "type": "survey-feature",
        "id": "690",
        "status": 200
      },
      {
        "_op": "PATCH",
        "type": "survey-feature",
        "id": "2483",
        "status": 200
      },
      {
        "_op": "POST",
        "type": "survey-feature",
        "id": "2484",
        "_locId": "_1",
        "status": 201
      }
    ]
  }
}
```

dgeb:
I have been struggling a bit with the name `tempid`. I think of this as a document-scoped secondary key that might be more clearly named `key`, which also has the advantage of not trying to fake a union of two words into one.

What do others think of calling this field `key`? I've been exploring this concept for use with my operations proposal, and would like to formalize it in the main 1.1 spec outside of the confines of sideposting.

valscion:
~~Calling it `key`, or something more specific (such as `batch-key` or similar) seems more intuitive to me than `temp-id`. :+1:~~

EDIT: The reasoning in comments below have changed my mind. Read them instead of my comment :relaxed:

DLiblik:
Another logistics challenge on the server side is cross-referencing POSTs with temp ids.
For example (contrived):

```
POST: blog-post.first-comment => comment
POST: comment.post => blog-post
```

If this is to be created on the server in one batch, then both objects
will have a temp id, and one of the objects must be created first - this
is further compounded if the first-comment property is required, but that
truly is a completely out-of-scope issue for this. 

This is not in scope for the message structure, I know, but one solution
is to 'split' one of the POST operations into a POST and a PATCH:

```
POST: blog-post.first-comment => null
POST: comment.post => blog-post
PATCH: blog-post.first-comment => comment
```

If this is a client responsibility, then the message format needs to support
having more than one operation for the same object (the blog-post in this
example).

There are other examples we've had where ordering of the operations
is important and also includes a need for multiple PATCHes to a pivot
object:

```
PATCH a to remove a=>b
DELETE b
POST c (b and c can't co-exist)
PATCH a to set a=>c
```

ethanresnick:
> What do others think of calling this field key?

Honestly, `key` seems less clear to me, because it's so generic. E.g., `key` could be permanent, whereas `tempid` better suggests that the id only applies for the duration of this request/response.

That said, I care way more about picking something than I do about the exact name, so I'm fine going with `key`, esp. if that would work better for operations.

DLiblik:
The reason we chose `_locId` (`local-id` would be clearer) is because in our
case the ID is assigned in the client when the object is created which can be
some time before it is submitted (i.e. the user creates something and edits
it for a while before submitting).

That's outside the scope of the issue here, but considering the context of the
use case here is perhaps valuable.

That said @ethanresnick , it's important to move this forward and `key`
would work just fine for us as well.

richmolj:
I agree with @ethanresnick - I prefer the more descriptive `tempid` than the generic `key`, but happy with anything that moves this along.

Rio517:
It seems like this discussions has stalled once again on picking a name for the temp ids. We were close to concluding `tid` was the way to go.  Can I continue to advocate for `tid`? I think it is more consistent other common id modifiers - `uuid` and `jid` coming to mind.

jgornick:
I prefer `tempid` as well. 

I look at this from another perspective too: Whatever is decided means the name will most likely have to be "reserved" similar to `type` and `id`. So, in theory one wouldn't include this field name in their attributes to avoid conflicts. I _feel_ like `key` (e.g. `key`/`value` objects) might conflict with existing implementations of JSON (just a feeling, no data to back this up).

dgeb:
> Whatever is decided means the name will most likely have to be "reserved" similar to type and id.

This is not necessarily true - we do not actually _need_ to reserve this as a field in the attributes/relationships/type/id namespace. It is analogous to allowing an attribute named `meta`, which is fine although perhaps odd. This member should never be used for sorting, filtering, etc. 

Thanks for everyone's feedback re: `key`. I will follow through with my operations proposal to share another perspective. I don't think we'll be merging either proposal until both are considered together.

In the mean time, there are some other general concerns I still have about this PR:

* We can't make the presence of `key` / `tempid` a requirement for linkage. Client-generated ids must be allowed for "included" resources, just as they can be allowed for primary resources. So `type` / `id` must also be allowed for linkage.

* This proposal provides very little utility without also allowing a corresponding update (i.e. `PATCH`) solution that can also create resources. Take for example an invoice that "contains" invoice items. If an invoice can be created together with invoice items, it should also be able to be updated together with new invoice items. The nuance of specifying invoice items to create vs update seems thorny, especially when paired with the need to allow for client-generated ids.

scott-mcdonald:
I also agree with @ethanresnick  that `key` is not a clear self-documenting name for the client supplied document scoped temporary identifier for creating related resources. I was thinking about this proposal over the Thanksgiving break and wanted to throw out some thoughts/ideas that may help or may not help and be total crap:

**Idea #1** - Keeping with the current design - what are we really asking the server to do? The client wants the server to return the original client created temporary identifier it created for document scoped linkage purposes with the newly created server identifiers. So with that said, any of the names like `temp-id`, `tempid`, `tid`, `original-id`, `client-id`, `local-id`, etc. in the created/updated resources would work and seem self-documenting.

**Idea #2** - Thinking differently - what are we really asking the server to do? The client wants the mappings of client generated identifiers to server generated identifiers. And BTW, not all clients care about the temporary client-generated identifier mappings being echoed back - so only if the "client" asks for the mappings having some new document-level "section" that contains the mappings that we have been debating about in this proposal. Not sure what the section would be called so just picking `mappings` to pick something for now, it would look something like the following:

````js
{
  "links": {...}
  "data": {...} or [...]
  "included": [...]
  "mappings": [{"type": "articles", "client-id": "123", "server-id": "567"}, or
               {"type": "articles", "client-id": "123", "id": "567"}, or
               {"type": "articles", "tid": "123", "id": "567"}, or
               {"type": "articles", "temp-id": "123", "id": "567"}, or whatever but something along these lines
               ...]
}
````

Maybe it goes beyond mappings but the results of the operation as well such as if it was created or updated or deleted(?) so something like the following:

````js
{
  "links": {...}
  "data": {...} or [...]
  "included": [...]
  "operations": [{"type": "articles", "client-id": "123", "server-id": "567", "result": "Created"}, or
               {"type": "articles", "client-id": "123", "id": "567", "result": "Created"}, or
                 {"type": "articles", "tid": "123", "id": "567", "result": "Updated"}, or
                 {"type": "articles", "temp-id": "123", "id": "567", "result": "Created"},or whatever but something along these lines
                 ...]
}
````
This type of document would only be returned as the response of POST, PATCH. This idea seems self-documenting and explicit but maybe is overkill. Also I am not sure how a client could "ask" for this or it is always returned by the server...

**Idea #3** - Maybe this "mapping" feature **is not** the responsibility of the specification at all but instead is an implementation detail of the hypermedia API server if this feature is desired. The hypermedia API domain model as designed could contain an attribute (or meta) that is intended to be a unique identifier initialized by a client just for this mapping purpose and is echoed back to the client or something along that lines.

**Comment by ethanresnick** [comment]
@dgeb Thanks for weighing in here, Dan. I think it's really great + important to get your input as this is evolving. A couple questions:

> We can't make the presence of key / tempid a requirement for linkage. Client-generated ids must be allowed for "included" resources, just as they can be allowed for primary resources. So type / id must also be allowed for linkage.

I'm not sure I follow this. Yes, we have to allow client-generated ids for all resources (even in `included`), but does that mean we can't also require that sideposted resources have a temp id, and that that temp-id be used for linkage, *whether or not they have a client-generated id*? That's what I proposed in my fourth point [here](https://github.com/json-api/json-api/pull/1197#issuecomment-318208843), and it has some other advantages.

> This proposal provides very little utility without also allowing a corresponding update (i.e. PATCH) solution that can also create resources... The nuance of specifying invoice items to create vs update seems thorny, especially when paired with the need to allow for client-generated ids.

Maybe this looks easier to specify in light of my answer above, but the permutations seem relatively straightforward:

- Create primary resource(s) and create related ones (the classic sideposting case): do a POST with the primary resource to create in `data` and related resources to create in `included` (with tempids).

- Update primary resource(s) and create related ones (first case you mentioned): same as above, except the method is PATCH.

- Update primary resource(s) and _update_ related ones (second case you mentioned): A PATCH, where the resources to update are in `included`. These resources to update are distinguished from resources to create (i.e., the case above), because they wouldn't have a tempid. This is the solution @beauby proposed [here](https://github.com/json-api/json-api/pull/1197#issuecomment-318705828). We could allow this case in 1.1, or just say that servers should error on `included` resources with no tempids; then, we can specify this behavior later.

- Create resources and update related ones: this is the last case in the matrix of (create, update) x (primary, included). If we wanted to support it -- and I'm not sure we really do -- we could use the same approach as above to differentiate the related resources that are updates vs creations.

Create resource and update an existing resource that's being linked to the new one:

**Comment by ethanresnick** [comment]
@scott-mcdonald Thanks for weighing in as well! I'm not opposed to the idea of having a single mapping of tempids to permanent ids in the response (I actually explored that [earlier](https://github.com/json-api/json-api/issues/795#issuecomment-132955057)). That said, I'm also fine with putting the tempid in each resource object in the response. 

I've added new items to the issue checklist to cover these two options. For now, though, I don't think where to put the tempids in the response is the real sticking point, so we can probably come back to that at the end.

**Comment by dgeb** [comment]
> Yes, we have to allow client-generated ids for all resources (even in included), but does that mean we can't also require that sideposted resources have a temp id, and that that temp-id be used for linkage, whether or not they have a client-generated id?

@ethanresnick it feels like a strange requirement to force the usage of temporary ids just to identify included resources that should be created, especially when paired with a primary resource that's created with just a `type` and `id`. And then those temporary ids would have to be managed and returned by the server in the response. And all this ceremony is just to mark included resources as "to be created", when none of this is necessary when posting resources as primary data to the same server that accepts client-generated ids.

**Comment by richmolj** [comment]
> And all this ceremony is just to mark included resources as "to be created"

@dgeb there is a separate reason for temp ids, aside from the one you note, explained [here](https://github.com/json-api/json-api/pull/1197#issuecomment-343179886).

On a separate note:

> when none of this is necessary when posting resources as primary data

This is why I propose adding `method` (open to other names) - pair a resource with a verb just like you would for primary data. This allows us to support `delete` and `disassociate` as well, explained at the same link above.

**Comment by ethanresnick** [comment]
> And all this ceremony is just to mark included resources as "to be created", when none of this is necessary when posting resources as primary data to the same server that accepts client-generated ids.

@dgeb It's also for backwards compatibility, as I mentioned upthread:

> if client side ids are in use and the pointer [to the included resource] is just {type: "...", id: "client-generated-id"}, the server should technically return a 404, because it should ignore the presence of `included`, try to link a related resources based on the type/id, and not find the related resource. Some servers may instead create a link to a resource that doesn't exist (if their db doesn't do a foreign key check) and return a 200, without having created the sideposted resource at all. To avoid this, and the 400 vs. 404 vs. 403 discrepancy, it might be good to always use a pointer with a temporary id, even if client-side ids are in use.

Edit: Per the quote above, there is a risk without temp ids that existing servers will return a 200 on sidepost requests even though they should error. However, the case I described in the quote above is arguably an implementation bug—though it still might be a common issue in APIs backed by non-relational dbs. A rarer, but perhaps more troubling case — because it can't be categorized as a bug — is:

```
POST /collection

{
   "data": [{ 
     /* resource object with client-generated id, 
         and a relationship/resource identifier object pointing to included[0] */
   }],
   "included": [{
     /* sideposted resource with a client-generated id 
         that an existing resource happens to be using */
   }]
```

This should 409, because the client is trying to create the sideposted resource with an id that's already in use. However, an existing server will instead 200 by ignoring `included` (per JSON:API's rule about ignoring unknown keys), and creating the primary resource with a relationship to the existing resource (which is wrong).

We may be ok with this error though, since it's infrequent. In that case, the other key argument is that we're gonna need to spec something like temp ids anyway, for APIs that use server-side ids. Given that, it probably doesn't make sense for client libraries to have to have two separate code branches for sideposting based on the server's permanent id strategy. Likewise, we probably don't want to have to spec/maintain two differently sets of sideposting rules based on which id strategy is in use.

~~Basically, we want old servers to hard error on the presence of sideposted resources, and a new pointer format is one way to do that. There are others (e.g., merge sideposting with bulk creation/patch, and always require data to be an array when sideposting is in use), but some of them have caveats (e.g., the "force `data` to an array" doesn't work for patching relationship endpoints where data can already be an array), and I kinda like the new pointer approach for how it doesn't make server implementation of sideposting dependent on the permanent id strategy in use.~~

@richmolj
> This is why I propose adding method (open to other names) - pair a resource with a verb just like you would for primary data. This allows us to support delete and disassociate as well, explained at the same link above.

That is, indeed, more powerful. Once we add all that power, though, it's starting to feel like the full operations extension might be more appropriate for those use casses. Still, I don't want to foreclose the `method` option in the future. (My "inline operations" proposal from #795 was in the same vein.) 

My preferred solution for 1.1 would be to always have the resources in `included` be created (i.e., only supporting the top two cases from [here](https://github.com/json-api/json-api/pull/1197#issuecomment-347310889)), and always have tempids. Then, if we want to support other actions on the included resources down the line, we can add a method key (which defaults to "create" if missing) or we can go the route of treating a missing tempid as a sign to update rather than create. Either way, we've preserved our options and the extensions compose nicely.

**Comment by dgeb** [comment]
> there is a separate reason for temp ids, aside from the one you note, explained here.

@richmolj I was specifically talking about a scenario in which a server supports client-generated ids. Sorry if I wasn't clear.

> Once we add all that power, though, it's starting to feel like the full operations extension might be more appropriate for those use cases.

@ethanresnick I will make this a public proposal within a week. I can't expect others to discuss something that's not concrete and public.

**Comment by ethanresnick** [comment]
> @ethanresnick I will make this a public proposal within a week. I can't expect others to discuss something that's not concrete and public.

@dgeb Sounds good. I also haven't seen the proposal, so I won't speak more for it :) Also, I updated my comment above to clarify my current position on tempids, since the old framing didn't really get at it. Lmk if that makes sense.

**Comment by dgeb** [comment]
> Also, I updated my comment above to clarify my current position on tempids, since the old framing didn't really get at it. Lmk if that makes sense.

@ethanresnick Thanks for clarifying. I should say that there would be no ambiguity if side-posting was opted into explicitly as an extension (as per #1207). A server that accepts the extension would know exactly what to expect.

**Comment by richmolj** [comment]
> That is, indeed, more powerful. Once we add all that power, though, it's starting to feel like the full operations extension might be more appropriate for those use casses. 

I look forward to seeing the full operations proposal. That said, you could make the case that this entire PR should go away in favor of the operations proposal - that was the direction https://github.com/json-api/json-api/issues/795 was going before https://github.com/json-api/json-api/issues/1089 was created to try and separate the two.

So what's the goal of this PR? I think there are three ways to define the goals of "sideposting" independent of operations:

* **Abstract**: Ability to perform the same actions we already perform on primary data (`GET/POST/PATCH/DELETE`) at any level in the graph.
* **Practical**: Support simple, common web forms like shown in https://github.com/json-api/json-api/issues/1089 and [this tutorial](https://jsonapi-suite.github.io/jsonapi_suite/tutorial). This is actually how I got involved - `ember-data` couldn't support simple forms, and all discussion of the issue led here.
* **Simple/Intuitive**: From what I've seen so far, the operations proposal will be more complex than sideposting (such is its nature, as it supports more complex functionality). Sideposting simply mirrors the same payload we see on `GET` (`tempid` tangent aside).

The first two bullet-points would require a way to delete data. After that, there's a clear separation from how I think of operations - more complex scenarios where I need partial transactions, ordering of operations, etc...RPCish stuff that diverges from the general premise of "mirror the existing spec".

To the third bullet-point - can we support `DELETE` while still keeping things simple? I believe so - what could be simpler than "secondary data works the same way as primary data"? We interact with primary data by pairing a resource with a verb. I suggest pairing sideposted data with a verb as well. This seems intuitive (a familiar REST pattern and no implicit side-meaning of `tempid`), supports more scenarios (`DELETE` at minimum), and draws a clear distinction from operations.

One final note. Consider the nested web forms I have referenced above. If this PR does not support `DELETE`, that means my form submission will submit one payload on create/update and *a significantly different payload* if the user marks an item for deletion. That's going to be a pain point for developers and library authors alike - at that point I'd prefer avoiding sideposting altogether just to keep things consistent.

**Comment by ethanresnick** [comment]
>  I should say that there would be no ambiguity if side-posting was opted into explicitly as an extension (as per #1207). A server that accepts the extension would know exactly what to expect.

@dgeb That's true, but, of course: 1) it would block sideposting on us figuring out that negotiation mechanism (which in turn might be blocked on other details of extensions and/or profiles) and 2) there's still the issue that APIs using server assigned ids are going to need temp ids anyway. For those reasons, I'd much, much prefer to keep @beauby's proposal basically as is, address the remaining checklist items, and merge in some version of sideposting (that addresses just creating related resources on POST and PATCH) ASAP, and call that 1.1 beta. Ideally, we could do that by year end. Thoughts?

Even if you disagree with the plan above, can we try to define a concrete timeline + next steps progression for sideposting? (And hopefully for extensions/profiles too, which I believe I owe you a message on — although that conversation may be more stuck than sideposting, so perhaps it makes sense to proceed with sideposting first, which actually has been requested much more frequently.)

> That said, you could make the case that this entire PR should go away in favor of the operations proposal - that was the direction #795 was going before #1089 was created to try and separate the two.

@richmolj I still think the reasons we split out this proposal out apply. Just to reiterate them for others in this thread (by quoting you and me from #795):

> most people are looking for a solution to [Creating/updating a single resource with its relations in one go]. I think a minority care about [generically grouping multiple operations into a single request]. But arriving at a solution on [the former] is being blocked by [the latter].

Moreover, re the argument that we should do only operations because sideposting is duplicative:

> we shouldn't take the "it's duplicative" argument to its extreme, as that would suggest removing most of the base spec, since all one really needs is operations.

I.e., *everything* can be read as duplicative of operations, but that's a whole different model, and it shouldn't prevent us from extending the existing payload structure to cover common cases.

> The first two bullet-points would require a way to delete data.

This idea — basically, that a sideposting MVP requires deletion support — seems to be at the heart your comment. I'm curious to hear what others think about this.

As I see it, a simple rule for now that says "create the client-provided resources in `included` on POSTs and PATCHes" would solve a bunch of use cases and doesn't preclude at all, or make any messier, expanding support to cover a proposal like #1089 down the line.

**Comment by richmolj** [comment]
@ethanresnick - well summarized. Let me take one more crack at clarifying my comment, since lack of delete support means my libraries wouldn't be able to use any of this (😭 ).

> most people are looking for a solution to [Creating/updating a single resource with its relations in one go]. I think a minority care about [generically grouping multiple operations into a single request]. But arriving at a solution on [the former] is being blocked by [the latter].

I think the _reason_ most people are looking for this solution is because it's very common to have a nested form in a single-page javascript application. It's also very common for those forms to have a delete button (it's in the examples I've given, [here](https://github.com/json-api/json-api/issues/1089) and [here](https://jsonapi-suite.github.io/jsonapi_suite/tutorial)). So lack of delete support, to me, means jsonapi still cannot support common web forms and the original problem is not being solved. I think it would be a poor developer experience to say, "if your form has a delete button do it this way, otherwise do it this completely different way".

This also significantly impacts library developers. Let's say you're writing a client and want an API for "save this object with its relationships":

```ts
employee.save({ with: { positions: 'department' } })
```

The library author needs to go down an entirely different codepath if a position or department has been deleted. At that point, this PR is actually counterproductive - a library author would want to generate the same operations-style payload every time, rather than switching payloads conditionally.

I still **very much want sideposting separate from operations**, because the payload would be simpler and more intuitive...but if delete is not supported I'm not sure that's a viable option.

I'd actually like to ask the opposite - why ***wouldn't*** we want to support delete? It means we no longer have the implicit meaning of `tempid` (meaning, no create/update inference because you'll have an explicit verb saying `create` or `update`), enables more server-side opportunities (we could have a base list of verbs, but stuff like `disassociate` could be optionally supported by the server), has symmetry with the way we interact with primary data (the RESTful paradigm of pairing a resource with a verb)  and a clear separation of sideposting vs operations ("use operations whenever pairing a resource with a verb is not enough, such as when you need to control the order of writes"). What's the downside?

ethanresnick:
@richmolj 

> I'd actually like to ask the opposite - why wouldn't we want to support delete?

Supporting delete = bigger scope = more to find consensus on = who the hell knows when this will ship. At least that's how I've been thinking about it; getting consensus on _anything_ re sideposting has been so hard that I really want the narrowest version possible.

> So lack of delete support, to me, means jsonapi still cannot support common web forms and the original problem is not being solved... The library author needs to go down an entirely different codepath if a position or department has been deleted. At that point, this PR is actually counterproductive - a library author would want to generate the same operations-style payload every time, rather than switching payloads conditionally.

I definitely don't want something that's counterproductive. But, I'm not sure it's as bad as you describe. Delete is really two things: 1) delete resource entirely, or 2) disassociate it from the relationship. In the case of 1), how often does that need to be done in the same request? IMO, the main point of sideposting is to let creations that need to happen transactionally happen transactionally, and it seems like deleting a related resource (and retrying delete if needed) can usually happen in a separate request without issue. If the delete needs to happen for (e.g.) foreign key integrity, the db may take care of it on its own. 

For the second case (disassociation), there's an obvious way to do that today, which is to just replace the whole relationship by sending back `data` with the new contents. That's slightly more likely to run into concurrency issues than having separate `updated_`/`removed_`/`added_data` buckets would, but it's not as though having the separate keys makes you immune and, to fully prevent the problem, you need something like #600 either way. Also, replacing `data` isn't practical in some cases when the relationships contents are huge, but these are relatively rare. 

Don't get me wrong: I'm not saying that the absolute minimal sideposting approach I'm proposing is a cure all, just that it seems like a win over the status quo, and something we can build on, and that it's narrow enough that we won't regret adding it.

> It means we no longer have the implicit meaning of tempid (meaning, no create/update inference because you'll have an explicit verb saying create or update)

I'm not proposing this switching logic for the minimal version. Quoting myself:

> My preferred solution for 1.1 would be to always have the resources in included be created (i.e., only supporting the top two cases from [here](https://github.com/json-api/json-api/pull/1197#issuecomment-347310889)), and always have tempids. Then, if we want to support other actions on the included resources down the line, we can add a method key (which defaults to "create" if missing) or we can go the route of treating a missing tempid as a sign to update rather than create. Either way, we've preserved our options and the extensions compose nicely.

Startouf:
Hey some feedback from someone using 
- noSQL (MongoDB) and Elasticsearch databases.
- a Rails 5 API  communicating with a Single Page React App using json:api and...
- ... @richmolj `jsonapi_suite` library featuring sideposting with `method` and `tempid` 

The `method` concept allows to clearly and intuitively make the distinction between 2 different types of data models and their associated operations. I'm quite in favor of it.

I'm using MongoDB/mongoid as my primary DB/ORM, so I'm already into JSON from read/persistence to response and my use cases also involve embedded data. Sometimes, there will be a need to mix embedded-relationships with foreign relationships and the `method` seems to be the best way to clear things up here

For embedded data like `person <-> educations` ...
```
person: { 
  educations: [
    { title: "Some fun at MIT !", school_id: '5a207c88aba9cfa1a2f5943f', start_date: "2017-11-05", ... }
    { title: "Bored in High School", school_id: '5a207c88aba9cfa1a2f5943f', start_date: "2015-11-05", ... }
  ]
}
```
the sideposting requirement is to be able to `create/destroy` those associations whereas for foreign key based relations like `education <-> school`

```
education: { 
  school_id: '5a207c88aba9cfa1a2f5943f'
}
schools: {
  id: '5a207c88aba9cfa1a2f5943f',
  name: "Massachussets Institute of Technology"
}
```

...the requirement is to be able to associate/disassociate. And that naming makes it super-clear. This could also leave some flexibility for some people who might need even more special types of persistence (I'm quite often struggling because people designed their libraries for SQL and there is little support to extend the behavior to support more fashioned types of persistence like embedded data, etc.)

Actually in some cases of nested relationships (which is already the case in my example above) it makes sense to keep the `method` even in included data payloads. I have actually almost got to the point where I actually needed that in my application. Here is a [gist](https://gist.github.com/Startouf/7d6aaa2a616d4f1355afddc618c09ae1) of what I'm thinking about 

And in those cases of `associate/disassociate`, **yes** it totally makes sense that Patch be able to updated those nested/embedded resources.

My answers to the draft suggestion would be

- Yes use `tempid`
- Use `included`, no need to invent something else
- `tempid`s in each resource as needed, and unicity of `tempid`s on the whole payload
- using `?include` in POST : it should behave like `?include` in normal GET operations IMO. It could be that sideposted resources are not persisted immediately (think of a 202 accepted). But actually I'd rather the spec stayed vague about that at this stage
- location of created sideposted resources in the response : depending on 201, 202, etc. the answer is not really easy. Leave it vague ?

Regarding linkage graphs, @ethanresnick made some good points. I haven't got anything else to say / haven't though a lot yet

- for PATCH + allow updates/destroy of related resources, this makes a lot of sense for embedded data. yes for me

- Bulk creation/deletions : I would like the json:api specs to go that way, but this feels quite too much already (usually bulk writes means there could be a top-level document (even if something entirely artificial) grouping sub-documents. Also this is outside the scope of `sideposting` ? (but naturally if json:api is extended to bulk writes and we have sideposting, we should have bulk-sideposting :-) )

As a final remark, transaction mechanisms are hard to enforce with noSQL databases like MongoDB especially when denormalization is involved. On some of cases it would be okay to have errors on some sideposts but still write the main resource. When reading the jsonapi spec I had a thought this could be a possibility "A server MAY choose to stop processing as soon as a problem is encountered, or it MAY continue processing and encounter multiple problems" but I understand it's probably too hard to make a proper spec without the advantages brought by enforcing transactions... But if still following this idea, it would have made sense that a jsonapi sidepost requests would be able to return both `data` and `errors` in the main payload : `data` would indicate the main resource was persisted and sideposting errors could be put in the `errors` key, and using relationship + attribute pointers to indicate which sideposts were not successful.

richmolj:
> Supporting delete = bigger scope = more to find consensus on = who the hell knows when this will ship.

@ethanresnick I could not be more sympathetic to this point ❤️ . I understand completely and agree - if we get stuck on this point, let's not hold up this PR. I hate that I'm even eating up comment space. That said - I've actually not heard any disagreement or downsides yet, and this was in scope for [the prior sideposting issue that got 30 upvotes](https://github.com/json-api/json-api/issues/1089) as well as the original scope of https://github.com/json-api/json-api/issues/795. We have something simple that enables more functionality and is already working for many real-world applications - let's at least ensure we have disagreement here before deciding it shouldn't be part of the debate.

Does anyone disagree with these points?:

* Conceptually, sideposting should support `DELETE`.
  * To disagree with this, your opinion would be that sideposting should support create/update but use the separate operations API for `DELETE`.
* If sideposting is going to support `DELETE`, the way to do so is to pair the sideposted resource with a verb.
  * I'm open to whatever naming and whitelisting of verbs here. The spec could even be super vague and leave verbs up to server implementation (which might prevent bogging this issue down with debate).

I would be surprised if there's disagreement on the first bullet-point. The second might be trickier, but until it bogs us down I believe it should remain in scope.

> I'm not proposing this switching logic for the minimal version.

Thanks for this note. So to clarify - this PR would actually only support `create`, but NOT support `update` or `delete`. Is that your position? So we would support web forms, but only for the initial create...? I sympathize with wanting to trim scope to ship something, but that doesn't reach the level of MVP in my opinion.

Regarding your specific points on `DELETE`: 

* One of the primary users of jsonapi is ember-data, which [has already been making decisions based around the concept of 'single request'](https://www.emberjs.com/blog/2015/09/02/ember-data-2-0-released.html#toc_unsaved-deleted-records).
* I have an ETL process sending large amounts of data to my API - separate requests would complicate the client code and slow down my ETL.
* I would have to write + test a separate endpoint just for the delete, versus piggybacking on the same sidepost request.
* Sometimes these *do* need to run in the same transaction, such as "the parent requires at least one child to be valid". For example, given a menu requires a fish of the day, you can only delete the prior fish of the day if you are also adding a new one.
* The `data` replacement is a non-starter for me, as it's immediately not viable with a basic "blog post has many comments" example (I had this problem the second time I wrote a JSONAPI, so it's hard for me to see this as rare).

This is all to say, of course we can come up with workarounds, but I have a hard time seeing these alternatives as *preferable* to a simple resource/verb pairing that's already the solution we use for primary data. My main argument here is "keep things simple".

> That's slightly more likely to run into concurrency issues than having separate updated_/removed_/added_data buckets would, but it's not as though having the separate keys makes you immune and, to fully prevent the problem, you need something like #600 either way.

To be clear, I don't think anyone is proposing separate buckets anymore. The thought is to take this same PR and add `method` (or equivalent) to the relationships payload, [as explained here](https://jsonapi-suite.github.io/jsonapi_suite/concepts#sideposting).

scott-mcdonald:
@richmolj Can you give an example of your proposed _sideposted_ `DELETE` request for my edification?

richmolj:
@scott-mcdonald you can see the full payload + explanation at https://jsonapi-suite.github.io/jsonapi_suite/concepts#sideposting

scott-mcdonald:
@richmolj Thank you! In that example the HTTP verb is `PATCH` and you have a `method` enumeration to specify the _verb_ to perform to make it **explicit** what the client is requesting on the related resource - I like this approach! It is additive to the current specification and in my opinion simple and explicit  - if this PR adopts this fundamental approach having `method` in `relationship` then it makes sense to add all the verbs you demonstrated in your example (`create`, `update`, `destroy`, `disassociate`). Anyways that is my opinion.

Also I advocate for `tid` over `tempid` or `temp-id` if anybody is asking.

Also if the entire request is an atomic **transaction**, can we loosen the primary data constraint of it being a single object to being an array of objects for bulk sideposting? Or am I reaching for the stars...

_BTW in your example you stated it is creating two related tags, but it looks like only one tag is in the json:api document._

richmolj:
>  can we loosen the primary data constraint of it being a single object to being an array of objects for bulk sideposting?

@scott-mcdonald though I certainly support this as a separate issue, going with the `method` approach would allow your primary data to be a dummy `entry` resource, then sidepost onto it. So the functionality would be enabled if not ideal.

scott-mcdonald:
@richmolj I think my use of `bulk sideposting` was not the best word choice. I simple meant the primary `data` can be an array of resources instead of a single resource for _bulk change_ as an atomic operation which isn't what this PR is about.

I am also wondering if this `method` approach is essentially the same as what the owners are calling `operations` and thus will be solved by that proposal? I'll guess we will find out soon, maybe...

sandstrom:
There is now a proposal for local ids, `lid`s, that will make side-posting easier.

If you like this to be implemented, please support it here: https://github.com/json-api/json-api/pull/1244

pcrglennon:
Have there been any recent updates on this or Operations (#1254)? It's not clear (at least, to someone new to all this) whether this is still being considered, or whether Operations would obviate the need for sideposting entirely.

simonc:
Hi there 😊 I would also be interested in knowing were we're at regarding this subject. Is work stil happening on this front? Anyway we can help? ❤️

sandstrom:
@simonc It's still in the plans. The next step is to add `lid` (local-id) to the spec, in 1.2.

You can follow along here:
https://github.com/json-api/json-api/pull/1244

(I know it's being actively worked on)

Shizzen83:
Hi there :)

First of all, thank you very much about your hard work which created that awesome spec. I discovered it a few months ago and I'm currently implementing it on Laravel to support both spec versions. That's why I wondered if there was some recent news about bulk operations.

Thanks in advance!

auvipy:
how does the atomic operation extension influence this?

GerritSe:
Is there any progress being made regarding this feature? Is there an active discussion group somewhere else?

freddrake:
How does the [Atomic Operations extension](https://jsonapi.org/ext/atomic/) not solve the problem sideposting was intended to achieve?  That extension is planned for specification version 1.1.

jelhan:
> How does the [Atomic Operations extension](https://jsonapi.org/ext/atomic/) not solve the problem sideposting was intended to achieve? That extension is planned for specification version 1.1.

I think it fully does.

It should be possible to implement sideposting as proposed here as an extension, but I wouldn't be sure why we would need a second solution for the same problem.

I think this proposal was valuable input. Exploring this option helped to iterate on the problem, which result in Atomic Operations. This merge request could be closed now in my opinion.

@beauby Do you have any concerns closing it? Does someone else has?

sandstrom:
I agree that the atomic operations extension could be used to accomplish the same thing. 

But I still think sideposting would be valuable for the following reasons:

1. Sideposting is easier to implement in client- and server-side libraries, due to its much smaller scope.
2. With atomic operations, all requests would often go through one endpoint, e.g. `/operations`. This breaks commonly used app performance monitoring systems, that use the endpoint name to separate out different resource types. 
    
    For example, Skylight, that we're using, will rank endpoints by performance out of the box. Would need extra work for it to analyze a single-path API.
3. With atomic operations, when all requests to be POSTs to `/operations`, you cannot use many of the inherent assumptions about GET, POST, DELETE, etc.
    
    For example, just looking at the method, you don't know if it's safe to retry the request. Similarly, our client is configured to send GET requests to replica slave nodes if no POST/PUT/DELETE has been sent in the past 10 seconds. Such logic would need to be customized.
    
    We're also using the method type for a bunch of other internal logic, that wouldn't work with operations extension using only POST.

*If I'd have to pick one, I'd pick proper sideposting over atomic operations.*

***

One example from our application is the `Invoice` model, with one or more embedded `InvoiceItem` models. These invoice items are embedded at the database level, so makes sense to create/update/delete them via the parent invoice. 

In this case, sideposting would be very useful.

```http
PUT /invoices/123

{
  "data": { 
    "id": "123", 
    "type": "invoices", 
    "attributes": { "invoice_date": "2022-01-15", "amount" : 250 },
    "relationships": {
      "invoice_items": {
        "data": [{ "type": "invoice_items", "id": "14" }, { "type": "invoice_items", "lid": "a1" }]
      }
    }
  },
  "included": [
    {
      "type": "invoice_items",
      "id": "14",
      "attributes": { "label": "Tax", "amount": 50 }
    },
    {
      "type": "invoice_items",
      "lid": "a1",
      "attributes": { "label": "Hammer", "amount": 200 }
    }
  ]
}
```

bradjones1:
`lid` is in 1.1, so I think this needs updated to use that instead of `temp-id`? https://github.com/json-api/json-api/pull/1244

jugaadi:
> With atomic operations, we're forcing all requests to go through one endpoint

@sandstrom Is it necessary?

My understanding is that atomic operations requires all requests to use POST and mention the extension name in the Content Type itself. Not sure if its limited to a single end-point.

cc: @dgeb

beauby:
Closing this as I am not actively working on it, and a lot has changed since I stopped.

PR #1198, opened 2017-07-26
Added php implementations

pmarien:
Added client implementation enm/json-api-client and server implementations enm/json-api-server and enm/json-api-server-bundle


steveklabnik:
Thanks!

Issue #683, opened 2015-05-27
Identify Errors by URI


ethanresnick:
Before I write up a PR for this, I want to see if there's interest. The idea is to identify particular error cases described by the spec with a URI, which would live at the `href` property in the returned error object. This would have a few advantages, as I see it: 
1. It would allow clients to distinguish between multiple errors that return the same status code. For example, a 403 Forbidden can occur on a request to replace a to-many relationship either because full replacement isn't allowed, in which case the client might be able to reformat its request successfully, or because the client simply isn't authorized to patch the resource, in which case it should give up. The same ambiguity can occur for a POST with a client-generated id, that is: the 403 could be coming because of a general authorization error, which the client can't do anything about, or because client-generated ids are unsupported, which it may be able to adjust to. As I alluded to in #673, there will also likely be different-but-same-status-code errors around extension negotiation that it would be valuable for clients to be able to distinguish as well. For example, the difference between extension negotiation not being supported at all vs. a particular combination of extensions not being supported.
2. It would be a usability win, in that each error uri could actually be dereferenceable to a page at jsonapi.org that has a human-readable description of the error and suggestions for fixing it. Angular does this for its errors and its incredibly helpful.

To get the full benefits of (1) for clients, I think this would ideally be a **MUST** for 1.0 (at least for certain errors), since it doesn't impose any real implementation burden on servers. That said, these URIs could be added as a SHOULD post 1.0, with almost the same value.

dgeb:
I think that we're going to have to accept SHOULDs post-1.0 as the most stringent possible requirements, given the lack of time available and experience with real world implementations in this regard. It would probably be best to start with recommendations and then work from there.

ethanresnick:
@dgeb Sounds good! I'll let this sit until after the 28th and we'll revisit it then.

hhware:
@ethanresnick, going one step further, could `links` in error objects be considered for this purpose?

ethanresnick:
@hhware Yes, it would be reasonable and more consistent to say `"links": {"describedby": "..."}` from within the error object in place of `"href": "..."`. But that would be a (small) breaking change. @dgeb @tkellen Would you be in favor of such an adjustment?

hhware:
@ethanresnick, indeed, I somehow missed that the spec already has `href` in `errors`! In this case, I believe it should be converted to `links`. Submitted #690 for that.

ethanresnick:
[...]

PR #1199, opened 2017-07-26
Error object: add type link

ethanresnick:
A minimal PR based on my latest comment in #683. See that first for some caveats on this. Otherwise, I think it's good to merge.

@jmeas, Github won't let me request a review from you for some reason, but you're welcome to weigh in on this too.

If anyone else has thoughts, feel free to comment here or in #683 too.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -504,10 +504,10 @@ For example:
 ### <a href="#document-links" id="document-links" class="headerlink"></a> Links
 
 Where specified, a `links` member can be used to represent links. The value
-of each `links` member **MUST** be an object (a "links object").
+of this member **MUST** be an object (a "links object").
 
-Each member of a links object is a "link". A link **MUST** be represented as
-either:
+<a href="#document-links-link" id="document-links-link"></a>
+Within this object, a link **MUST** be represented as either:
 
 * a string containing the link's URL.
 * <a id="document-links-link-object"></a>an object ("link object") which can 
@@ -516,7 +516,11 @@ either:
   * `meta`: a meta object containing non-standard meta-information about the
     link.
 
-The following `self` link is simply a URL:
+Except for the `type` key in an [error object]'s links object, each key present
+in a links object  **MUST** have a single link as its value. The  `type` key, 
+if present, **MAY** hold an array of links.
+
+For example, The following `self` link is simply a URL:
 
 ```json
 "links": {
@@ -540,8 +544,7 @@ about a related resource collection:
 
 > Note: Additional members may be specified for links objects and link
 objects in the future. It is also possible that the allowed values of
-additional members will be expanded (e.g. a `collection` link may support an
-array of values, whereas a `self` link does not).
+additional members will be further expanded.
 
 ### <a href="#document-jsonapi-object" id="document-jsonapi-object" class="headerlink"></a> JSON API Object
 
@@ -1801,11 +1804,12 @@ An error object **MAY** have the following members:
 
 * `id`: a unique identifier for this particular occurrence of the problem.
 * `links`: a [links object][links] containing the following members:
-  * `about`: a [link][links] that leads to further details about this
-    particular occurrence of the problem.
-  * `type`: a [link][links] whose URI that identifies the type of error that 
-     this particular error is an instance of. When dereferenced, this URI 
-     **SHOULD** provide a human-readable explanation of the general error.
+  * `about`: a [link][link] that leads to further details about this
+    particular occurrence of the problem. When derefenced, this URI **SHOULD**
+    return a human-readable description of the error.
+  * `type`: a [link][link] or array of links that identifies the type of error 
+    that this particular error is an instance of. This URI **SHOULD** be 
+    dereferencable to a human-readable explanation of the general error.
 * `status`: the HTTP status code applicable to this problem, expressed as a
   string value.
 * `code`: an application-specific error code, expressed as a string value.
@@ -1837,6 +1841,8 @@ An error object **MAY** have the following members:
 [compound document]: #document-compound-documents
 [meta]: #document-meta
 [links]: #document-links
+[link]: #document-links-link
+[error object]: #error-objects
 [error details]: #errors
 [member names]: #document-member-names
 [pagination]: #fetching-pagination
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -2303,7 +2303,7 @@ An error object **MAY** have the following members:
   * `about`: a [link][link] that leads to further details about this
     particular occurrence of the problem. When derefenced, this URI **SHOULD**
     return a human-readable description of the error.
-  * `type`: a [link][link] or array of links that identifies the type of error 
+  * `type`: an array of [links][link] that identify the type of error 
     that this particular error is an instance of. This URI **SHOULD** be 
     dereferencable to a human-readable explanation of the general error.
 * `status`: the HTTP status code applicable to this problem, expressed as a
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1803,6 +1803,9 @@ An error object **MAY** have the following members:
 * `links`: a [links object][links] containing the following members:
   * `about`: a [link][links] that leads to further details about this
     particular occurrence of the problem.
+  * `type`: a [link][links] whose URI that identifies the type of error that 
+     this particular error is an instance of. When dereferenced, this URI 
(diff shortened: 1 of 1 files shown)


jamesplease:
Thanks for putting this together, @ethanresnick! I think adding something like this to the errors link object makes a lot of sense. I'm not as optimistic that this could obviate the need for the `code` property, but there's no question that this will nicely complement the `code`.

One concern I have is that the difference between the `about` and `type` links may come across as too subtle right now. Their similarity may make it hard for implementers to use them confidently. 

The differences, according to this PR, are:

- `about` references more information about the _specific_ occurrence of the error,`type` includes more information about the _type_ of error
- `type` must be human-readable

Limiting the conversation to just human-readable links, I've seen two types of links that I might consider using in my errors. One is just:

`thing.com/errors/the-problem-you're-facing`

These sorts of errors are usually human-readable, yet generic, in that the same message is displayed for all occurrences of the issue. So based on my reading of the spec, these fall under `type`. Another type of message I've seen is:

`thing.com/errors/the-problem-you're-facing?component=Table&method=doSomething`

This is a templated human-readable description of the problem that uses the query params to provide a detail specific to the occurrence of the problem. The fact that this is human-readable suggests that this might be a `type`, but considering that it's been tailored to the specific occurrence of the problem, it seems that this is better suited as an `about` link.

I may be misunderstanding the difference, or perhaps we would want both of these to fall under `type` links?

Either way, I wonder if we can emphasize the difference between these two links more in the spec. Perhaps through examples, or comparing/contrasting the two?

_format/1.1/index.md
```diff
@@ -1803,6 +1803,9 @@ An error object **MAY** have the following members:
 * `links`: a [links object][links] containing the following members:
   * `about`: a [link][links] that leads to further details about this
     particular occurrence of the problem.
+  * `type`: a [link][links] whose URI that identifies the type of error that
```
dgeb: > whose URI that identifies

Use of `whose` + `that` is a little awkward. Could we just strike "whose URI" entirely?

ethanresnick:
@jmeas @dgeb Hey guys, sorry I haven't responded here (after, I know, being very anxious to make this change happen). I think James raised some really good points that I want to think about more, but I'll try to get back to this ASAP

ethanresnick:
Hey all! Finally following up on this, and hoping we can get it merged.

To James' questions:

> The differences, according to this PR, are:
> - `about` references more information about the specific occurrence of the error, `type` includes more information about the type of error
> - type must be human-readable

The first bullet point is meant to be the only difference. In other words, `type` doesn't *need* to be dereferencable to a human readable description (the PR makes that a SHOULD). Conversely, `about` should point to a human-readable description just like `type`. I've updated the PR to add that "SHOULD" to about, so that it's now clearer that the only difference is: type = class of error, about = instance of error. 

Given that, you'd have:

```json
{ 
  "type": "thing.com/errors/the-problem-you're-facing",
  "about": "thing.com/errors/the-problem-you're-facing?component=Table&method=doSomething"
}
```

I've also updated this PR to have `type` be our first link that can hold an array of links. We're definitely going to need links that can take an array (e.g., for `profile`) and we already called out in 1.0 that we'd add links like this in the future, so I think this is a fully compatible addition that implementations have been adequately warned about:

> Note: Additional members may be specified for links objects and link objects in the future. It is also possible that the allowed values of additional members will be expanded (e.g. a collection link may support an array of values, whereas a self link does not).

ethanresnick:
@dgeb I think we were all on board about this a while back, but I was hesitant to merge it for various reasons. I'd like to merge it now, though, so that profiles can define their own recognizable errors (e.g. "filter using invalid operator") and identify them by unique `type` urls (which would be sub paths of the profile's main url; e.g. `` `${profile_url}/errors/invalid-operator` ``), rather than with `code` values that might conflict. Are you still game to merge this?

_format/1.1/index.md
```diff
@@ -2299,8 +2300,12 @@ An error object **MAY** have the following members:
 
 * `id`: a unique identifier for this particular occurrence of the problem.
 * `links`: a [links object][links] containing the following members:
-  * `about`: a [link][links] that leads to further details about this
-    particular occurrence of the problem.
+  * `about`: a [link][link] that leads to further details about this
+    particular occurrence of the problem. When derefenced, this URI **SHOULD**
+    return a human-readable description of the error.
+  * `type`: a [link][link] or array of links that identifies the type of error
```
dgeb: Why allow an array of links for `type` but not for other links?

ethanresnick: Allowing an array for `about` and other existing links would be a breaking change (probably + unfortunately), but `type` should be an array because an error can obviously be of multiple types, and there's precedent for array-shaped links with `profile`. Ideally, I'd like to have all new links going forward hold an array for consistency

dgeb: I am not sure it's a benefit to allow all links to hold an array - consider the `self` link for instance.

And there is a singular aspect to defining each error in an error object. `code` is singular as are the other links. Furthermore, `type` throughout the spec is used to define a singular type of resource, so it seems a nice parallel to define a singular error link. Having a single link to turn to for info can be a nice simplification.

Perhaps I'm missing something. I'm definitely open to hearing some use cases that make you think an array here would be desirable.

_format/1.1/index.md
```diff
@@ -536,9 +536,10 @@ Within this object, a link **MUST** be represented as either:
     link.
   * Any link-specific target attributes described below.
 
-Except for the `profile` key, each key present in a links object **MUST** have
-a single link as its value. The `profile` key, if present, **MUST** hold an
-array of links.
+Except for the `profile` key in the top-level links object and the `type` 
+key in an [error object]'s links object, each key present in a links object 
+**MUST** have a single link as its value. The aforementioned `profile` and
+`type` keys, if present, **MUST** hold an array of links.
```
dgeb, line 542: This phrasing has gotten a bit awkward, and is not entirely consistent with error `type` (which can contain a singular link or an array). I'd prefer defining `profile`'s requirements separately and then using general wording such as "Except as explicitly noted elsewhere in this specification, each key present in a links object **MUST** have a single link as its value."

_format/1.1/index.md
dgeb: Having discussed this a bit more with @ethanresnick, I am now ok with allowing a singular link or array of links here. The array could be used for further refinements of a `type` - from general to specific. This could allow error objects in an API to evolve while still supporting old clients looking for a particular link.

_format/1.1/index.md
ethanresnick: Thanks for catching this! I totally didn't realize that I was saying `type` could be a single link in one place, but that it had to be an array in another place.

If we allow `type` to be a singular link or an array, we end with two formats that clients have to support (array and object), which is more total complexity than just having one format, for no real benefit. So I'd prefer requiring `type` to be an array since, of course, a server can always send an array of one item.

dgeb: Yeah, that's fine by me - and is consistent with `profile`. I suppose when we allow an array, we should just require it for simplicity.

Just to reiterate: I really don't want all new links we introduce to be arrays by default - sometimes there's value in a single canonical link (like `self`)

ethanresnick: > Yeah, that's fine by me - and is consistent with `profile`.

Awesome.

> I suppose when we allow an array, we should just require it for simplicity... Just to reiterate: I really don't want all new links we introduce to be arrays by default - sometimes there's value in a single canonical link (like self)

Yup, agreed. That is, if array is allowed, it's required, but there could be links where an array isn't allowed (like `self`). 

I think the number of link types that are truly singular is pretty small, tbh. So, if we were starting from scratch, I'd say to have only one format (arrays), and then enforce that certain link types (`self`) can only hold one link through a normative MUST. But that ship has sailed, and there's (probably) not an elegant, backwards-compatible way to get there.

wimleers:
I think it'd be great to have this crystal clear example from 
https://github.com/json-api/json-api/pull/1199#issuecomment-375452664 in the spec itself:
```
{ 
  "type": "thing.com/errors/the-problem-you're-facing",
  "about": "thing.com/errors/the-problem-you're-facing?component=Table&method=doSomething"
}
```

ethanresnick:
@wimleers Sure; please open a PR! Although note that type is now an array, so it'd be:

```
{ 
  "type": ["thing.com/errors/the-problem-you're-facing"],
  "about": "thing.com/errors/the-problem-you're-facing?component=Table&method=doSomething"
}
```

PR #1201, opened 2017-07-31
Updated implementations index

DavidDuwaer:
I've built this library that can be included in the list. I've put it after Devour in the list because its usage syntax is somewhat similar.


steveklabnik:
Thanks!

> I've put it after Devour in the list because its usage syntax is somewhat similar.

In order to not play favorites, we add names to the list chronologically right now; could you move this to the bottom? Thanks!

DavidDuwaer:
Ah sorry, I'll move it down.

steveklabnik:
No worries, you couldn't have known. Thanks!

PR #1202, opened 2017-08-02
Remove RestPack::Serializer from implementations

beauby:
Remove `RestPack::Serializer` from implementations as it is unmaintained and non-conformant to the spec.
Refs: https://github.com/RestPack/restpack_serializer/issues/128, https://github.com/RestPack/restpack_serializer/issues/123


steveklabnik:
Thanks!

PR #1203, opened 2017-08-02
Remove RABL from implementations

beauby:
Remove RABL from implementations as the page about "emitting conforming JSON" is non-conformant to the 1.0 spec.

### DIFF
--- a/implementations/index.md
+++ b/implementations/index.md
@@ -159,8 +159,7 @@ empowers your JSONAPI compliant [Rails](http://rubyonrails.org/) APIs. Implement
 * [ActiveModel::Serializers](https://github.com/rails-api/active_model_serializers)
 is one of the original exemplar implementations, but is slightly out of date at
 the moment.
-* [The rabl wiki](https://github.com/nesquena/rabl/wiki/Conforming-to-jsonapi.org-format)
-has a page describing how to emit conformant JSON.
+* [RestPack::Serializer](https://github.com/RestPack/restpack_serializer) implements the read elements of json-api. It also supports paging and side-loading.
 * [Oat](https://github.com/ismasan/oat#adapters) ships with a JSON API adapter.
 * [JSONAPI::Resources](https://github.com/cerebris/jsonapi-resources) provides a complete framework for developing a JSON API server. It is designed to work with Rails, and provides routes, controllers, and serializers.
 * [Yaks](https://github.com/plexus/yaks) Library for building hypermedia APIs, contains a JSON API output format.


ethanresnick:
Maybe we should open an issue on these respective repos notifying the maintainer before we remove them?

beauby:
@ethanresnick Sure, I actually just opened PRs to remove implementations that were clearly far away from the spec for a long time, as I believe it adds a lot of confusion to newcomers.

Incidentally, I started working on a simple test suite with a bunch of JSON API document exhibiting support for certain parts of the spec so that library maintainers can somehow "prove compliance" (probably with a score, rather than a binary outcome).

implementations/index.md
```diff
@@ -159,8 +159,7 @@ empowers your JSONAPI compliant [Rails](http://rubyonrails.org/) APIs. Implement
 * [ActiveModel::Serializers](https://github.com/rails-api/active_model_serializers)
 is one of the original exemplar implementations, but is slightly out of date at
 the moment.
-* [The rabl wiki](https://github.com/nesquena/rabl/wiki/Conforming-to-jsonapi.org-format)
-has a page describing how to emit conformant JSON.
+* [RestPack::Serializer](https://github.com/RestPack/restpack_serializer) implements the read elements of json-api. It also supports paging and side-loading.
```
steveklabnik: this seems to be adding one too?

beauby: fixed

ethanresnick:
@beauby yup, I definitely think these should be removed too. I just want to make sure we're giving the owners a heads up. I noticed the other PRs you opened already refenced issues on the maintainers' repo, so this concern really only applies here.

Re a compliance checklist/test suite/score: I think that's a wonderful idea as well. I'm excited to see what you make! I also recall @hhware proposed something similar way back in #733, so taking a look at that might offer some ideas for extra things to include

beauby:
@ethanresnick Just made a PoC for the compliance test:

It's a little gem called `jsonapi-grader`, which ships with a binary of the same name. Source code is available [here](https://github.com/beauby/jsonapi-grader). 
The idea is pretty simple: we have a bunch of [scenarii](https://github.com/beauby/jsonapi-grader/tree/master/scenarii), which are just JSON API documents, and the goal is that each implementation provides an executable to generate each of those documents. 
The grader then simply loops through the scenarii (each of which is attributed a score, and a boolean stating whether it is compulsory for compliance, see [this manifest](https://github.com/beauby/jsonapi-grader/tree/master/scenarii.json)), and grades each of them (if implemented). It then outputs something along the lines of:
> $ jsonapi-grader ./jsonapi-rb/compliance/
Running scenario null_data... passed
Running scenario empty_collection... passed
Running scenario simple_resource... passed
Running scenario simple_resource_attributes... passed
Running scenario simple_resource_meta... passed
Running scenario simple_resource_jsonapi_object... passed
Compliance: true
Score: 420 / 420 (100.0%)

For an example of the work library implementers would have to provide to ensure compliance, see [here](https://github.com/jsonapi-rb/jsonapi-rb/pull/3/files).

PR #1206, opened 2017-08-02
Remove outdated RestPack::Serializer examples.

beauby:
I suspect Endpoints' examples are outdated too.


steveklabnik:
I believe that endpoints should be compatible, but am not 100% sure (I just asked one of the original authors)

PR #1210, opened 2017-08-04
Reorganize ruby implems by framework.

beauby:


dgeb:
These subcategories seem useful to me 👍

ethanresnick:
Yeah, I think this is an improvement too. Thanks @beauby! Longer term, I think we should definitely come up with a new way to organize all implementations to help users find good ones that are appropriate for their project.

PR #1212, opened 2017-08-23
update link to crnk.io implementation

remmeier:
move from https to http as github pages does not properly support https.


steveklabnik:
Thanks!

Issue #1209, opened 2017-08-04
Compliance test suite


beauby:
## The issue
This topic has been brushed a few times already (#1098, #892, #733). The idea is to provide a test suite so that library implementers can 1. prove compliance to the spec and 2. uniformly showcase their abilities.

## The solution
I went ahead and created some basic infrastructure: [jsonapi-grader](https://github.com/beauby/jsonapi-grader), which is mainly a bunch of request specs. 

The idea with the validation of server libraries is that each maintainer would set up a service (on a free heroku instance or similar) against which users could run the test suite (and possibly the jsonapi.org maintainers would periodically run it as well). The test scenarii would possibly be documented on the jsonapi.org website.

## Making it happen
Sounds interesting? Get involved and help build the actual test suite (suggesting scenarii and/or implementing them).

beauby:
This could be a helpful reference: https://json-ld.org/test-suite/

evolve2k:
Since this is the new home for this discussion I'll bring my recommendation here from https://github.com/json-api/json-api/issues/1098

I'm proposing we use '**Author > Books > Reviews**' as the basis of an example app.

Lets make it so that **all resources in the domain have names that are very easy to grok** and can not be in any way misunderstood for code related terms.

beauby:
@evolve2k Thanks for the input. I agree that we should keep the test scenarii and the tested resources/endpoints simple to reason about. 
However, I believe it would be easier for implementers that each scenario tests a **specific thing** (which is not opposed to having resources/endpoints named as you suggest, but I'd like to avoid having 3 endpoints that have to comply to many different scenarii at the same time). 
One possible solution would be to have each test scenario hit namespaced endpoints (e.g. for the first test `http://compliance.example.com/test1/books`).

fotinakis:
One thing I think should be in there: a model with a two-word name, like `CustomerReviews`, so that we can test serializing the type to `customer-reviews`.

jaredcnance:
@fotinakis I'm not certain we should be concerned with the casing choice of the Server since, although it is a recommendation, is not required. For example, an implementation _may_ only support camelCased or snake_cased names. Compound resource names introduce, what I would consider to be, an unnecessary complexity. It makes result validation a bit more complicated since now we need to check the keys of the returned document against several different casing options. But, I also see the value in being able to state what casing mode the Server uses by default. If we did use compound names, I believe it should be isolated to a single test, validating ONLY that.

ethanresnick:
[...]

PR #1214, opened 2017-08-29
[RFC] Compliance specs.

beauby:
This document lists a bunch of test cases for ensuring compliance of implementations. Each test case will be an HTTP request/response.

Ref: #1209.


jelhan:
I like the idea to provide a reference test set for library authors. Sadly it seems that it became stale. I'm going to close this PR. Let's pick up the idea in #1209.

PR #1217, opened 2017-09-06
Update description of JsonApiFramework client and server side implementations to latest version

scott-mcdonald:
Signed-off-by: Scott McDonald <scott@quantumsoft.com>


steveklabnik:
Thanks!

PR #1222, opened 2017-09-26
Update Implementation description

DavidDuwaer:
Updated description of Coloquent


dgeb:
Thanks for the update!

PR #1225, opened 2017-10-12
Add elixir client implementation

bundacia:
Adding a new elixir client library for JSON API that we're developing at Decisiv.


steveklabnik (approved):
Thanks a ton!

Issue #1223, opened 2017-10-10
Clarify asynchronous-processing recommendations


TrentBartlem:
The [recommendations ](http://jsonapi.org/recommendations/#asynchronous-processing) show example response for creating an async task, and what to return when the task completes, but no hints about what the response should be in the interim, while the task is still pending.

benoittgt:
What do you think should be a good idea for this?

TrentBartlem:
One option is a HTTP 202 Accepted response with a body or header to provide guidance to the client (e.g `Retry-After` to suggest when next to check, or `Last-Modified` to enable the option of 304 Not Modified during subsequent requests). If you're following HATEOS principles then a link to cancel the request can also go in the body.

Alternatively, HTTP 200, which is also arguably correct, since the response _is_ okay. The sources I've read have been a bit split on this issue.

dopin:
@TrentBartlem I would add `Location` to let user know where to get next time.

TrentBartlem:
@dopin The location header is already on that page;  once a real response is available a 303 See Other response is returned with a `Location` header referencing the final result.

What's in flux is what the task's resource should return when the async request is still being worked on.

ethanresnick:
@TrentBartlem Wouldn't the response body be the same as the body that came back with the original `202 Accepted`? I.e.:

```
GET /photos/queue-jobs/5234

{
  "data": {
    "type": "queue-jobs",
    "id": "5234",
    "attributes": {
      "status": "Pending request, waiting other process"
    },
    "links": {
      "self": "/photos/queue-jobs/5234"
    }
  }
}
```

Edit: I suppose recommending that `Retry-After` also be included isn't a bad idea. I'm open to that. Recommending `Last-Modified` seems a bit overkill, in that using it (or an ETag) to support 304 applies to all responses, not just async tasks.

TrentBartlem:
Seems like a reasonable http body, @ethanresnick . Just wasn't sure if json-api had an opinion about whether that response would be HTTP 200 or HTTP 202. As I said above, opinions differ so having an example might guide new devs.

(Note that Retry-After only has to-the-second precision so it's not particularly fine-grained. I guess it's okay to set `Retry-After: 0` if your endpoint is okay with more frequent calls)

ethanresnick:
> Just wasn't sure if json-api had an opinion about whether that response would be HTTP 200 or HTTP 202. 

[...]

PR #1226, opened 2017-10-14
Clarify pending async process response #1223

TrentBartlem:
see https://github.com/json-api/json-api/issues/1223


ethanresnick:
👍 Thanks!

PR #1227, opened 2017-10-14
Better multiple related resources request example

bf4:
> GET /articles/1?include=author,comments.author

is confusing since `author` and `comments.author` both reference the `author` resource', 
but with different meaning.

It looks a lot like how

> GET /users/1?include=articles,articles.comments

is identical to

> GET /users/1?include=articles.comments

since full linkage requires intermediate resources to be returned

So, I've changed the the related resources to be more obviously different

### DIFF
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -971,7 +971,7 @@ resources.
 Multiple related resources can be requested in a comma-separated list:
 
 ```http
-GET /articles/1?include=author,comments.author HTTP/1.1
+GET /articles/1?include=ratings,comments.author HTTP/1.1
 Accept: application/vnd.api+json
 ```
 
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -971,7 +971,7 @@ resources.
 Multiple related resources can be requested in a comma-separated list:
 
 ```http
-GET /articles/1?include=ratings,comments.author HTTP/1.1
+GET /articles/1?include=comments.author,ratings HTTP/1.1
 Accept: application/vnd.api+json
 ```
 


_format/1.0/index.md
```diff
@@ -971,7 +971,7 @@ resources.
 Multiple related resources can be requested in a comma-separated list:
 
 ```http
-GET /articles/1?include=author,comments.author HTTP/1.1
+GET /articles/1?include=ratings,comments.author HTTP/1.1
```
bf4: or maybe `author,comments.rating` ¯\_(ツ)_/¯

bf4: It might also be worth noting that the included resources are named by their relationship names, not by their resource types

jelhan: Do you mind switching the order? I think it is easier if the new part (compared to the previous examples) is added at the end.

```suggestion
GET /articles/1?include=comments.author,ratings HTTP/1.1
```

jelhan (changes requested):
Thanks a lot for this proposal. I agree that having `author` twice could be confusing. Additionally I would switch the order. Than it would directly add to the examples given before. In order they would be the following:

```http
GET /articles/1?include=comments HTTP/1.1
GET /articles/1?include=comments.author HTTP/1.1
GET /articles/1?include=comments.author,ratings HTTP/1.1
```

dgeb (approved):
Thanks @bf4 and @jelhan. I can see how this is subtly more clear.

dgeb:
@bf4 / @jelhan would either of you be up for applying this same change to the 1.1 spec in another PR?

jelhan:
> @bf4 / @jelhan would either of you be up for applying this same change to the 1.1 spec in another PR?

Done in  #1629.

PR #1232, opened 2017-10-24
Add raml-json-api to implementations

mblackritter:
[raml-json-api](https://github.com/RJAPI/raml-json-api) looks like a very promising timesaver and sophisticated solution to bootstrap this whole workflow for Laravel, it likely should be listed. 🤗


dgeb:
Great - thanks!

Issue #1230, opened 2017-10-23
Requests with `include` should always return `included`


noetix:
When requesting results with included data, the response should always have the `included` top-level key. The response structure should not change depending on the data sourced to generate the response.

e.g. `GET /todos?include=comments`

The expectation is that the response should have a list of todo items under `data` and any included data under `included`, but depending on whether there comments or not will determine whether `included` as a top-level key exists or not.

As a client trying to use this data I have to first check that the array exists then try find my result.

beauby:
The idea is that you would use come kind of client library to take care of that on the client side. If you do not want to commit to a full-fledged client library because you feel your use-case is too simple for that, you can simply define a helper function that does the checking for you.

noetix:
I do use a client library to facilitate these requests and I could have posted this issue there. However, aside from the fact that the client library author would turn to the spec for guidance -- to which he would receive none -- the issue is about returning a consistent message format.

When my request has specifically asked for included data to be returned, I expect that the `included` top-level member to be there regardless of whether there are any results. Exactly the same as `data` -- we don't optionally include that depending on result set.

seanrkerr:
👍

ethanresnick:
I would be open to a PR mandating that, if the `include` parameter is in use, `included` must be present in the response document (even if empty). Please feel free to submit.

ethanresnick:
I’ve merged #1236 to address this. If anyone objects, we can discuss further.

PR #1236, opened 2017-11-11
Always return "included" where "include" has been requested

noetix:
https://github.com/json-api/json-api/issues/1230

One additional thing that may be worthwhile is reiterating that the `included` param **must** always be an array, and that an empty array should be returned when no additional resources are available to add to the response.

Open to feedback. :)

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -943,6 +943,8 @@ client to customize which related resources should be returned.
 If an endpoint does not support the `include` parameter, it **MUST** respond
 with `400 Bad Request` to any requests that include it.
 
+If an endpoint supports the `include` parameter and a client supplies it, the server **MUST** respond with the `included` section of the [compound document].
+
 If an endpoint supports the `include` parameter and a client supplies it,
 the server **MUST NOT** include unrequested [resource objects] in the `included`
 section of the [compound document].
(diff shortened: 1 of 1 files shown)


ethanresnick:
This is great. Just so we're not repeating the clause "If an endpoint supports the `include` parameter and a client supplies it", maybe we can merge the paragraph you added with the one that follows?

Something like:

> If an endpoint supports the `include` parameter and a client supplies it: 
> 
>    1. The server's response  **MUST** be a [compound document] with an `included` key—even if that `included` key holds an empty array (because the requested relationships are empty). 
>
>   2. This `included` section **MUST NOT** include [resource objects] not requested through the `include` parameter's value.

noetix:
Opted to leave the second list item as-is.

Good call on the repetition.

ethanresnick:
> The server **MUST** respond with the `included` section of the [compound document] — even if

I realize I'm really nitpicking here, and it's probably more important to just get this merged, but something about the wording quoted above bothers me. Specifically, I think having an `included` section is part of what it means to be a compound document (i.e., you can't be a compound document without one). 

Your wording seems to suggest, though, that the response is already a compound document, which it isn't until the `included` section is added. That's why, in my wording, I tried to say that the the response must *be* a compound document. Then, I added "with an `included` key"—even though that's technically redundant—just to maximize readability in case the full implications of "being a compound document" weren't obvious.

Can we stick with my wording for the first bullet, and then I'll get this merged?

noetix:
@ethanresnick Agreed.

ethanresnick:
Great. Merged!

Issue #887, opened 2015-09-17
Spec contradiction: `null` as link value


ethanresnick:
Hey look, our first erratum! :tada: 
### The contradiction

In the Links section, we say:

> Each member of a links object is a "link". A link MUST be represented as either:
> - a string containing the link's URL.
> - an object ("link object") which can contain the following members...

But then, in the pagination section we say:

> [Pagination keys] MUST either be omitted or have a `null` value to indicate that a particular link is unavailable.

This is a contradiction because the pagination keys are clearly members of the links object, so, by the first quote, `null` shouldn't be an allowed value.
### Goals for a solution
1. (imo) The spec should continue to provide a mechanism for a server to affirmatively assert that a link (i.e. a resource with a particular relation) does not exist. This is a stronger statement than the server just omitting a link, which only tells the client that it might or might not exist. This mechanism should work with as many links as possible.
2. Uniform processing rules. If some links need to be processed specially, I think that'll lead to pain and WTF moments for implementors/users.
3. Backwards compatibility. This has two parts: 1) not violating any of our formal compatibility promises and 2) not breaking things in reality. Personally, I think we should focus more on whether we're breaking things in reality than in theory. (My reference for this is javascript, which, with the introduction of the `let` keyword and destructuring syntax actually changed the meaning of statements like `let[i] = [1];` from being an assignment to an array index to a destructuring assignment. This change was made because the spec committee scanned JS code online and found that it broke basically nothing in practice.) Then again, I also understand the importance of us keeping our formal commitments.

So, with those goals in mind...
#### No longer allow `null` for any links; introduce an alternate "missing link" mechanism".

Here, I'm imagining something like this:

``` js
{
  "links": {
    "self": "http://example.com/people?page=1",
    "next": "http://example.com/people?page=2"
  },
  "missing-links": ["prev"], //before, we'd have had "prev": null above
  "data": []
}
```

Analysis:
- [x] Offers a mechanism for indicating a missing link: yes.
- [x] Creates uniform processing rules: yes.
- [ ] Satisfies backwards compatibility: sort of. I think it satisfies our formal commitments to backwards compatibility, because we can't be accused of breaking a behavior that was specified contradictorily in the first place. Such a behavior should be considered undefined, and now we're just defining it. However, while this satisfies our formal BC commitments imo, it might break some real world implementations that are already using null in their pagination links.
#### Allow `null` for all links except the few where it's currently prohibited

Basically, we'd say something like:

> ... A link MUST be represented as either:
[...]

PR #1243, opened 2017-12-08
Schema should allow pagination links to be null

DamirSvrtan:
I have cases where my pagination links `next` and `prev` are null. The current schema does not allow that, the proposed change does.

### DIFF
--- a/schema
+++ b/schema
@@ -37,7 +37,7 @@
         },
         "links": {
           "description": "Link members related to the primary data.",
-          "allOf": [
+          "anyOf": [
             {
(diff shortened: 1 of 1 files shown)


ethanresnick:
Ref https://github.com/json-api/json-api/issues/887

Relequestual:
It's not quite as simple as it might seem.
Suggest closing in favour of #1555

dgeb:
Thanks for weighing in @Relequestual. Closing in favor of your schema update.

PR #1245, opened 2017-12-09
Introduce `v` media type parameter to specify media type version

dgeb:
Require that the `v` media type parameter be present to identify a specific version of the JSON API media type. If this parameter is not present, version 1.0 of the spec must be assumed.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -24,38 +24,55 @@ interpreted as described in RFC 2119
 
 ## <a href="#content-negotiation" id="content-negotiation" class="headerlink"></a> Content Negotiation
 
+### <a href="#content-negotiation-version" id="content-negotiation-version" class="headerlink"></a> Version
+
+The base JSON API media type **MUST** be associated with the media type
+parameter `v` to specify the version of the media type. This document describes
+version 1.1 of the media type. In order for this version of the JSON API
+specification to apply, the `v` media type parameter **MUST** be present and
+have a value of `1.1`.
+
+If the media type parameter is not specified, version 1.0 of this specification
+**MUST** be assumed. This version is described in an earlier version of this
+document.
+
 ### <a href="#content-negotiation-clients" id="content-negotiation-clients" class="headerlink"></a> Client Responsibilities
 
 Clients **MUST** send all JSON API data in request documents with the header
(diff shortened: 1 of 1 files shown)


dgeb:
I've been discussing this approach with @wycats. We agree there's value in a separate PR that allows for _feature_ instead of _version_ based negotiation to allow for comparison / discussion of the two alternatives. I'll start with [local IDs](https://github.com/json-api/json-api/pull/1244) as an example feature. PR coming soon ...

ethanresnick:
I 100% agree that feature negotiation makes more sense than numeric version negotiation (which just seems like an anti-pattern). That said, even feature negotiation might be an anti-pattern, compared to just producing documents with new features in a way that will be ignored by old consumers (if the features can be ignored) and using a new payload structure to force an error (if the features can't be ignored). I'm not sure what's better.

I can imagine content negotiation producing clearer error messages, but the "just fail" approach seems to have pros too — it can save servers from having `Vary: Accept`, which destroys caching; we don't have to worry about the header getting huge over time; and its consistent with the current spec, where optional features fail if unsupported without the need for content negotiation.

Regardless, I'd really like us to be on the same page in terms of terminology + the historical attempts at versioning before we try to figure this out. Since JSON:API is part protocol and part document format, idk if there's a perfect historical analogue (AtomPub?), but Javascript and HTTP obviously offer some useful experience. XML, as another document format, also has a model that's relevant and that's influenced my thinking a lot; I described the XML model, and linked to some relevant W3C docs, back in https://github.com/json-api/json-api/issues/915#issuecomment-165319520 [the "Versioning and Extensibility" section], which I'd recommend reading if y'all aren't already familiar with their approach. Also, if you guys are familiar with relevant technologies/versioning approaches beyond the three I mentioned above, let's bring those into the conversation too; I'd love to learn more about their models.

dgeb:
@ethanresnick I've been sitting on this for a while and have come around to agreeing with you that feature negotiation might be an anti-pattern as well and inconsistent with v1.0 of the spec (which of course has plenty of optional "features"). That's why I've held off on the PR I mentioned above. I'm currently doing more research ... hope we can discuss soon.

dgeb:
Closing this as I am no longer in favor of this hard requirement to use post-1.0 features.

Issue #1242, opened 2017-12-02
gRPC Proto + GraphQL Schema Standard


disturbing:
I'm about to build a schema for protobuf by following json-api. The way this is formatted, it's really going to make grpc uber flexible for building apis quickly.

Has anyone done this before? If not, how could we go about getting a standard proto file and graphql schema here and potentially in the 1.1 release.

dgeb:
Sounds interesting. It would be good to get this discussion started over at http://discuss.jsonapi.org/

Issue #850, opened 2015-08-21
Recommendation for dasherizing (kebab-case-ing) namespaced object, such as `Admin::User`


bf4:
JSON API recommends dasherized keys, so an `AdminUser` would be represented by `admin-user`. But, does/should JSON API have an opinion on namespaced objects? That `Admin::User` would be dasherized as `admin/user`?  (That's the result in Rails of `"Admin::User".underscore.dasherize`)

joaomdmoura:
+1

bf4:
I suppose this would be a PR to http://jsonapi.org/recommendations/#naming ?

bf4:
Do I understand correctly from the spec that `SOLIDUS, "/"` is disallowed in a member name, so I'm thinking maybe it has to be `_`?

Spec recommendations http://jsonapi.org/recommendations/#naming

> Member names SHOULD contain only the characters "a-z", "0-9", and the hyphen minus  ("-") as separator between multiple words.

Spec Rules http://jsonapi.org/format/#document-member-names, basically:

> A-Z, a-z, 0-9, -, _,  SPACE (not recommended), any UNICODE character except U+0000 to U+007F (not recommended)
> With some disallowed characters http://jsonapi.org/format/#document-member-names-reserved-characters

ethanresnick:
Yes, the solidus is not allowed and a hyphen, double hyphen, or underscore could be used instead. But I'm hesitant to make any sort of recommendation here, because, given the complexities around hierarchical types, I don't want clients to start inferring any meaning from whatever separator pattern we might recommend.

bf4:
So, this Ember code not per JSON API?

https://github.com/emberjs/data/blob/aeccf38a5193a9fbf763a835588cc4ad315324b7/packages/ember-data/lib/serializers/json-api-serializer.js#L319-L351

>    By default `JSONAPISerializer` follows the format used on the examples of
>    http://jsonapi.org/format and uses dashes as the word separator in the JSON
>    attribute keys.
> 
> import { pluralize, singularize } from 'ember-inflector/lib/system/string';
> 
> var dasherize = Ember.String.dasherize;
> 
>   keyForAttribute: function(key, method) {
>     return dasherize(key);
>   },

Where dasherize is https://github.com/emberjs/ember.js/blob/v2.0.1/packages/ember-runtime/lib/system/string.js#L242-L258

>  Replaces underscores, spaces, or camelCase with dashes.
> 
[...]

Issue #525, opened 2015-03-29
Dasherized names


e-oz:
It's just recommendation, but just for curiosity, please let me know how such names supposed to use in JS? `obj['dash-name']` instead of `obj.camelCase` or `obj.underlined_name`? And do you really use "dasherized" names for objects fields on the server side?

tkellen:
You are correct, dasherized properties in javascript can be accessed as you've shown (e.g. `obj['dashed-name']`) Yes, some (but not all) implementations use dasherized names on the server side.

notslang:
I found this to be pretty strange... since JSON is derived from JS, I would expect it to follow the same camelCase naming that JS uses.

notslang:
Oh, and this is a duplicate of #323

PR #1247, opened 2017-12-25
Proposal to follow Google and Schema.org's JSON Style Guide: Use "camelCase" Naming Convention

garrettmac:
Updating Naming Rules.
<hr>

* JSON <strike>**MUST**</strike> **SHOULD** use camelCase 
* URLs **SHOULD** use kebab-case.

with a emphasize on the first
> JSON <strike>**MUST**</strike> **SHOULD**  use camelCase 
<hr>

Here goes.

**I think it should be changed or dropped.**

I know this is a heavily opinionated topic, but here goes.
The JSON object property naming conventions seem to favor camelCase.

#### [Google JSON Style Guide](https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format) (they follow the [JSON.org](http://json.org) spec):

> * Property names must be camel-cased, ascii strings.

 here are a few more:

####  [W3resource](https://www.w3resource.com/slides/json-style-guide.php)(same thing) state:

> Property names must be camel-cased, ascii strings.

#### [Crockford](http://crockford.com/javascript/code.html#names) state:

> * Most variables and functions should start with a lower case letter.
> * Names should be formed from the 26 upper and lower case letters (A .. Z, a .. z), the 10 digits (0 .. 9) 

#### [Schema.org](https://schema.org/docs/old_extension.html) 

> Properties start with a lower case letter and are also camelCase.

 **`camelCase` in the wild:**
*  [Mozilla Objects and properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects#Objects_and_properties)
* [W3School's [JS JSON](https://www.w3schools.com/js/js_json.asp) and [JS Conventions](https://www.w3schools.com/js/js_conventions.asp)
* [Codecademy](https://www.codecademy.com/courses/javascript-beginner-en-xTAfX/0/1)
* **graphql** spec uses `camelCase` for fields.

# Using `kebab-case`

`kebab-case` notation has collisions (or does not play nicely) with:
* The JavaScript Language (you cannot do `person.first-name`)
* EmberJs Framework
* Schema.org / JSON-LD
* GraphQL / GraphQL Queries? (admittedly haven't done much here)

# Conclusion

**Issues as a result or related to the spec using saying to use `kebab-case`** are as follows:
https://github.com/json-api/json-api/issues/323
https://github.com/json-api/json-api/issues/571
https://github.com/json-api/json-api/issues/191
https://github.com/json-api/json-api/issues/127
https://github.com/json-api/json-api/issues/1242
https://github.com/json-api/json-api/issues/850
https://github.com/json-api/json-api/issues/525
https://github.com/json-api/json-api/pull/341

I'd expect they will continue until this issue is resolved, especially as GraphQL becomes more popular.

I think it's safe to assume `camelCase` won this battle, especially with JSON-LD, schema.org and as GraphQL continues to grow.

I think this spec should it be changed or dropped, curious on your take @dgeb @ethanresnick @tkellen 

# More Resource and Discussion on web

* https://en.wikipedia.org/wiki/Naming_convention_%28programming%29#Multiple-word_identifiers
* https://stackoverflow.com/a/19287394/4642530
* https://github.com/emberjs/data/blob/aeccf38a5193a9fbf763a835588cc4ad315324b7/packages/ember-data/lib/serializers/json-api-serializer.js#L319-L351
* https://github.com/interagent/http-api-design/issues/18
* https://tools.ietf.org/html/rfc3986
* https://ember-cli.com/naming-conventions
* https://github.com/holidayextras/jsonapi-server/issues/206
* https://github.com/kurko/ember-json-api
* https://discuss.codecademy.com/t/object-property-naming-conventions-camelcase-or-under-score/36878/3
* https://discuss.emberjs.com/t/ams-with-jsonapi-underscored-vs-dasherized-types/8930/2
* https://github.com/interagent/http-api-design/issues/18
* https://github.com/imazen/imageflow/issues/99
* https://github.com/kurko/ember-json-api/issues/53

### DIFF
--- a/recommendations/index.md
+++ b/recommendations/index.md
@@ -20,16 +20,14 @@ The allowed and recommended characters for an URL safe naming of members are def
 
 ### JSON  
 
-Naming format should follows the [Google JSON Style Guide](https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format), where:
+It is important to adhere to a defined convention. Recommended property names **SHOULD** conform to the following guidelines:
 
-> Property names must conform to the following guidelines:
+- Property names **SHOULD** be camel-cased, ascii strings.
+- First characters **SHOULD** be "a-z" (U+0061 to U+007A).
+- Property names **SHOULD** be meaningful names with defined semantics.
 
-> * Property names should be meaningful names with defined semantics.
-> * Property names must be camel-cased, ascii strings.
-> * The first character must be a letter, an underscore (_) or a dollar sign ($).
-> * Subsequent characters can be a letter, a digit, an underscore, or a dollar sign.
-> * Reserved JavaScript keywords should be avoided (A list of reserved JavaScript keywords can be found below).
-> * These guidelines mirror the guidelines for naming JavaScript identifiers. This allows JavaScript clients to access properties using dot notation. (for example, result.thisIsAnInstanceVariable). 
+
+These guidelines mirror the guidelines for naming JavaScript identifiers. This allows JavaScript clients to access properties using dot notation.
 
 > Here's an example of an object with one property:
 
--- a/recommendations/index.md
+++ b/recommendations/index.md
@@ -10,10 +10,35 @@ are beyond the scope of the base JSON API specification.
 
 ## <a href="#naming" id="naming" class="headerlink"></a> Naming
 
+
+### URLs
+
 The allowed and recommended characters for an URL safe naming of members are defined in the format spec. To also standardize member names, the following (more restrictive) rules are recommended:
 
 - Member names **SHOULD** start and end with the characters "a-z" (U+0061 to U+007A)
 - Member names **SHOULD** contain only the characters "a-z" (U+0061 to U+007A), "0-9" (U+0030 to U+0039), and the hyphen minus (U+002D HYPHEN-MINUS, "-") as separator between multiple words.
+
+### JSON  
+
+Naming format should follows the [Google JSON Style Guide](https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format), where:
+
+> Property names must conform to the following guidelines:
+
+> * Property names should be meaningful names with defined semantics.
+> * Property names must be camel-cased, ascii strings.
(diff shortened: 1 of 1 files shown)


dgeb:
I appreciate the effort that you put into documenting this issue. I've personally been leaning toward changing the default recommendation for JSON members for a while now. We may have placed too much emphasis on correlating case between URLs and JSON members in forming our current recommendations.

However, we can't use language as strong as `**MUST**`. At this point, camelCase could only be a _recommendation_ (and even if we were pre-1.0, I would still favor keeping this a recommendation instead of a requirement).

I'll discuss this with the other editors and of course the community should chime in.

dgeb:
By the way, I'm strongly opposed to referencing the Google Style Guide in the spec. But I am open to discussing its recommendations as related to JSON:API.

Would you mind reopening this as an issue instead of a PR, and focus the discussion purely on recommending camel-case for JSON members? After all, we already recommend kebab-case for URLs, so that's not even a change you're proposing.

ethanresnick:
> focus the discussion purely on recommending camel-case for JSON members?

I would be 100% in favor of this as well.

garrettmac:
@dgeb  https://github.com/json-api/json-api/pull/1247#issuecomment-362636471

> By the way, I'm strongly opposed to referencing the Google Style Guide in the spec.

Why is that? Just curious?

garrettmac:
@dgeb  https://github.com/json-api/json-api/pull/1247#issuecomment-362636471

> By the way, I'm strongly opposed to referencing the Google Style Guide in the spec.

<strike>Why is that? Just curious?</strike>.

I see what you're saying. I actually did reference the Google Style Guide in the JSON recommends. Good call. Removed it.

garrettmac:
Also moved into an issue https://github.com/json-api/json-api/issues/1255

antonkomarev:
@garrettmac because it's dependency. If Google will decide to change their naming convention, this will means that {json:api} require to follow it too.

rtablada:
I think that this is an area where better delineation of spec vs recommendations and the community around it is essential.

That said right now the documentation is a bit hazy at best that "members" names apply to attributes, relations, urls, and types.

PR #1248, opened 2017-12-30
JSON API Playground link added to implementations

pablorsk:
lets http://jsonapiplayground.reyesoft.com/ and play :ok_hand:


dgeb:
Very cool! Thanks @pablorsk!

PR #1251, opened 2018-02-07
Clarify pagination links in relationship object

### LINKED ISSUE #509: Pagination of to-many relationships is underspecified
There's no clarification or example for this line:

> A link object that represents a to-many relationship MAY also contain pagination links, as described below.

I _think_ that means something like this: 

``` json
// ...
{
  "type": "articles",
  "id": "1",
  "title": "Rails is Omakase",
  "links": {
    "self": "http://example.com/articles/1",
      "comments": {
        "self": "http://example.com/articles/1/links/comments",
        "resource": "http://example.com/articles/1/comments",
        "linkage": [{"type": "comments", "id": "5"}, {"type": "comments", "id": "12"}],
        "next": "http://example.com/articles/1/comments?page[cursor]=XYZCAFE"
      }
  }
}
// ...
```

but I'm not completely sure that's correct. Is that `next` link what the spec intended?

Also: should that `next` link be more like a `self` or a `resource` link for the relationship? I'm actually not completely sure I even know the difference between the two kinds of links in the case of a To-Many relationship :-/

dgeb:
This is a clarification, not a change, so I'm comfortable using MUST here.

I believe this is just a refinement of the [pagination](http://jsonapi.org/format/#fetching-pagination) section which already contains statements such as:

> A server MAY provide links to traverse a paginated data set (“pagination links”).

> Pagination links MUST appear in the links object that corresponds to a collection. 

In other words, the pagination links clearly correspond to the collection. And for relationship objects, the collection is relationship data, not related resources.

Closes #509

### DIFF
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -262,7 +262,9 @@ A "relationship object" **MUST** contain at least one of the following:
   relationship.
 
 A relationship object that represents a to-many relationship **MAY** also contain
-[pagination] links under the `links` member, as described below.
+[pagination] links under the `links` member, as described below. Any
+[pagination] links in a relationship object **MUST** paginate the relationship
+data, not the related resources.
 
 > Note: See [fields] and [member names] for more restrictions on this container.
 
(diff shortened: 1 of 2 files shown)


ethanresnick:
👍 

I'm gonna merge this. I think this addresses #509, though there were also some requests in that issue for us to add more examples, so I'll let you decide whether to close that issue now or add the examples first.

PR #1256, opened 2018-02-12
Move fast_jsonapi to it\'s proper place in the ruby server list

DVG:
Fast JSONApi was incorrectly listed as a client implementation.


dgeb:
Ah, good catch. Thanks @DVG!

PR #1258, opened 2018-02-17
Removed py-jsonapi

AdorablePotato:
I don't maintain the library anymore and there are already (too) many implementations for Python :)


dgeb:
Thanks for helping us keep the list updated!

PR #1268, opened 2018-04-15
Introduce Profiles to v1.1

### LINKED ISSUE #762: Content-Negotiation Clarifications
The content negotiation section could benefit from examples, and "in-layman's-terms" explanation. There's a lot of "with" and "without" around media type parameters that can be confusing if you have no idea what they are.

### LINKED ISSUE #1207: Overview: Extending and describing the spec
Apologies for the broadness of this issue, which overlaps with many other issues
(see #1195 for perhaps the most up to date history and list of issues). However,
I think it's vital that we agree on some fundamentals before moving forward to
accept means to describe and extend usage of the base specification.

---

When viewed from the perspective of conformance, there are effectively
three ways to consider expressing additional semantics for the base
specification:

* **conformant** - Semantics are expressed within the areas of the spec
  explicitly reserved for implementations.

* **additive** - Semantics are assigned to structural or procedural
  elements that are reserved but not specified.

* **breaking** - Structural or procedural elements are assigned semantics that
  conflict with the spec.

Each category is discussed below.

### Breaking extensions

Although the concept of breaking extensions was considered in 2014-2015, there
has been significant pushback (see #614) against this concept. I believe there's
now fairly unanimous agreement from the editors and the community that creating
a new derivative media type would be preferable to allowing breaking extensions in
the spec.

### Additive extensions

Additive extensions would cover allowing additional member(s) in the reserved
areas of the document structure. It's also possible for additive extensions to
dictate how clients and servers should interpret those additional members.

Let's look at a few examples of additive extensions that have been proposed:

* **Bulk posting** - This extension would allow multiple resources to be created
  in a single request. It would expect that top-level `data` be passed in as
  an array, and that servers create all resources or none. Created resources
  would be returned in an array.

* **Side posting** - This extension is being explored in #1197, which
  allows related resources to be created in the `included` array.

* **Additive / subtractive fields** - This extension has been proposed in
  #1176, which allows fields to be specified relative to an understanding
  of a default set of fields per type.

* **Operations** - I'll be proposing this extension in the next several days.
  It will introduce a new `operations` array that can be used to process
  multiple actions, including fetching and mutating resources, serially and
  transactionally.

We need to be _extremely_ careful when discussing additive extensions. If
additive extensions are not curated by the editors of the spec itself, it is
inevitable that changes to the spec will break additive extensions.

A key requirement for additive extensions is that it would be technically
possible for them to eventually become part of the base spec, even if we choose
to keep them as extensions indefinitely. Therefore, additive extensions can
not express any demands on non-normative areas of the spec, such as `meta`
values.

To avoid fragmentation in implementations of the spec, we should not introduce
extensions with largely overlapping concerns. We should ensure that any additive
extensions can work together.

The usage of additive extensions must be negotiated, probably through a media
type parameter. For an overview of how media type parameter negotiation might
occur, see [this deprecated page on extensions](https://github.com/json-api/json-api/blob/9c7a03dbc37f80f6ca81b16d444c960e96dd7a57/extensions/index.md#-extension-negotiation).

### Conformant mechanisms

Mechanisms to describe spec structure and usage that fully conform to the spec
are tightly constrained. These constraints free us (the editors and the
community) from concerns about managing different implementations, ensuring the
reasoning behind each is sound, and restricting overlap.

Even within the constraints of the spec, there's still a lot of flexibility
available to implementations. There can be benefits to codifying and sharing
the choices made by implementations.

Although we do not need to be concerned about how Organization A is
using the `page` parameter, it would be nice if Organization A could share
their methodologies with the broader community so that Organization B might
benefit. And it would be even nicer if tooling were developed that shared these
understandings so that Organization C could benefit without thinking deeply
about the implementation.

This aligns quite well with the concept of a "profile" as described in
[RFC 6906](https://tools.ietf.org/html/rfc6906):

> A media type defines both the semantics and the serialization of a
  specific type of content.  In many cases, media types have some
  built-in extensibility or openness, so that specific instances of the
  media type can layer additional semantics on top of the media type's
  foundation.  In this case, a profile is the appropriate mechanism to
  signal that the original semantics and processing model of the media
  type still apply, but that an additional processing model can be used
  to extract additional semantics.

@ethanresnick has explored these concepts these concepts in detail and has
proposed that these be called "profile extensions", as outlined in
https://github.com/json-api/json-api/pull/1195

Ethan's proposal allows profile extensions to be associated with structural
elements in the document. However, I'm also concerned about allowing for
specification of _any_ unreserved areas of the spec, including structural
elements like attributes, relationships, and meta, but also query params like
`filter`, `page`, etc. It also seems in keeping with the spirit of a profile to
allow the combination of additional semantics in a single profile - including
both usage and structural expectations.

For instance, let's take the `page` example above. Imagine developers want to
share a common usage of offset / limit pagination that uses the `page`
parameter (e.g. `/articles?page[offset]=0&page[limit]=10`). And they want to
allow servers to specify a `page` member in the top-level `meta` with the
`total` member. It seems beneficial that these concepts could be wrapped
together in a single "profile" or "profile extension", so that
implementations can agree on both structure and usage atomically.

Are there other places better suited for the expression of capabilities? For
instance, we have discussed the "home document" proposal, as described in
https://mnot.github.io/I-D/json-home/  Are there other options we have not yet
considered?

----

I'm opening this issue to discuss a number of matters:

* Do we agree on the basic classifications above?

* Should negotiation mechanisms differ for additive vs. conformant
  extensions (i.e. "profiles")?

* What are the appropriate boundaries for a profile extension? Should they
  include structure + usage?

* Are there other standards, such as home documents and schemas, that should
  be considered?

* Is consensus strong enough re: additive extensions that we should consider
  focusing on the extension mechanism before anything profile-related?

I'd like to start as simply and conservatively as possible. But we also need to
think wholistically. We should establish a path by which every choice made by a
particular implementation of the spec can be described, standardized, and
shared.

This is a great time to express your questions, concerns, and opinions about
extending the specification.

dgeb:
_Note: Although this PR has my name attached to it, it represents a lot of hard work by @ethanresnick in particular over the past several years. Ethan, @wycats, and I have closely collaborated to reach a mutually acceptable solution in an attempt to finally push this over the finish line. This may yet be tweaked a little as some last minute adjustments have yet to be reviewed by Ethan and Yehuda. And of course we welcome feedback from the rest of the community._

-----

Profiles are a means to describe additional semantics the JSON API media type, without altering the basic semantics. In other words, a profile describes a fully-compliant, yet more opinionated, implementation of the spec. It is hoped that profiles will unlock new layers of shared conventions.

A profile can essentially turn every MAY or SHOULD in the base spec into a MUST. Profiles might describe how the `filter` or `page` params are used. They might specify that every resource contains a `self` link, or that resources should include timestamps or versions as either attributes or in `meta`.

Profiles are identified by URI, which ideally should dereference to a document that describes the semantics of the profile.

One or more profiles may be associated with the JSON API media type through the `profile` media type parameter. The application of profiles to a particular document can be specified by clients and servers  via the `Content-Type` header. The application of one or more profiles to a response can be requested by a client via the `Accept` header.

In order to _require_ the application of one or more profiles, the profile(s) must be specified with the `profile` query parameter. This allows clients to request a profile or profiles and receive a hard failure if they can't be applied. 

Any structural elements introduced by a profile can be aliased. This allows for better adaptability and composition of profiles. A new profile descriptor object is introduced which allows for declarations of aliases in a particular document.

Closes #1207
Closes #1195 
Closes #762

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1859,7 +1859,7 @@ add new features over time.
 
 A profile **SHOULD** explicitly declare "keywords" for any elements that it
 introduces to the document structure. If a profile does not explicitly declare a
-keyword for an element, then the name of the element itself can be considered a
+keyword for an element, then the name of the element itself is considered to be its
 keyword. Keywords **MAY** be aliased in any representation through the use of
 [profile descriptors](profile-descriptors), as described below.
 
@@ -1873,9 +1873,9 @@ parameter **MUST** equal a whitespace-separated list of profile URIs.
 > specification requires that its value be surrounded by quotation marks 
 > (U+0022 QUOTATION MARK, "\"") if it contains more than one URI.
 
-Clients and servers **SHOULD** use the `profile` media type parameter in
-conjunction with the JSON API media type in a `Content-Type` header to specify
-the application of one or more profiles to a JSON API document.
+Clients and servers **MUST** include the `profile` media type parameter in
+conjunction with the JSON API media type in a `Content-Type` header to indicate
+that they have applied one or more profiles to a JSON API document.
 
 A client **MAY** use the `profile` media type parameter in conjunction with the
 JSON API media type in an `Accept` header to _request_, but not _require_, that
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1869,6 +1869,10 @@ The `profile` media type parameter is used to describe the application of
 one or more profiles to the JSON API media type. The value of the `profile`
 parameter **MUST** equal a whitespace-separated list of profile URIs.
 
+> Note: When serializing the `profile` media type parameter, the HTTP 
+> specification requires that its value be surrounded by quotation marks 
+> (U+0022 QUOTATION MARK, "\"") if it contains more than one URI.
+
 Clients and servers **SHOULD** use the `profile` media type parameter in
 conjunction with the JSON API media type in a `Content-Type` header to specify
 the application of one or more profiles to a JSON API document.
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1840,24 +1840,10 @@ identification:
 However, to aid human understanding, profile URIs **SHOULD** return
 documentation of the profile.
 
-A profile **MAY** assign meaning to particular elements of the document
-structure, such as resource attributes or members of meta objects, and even
-to query parameters whose behavior is left up to each implementation, such
-as `filter` and all those that contain a non a-z character.
-
-The scope of a profile **MUST** be clearly delineated. The elements reserved by
-a profile, and the meaning assigned to those elements, **MUST NOT** change over
-time or else the profile **MUST** be considered a new profile with a new URI.
-
-Profiles **MAY** be updated over time to add new capabilities. However, any such
-changes **MUST** be [backwards and forwards compatible](http://www.w3.org/2001/tag/doc/versioning-compatibility-strategies#terminology),
-in order to not break existing users of the profile. This requirement usually
-limits changes to adding optional keys within objects specified in the profile's
-original definition.
-
 The following example profile reserves a `timestamps` member in the `meta`
 object of every resource:
 
+<a id="profiles-timestamp-profile"></a>
 ```text
 # Timestamps profile
 
@@ -1888,20 +1874,6 @@ This profile defines the following keywords:
 
 * `timestamps`
 ```
-
-This profile could evolve to allow other optional members, such as `deleted`,
-in the `timestamps` object. But it could not make that member required, nor
-could it introduce a new sibling to `timestamps`.
-
-To aid evoluation and interoperability, profiles **SHOULD** reserve an
-object-valued member anywhere they expect to potentially add new features over
-time.
-
-> Note: When a profile changes its URI, a huge amount of interoperability is lost.
-> Users that reference the new URI will have their messages not understood by
-> implementations still aware only of the old URI, and vice-versa.
-> Accordingly, the advice above is aimed at allowing profifiles to grow
-> without needing to change their URI.
 
 ### <a href="#profile-media-type-parameter" id="profile-media-type-parameter" class="headerlink"></a> `profile` Media Type Parameter
 
@@ -2068,6 +2040,36 @@ key `version` described in the profile:
 }
 ```
 
+### <a href="#profiles-authoring" id="profiles-authoring" class="headerlink"></a> Authoring Profiles
+
+A profile **MAY** assign meaning to particular elements of the document
+structure, such as resource attributes or members of meta objects, and even
+to query parameters whose behavior is left up to each implementation, such
+as `filter` and all those that contain a non a-z character.
+
+The scope of a profile **MUST** be clearly delineated. The elements reserved by
+a profile, and the meaning assigned to those elements, **MUST NOT** change over
+time or else the profile **MUST** be considered a new profile with a new URI.
+
+Profiles **MAY** be updated over time to add new capabilities. However, any such
+changes **MUST** be [backwards and forwards compatible](http://www.w3.org/2001/tag/doc/versioning-compatibility-strategies#terminology),
+in order to not break existing users of the profile. This requirement usually
+limits changes to adding optional keys within objects specified in the profile's
+original definition.
+
+The hypothetical [timestamps profile], for example, could evolve to allow other 
+optional members, such as `deleted`, in the `timestamps` object. But it could 
+not make that member required, nor could it introduce a new sibling to `timestamps`.
+
+To aid evoluation and interoperability, profiles **SHOULD** reserve an
+object-valued member anywhere they expect to potentially add new features over
+time.
+
+> Note: When a profile changes its URI, a huge amount of interoperability is lost.
+> Users that reference the new URI will have their messages not understood by
+> implementations still aware only of the old URI, and vice-versa.
+> Accordingly, the advice above is aimed at allowing profifiles to grow
+> without needing to change their URI.
 ## <a href="#errors" id="errors" class="headerlink"></a> Errors
 
 ### <a href="#errors-processing" id="errors-processing" class="headerlink"></a> Processing Errors
@@ -2129,6 +2131,7 @@ An error object **MAY** have the following members:
 [link]: #document-links-link
 [link object]: #document-links-link-object
 [profiles]: #profiles
+[timestamps profile]: #profiles-timestamp-profile
 [profile aliases]: #profile-keywords-and-aliases
 [error details]: #errors
 [error objects]: #errror-objects
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1863,6 +1863,8 @@ keyword for an element, then the name of the element itself is considered to be 
 keyword. Keywords **MAY** be aliased in any representation through the use of
 [profile descriptors](profile-descriptors), as described below.
 
+All profile keywords **MUST** meet this specification's requirements for [member names].
+
 ### <a href="#profile-media-type-parameter" id="profile-media-type-parameter" class="headerlink"></a> `profile` Media Type Parameter
 
 The `profile` media type parameter is used to describe the application of
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1809,34 +1809,34 @@ to standard query parameters without conflicting with existing implementations.
 
 ## <a href="#profiles" id="profiles" class="headerlink"></a> Profiles
 
-JSON API supports the use of "profiles" as a means to describe additional
-semantics to the media type, without altering the basic semantics described in
-this specification.
+JSON API supports the use of "profiles" as a way to indicate additional
+semantics that apply to a JSON:API request/document, without altering the 
+basic semantics described in this specification.
 
 [RFC 6906](https://tools.ietf.org/html/rfc6906) covers the nature of profile
 identification:
 
-> Profiles are identified by URI.  However, as is the case with, for
+> Profiles are identified by URI. However, as is the case with, for
   example, XML namespace URIs, the URI in this case only serves as an
   identifier, meaning that the presence of a specific URI has to be
   sufficient for a client to assert that a resource representation
-  conforms to a profile.  Thus, clients SHOULD treat profile URIs as
-  identifiers and not as links, but profiles MAY be defined in a way
-  that the URIs do identify retrievable profile description and thus
-  can be accessed by clients by dereferencing the profile URI.  For
-  profiles intended for use in environments where clients may encounter
-  unknown profile URIs, profile maintainers SHOULD consider to make the
-  profile URI dereferencable and provide useful documentation at that
-  URI.  The design and representation of such profile descriptions,
-  however, is outside the scope of this specification.
+  conforms to a profile. Thus, clients SHOULD treat profile URIs as
+  identifiers and not as links.
+
+However, to aid human understanding, profile URIs **SHOULD** return 
+documentation of the profile.
 
 A profile **MAY** assign meaning to particular elements of the document
 structure, such as resource attributes or members of meta objects, and even
-reserved query parameters, such as `filter`. The scope of a profile **MUST** be
-clearly delineated. The elements reserved by a profile, and the meaning assigned
-to those elements, **MUST NOT** change over time or else the profile **MUST** be
-considered a new profile with a new URI. However, a profile **MAY** evolve
-additively within the scope originally claimed by the profile.
+to query parameters whose behavior is left up to each implementation, such 
+as `filter` and all those that contain a non a-z character.
+
+The scope of a profile **MUST** be clearly delineated. The elements reserved by 
+a profile, and the meaning assigned to those elements, **MUST NOT** change over 
+time or else the profile **MUST** be considered a new profile with a new URI. 
+
+However, a profile **MAY** evolve additively within the scope originally claimed 
+by the profile.
 
 For example, let's say that a profile reserves a `timestamps` member in the
 `meta` object of every resource. Originally, this profile defines the value
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1945,7 +1945,41 @@ Bad Request` status code.
 > with some extra information, as described in the requested profile. By
 > contrast, when a client lists a profile in the `profile` *query parameter*,
 > it's asking the server to *process the incoming request* according to the
-> rules of the profile.
+> rules of the profile. This can fundamentally change the meaning of the 
+> server's response.
+
+#### <a href="#profile-query-parameter-omitting" id="profile-query-parameter" class="headerlink"></a> Omitting the `profile` Query Parameter
+
+A server **MAY** define an internal mapping from query parameter names to 
+profile URIs. The profile URI for a query parameter name in this mapping 
+**MUST NOT** change over time.
+
+If a requested URL does not contain the `profile` query parameter and does 
+contain one or more query parameters in the server's internal mapping, the 
+server may act as though the request URL contained a `profile` query parameter 
+whose value was the comma-separated list of each profile URI found in the 
+server's internal mapping for the query parameters in use on the request.
+
+For example, the server might support a profile that defines a meaning for the
+values of the `filter` query param. Then, it could define its internal 
+param name to profile uri mapping like so:
+
+```json
+{ "filter": "https://example.com/my-filter-profile" }
+```
+
+Accordingly, when a request for 
+
+```
+https://example.com/?filter=xyz
+```
+
+would be interpreted by the server as:
+
+```
+https://example.com/?filter=xyz&profile=https://example.com/my-filter-profile
+```
+
 
 ### <a href="#profile-keywords-and-aliases" id="profile-keywords-and-aliases" class="headerlink"></a> Profile Keywords and Aliases
 
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1831,18 +1831,16 @@ A profile is a separate specification defining these additional semantics.
 [RFC 6906](https://tools.ietf.org/html/rfc6906) covers the nature of profile
 identification:
 
-> Profiles are identified by URI. However, as is the case with, for
-  example, XML namespace URIs, the URI in this case only serves as an
-  identifier, meaning that the presence of a specific URI has to be
-  sufficient for a client to assert that a resource representation
-  conforms to a profile. Thus, clients SHOULD treat profile URIs as
-  identifiers and not as links.
+> Profiles are identified by URI... The presence of a specific URI has to be
+  sufficient for a client to assert that a resource representation conforms to 
+  a profile [regardless of any content that may or may not be available at that 
+  URI].
 
-However, to aid human understanding, profile URIs **SHOULD** return
+However, to aid human understanding, visiting a profile's URI **SHOULD** return
 documentation of the profile.
 
 The following example profile reserves a `timestamps` member in the `meta`
-object of every resource:
+object of every resource object:
 
 <a id="profiles-timestamp-profile"></a>
 ```text
@@ -1865,8 +1863,10 @@ Every resource object **MAY** include a `timestamps` member in its associated
 * `created`
 * `updated`
 
-The value of each member **MUST** comply with the ISO 8601 standard. Any 
-unrecognized members in this object **MUST** be ignored by the application 
+The value of each member **MUST** comply with the variant of ISO 8601 used by 
+JavaScript's `JSON.stringify` method to format Javascript `Date` objects. 
+
+Any unrecognized members in this object **MUST** be ignored by the application 
 processing the document. 
 
 ## Keywords
@@ -1906,7 +1906,7 @@ Accept: application/vnd.api+json;profile="http://example.com/extensions/last-mod
 Servers **MAY** add profiles to a JSON API document even if the client has not
 requested them. The recipient of a document **MUST** ignore any profile extensions
 in that document that it does not understand. The only exception to this is profiles
-whose support is required using the `profile` query parameter, as described below.
+whose support is required using the `profile` query parameter, as described later.
 
 #### <a href="#profiles-sending" id="profiles-sending" class="headerlink"></a> Sending Profiled Documents
 
@@ -2052,10 +2052,10 @@ For example, it would be illegal for a profile to define a new key in a
 document's [top-level][top level] object, or in a [links object], as JSON API 
 implementations are not allowed to add custom keys in those areas.
 
-Likewise, a profile **MAY** assign a meaning to query parameters whose behavior 
-is left up to each implementation, such as `filter` and all those that contain a 
-non a-z character. However, profiles **MUST NOT** assign a meaning to query 
-parameters that [are reserved](#query-parameters).
+Likewise, a profile **MAY** assign a meaning to query parameters or parameter 
+values whose details are left up to each implementation, such as `filter` and 
+all those that parameters that contain a non a-z character. However, profiles 
+**MUST NOT** assign a meaning to query parameters that [are reserved](#query-parameters).
 
 The meaning of an element defined by a profile **MUST NOT** vary based on the 
 presence or absence of other profiles.
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -111,8 +111,8 @@ The top-level [links object][links] **MAY** contain the following members:
 * `self`: the [link][links] that generated the current response document.
 * `related`: a [related resource link] when the primary data represents a
   resource relationship.
-* `profile`: an array of [links][link], each giving the URI of a
-  [profile][profiles] in use in the document.
+* `profile`: an array of [links][link], each specifying a [profile][profiles]
+  in use in the document.
 * [pagination] links for the primary data.
 
 The document's "primary data" is a representation of the resource or collection
@@ -525,10 +525,10 @@ of this member **MUST** be an object (a "links object").
 <a href="#document-links-link" id="document-links-link"></a>
 Within this object, a link **MUST** be represented as either:
 
-* a string containing the link's URL.
+* a string containing the link's URI.
 * <a id="document-links-link-object"></a>an object ("link object") which can
   contain the following members:
-  * `href`: a string containing the link's URL.
+  * `href`: a string containing the link's URI.
   * `meta`: a meta object containing non-standard meta-information about the
     link.
 
@@ -536,7 +536,7 @@ Except for the `profile` key, each key present in a links object **MUST** have
 a single link as its value. The `profile` key, if present, **MUST** hold an
 array of links.
 
-For example, the following `self` link is simply a URL:
+For example, the following `self` link is simply a URI:
 
 ```json
 "links": {
@@ -544,7 +544,7 @@ For example, the following `self` link is simply a URL:
 }
 ```
 
-By contrast, the following `related` link includes a URL as well as
+By contrast, the following `related` link includes a URI as well as
 meta-information about a related resource collection:
 
 ```json
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -2157,6 +2157,12 @@ The scope of a profile **MUST** be clearly delineated. The elements reserved by
 a profile, and the meaning assigned to those elements, **MUST NOT** change over
 time or else the profile **MUST** be considered a new profile with a new URI.
 
+> Note: When a profile changes its URI, a huge amount of interoperability is lost.
+> Users that reference the new URI will not have their messages understood by
+> implementations still aware only of the old URI, and vice-versa. Accordingly,
+> it's important to design your profile so that it can evolve without its URI
+> needing to change. See ["Revising a Profile"](#profiles-updating) for details.
+
 Finally, a profile **MUST NOT**:
 
 1. assume that, if it is supported, then other specific profiles will be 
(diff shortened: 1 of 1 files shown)


_format/1.1/index.md
```diff
@@ -1781,6 +1807,122 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+parameter **MUST** equal a whitespace-separated list of profile URIs.
+
+Clients and servers **SHOULD** use the `profile` media type parameter in
+conjunction with the JSON API media type in a `Content-Type` header to specify
+the application of one or more profiles to a JSON API document.
+
+A client **MAY** use the `profile` media type parameter in conjunction with the
+JSON API media type in an `Accept` header to _request_, but not _require_, that
+the server apply one or more profiles to the response. When such a request is
+received, a server **SHOULD** attempt to apply the requested profiles in
```
joukevandermaas: How would a server signal that it did or did not succeed in this attempt?

dgeb: Profiles requested via `Accept` are considered optional. A server may apply a different set of profiles, or a limited set, to the response document. The actual applied profiles can be inspected by the client in the `Content-Type` header of the response.

If a client _requires_ that a profile be applied, then that profile should be requested via the `profile` query parameter. The server will hard error if it can't apply all the required profiles. Profiles that affect processing rules, and not just representation, should be required this way.

_format/1.1/index.md
```diff
@@ -1781,6 +1807,122 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+### <a href="#profile-query-parameter" id="profile-query-parameter" class="headerlink"></a> `profile` Query Parameter
+
+A client **MAY** use the `profile` query parameter to _require_ the application
+of one or more profiles to the JSON API media type in processing a request. The
+value of the `profile` query parameter **MUST** equal a URI-encoded
+whitespace-separated list of profile URIs.
+
+If a server receives a request requiring the application of a profile or
+combination of profiles that it can not apply, it **MUST** respond with a `400
+Bad Request` status code.
```
marcoow: The fact that the `profile` query param has different semantics than the media type parameter seems a bit confusing to me. Are there any other reasons for this besides the need of having ways for expressing both requesting and requiring a profile? It might be an option to somehow encode these semantics (request vs. require) in the query/media type parameter that's passed vs. where it's passed (header vs. query param)?

dgeb: @marcoow We had quite the discussion in #1207 that largely centered upon this matter, and it took me a while to accept the current proposal, so I certainly understand your perspective. Please see the [note](https://github.com/json-api/json-api/pull/1268/commits/510e783b54917ed15b13931dd07dff0e1d552f27#diff-571f6cee9a492e0e177801c21a4b2b2aR1896) that @ethanresnick just added to expand upon the distinction and let us know if this clarifies the matter for you.

marcoow: Ah sorry - I had missed the discussion around the topic in #1207. I guess if the bike-shedding has been done there already that's great and I don't want to open up another discussion around the topic again here 🙊. I'll read up on the discussion later!

I think the note makes it clearer and will probably help preventing people from running into problems when they're not aware of the difference 👍

_format/1.1/index.md
```diff
@@ -1781,6 +1807,178 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+apply one or more profiles when processing the request. The value of the `profile` 
+query parameter **MUST** equal a URI-encoded whitespace-separated list of profile URIs.
+
+If a server receives a request requiring the application of a profile or
+combination of profiles that it can not apply, it **MUST** respond with a `400
+Bad Request` status code.
+
+> Note: When a client lists a profile in the `Accept` header, it's asking the server
+> to compute its response as normal, but then send the response document with some
+> extra information, as described in the requested profile. By contrarst, when a client
```
marcoow: typo: ~contrarst~contrast

_format/1.1/index.md
```diff
@@ -1781,6 +1822,183 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+  conforms to a profile. Thus, clients SHOULD treat profile URIs as
+  identifiers and not as links.
+
+However, to aid human understanding, profile URIs **SHOULD** return 
+documentation of the profile.
+
+A profile **MAY** assign meaning to particular elements of the document
+structure, such as resource attributes or members of meta objects, and even
+to query parameters whose behavior is left up to each implementation, such 
+as `filter` and all those that contain a non a-z character.
```
wimleers: I don't think `a non a-z character` is precise enough. Perhaps it's better to point to http://jsonapi.org/format/#document-member-names-allowed-characters?

_format/1.1/index.md
```diff
@@ -1781,6 +1822,183 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+
+The scope of a profile **MUST** be clearly delineated. The elements reserved by
```
wimleers: This is the first time the spec would be referring to `elements` as a concept. There's only _one_ mention of "element" in http://jsonapi.org/format/ today, and it's only mentioned in passing:
> If sorting is supported by the server and requested by the client via query parameter sort, the server MUST return *elements of the top-level data array* …
(Emphasis mine.)

I think it would be valuable to disambiguate this.

wimleers: Never mind; it's being defined earlier in this PR:
> particular elements of the document structure, such as resource attributes or members of meta objects, and …

_format/1.1/index.md
```diff
@@ -1781,6 +1822,183 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+original definition.
+
+For example, let's say that a profile reserves a `timestamps` member in the
+`meta` object of every resource. Originally, this profile defines the value
+of `timestamps` as an object that must contain one member: `created`. The
+profile could evolve to allow an optional member, `updated`, in the `timestamps`
+object. But it could not make that member required, nor could it introduce
+a new sibling to `timestamps`.
+
+Profiles **SHOULD** reserve at least one object-valued member, and **SHOULD**
```
wimleers: This comes as a surprise to me. This means that a spec like Drupal's [Fancy Filters](https://gist.github.com/e0ipso/efcc4e96ca2aed58e32948e4f70c2460) is violates this recommendation. OTOH, [Partial Success](https://gist.github.com/e0ipso/732712c3e573a6af1d83b25b9f0269c8) would comply (it reserves the `errors` key in meta objects).

So what about profiles that define behavior for reserved query parameters, such as `filter`?

_format/1.1/index.md
```diff
@@ -1781,6 +1822,183 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+> Users that reference the new URI will have their messages not understood by
+> implementations still aware only of the old URI, and vice-versa. 
+> Accordingly, the advice above is aimed at allowing profifiles to grow
+> without needing to change their URI.
+
+### <a href="#profile-keywords" id="profile-keywords" class="headerlink"></a> Profile Keywords
+
+A profile **SHOULD** explicitly declare "keywords" for any elements that it
+introduces to the document structure. If a profile does not explicitly declare a
+keyword for an element, then the name of the element itself is considered to be its
```
wimleers: I don't think I fully comprehend this, nor its intent. I think a concrete example would help.

_format/1.1/index.md
```diff
@@ -1781,6 +1822,183 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+
+### <a href="#profile-media-type-parameter" id="profile-media-type-parameter" class="headerlink"></a> `profile` Media Type Parameter
+
+The `profile` media type parameter is used to describe the application of
+one or more profiles to the JSON API media type. The value of the `profile`
+parameter **MUST** equal a space-separated (U+0020 SPACE, " ") list of profile URIs.
+
+> Note: When serializing the `profile` media type parameter, the HTTP 
+> specification requires that its value be surrounded by quotation marks 
+> (U+0022 QUOTATION MARK, "\"") if it contains more than one URI.
```
wimleers: Excellent note!

_format/1.1/index.md
```diff
@@ -1781,6 +1822,183 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+
+> The most likely other causes of a 415 error are that the server doesn't
+support JSON API at all or that the client has failed to provide a required 
+profile.
+
+### <a href="#profile-query-parameter" id="profile-query-parameter" class="headerlink"></a> `profile` Query Parameter
+
+A client **MAY** use the `profile` query parameter to _require_ the server to
+apply one or more profiles when processing the request. The value of the `profile` 
+query parameter **MUST** equal a URI-encoded whitespace-separated list of profile URIs.
```
wimleers: This is consistent with the format for the `Accept`/`Content-Type` header value. But wouldn't in this case a comma-separated list be more logical?

Additionally: I tried to find the RFC that prescribes the use of space-separated values but could not find it. Perhaps _both_ are better served by using commas?

wimleers: D'oh, I made a silly mistake: URLs can contain commas, therefore using that as a separator does _not_ make sense at all!

_format/1.1/index.md
```diff
@@ -1781,6 +1822,183 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+JSON API media type in an `Accept` header to _request_, but not _require_, that
+the server apply one or more profiles to the response document. When such a 
+request is received, a server **SHOULD** attempt to apply the requested profiles 
+to its response.
+
+For example, in the following request, the client asks that the server apply the
+`http://jsonapi.org/extensions/last-modified` profile if it is able to.
+
+```http
+Accept: application/vnd.api+json;profile="http://jsonapi.org/extensions/last-modified", application/vnd.api+json
```
wimleers: `profile="URL1 URL2 URL3"` is very verbose. In an `Accept` or `Content-Type` header, this is annoying but manageable. Once you're using the `profile` query parameter, it becomes rather unwieldy.
Wouldn't it be reasonable to rely on IDs?

I'm guessing that long-term, you want to see this supported, but that would require a JSON API profile registry, which doesn't exist yet.

_format/1.1/index.md
```diff
@@ -1781,6 +1822,183 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+> Note: The second instance of the JSON API media type in the example above is
+  required under the [client's content negotiation responsibilities](#content-negotiation-clients).
+  It is used to support old servers that don't understand the profile parameter.
+
+Servers **MAY** add profiles to a JSON API document even if the client has not
+requested them. The recipient of a document **MUST** ignore any profile extensions 
+in that document that it does not understand. The only exception to this is profiles
+whose support is required using the `profile` query parameter, as described below.
+
+#### <a href="#profiles-sending" id="profiles-sending" class="headerlink"></a> Sending Profiled Documents
```
wimleers: I think this section would be better placed _after_ the next one; I think reading is more common than mutating. I first didn't realize that this section was only about mutating.

_format/1.1/index.md
```diff
@@ -1781,6 +1822,183 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+
+Clients and servers **MUST** include the `profile` media type parameter in
+conjunction with the JSON API media type in a `Content-Type` header to indicate
+that they have applied one or more profiles to a JSON API document.
+
+Likewise, clients and servers applying profiles to a JSON API document **MUST** 
+include a [top-level][top level] [`links` object][links] with a `profile` key, 
+and that `profile` key **MUST** include a [link] to the URI of each profile
```
wimleers: When no profiles are specified in the request document, but they _are_  specified in the `Content-Type` request header or the `profile` URL query parameter, should a `400` error response be sent?

IOW: are JSON API documents in request bodies required to match the `Content-Type` and/or `profile` URL query argument?

wimleers:
First off: thanks to everyone for this epic work, in this issue and related issues!

I'm reviewing this as one of the people doing the work to bring *JSON API support to Drupal core*. Note that there are already [>5000 Drupal 8 sites using JSON API](https://www.drupal.org/project/usage/jsonapi) (that usage count is an underestimation: it's the lower bound).

As mentioned in #1207, we're working to [minimize/reduce risk as much as possible](https://www.drupal.org/project/jsonapi/issues/2931785), by having detailed attention to [API surface and its implications: backwards compatibility, maintainability, supportability](https://wimleers.com/talk/bc-burden-benefit). We managed to achieve zero configuration: install this module and all Drupal data is instantly exposed via JSON API, respecting the existing access control logic. If users can view `foobar` resources in Drupal (as HTML), they can also `GET` them via JSON API.

Specifically, the https://www.drupal.org/project/jsonapi module for Drupal (that we're working to move into Drupal core itself) uses two extensions: "Fancy Filters" and "Partial Success" — see https://www.drupal.org/project/jsonapi/issues/2955020. Without the former, no querying of resources (e.g. only published articles written by Wim), without the latter, impossible to do per-resource access control (e.g. Wim can see unpublished articles, but Gabe can't).

Today, and for the past ~2 years, those extensions are always running, but their existence is not communicated explicitly, because the JSON API 1.0 spec does not support extensions, per http://jsonapi.org/extensions/.

Therefore this issue/PR is _essential_ for JSON API support to land in Drupal core. Otherwise we risk future JSON API specs moving in directions incompatible with what Drupal's JSON API implementation does … and then we would have a _massive_ backwards compatibility problem.

---

@effulgentsia, @gabesullice and I commented in #1207.

I wanted to call out a two specific concerns they raised, and whether this PR addresses them:

1.  From the OP of this PR:
> The application of one or more profiles to a response can be requested by a client via the `Accept` header.
> In order to require the application of one or more profiles, the profile(s) must be specified with the `profile` query parameter.

This addresses @effulgentsia's concern in https://github.com/json-api/json-api/issues/1207#issuecomment-380583914.

This would mean no BC break in the current `jsonapi` module (version 1.x), because everything would continue to work. We're about to create version 2.x, with many soft BC breaks to address shortcomings in the original implementation. In version 2.x, I don't think we'd require clients to specify `?profile=fancyfilters,partialsuccess` for every request (because bad Developer Experience)

2. Gabe's concern at https://github.com/json-api/json-api/issues/1207#issuecomment-379840177
> Request
> GET /valid-url?unknownProfileParameter
> Accept: json-api
> Link: <profile-defining-unknown-parameter>; rel=profile
> 
> Response
> 404 (if server doesn’t support the profile);
> 400 if the server doesn't understand the qp value
> 200 otherwise.
> What the client provides in Accept doesn't actually influence this response.
>
> I see this as a minor improvement to avoid the conflicts that were worked around in this sentence:
> > So, the first/primary extension the server wants to support for filtering can claim ?filter, and the second ends up with ?profiles[other-filter]

I don't think this concern is fully addressed? I think the conflict @gabesullice explained/predicted there still exists. I can understand that his `Link` request header proposal may not be accepted, but I don't see a reason why his `404` proposal (in case the server doesn't support the profile) is not in the PR.

Looking forward to your thoughts about this :)

---

Review of the actual PR to follow.

wimleers (commented):
Several nitpicks, but also some actual questions.

_format/1.1/index.md
wimleers: On second thought, I don't think this makes sense, because it's already valid to just send
```
{
  "data": { … }
}
```

Because http://jsonapi.org/format/#document-top-level already says that only one of `data`, `errors` and `meta` is a MUST-have.

_format/1.1/index.md
```diff
@@ -540,6 +558,14 @@ about a related resource collection:
 }
 ```
 
+Here, the `profile` key specifies an array of `profile` links:
+
+```json
+"links": {
+  "profile": [{ "href": "http://jsonapi.org/ext/example-ext" }]
```
wimleers: Nit: Still mention of `ext` instead of `profile`.

_format/1.1/index.md
```diff
@@ -1781,6 +1822,183 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+
+> Note: When a profile changes its URI, a huge amount of interoperability is lost.
+> Users that reference the new URI will have their messages not understood by
+> implementations still aware only of the old URI, and vice-versa. 
+> Accordingly, the advice above is aimed at allowing profifiles to grow
+> without needing to change their URI.
+
+### <a href="#profile-keywords" id="profile-keywords" class="headerlink"></a> Profile Keywords
+
+A profile **SHOULD** explicitly declare "keywords" for any elements that it
```
wimleers: *How* can a profile explicitly declare a keyword?

_format/1.1/index.md
```diff
@@ -1781,6 +1822,183 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+Each profile descriptor object **MAY** contain an `aliases` object. The key of
+each member in `aliases` **MUST** be a [keyword](profile-keywords) explicitly
+declared by the profile, and the value **MUST** be an alias that applies to this
+particular representation. This aliasing mechanism allows profiles to be applied
+in a way that is both consistent with the rest of the representation and does
+not conflict with other profiles.
+
+For instance, the following document includes a single profile descriptor for a
+profile that, let's say, assigns meaning to a `version` keyword that can be
+included in any resource's `meta` object. This descriptor provides an alias for
```
wimleers: Once we have a mechanism to explicitly declare a keyword, that also has as a side benefit that that declaration could also state whether the keyword applies to `meta`, a query param, or something else still.

wimleers:
FYI: this is being implemented for Drupal at https://www.drupal.org/project/jsonapi/issues/2955020.

_format/1.1/index.md
```diff
@@ -1781,6 +1822,183 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
...
+As described above, the top-level [`jsonapi`](document-jsonapi-object) object
+**MAY** contain a `profiles` member that contains a map of profile descriptor
+objects, keyed by profile URI.
+
+Each profile descriptor object **MAY** contain an `aliases` object. The key of
+each member in `aliases` **MUST** be a [keyword](profile-keywords) explicitly
+declared by the profile, and the value **MUST** be an alias that applies to this
+particular representation. This aliasing mechanism allows profiles to be applied
+in a way that is both consistent with the rest of the representation and does
+not conflict with other profiles.
```
wimleers: Does this mean JSON API servers are required to resolve/process the aliases defined in a request document? (It's possible that the client uses different aliases than the server.)

_format/1.1/index.md
ethanresnick: The text might warrant some clarification, but I believe "those" refers to "query parameters left up to each implementation", the rule for which is non a-z (see http://jsonapi.org/format/#query-parameters)

_format/1.1/index.md
ethanresnick: Yeah, this was written with profiles that add data to the document in mind, not those that just use query parameters. Although, on second thought, it's probably a good idea even for a profile like fancy filters to reserve a key (probably somewhere in `meta`) so that, if it ever wants to add some metadata to the response (e.g., the size of the unfiltered set, or a deprecation warning or something), it can do so.

_format/1.1/index.md
ethanresnick: The profile would declare the keyword in its specification. Part of the problem is that this PR never actually says what a profile *is*. We go straight from talking about what they do ("a way to indicate additional semantics") to how they're identified ("by URI"). If we import some language from #1195:

> A [profile] is a small, separate specification that defines...

(We have to update what's after the ...)

Then, it's clear that you declare keywords in that specification

_format/1.1/index.md
ethanresnick: > Once you're using the `profile` query parameter, it becomes rather unwieldy.

It is a bit unweildy, and I saw that in [an earlier comment](https://github.com/json-api/json-api/pull/1268#issuecomment-394334144) you said: 

> In version 2.x, I don't think we'd require clients to specify ?profile=fancyfilters,partialsuccess for every request (because bad Developer Experience)

I want to think this over and take it seriously. If the query parameter is too unwieldy that people don't bother, then it's really problematic to hinge negotiation on it. IDs could be part of the answer, but they'd seem to have the inadvertent side effect of creating two different identifiers for each profile (its URI and registry id), which is probably bad for caching + implementations would have to remember/be careful to handle both identifiers the same way.

_format/1.1/index.md
ethanresnick: > When no profiles are specified in the request document, but they are specified in the Content-Type request header or the profile URL query parameter, should a 400 error response be sent?

This is a good point. I think we want to say that, on requests, if the profiles listed in `Content-Type` don't match those listed in the document body, the server should return a 400. 

> On second thought, I don't think this makes sense, because it's already valid to just send

I don't think that prevents us from requiring a 400 in this case, because none of the requests that are valid today will contain `profile` in the `Content-Type`.

_format/1.1/index.md
ethanresnick: > Additionally: I tried to find the RFC that prescribes the use of space-separated values but could not find it. 

I believe this comes from the RFC that defines the profile parameter https://tools.ietf.org/html/rfc6906

> D'oh, I made a silly mistake: URLs can contain commas, therefore using that as a separator does not make sense at all!

Yeah, though, even with spaces, we're gonna have to figure out some encoding concerns here I suspect.

_format/1.1/index.md
ethanresnick: Yes

ethanresnick:
@wimleers Thanks for the review, and sorry for my delay in addressing your comments. I've been swamped. 

I don't think any of your comments suggest that this PR needs serious changes, which is reassuring. The possible exception is the around the developer experience of requiring long urls in the `profile` query parameter for every request. I'll bring that up with @dgeb, though I imagine he already thought about it when making the proposal, and I know that that was an area where it was tricky to find any consensus.

**EDIT**: I think the solution here is simple: the server can interpret each query parameter as coming from the profile of its choice by default (though it can't change that choice over time). In other words, it can implicitly interpret `?filter=.....` as coming from fancy filters in the absence of a `?profiles` parameter to the contrary, as long as it always interprets it that way. If the `?profile` parameter is used, though, it still must be respected (e.g., `?filter=....&profile=https://drupal.org/filter-spec-2` is no longer a fancy-filters request). Finally, `?filter=....&profile=https://drupal.org/fancy-filters` is just equivalent to the request without the `?profile` param.

Does that approach work for everyone?

> I don't think this concern is fully addressed? I think the conflict @gabesullice explained/predicted there still exists. I can understand that his Link request header proposal may not be accepted, but I don't see a reason why his 404 proposal (in case the server doesn't support the profile) is not in the PR.

I'm not sure I understand this. Can you give an example?

dgeb:
> I think the solution here is simple: the server can interpret each query parameter as coming from the profile of its choice by default (though it can't change that choice over time). 

@ethanresnick I agree that this is the most simple and elegant solution. The critical part here is that the JSON:API resource(s) (i.e. HTTP resource) identified by a request does not change for a given URL. If usage of a profile targets different resource(s), then the same URL must not be used. And we explicitly provide the `profile` query param for the purpose of guaranteeing a unique URL when this is the case. I think it's probably ideal for a given API to choose a default set of profiles carefully, so that conventional usage of that API requires no `profile` query param at all. I'm going to attempt to rework the phrasing in this PR to this effect over the next day or two.

_format/1.1/index.md
dgeb: That's fair. An example would probably go a long way here.

The point of keywords is to allow any member that's defined by profile to be aliased in a particular implementation. This aliasing might be as simple as ensuring that an API that uses camelCase throughtout does not have to use dasherized-case to comply with a particular profile (e.g. `totalCount` vs. `total-count`)

_format/1.1/index.md
dgeb: I think your suggestion [below](https://github.com/json-api/json-api/pull/1268#issuecomment-397085657) is the right direction here @ethanresnick.

wimleers:
>  I've been swamped.

No need to apologize — I know the feeling all too well!

> I don't think any of your comments suggest that this PR needs serious changes, which is reassuring. 

Indeed! :) 👍 

> I'm not sure I understand this. Can you give an example?

I'll let @gabesullice reply to this — it's possible I'm misinterpreting what he wrote.

> I think the solution here is simple: the server can interpret each query parameter as coming from the profile of its choice by default (though it can't change that choice over time).

Agreed!

@ethanresnick++
@dgeb++

_format/1.1/index.md
dgeb: We can probably clarify intent here by saying something like:

> In order to support the potential to add features over time, profiles **SHOULD** reserve at least one object-valued member in any portion of the document in which they may realistically expand in the future.

dgeb:
After discussing with @ethanresnick, I've been convinced that the `profile` links objects can use target attributes to express `aliases`, as per similar usages in [RFC 5988](https://tools.ietf.org/html/rfc5988).

This means that we're able to completely remove the `jsonapi/profiles` member originally described in this PR.

I've updated this PR to reflect these changes and added some example profiles as well.

wimleers:
That looks like a good step forward 👍

_format/1.1/index.md
```diff
@@ -1778,8 +1816,456 @@ If a server encounters a query parameter that does not follow the naming
...
+
+For example, the server might support a profile that defines a meaning for the
+values of the `filter` query param. Then, it could define its internal 
+param name to profile uri mapping like so:
+
+```json
+{ "filter": "https://example.com/my-filter-profile" }
+```
+
+Accordingly, when a request for
```
dgeb: Nit regarding wording. Perhaps: "Accordingly, a request for:"

_format/1.1/index.md
```diff
@@ -1778,8 +1816,456 @@ If a server encounters a query parameter that does not follow the naming
...
+> Note: When a client lists a profile in the `Accept` header, it's asking the
+> server to compute its response as normal, but then send the response document
+> with some extra information, as described in the requested profile. By
+> contrast, when a client lists a profile in the `profile` *query parameter*,
+> it's asking the server to *process the incoming request* according to the
+> rules of the profile. This can fundamentally change the meaning of the 
+> server's response.
+
+#### <a href="#profile-query-parameter-omitting" id="profile-query-parameter" class="headerlink"></a> Omitting the `profile` Query Parameter
+
```
dgeb, line 1952: I wonder if this section would be more clear with a motivating lead-in, such as "A server may apply a set of default profiles without requiring them to be explicitly requested."

_format/1.1/index.md
```diff
@@ -1778,8 +1816,456 @@ If a server encounters a query parameter that does not follow the naming
...
+### <a href="#profile-keywords-and-aliases" id="profile-keywords-and-aliases" class="headerlink"></a> Profile Keywords and Aliases
+
+A profile **SHOULD** explicitly declare "keywords" for any elements that it
+introduces to the document structure. If a profile does not explicitly declare a
+keyword for an element, then the name of the element itself (i.e., its key in
+the document) is considered to be its keyword. All profile keywords **MUST** 
+meet this specification's requirements for [member names].
+
+For the purposes of aliasing, a profiles elements are defined shallowly.
```
dgeb: "a profile's elements"

_format/1.1/index.md
```diff
@@ -1778,8 +1816,456 @@ If a server encounters a query parameter that does not follow the naming
...
+values whose details are left up to each implementation, such as `filter` and 
+all those that parameters that contain a non a-z character. However, profiles 
+**MUST NOT** assign a meaning to query parameters that [are reserved](#query-parameters).
+
+The meaning of an element or query parameter defined by a profile **MUST NOT** 
+vary based on the presence or absence of other profiles.
+
+The scope of a profile **MUST** be clearly delineated. The elements and query 
+parameters specified by a profile, and their meanings, **MUST NOT** change over
+time or else the profile **MUST** be considered a new profile with a new URI.
```
dgeb, line 2203: @ethanresnick - do you think it is clear enough that objects defined by a profile can expand to include new members over time?

dgeb: This is covered in the section below - perhaps that's sufficient without adding another phrase here.

_format/1.1/index.md
ethanresnick: Fixed!

_format/1.1/index.md
ethanresnick: I like the idea of a motivating lead in makes sense to me, and I've added one in the latest commit. I used slightly different wording because I didn't want to imply that the server could alter the HTTP resource's behavior by applying a profile if _no_ query param is specified.

_format/1.1/index.md
ethanresnick: 👍

_format/1.1/index.md
ethanresnick: Yeah, I think that's sufficient. Happy to add expand though if you have specific wording in mind.

dgeb:
@ethanresnick Thanks for making those adjustments. I think this is finally ready to squash and merge! 🚀

ethanresnick:
@dgeb AWESOME. I'm going squash this down a bit (though maybe keep more than one commit, as having some distinct history is useful I think) and then will merge.

ethanresnick:
Actually, jk, there aren't really neat commit divisions anymore anyway, so squash and merge it is.

dgeb:
Woot! Thanks for all your work on profiles over the years, @ethanresnick. This one is more yours than mine. :shipit:

ethanresnick:
Thanks Dan! Next up: operations :)

wimleers:
Exciting! 🎉  Congrats! And thanks :)

This means that the Drupal JSON API module — which is currently on a path to inclusion into Drupal core itself — now actually _could_ implement this. How certain are we that there will be no more breaking changes in the spec wrt profiles?

I reopened the Drupal.org issue: https://www.drupal.org/project/jsonapi/issues/2955020#comment-12781015.

ethanresnick:
> Exciting! 🎉 Congrats! And thanks :)
Thanks!

> This means that the Drupal JSON API module — which is currently on a path to inclusion into Drupal core itself — now actually could implement this. 
🎉 

> How certain are we that there will be no more breaking changes in the spec wrt profiles?

I can't forsee any changes, except if implementation experience proves that some part of this design is really unworkable.

PR #1270, opened 2018-04-23
Add swisnl/json-api-client to PHP client implementations

bbrala:
We created an implementation for PHP which helps using a JSON API as your data source for your application. The beauty is it helps you talk to the API with your own models.


ethanresnick:
Sorry this took so long! Not sure how it slipped through the cracks

bbrala:
Thanks, no problem :)

PR #1278, opened 2018-05-26
Add JSORM and JSONAPI Suite implementations

richmolj:


dgeb:
@richmolj I'm surprised these weren't listed yet. Thanks for all your work!

PR #1279, opened 2018-05-27
Change RFC2119 wording to be consistent with RFC8174

### LINKED ISSUE #1277: Change RFC2119 wording to be consistent with RFC8174
See [RFC8174](https://tools.ietf.org/html/rfc8174), which contains a clarification/update of the RFC2119 wording that should ideally be used instead of the old one.

cmeeren:
Fixes #1277


ethanresnick:
LGTM. Thanks

Issue #1255, opened 2018-02-10
Recommending camelCase for JSON members


garrettmac:
As suggested by @dgeb I'm moving my PR https://github.com/json-api/json-api/pull/1247 into an issue for recommending camelCase for JSON members. 

The gist of my argument couldn't have been said better than the point @dgeb mentions in https://github.com/json-api/json-api/pull/1247
> We may have placed too much emphasis on correlating case between URLs and JSON members in forming our current recommendations.

<br>

I wanted to first mention it's JSON, which stands for **Javascript** Object Notation,  therefor should be in **camelCase**. 

In addition I also wanted to share the following:

**Several well-established style guides recommend camelCase:**

<hr>

#### [JSON.org](http://json.org):

> * Property names must be camel-cased, ascii strings.

#### [Google JSON Style Guide](https://google.github.io/styleguide/jsoncstyleguide.xml?showone=Property_Name_Format#Property_Name_Format) (they follow the [JSON.org](http://json.org) spec):

> * Property names must be camel-cased, ascii strings.

#### [Crockford](http://crockford.com/javascript/code.html#names) state:

> * Most variables and functions should start with a lower case letter.
> * Names should be formed from the 26 upper and lower case letters (A .. Z, a .. z), the 10 digits (0 .. 9) 

#### [Schema.org](https://schema.org/docs/old_extension.html) 

> Properties start with a lower case letter and are also camelCase.

####  [W3resource](https://www.w3resource.com/slides/json-style-guide.php)(same thing) state:

> Property names must be camel-cased, ascii strings.

 **`camelCase` in the wild:**
*  [Mozilla Objects and properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Working_with_Objects#Objects_and_properties)
* [W3School's [JS JSON](https://www.w3schools.com/js/js_json.asp) and [JS Conventions](https://www.w3schools.com/js/js_conventions.asp)
* [Codecademy](https://www.codecademy.com/courses/javascript-beginner-en-xTAfX/0/1)
* **graphql** spec uses `camelCase` for fields.

<hr>

**`kebab-case` does not play nicely with JSON. Some examples include:**
* JavaScript (you cannot do `person.first-name`)
* EmberJs Framework
* Schema.org / JSON-LD
and
* GraphQL / GraphQL Queries? (maybe? not actually sure on this one, admittedly haven't done much here)

I would enjoy feedback from the maintainers of the project and the community.

[...]

PR #1297, opened 2018-09-29
Recommend camel cased member names

ethanresnick:
@dgeb and I would like to change our current member naming recommendations from kebab-cased to camel cased. 

This recommendation is somewhat more important with the introduction of profiles, as it would be ugly if some profiles used one convention and others used another; authors would be stuck with inconsistent naming in their profiled documents.

My reasons for supporting this change to camel-case are:

- camelCased names can be used directly as identifiers in almost all programming languages, making json:api easier to get started with and to work with over time. Dasherized names are usable as identifiers pretty much only in Lisps.

- camelCased names are the most common convention in the JS community, and JS is the biggest user of JSON:API

- camelCased names are slightly shorter than dasherized (or snake-case) names, which could help with url character limits in the case of complex filters or other types of complex queries (e.g., some GraphQL like “deep querying” feature) that we might serialize into urls in the future.

Closes #1247, #1255 

Note: 1255 talked about also changing the recommendation for type names to be camel-cased, which would ripple through to URLs. I support this --- except that I couldn't find anywhere where we actually currently recommend kebab-case for `type`, so there's nothing to change. Maybe we want a new recommendation here?

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -443,8 +443,8 @@ A complete example document with multiple included relationships:
     "type": "people",
     "id": "9",
     "attributes": {
-      "first-name": "Dan",
-      "last-name": "Gebhardt",
+      "firstName": "Dan",
+      "lastName": "Gebhardt",
       "twitter": "dgeb"
     },
     "links": {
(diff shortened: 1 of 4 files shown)


dgeb (approved):
As @ethanresnick mentioned, I approve of the change to recommending camelCase. This PR LGTM.

I think it's probably worthwhile to add a recommendation re: `type` as well, but it seems better to handle that separately since no recommendation exists currently. Furthermore, there can be other aspects to `type`, such as namespacing/scoping, that will probably be beyond the scope of recommendations.

ethanresnick:
Awesome! I'm happy to have another PR for `type` recs too.

antonkomarev:
Thanks a lot!

Issue #1304, opened 2018-10-07
Draw out distinct concepts of a query param and "parameter family"


ethanresnick:
Right now, the JSON:API spec uses the terms "query parameter" and "query parameter name" in a way that's inconsistent with other web standards, and internally inconsistent throughout the spec.

Consider the URL `/articles?page[cursor]=xyz&page[size]=10`. Does that URL have one query parameter (named `page`) or two (one named `page[cursor]` and one named `page[size]`)? 

From a standards POV, the best/only answer comes from the WHATWG Url standard, which [says (see step 4)](https://url.spec.whatwg.org/#concept-urlsearchparams-new) that it has two. Accordingly, the following code logs first `"page[cursor]"` and then`"page[size]"`:

```js
for(let [k,v] of new URLSearchParams("?page[cursor]=xyz&page[size]=10")) {
  console.log(k);
}
```

(Other standards, like RFC 3986 and the HTTP RFCs don't have a concept of distinct query parameters at all; they only have the idea of a single query string.)

When the JSON:API spec says, though, that: "The `page` query parameter is reserved for pagination", we're clearly intending to cover the whole set of `page[offset]`, `page[cursor]`, `page[xxx][yy]` parameters with that statement — not just the literal parameter `page`. 

Therefore, I propose that we explicitly define the concept of a "query parameter family", distinct from an individual query parameter. That will let us use the term "query parameter" in a way that's consisten with WHATWG and, perhaps more importantly, it'll clear up a few spec ambiguities:

1. In error objects, we can clarify that the `"parameter"` key is supposed to refer to the specific parameter, and not the parameter family's name. For example:

    The request `GET /xxx?fields[books]=tilte&fields[people]=name` has a typo: the `title` field has been misspelled as `tilte`. In the resulting error object, I think we want something like:

    ```js
    { /* ... */  "source": { "parameter": "fields[books]" } }
    ```

    Having the parameter be `fields[books]` is more useful than having it just be `fields` but, at the moment, it's deeply ambiguous which is correct. In the sparse fieldsets section of the spec, for example, we manage to refer to both "a `fields[TYPE]` parameter" (which implies `field[books]` could be the source) and "the `fields` parameter" (which seems to imply the opposite). 

[...]

Issue #1302, opened 2018-10-07
Clarify encoding of square brackets in query param names


ethanresnick:
JSON:API requires the use of square brackets in `?fields[TYPE]` query parameter names. But we don't say in any normative text whether these square bracket should be percent-encoded. 

We strongly imply that they should be, though, through non-normative notes like:

> Note: The above example URI shows unencoded `[` and `]` characters simply for readability. In practice, these characters must be percent-encoded, per the requirements in RFC 3986.

The idea implicit in these notes seems to be that "the syntax for URLs is defined by RFC 3986 so, if that RFC requires encoding, then of course these characters must be encoded".

The issue, which I didn't realize at the time, is that there are actually two competing standards around the format of URLs; the other is the WHATWG Url spec, which browsers follow.

Per the WHATWG Url spec, browsers do _not_ percent-encode `[` and `]` in query strings by default.

Therefore, in an ideal world, I think we'd define field parameter names with encoded square brackets (i.e., `fields%5BTYPE%5D`) and those with un-encoded square brackets as equivalent aliases for one another, and would require servers to support both.

In practice, I'm pretty sure we can — and, for clarity, I think we should — turn our current notes into a *normative MUST* requiring that servers support the percent-encoded form. Since HTTP's specification incorporates RFC 3986 as its normative definition of a URL, and we've always said that JSON:API should be read to comply with HTTP, and because we have the existing notes we do referencing RFC 3986, I don't think this MUST would be a breaking change.

Further, I think we should add a **SHOULD** that servers support the unencoded versions — since this is what browsers will send by default — and that these unencoded parameter names should be treated as identical to the percent-encoded variants.

dgeb:
@ethanresnick I agree with this analysis. The spec could use the clarifying statements you suggest 👍

ethanresnick:
Closed by #1308

Issue #1266, opened 2018-04-07
Undefined grammar for sort query parameter


ethanresnick:
The spec for the `sort` query parameter doesn't specify what format/grammar a "sort field" must follow. However, it might be good to define a grammar now that works with existing implementations, to preserve our extensibility options in the future.

The genesis of this issue is that I wanted my API to support sorting the returned resources by their distance from some point. To accomplish that, I was thinking about having a `?sort=distanceFrom[lng,lat]` syntax or similar. This syntax clearly would involve a minus sign (if the longitude or latitude is negative), but the minus sign would only appear in a context where my parser can distinguish it from the minus sign that can prefix a sort field. Is that valid? [Edit: it probably has to be, as disallowing the minus sign within sort fields would make it impossible to filter on kebab-case field names.]

What about the comma in my proposed syntax, which can likewise be contextually distinguished from the comma that separates sort fields? 

The current spec says that sort fields don't need to correspond to field names, so it's not clear that the member naming rules would automatically apply to sort fields. Also, strictly speaking, the spec shows an example of a sort field that contains a "." in it, which would be invalid as a member name.

I guess the most minimal thing we could do is say that the comma can't appear directly in a sort field, but can appear urlencoded. ~~The problem with that is, I suspect, that some implementations are going to unconditionally percent encode the comma that separates sort fields (even though that's not strictly necessary), which would then make it impossible to use whether a comma is url encoded as a way to determine whether the comma is part of the sort field or a separator between sort fields.~~ [Edit: this is wrong.] So, a better solution might be to disallow commas within a sort field altogether, which seems quite reasonable given the current spec (even though it would be quite convenient for my use case). If we do that, I guess the question is also just whether we want to reserve some other characters at the same time for future extensions.

ethanresnick:
One note about disallowing minus signs: we probably only want to disallow them at the start of the query string, or immediately after a comma [i.e., as the first character of a sort field], as otherwise it'd be impossible to filter on kebab-case field names

dgeb:
It seems fine to spell out that url-encoded commas may be used _within_ sort fields. 

I don't think we need to specify anything else re: minus signs, because the spec is pretty clearly speaking only about their significance at the beginning of sort fields:

> The sort order for each sort field MUST be ascending unless it is prefixed with a minus (U+002D HYPHEN-MINUS, “-“), in which case it MUST be descending.

I think it would be breaking, and unnecessary, to try to disallow minuses within sort fields at this point.

[...]

PR #1308, opened 2018-10-20
feat(query-params): clarify parsing/serializing, individual params vs family

ethanresnick:
Addresses #1302, #1304.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1806,7 +1806,7 @@ Although "query parameter" is a common term in everyday web development, it is
 not a well-standardized concept. Therefore, JSON:API provides its own 
 [definition of a query parameter](#appendix-query-details).
 
-For the most part, JSON:API's definition coincides with coloquial usage, and its 
+For the most part, JSON:API's definition coincides with colloquial usage, and its 
 details can be safely ignored. However, one important consequence of this 
 definition is that a URL like the following is considered to have two distinct
 query parameters:
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1836,7 +1836,7 @@ etc. However, `filter[_]` is not a valid parameter name in the family, because
 ### <a href="#query-parameters-custom" id="query-parameters-custom" class="headerlink"></a>   Implementation-Specific Query Parameters
 Implementations **MAY** support custom query parameters. However, the names of 
 these query parameters **MUST** come from a [family][query parameter family] 
-whose base name is a legal [member names] and also contains at least 
+whose base name is a legal [member name][member names] and also contains at least 
 one non a-z character (i.e., outside U+0061 to U+007A).
 
 It is **RECOMMENDED** that a capital letter (e.g. camelCasing) be used to 
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -644,7 +644,7 @@ first or last character:
 
 The following characters **MUST NOT** be used in member names:
 
-- U+002B PLUS SIGN, "+" _(used for ordering)_
+- U+002B PLUS SIGN, "+" _(has overloaded meaning in URL query strings)_
 - U+002C COMMA, "," _(used as a separator between relationship paths)_
 - U+002E PERIOD, "." _(used as a separator within relationship paths)_
 - U+005B LEFT SQUARE BRACKET, "[" _(used in sparse fieldsets)_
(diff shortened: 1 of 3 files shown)


_format/1.1/index.md
```diff
@@ -1809,9 +1801,43 @@ responses, in accordance with
...
 
-Implementation-specific query parameter names **MUST** adhere to the same
-constraints as [member names], with the additional requirement that they
-**MUST** contain at least one non a-z character (i.e., outside U+0061 to U+007A).
+### <a href="#query-parameters-families" id="query-parameters-families" class="headerlink"></a>  Query Parameter Families
+Although "query parameter" is a common term in everyday web development, it is
+not a well-standardized concept. Therefore, JSON:API provides its own 
+[definition of a query parameter](#appendix-query-details).
+
+For the most part, JSON:API's definition coincides with coloquial usage, and its
```
wimleers:
```suggestion
For the most part, JSON:API's definition coincides with colloquial usage, and its 
```

_format/1.1/index.md
```diff
@@ -1809,9 +1801,43 @@ responses, in accordance with
...
+
+For example, the `filter` query parameter family includes parameters named:
+`filter`, `filter[x]`, `filter[]`, `filter[x][]`, `filter[][]`, `filter[x][y]`, 
+etc. However, `filter[_]` is not a valid parameter name in the family, because
+`_` is not a valid [member name][member names].
+
+### <a href="#query-parameters-custom" id="query-parameters-custom" class="headerlink"></a>   Implementation-Specific Query Parameters
+Implementations **MAY** support custom query parameters. However, the names of 
+these query parameters **MUST** come from a [family][query parameter family] 
+whose base name is a legal [member names] and also contains at least
```
wimleers:
```suggestion
whose base name is a legal [member name][member names] and also contains at least 
```

_format/1.1/index.md
```diff
@@ -1962,6 +1988,10 @@ To do so, a server **MAY** define an internal mapping from query parameter names
 to profile URIs. The profile URI for a query parameter name in this mapping 
 **MUST NOT** change over time.
 
+> Note: the server may choose to map all query parameter names from the same
```
wimleers, line 1991: Doesn't this allow two servers A an B that support the same profiles X, Y and Z to reach different conclusions wrt parameter handling?

_format/1.1/index.md
```diff
@@ -2326,6 +2356,51 @@ An error object **MAY** have the following members:
...
+A query parameter is a name–value pair extracted from, or serialized into, a 
+URI's query string. 
+
+To extract the query parameters from a URI, an implementation **MUST** run the 
+URI's query string, excluding the leading question mark, through the 
+[`application/x-www-form-urlencoded` parsing algorithm](https://url.spec.whatwg.org/#urlencoded-parsing),
+with one exception: JSON:API allows the specification that defines a query 
+parameter's usage to provide its own rules for parsing the parameter's value 
+from the `value` bytes identified in steps 3.2 and and 3.3 of the `application/x-www-form-urlencoded` 
+parsing algorithm. The resulting value might not be a string.
```
wimleers, line 2371:
```suggestion
parsing algorithm. The resulting value might not be a string, but could be a (nested) array of strings instead.
```
Small clarification, because "might not be a string" sounds rather ambiguous?

wimleers (changes requested):
A few small things :) Looking great!

_format/1.1/index.md
ethanresnick: Not sure I follow... since every server got to define it's own mapping of parameters to profile urls even before we introduced the parameter vs. parameter family distinction, hasn't that always been the case?

_format/1.1/index.md
ethanresnick: The current language is intentionally open ended. 

In https://github.com/json-api/json-api/issues/1266#issuecomment-379600455, I gave the example of a URL like `?sort=distanceFrom(lat,lng),otherField`. If a profile were to specify using the sort parameter like that (which I think would be legal), the parse result for that sort field would likely be some sort of object/tree, so just saying that the result could be a (nested) array of strings is insufficient.

_format/1.1/index.md
wimleers: I misinterpreted this.

I thought this was referring to the case when multiple profiles overlap in which query parameter names they react to/define behavior for — say `page[special]` is defined tiwce. "server may choose" then made me think "but then how can the client know which profile's behavior is going to apply?"

_format/1.1/index.md
wimleers: 👍

_format/1.1/index.md
ethanresnick: Ok, cool. Happy to clarify the text too if you think it's warranted.

dgeb (approved):
Some long overdue clarifications. Great work @ethanresnick!

ethanresnick:
Thanks @dgeb!

PR #1309, opened 2018-10-22
Add ruby client libraries

pcriv:
- Add jsonapi-simple_client
- Add jsonapi-record


ethanresnick:
Thanks for submitting these! 

At least for now, we're adding implementations to the list in the order they're submitted, so please update your PR to list these at the bottom of the appropriate section, rather than the top.

pcriv:
@ethanresnick done :)

ethanresnick:
Thanks!

PR #1311, opened 2018-10-22
Profile registration

ethanresnick:
This is (finally!) the PR that will allow people to register profiles and have them showcased on the JSON:API website, with a url on jsonapi.org. 

It includes:

- On the existing extensions page, submitted profiles are listed, grouped by category, using categories listed in each profile's yaml front-matter. (The list of available categories is set in `_config.yml`.) The extensions page now also has a description of the registration process and a link to a Markdown template, which authors can fill out with their profile's specification and submit with their PR.

- There's a new `profile.html` layout file, which is used to show off the user-submitted profiles. The profile specs live in the `_profiles` directory/collection. (Lots of HTML from the `page.html` was pulled out into files in the `_includes` directory so that it could be reused as snippets in this new `profile.html` layout.)

- There's an example cursor-based pagination profile. (If people are interested in implementing this profile — @wimleers? — I'd love to get implementation/design feedback on it before shipping it. But this feedback isn't blocking 1.1.)


extensions/index.md
```diff
@@ -2,22 +2,101 @@
...
+
+### <a href="#profile-creation-before" id="profile-creation-before" class="headerlink"></a> Before Creating a Profile
+
+Please **check whether an existing profile fits your needs or could be amended
+to fit your needs** before developing a new one.
+
+- If a suitable profile already exists, consider using it. Having fewer, more
+widely-deployed profiles makes it easier to create shared tooling.
+
+- If there's an existing profile that could be amented to fit your needs,
```
gabesullice:
```suggestion
- If there's an existing profile that could be amended to fit your needs,
```

extensions/index.md
```diff
@@ -2,22 +2,101 @@
...
+needed. Contact them through the information in their profile's registration
+and give them some time to reply.
+
+### <a href="#profile-creation" id="profile-creation" class="headerlink"></a> Authoring Your Profile
+
+To author your profile, [download and fill out the template](/profile_template.md).
+
+### <a href="#profile-register" id="profile-register" class="headerlink"></a> Register & Use Your Profile
+
+Once it's ready, submit your profile to the JSON:API profile registry by creating
```
gabesullice:
```suggestion
Once your profile is ready, submit it to the JSON:API profile registry by creating
```

extensions/index.md
```diff
@@ -2,22 +2,101 @@
...
+a pull request that contains your filled out template as an `index.md` file in a 
+uniquely-named folder in the [`_profiles`](https://github.com/json-api/json-api/tree/gh-pages/_profiles) 
+directory.
+   
+Your submission will be **approved within a week** by one of JSON API's
```
gabesullice: Really? I think that's an ambitious promise! Are you sure?

extensions/index.md
```diff
@@ -2,22 +2,101 @@
...
+feedback.
+
+One caveat: if your profile defines a new, fundamental mechanism for doing
+something "architectural" that other profiles may need to do too, it may be
+held for extra consideration. This extra review is designed to check that the
+proposed mechanism, were it to become a convention, wouldn't have problematic
+ramifications.
+
+Also, note that if the author of a registered profile dies, moves out of contact,
+or otherwise can't or won't make changes that are important to the community,
```
gabesullice: I would simply say "becomes non-responsive for an extended period of time"

extensions/index.md
```diff
@@ -2,22 +2,101 @@
...
+Once it's ready, submit your profile to the JSON:API profile registry by creating
+a pull request that contains your filled out template as an `index.md` file in a 
+uniquely-named folder in the [`_profiles`](https://github.com/json-api/json-api/tree/gh-pages/_profiles) 
+directory.
+   
+Your submission will be **approved within a week** by one of JSON API's 
+editors, assuming it meets the [profile extension requirements](/format/1.1/#profiles-authoring)
+and follows the template above. The community may give you feedback on your
+submission before it's approved, but you are not required to act on this 
+feedback.
```
gabesullice: To a casual reader I think this might be understood as: "you are not required to act on feedback for your submission to be approved"

Perhaps something more like this: "the community often provides feedback on new and existing profiles. You are encouraged to accept this feedback, unless it fundamentally alters your profile. This will help the community coalesce around fewer, more powerful profiles."

gabesullice (commented):
Haven't had a chance to read it all, but so far this looks great!

extensions/index.md
ethanresnick: Your suggestion has a typo (“submit it profile”); fix that and I’ll gladly pull this in.

extensions/index.md
ethanresnick: Yes. 

My biggest hope for profiles is that they will unblock the community to rapidly experiment with new features, without having to first build consensus for a particular design. The current consensus process has been a huge bottleneck imo.

So, I want these reviews to be quick and not too demanding of authors, which I think means limiting their scope. In particular, my vision for these reviews is that they assess only: 1) is this profile compliant with JSON:API's requirements for profiles?; and 2) if this profile did become widely adopted, would that cause any sort of problems? I don't think the goal should be to assess whether the proposed profile is well-designed or serves a meaningful use case, since that would take too long. If the result is that some poorly-designed profiles get registered, I don't think that's likely to be a big problem.

extensions/index.md
ethanresnick: > I think this might be understood as: "you are not required to act on feedback for your submission to be approved"

That's exactly what I'm trying to convey :) See my comment above.

extensions/index.md
gabesullice: Fair enough!

extensions/index.md
gabesullice: Whoops! 😅 

Fixed.

extensions/index.md
ethanresnick: I agree that this language could be a bit softer, if that's what you're going for.

For reference, I based it off the language used for media type registrations:

> The IESG may reassign responsibility for a media type. The most common case of this will be to enable changes to be made to types where the author of the registration has died, moved out of contact or is otherwise unable to make changes that are important to the community.

Maybe I should just copy that language directly?

gabesullice: I just don't think it's necessary to give specific examples at all. "Legally" it's better to define a class of scenarios which trigger an action than it is to define specific scenarios and risk missing something thereby forcing yourself to interpret "the spirit of the law".

extensions/index.md
gabesullice: WFM.

extensions/index.md
dgeb: Personally, I'm unable to make any promise to review profiles within a short timeframe, and I'm not sure it's desirable to force this either. If we or the community see an obvious design flaw in a profile, isn't it preferable to point out the flaw rather than to accept it? (But then again, I'm not comfortable allowing too much extra noise into this issue tracker by using it for extensive reviews.)

All of this to say that I feel that we shouldn't be guaranteeing "approval within a week" let alone "review within a week". I'm more comfortable saying something along the lines of "Your submission will be reviewed in a timely manner by JSON:API's editors and the community."

extensions/index.md
```diff
@@ -2,22 +2,101 @@
 layout: page
 title: Extensions
 show_sidebar: true
+redirect_from: /profiles
 ---
 
-## <a href="#status" id="status" class="headerlink"></a> Status
+JSON:API can be extended with profiles. These profiles enable an API to 
+provide clients with information or functionality beyond that described 
+in the base JSON API specification.
```
dgeb:
```suggestion
in the base JSON:API specification.
```

dgeb:
@ethanresnick I have a couple general concerns:

* Allowing registrations by name, such as `/profiles/cursor-pagination/`, provides an official level of approval that we won't be able to reverse. This seems especially dangerous when combined with an expedient review process that doesn't include a quality review.

* Even in the initial version, I'm not sure that profiles should be listed under `/extensions`. Perhaps we should just have a top-level `/profiles` index? And I'm not sure whether to keep the "Extensions" menu item at all. Perhaps linking to "how to create a profile" from `/profiles` would be adequate?

extensions/index.md
ethanresnick: Yes, if we see a glaring flaw in a profile's design, I definitely think we should point it out. And I imagine that, if we point out a major flaw, the profile's author will generally address it before registering the final version. 

I can clarify the text in this PR to say that, if the community gives feedback, the author can of course take time to act on that feedback, which then would prompt a new round of review, and that this whole process could take more than a week.

I'm really worried, though, about us coming to see the approval process as a "design review", because design reviews tend to take a lot of time/attention, and can get nitpicky. That's frustrating to authors and is likely not a good use of our very-limited bandwidth as editors.

Because profiles are decentralized, I don't see a lot of harm if a poorly-designed profile get registered; people can just not use that profile. It doesn't bind new APIs or hold back the spec. The second review criterion I proposed above ("if this profile did become widely adopted, would that cause any sort of problems?"), which I'm also happy to make more explicit in the PR's text, is meant to ensure that a registered profile wouldn't hold back the ecosystem either.

Given the limited harm, I'd much rather have some bad profiles get through, then have some basically well-designed profiles languish for weeks because we don't have time to review their designs. At least at first, people will likely be developing profiles for their project's immediate use case, and I don't think we can ask them to put their project on hold for more than a week while we're reviewing/asking for changes. 

If the issue is how to help implementers avoid using bad profiles, perhaps we can do something like link to the each profile's review thread next to any links to the its spec. Or, we can exercise our editorial discretion to highlight profiles we think our well designed by listing them above others. Whatever the approach, I don't think this is a hard problem to solve.

All told, I feel very strongly about the importance of quick review, and I'm happy to take on the task of reviewing all incoming profiles according to the criteria I gave above. If the one week timeline proves impossible, we can always change things then. 

Again, I think a critical value of profiles is their potential to unblock the community and get people sharing solutions — which I think means showing the profiles on jsonapi.org — and the last thing I want is for review to become a new bottleneck.

ethanresnick:
> Allowing registrations by name, such as `/profiles/cursor-pagination/`, provides an official level of approval that we won't be able to reverse. 

Is there an alternative to registration by name? The only ones I can think of seem worse (e.g., randomly assigning ids). Moreover, I'm pretty confident that we can decouple whether a profile has a nice URL from whether it seems to carry official approval. Right? After all, the profile's URL isn't shown until the pretty late in the profile discovery process — once the user is on the page for the profile's spec — so there are plenty of UI opportunities before then for us to highlight choice profiles and downplay poor ones. Or, for profiles where we don't want to takes an editorial view about their quality, we could even (if we really wanted to) randomize the order in they appear on the site, e.g.

> Even in the initial version, I'm not sure that profiles should be listed under `/extensions`. Perhaps we should just have a top-level `/profiles` index? 

Are you saying to the URL of the page in this PR should be `/profiles` instead of `/extensions`? If so, I'm 100% on board with that, and can update it accordingly.

> And I'm not sure whether to keep the "Extensions" menu item at all. Perhaps linking to "how to create a profile" from /profiles would be adequate?

I'm not sure I understand this. If we didn't have a top-level menu item, how would people find registered profiles? Are you saying you don't want to show the list of registered profiles on the site? If so, why?

dgeb:
> Are you saying to the URL of the page in this PR should be /profiles instead of /extensions? If so, I'm 100% on board with that, and can update it accordingly.

Great, sounds like we're in agreement 👍 

> I'm not sure I understand this. If we didn't have a top-level menu item, how would people find registered profiles? Are you saying you don't want to show the list of registered profiles on the site? If so, why?

I was just saying that we drop the `Extensions` menu item and replace it with `Profiles`. And I would prefer that the content at `/profiles` focus on listing the registered profiles rather than discussing the registration process. We can provide a short section at the top that links to more details about profile registration. So I'm proposing a shuffle of the content you have currently, not the removal of any of it.

dgeb:
> Moreover, I'm pretty confident that we can decouple whether a profile has a nice URL from whether it seems to carry official approval. Right? After all, the profile's URL isn't shown until the pretty late in the profile discovery process — once the user is on the page for the profile's spec — so there are plenty of UI opportunities before then for us to highlight choice profiles and downplay poor ones. 

One option here would be to have two classes of profiles - "official" and "community". Each would reflect their "class" in the directory, but not in the URL (otherwise a community profile could never be "promoted"). In this way we could allow profiles to prove themselves as community profiles first, and then eventually promote the most useful and popular ones.

_Follow up:_ I think the major weakness to my proposal is that we have to kind of "prejudge" the quality of community profiles to decide if they're solid enough to grab a significant URL. Maybe a good approach here would be to prefix all profile URLs with an index, such as the PR number used to add them. This approach is taken in the emberjs/rfcs repo - the only downside is that it requires a manual edit step upon merging.

ethanresnick:
> I was just saying that we drop the Extensions menu item and replace it with Profiles.

For context, I kept the menu item as "Extensions" because I thought the term "Profiles" would be unfamiliar to someone new to JSON:API. Basically, I wanted someone who'd just discovered the JSON:API site to see "Oh, there's this nice spec that covers common cases, and also all these off-the-shelf extensions". 

Once people are familiar with the spec and know what a profile is, I don't think label "Extensions" will confuse them — plus, we say as the first sentence on that page "JSON:API can be extended with profiles" — so I figured that directing the label at new users made more sense. Thoughts on this approach?

> I would prefer that the content at `/profiles` focus on listing the registered profiles rather than discussing the registration process. We can provide a short section at the top that links to more details about profile registration. 

I'm not necessarily opposed to this, but I don't really understand what problem it would be solving. Can you explain the motivation? My biggest goal is to keep the registration process easy to discover, and I think having the link to it in the main Extensions page's left nav is a pretty prominent place for it.

> Maybe a good approach here would be to prefix all profile URLs with an index, such as the PR number used to add them. This approach is taken in the emberjs/rfcs repo - the only downside is that it requires a manual edit step upon merging.

I think that could work. My one concern is that it could be tricky to author a profile when you don't know what its URL is gonna end up being. E.g., in my cursor pagination example profile, there are hardcoded references to various error URLs in the spec, but those URLs wouldn't exist at spec authoring time with this approach. So either the editors would have to edit all those URLs upon merging too, or encourage authors to use the `{% include profile_url.md page=page %}` helper in their spec text... but that seems very cumbersome.

(Also, as an aside, I'd love for the index to be a different path segment, rather than a hyphen-separated prefix — e.g. `/profiles/123/cursor-pagination` instead of `/profiles/123-cursor-pagination` — although that would complicate the merge step further.)

Another possible approach would be to use the author's github name as a namespace. So, the URL (and directory structure) would be, e.g., `/profiles/dgeb/cursor-pagination`. This has the huge advantage of no merge step and predictable URLs for authoring, since Github usernames are already a managed namespace for us. But idk how you or others feel about having the username in the URL; personally, I see pros and cons. Thoughts?

gabesullice:
> I kept the menu item as "Extensions" because I thought the term "Profiles" would be unfamiliar to someone new to JSON:API.

+1

>> I would prefer that the content at `/profiles` focus on listing the registered profiles rather than discussing the registration process. We can provide a short section at the top that links to more details about profile registration.

> I'm not necessarily opposed to this, but I don't really understand what problem it would be solving. Can you explain the motivation?

I think I agree with @dgeb. The vast majority of users will be searching for applicable profiles and how to use them, not how to register new ones. If you (1) understand profiles (2) want to author one, I think you're already familiar/motivated enough to go looking for a link.

> Another possible approach would be to use the author's github name as a namespace. ... But idk how you or others feel about having the username in the URL; personally, I see pros and cons. Thoughts?

This is where my mind went as well. Having a deterministic namespace prior to authoring the profile is desirable. However, I think it's important that there be _a_ namespace, but not that their be a strict definition of what the namespace must be.

The language could be something like this: "a profile URL must take the form `/profiles/{namespace-id}/{profile-short-name}` where `{namespace-id}` is a URL safe string which uniquely identifies either the author, organization or project registering the profile." After that, provide a recommendation that one use a Github username/org, email, personal/company TLD, etc.

For example, I'd like to author profiles that we've developed with lots of community involvement. I would not like to tie that profile to my name because I don't deserve all the credit. So, in my case, I'd like to be able to submit a profile like `/profiles/drupal/fancy-filters` instead of `/profiles/gabesullice/fancy-filters`.

dgeb:
I talked with @ethanresnick and we agree with the basic form of your proposal @gabesullice 

> The language could be something like this: "a profile URL must take the form /profiles/{namespace-id}/{profile-short-name} where {namespace-id} is a URL safe string which uniquely identifies either the author, organization or project registering the profile." 

We're going to strongly recommend that the author use either their username or an org to which they belong (which are both easy to verify of course) for a namespace. Exceptions can be allowed for unusual cases.

We also discussed introducing a new `Guide` section to consolidate a number of sections covering usage of the spec - like Recommendations and Examples. We can also provide directions for creating and registering profiles in that section. But this can be done after this PR is merged as a general content reorg - so no need to block.

extensions/index.md
ethanresnick: See updated text.

extensions/index.md
ethanresnick: See updated test.

ethanresnick:
Ok y'all, I've updated the PR per the comments above and my discussion with @dgeb. See the most recent commit.

dgeb (approved):

ethanresnick:
Beautiful. Rebased to keep only the relevant/distinct commits and merged.

PR #1314, opened 2018-10-29
Remove deprecated package

lindyhopchris:
Remove the `cloudcreativity/json-api` package from the list of PHP server implementations as it is deprecated.

For info I'm the maintainer of the deprecated package!


ethanresnick:
Thanks!

PR #1325, opened 2018-11-13
JsonApiParser is deprecated

philsturgeon:
Their README says the project has been deprecated. 

They advertised jsonapi-rb / jsonapi-parser as a replacement, but that's more of a validator than a usable client-library, so I just deleted this entry.


ethanresnick:
Ping @beauby. If we remove this, do you want to link any replacements instead?

Sonberg:
I made one that could replace JsonApiParser. https://github.com/json-api/json-api/pull/1586

jelhan (approved):

PR #1326, opened 2018-11-14
Add reststate implementations

CodingItWrong:
This adds a few new JavaScript client libraries I've been working on: a low-level stateless `client` library, and higher-level libraries that store state in the `mobx` and `vuex` state stores.


ethanresnick:
Thanks!

PR #1327, opened 2018-11-15
Fix Reststate typo

CodingItWrong:
Fixes a typo in "JSON:API" in my newly-added implementation entries. I want to at least be consistent across the three of those! :-D


dgeb:
@CodingItWrong thanks for the attention to detail 💯

PR #1328, opened 2018-11-30
1.1: update site's extensions page + announce release date

ethanresnick:
Note: only the last two commits are new here. (The others were already reviewed and merged in #1311, we just merged them into 1.1 branch rather than gh-pages.)


kocsismate:
Just a question: shouldn't the release of v1.1 wait until implementations start supporting the new features? This way, people could play with profiles and see if it is really usable in the wild.

I say this because this is what the standard committee of PHP (PHP-FIG) also does: they require 2+ implementations before a standard can be accepted, and I believe it's very useful to test ideas before setting them in stone.

Anyway, I have already started the adaption of JSON:API 1.1 in my PHP projects:
- Client-side: https://github.com/woohoolabs/yang (I believe there is not much to do before a beta release)
- Server-side: https://github.com/woohoolabs/yin/tree/4.0 (there are still plenty of things to do)

dgeb:
@kocsismate Thanks for raising this point. In the past, we've discussed adopting the same policy for finalizing new versions of the spec; specifically, at least two compliant implementations must exist for both the client and server side. I still  think any final release date should be contingent upon this requirement. With that said, I think it's also important to set a target date to provide some motivation for implementors.

ethanresnick:
> Just a question: shouldn't the release of v1.1 wait until implementations start supporting the new features? This way, people could play with profiles and see if it is really usable in the wild.

In short, yes. 

My thinking was that 6 weeks should be enough time for that (especially as some implementers have already been working with profiles since #1268 was merged). However, I'm happy to use the criteria @dgeb proposed, and push the date back if we don't have two compliant implementations by January 15.

ethanresnick:
Also, I've push another commit here that adds a changelog entry for the new features/changes in 1.1, since I imagine that will be helpful to implementers. (I'm pretty sure this list is complete, as I went through all the diffs since 1.0 came out.)

ethanresnick:
Also, per conversation with @dgeb, I've pushed the date back to 1/31 and added an explicit mention of the two implementation requirement everywhere that the date occurs.

dgeb:
Great job @ethanresnick - 1.1 here we come! 🎉

wimleers:
>  require 2+ implementations before a standard can be accepted, and I believe it's very useful to test ideas before setting them in stone.

I too like this :)

PR #1330, opened 2018-12-03
feat(format): better define unsupported client-required profile error

ethanresnick:
When the client requires a profile that the server doesn't support (using the `?profile` param), return a more-specific error response leveraging the new `type` mechanism.


dgeb (approved):
This seems good to me. It's a nice use of the new error `type` and more useful to a client than a plain `400` response.

ethanresnick:
👍

wimleers:
Nice!

I totally agree with
> It's a nice use of the new error `type` and more useful to a client than a plain `400` response.

I'm looking forward to updating the Drupal JSON:API implementations with concrete `type` URLs — we already have helpful messages, but this will make those errors easier to document!

ethanresnick:
> I'm looking forward to updating the Drupal JSON:API implementations with concrete type URLs — we already have helpful messages, but this will make those errors easier to document!

@wimleers Same. I'd love to create a set of error URIs on jsonapi.org that show canonical docs when dereferenced. Then, all the server implementations can point to those and get docs for free. I'm gonna open a PR with the errors that my implementation already has, and give you push access to the branch... hopefully you can add to it any errors from Drupal's implementation.

PR #1331, opened 2018-12-05
Profile changelog

ethanresnick:
Link to the git history for each registered profile from its profile spec page.

I realized that it might be marginally useful for implementors to know what a profile looked like at a given point in time. (E.g., so that they can choose not to implement support for specific profile features if they only care about interacting with a server that they know was built before those features existed.)

I'm not sure if this is the best way to handle this, though, compared to, e.g. a changelog whose text is directly part of the spec page. I'm interested in what others think.

### DIFF
--- a/_layouts/profile.html
+++ b/_layouts/profile.html
@@ -47,6 +47,12 @@
           {% include header_offset_2.md content=content %}
         </section>
         <section>
+          <h2><a href="#profile-versions" id="profile-versions" class="headerlink"></a>Version History</h2>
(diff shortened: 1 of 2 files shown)


wimleers:
Ideally we'd get a list of all published versions of a particular profile, along with diffs between each version pair.

ethanresnick:
> Ideally we'd get a list of all published versions of a particular profile, along with diffs between each version pair.

I think if we always squash and merge any PRs adding or revising a profile, the git history should offer that, right? So we just have to be consistent about squash and merge.

wimleers:
Indeed!

Issue #1332, opened 2018-12-05
Allow duplicated resources in compound documents?


ethanresnick:
Right now, the spec forbids multiple resource objects with same (type, id) from appearing together in a compound document. I think the theory is that allowing duplication could result in the same resource being sent with two different sets of fields: 

> Note: This approach ensures that a single canonical resource object is returned with each response, even when the same resource is referenced multiple times.

Also, deduping saves data over the wire.

However, neither of those rationale make much sense to me: if the spec can require that a resource not appear twice, it could just as easily require that, if it appears more than once, it must always appear with identical content; and, with gzipping, the extra data from duplication is probably not much of a problem.

Moreover, my sense is that this forced de-duping comes with a real cost for server implementations, which have to track all the resources in the response document as they build it, or do an O(n) traversal through the response before sending it. As an implementer, this has always seemed like more trouble than it's worth, and I've simply skipped doing that check.

So, what are we actually getting from this restriction? Is it really worth keeping?

Btw, re the spec requiring instead that "if a resource object appears more than once, it must always appear with identical content", that's only one option (and would be very easy for servers that do all their queries inside a transaction to guarantee, without extra overhead). Another option would simply be to tell the client to take the representation that occurs latest in the document.

kocsismate:
Just another implementor's experience: in my server implementation, I solved this problem efficiently using a Map data structure where keys are the combination of the types and IDs (https://github.com/woohoolabs/yin/blob/4ad1bbe120ba4ba2e21dde77e63deb854cc87bc4/src/JsonApi/Schema/Data/AbstractData.php#L114).

Personally, I like this forced de-duping (however, it wasn't easy to achieve). But I think I'd have to track the resources anyway even if this rule wasn't in effect, because multiple resources can include the same related resource, so it makes sense to only do so once, and tracking is the way to go as far as I know.

ethanresnick:
> I solved this problem efficiently using a Map data structure... I think I'd have to track the resources anyway even if this rule wasn't in effect, because multiple resources can include the same related resource, so it makes sense to only do so once

Even using a Map, you're still doing one check for every resource in the document (it's just that now that check is ~O(1) instead of O(n)). Maybe these checks are worth it; as you imply, there could be pretty big duplication otherwise, but I'm still not sure. I appreciate more implementer perspectives, though... that was my hope in opening this issue

wimleers:
[...]

PR #1333, opened 2018-12-13
Resource Versioning Profile

gabesullice:
First! 😁 (well, sorta... @ethanresnick submitted one first, but does that really count?)

This is a profile extracted from Drupal's JSON:API implementation.

Drupal has long had support for moderated and revisioned content. We plan to support this in our project and we hope to land [the feature quite soon](https://www.drupal.org/project/jsonapi/issues/2992833)—we're just cleaning up nitpicks and then it will land after we're out of a feature freeze (we're stabilizing a v2 right now).

We figured this would be a great way to pilot the new profile process.

<hr/>

I tried to make this profile very flexible while taking _heavy_ inspiration from [RFC 5829 - Link Relation Types for Simple Version Navigation between Web Resources](https://tools.ietf.org/html/rfc5829).

~This PR is still a work-in-progress and is not yet finalized, but I'm hoping to get a jump on any egregious oversights.~

~Before it is finalized, I would like for this profile to also include a set of recommended links and for it to include language about providing a version history endpoint. I don't think either of those things will be a requirement for adopting the profile though.~ Done!

/cc @wimleers @e0ipso

### DIFF
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -122,11 +122,16 @@ parameter value is known as the _version identifier_.
 
 ## Server Responsibilities
 
-A server **MUST** respond with `400 Bad Request` if a version negotiator is not
-supported.
+<a id="bad-version-negotiator"></a>A server **MUST** respond with `400 Bad Request` if a version negotiator is not
+supported. In this case, an error object that includes a `type` link to
+`https://jsonapi.org/profiles/drupal/resource-versioning#bad-version-negotiator`
+**MUST** be included in the response document.
    
-If a server cannot process the given version argument for the given negotiation
-mechanism, it **MUST** respond with a `400 Bad Request`.
+<a id="bad-version-argument"></a>If a server cannot process the given version argument for the given negotiation
+mechanism, it **MUST** respond with a `400 Bad Request`. In this case, an error
+object that includes a `type` link to
+`https://jsonapi.org/profiles/drupal/resource-versioning#bad-version-argument`
+**MUST** be included in the response document.
 
 If a server is able to process the version argument but an appropriate version
 cannot be located, the server **MUST** respond with a `404 Not Found`.
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -124,13 +124,13 @@ parameter value is known as the _version identifier_.
 
 <a id="bad-version-negotiator"></a>A server **MUST** respond with `400 Bad Request` if a version negotiator is not
 supported. In this case, an error object that includes a `type` link to
-`https://jsonapi.org/profiles/drupal/resource-versioning#bad-version-negotiator`
+`https://jsonapi.org/profiles/drupal/resource-versioning/#bad-version-negotiator`
 **MUST** be included in the response document.
    
 <a id="bad-version-argument"></a>If a server cannot process the given version argument for the given negotiation
 mechanism, it **MUST** respond with a `400 Bad Request`. In this case, an error
 object that includes a `type` link to
-`https://jsonapi.org/profiles/drupal/resource-versioning#bad-version-argument`
+`https://jsonapi.org/profiles/drupal/resource-versioning/#bad-version-argument`
 **MUST** be included in the response document.
 
 If a server is able to process the version argument but an appropriate version
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -294,8 +294,8 @@ identifier, the server **MUST** respond with a `501 Not Implemented`:
   - `prior-working-copy`
   - `subsequent-working-copy`
 
-If any version argument is receieved other than the version arguments in this
-section, a server **MUST** respond appropriately for a [bad version argument](#bad-version-argument).
-
 > Note: Future versions of this profile may define the behavior of these version
 > arguments.
+
+If any version argument is receieved other than the version arguments in this
+section, a server **MUST** respond appropriately for a [bad version argument](#bad-version-argument).
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -21,17 +21,17 @@ minimum_jsonapi_version_explanation:
 discussion_url: https://www.drupal.org/project/issues/jsonapi
 
 editor_name: |
- Gabriel Sullice
- Mateu Aguiló Bosch
- Wim Leers
+  Gabriel Sullice
+  Mateu Aguiló Bosch
+  Wim Leers
 editor_email: |
- gabriel@sullice.com
- mateu.aguilo.bosch@gmail.com
- work@wimleers.com
+  gabriel@sullice.com
+  mateu.aguilo.bosch@gmail.com
+  work@wimleers.com
 editor_website:
- n/a
- http://humanbits.es/
- https://wimleers.com
+  n/a
+  http://humanbits.es/
+  https://wimleers.com
 
 categories:
   - Resource Versioning
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -87,8 +87,6 @@ it:
 
   - The server’s response **MUST** contain the most appropriate version of the
     resource requested.
-  - The server **MUST NOT** include a version of a resource inappropriate for the
-    requested version.
   - The server **MUST** respond with `404 Not Found` if an appropriate version of
     the resource requested cannot be located. 
     
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -80,7 +80,9 @@ An endpoint **MAY** support a `resourceVersion` query parameter to allow a
 client to indicate which version(s) of a resource should be returned.
 
 If an endpoint does not support the `resourceVersion` parameter, it **MUST**
-respond with `400 Bad Request` to any requests that include it.
+respond with `400 Bad Request` to any requests that include it. An error object
+detailing the source of the error **SHOULD** include a `source` member referencing
+the `resourceVersion` query parameter.
 
 If an endpoint supports the `resourceVersion` parameter and a client supplies
 it:
@@ -127,12 +129,16 @@ identifier_.
 <a id="bad-version-negotiator"></a>A server **MUST** respond with `400 Bad Request` if a version negotiator is not
 supported. An error object detailing the source of the error **SHOULD** include a
 `type` link to
-`https://jsonapi.org/profiles/drupal/resource-versioning/#bad-version-negotiator`.
+`https://jsonapi.org/profiles/drupal/resource-versioning/#bad-version-negotiator`
+and **SHOULD** include a `source` member referencing the `resourceVersion` query
+parameter.
    
 <a id="bad-version-argument"></a>If a server cannot process the given version argument for the given negotiation
 mechanism, it **MUST** respond with a `400 Bad Request`. An error object detailing the
 source of the error **SHOULD** include a `type` link to
-`https://jsonapi.org/profiles/drupal/resource-versioning/#bad-version-argument`.
+`https://jsonapi.org/profiles/drupal/resource-versioning/#bad-version-argument`
+and **SHOULD** include a `source` member referencing the `resourceVersion` query
+parameter.
 
 If a server is able to process the version argument but an appropriate version
 cannot be located, the server **MUST** respond with a `404 Not Found`.
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -96,12 +96,13 @@ it:
 ## Format
   
 The value of the `resourceVersion` parameter **MUST** be a colon-separated
-(U+003A COLON, “:”) string. The first segment of the string **SHOULD** be
-interpreted as an identifier for a _version negotiation mechanism_. A version
-negotiation mechanism defines how a server will locate an appropriate resource
-version. Subsequent segments of the string **SHOULD** be interpreted as version
-negotiation arguments for the preceding mechanism. Collectively, this query
-parameter value is known as the _version identifier_.
+(U+003A COLON, “:”) string composed of one or more segments. The first segment
+of the string **SHOULD** be interpreted as an identifier for a _version
+negotiation mechanism_. A version negotiation mechanism defines how a server will
+locate an appropriate resource version. Subsequent segments of the string
+**SHOULD** be interpreted as version negotiation arguments for the preceding
+mechanism. Collectively, this query parameter value is known as the _version
+identifier_.
 
 > Note: For example, a server may support both ID-based and time-based
 > mechanisms for requesting a resource version. The former mechanism would be
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -114,14 +114,14 @@ identifier_.
 > attempt to define every possible mechanism for versioning resources.
 
 ```
-                   version-identifier
-                   _______|_________
-                  /                \
+                 version-identifier
+                  _______|________
+                 /                \
 ?resourceVersion=rel:latest-version
-                  \_/ \____________/
-                   |        |
-         version-negotiator |
-                     version-argument
+                 \_/ \____________/
+                  |        |
+        version-negotiator |
+                    version-argument
 ```
 
 ## Server Responsibilities
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -177,7 +177,7 @@ same:
 ```
 
 A server **MAY** provide a `version-history` link in a resource object's links
-object if a [version history](version-history) endpoint is available for the context resource
+object if a [version history](#version-history) endpoint is available for the context resource
 object.
 
 A server **MAY** provide the following resource object links so that a client may
@@ -293,6 +293,9 @@ identifier, the server **MUST** respond with a `501 Not Implemented`:
   - `successor-version`
   - `prior-working-copy`
   - `subsequent-working-copy`
+
+If any version argument is receieved other than the version arguments in this
+section, a server **MUST** respond appropriately for a [bad version argument](#bad-version-argument).
 
 > Note: Future versions of this profile may define the behavior of these version
 > arguments.
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -100,8 +100,8 @@ it:
   
 The value of the `resourceVersion` parameter **MUST** be a colon-separated
 (U+003A COLON, “:”) string composed of one or more segments. The first segment
-of the string **SHOULD** be interpreted as an identifier for a _version
-negotiation mechanism_. A version negotiation mechanism defines how a server will
+of the string **MUST** be interpreted as an identifier for a _version negotiation
+mechanism_. A version negotiation mechanism defines how a server will
 locate an appropriate resource version. Subsequent segments of the string
 **SHOULD** be interpreted as version negotiation arguments for the preceding
 mechanism. Collectively, this query parameter value is known as the _version
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -63,7 +63,7 @@ would have been known as the working copy as each one was created.
 If the version is then checked out for further changes each new revision becomes
 the working copy. When the working copy is checked in and made the default
 revision, this revision becomes the latest version and the version before it is
-its _predecessor version _. If no other revisions come after the latest version,
+its _predecessor version_. If no other revisions come after the latest version,
 it may also be considered the working copy.
 
 Sophisticated servers may support multiple versions of a resource as well as
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -203,14 +203,14 @@ links could be provided:
 
 | Revision | `latest-version` | `working-copy` | `predecessor-version` | `successor-version` | `prior-working-copy` | `subsequent-working-copy` |
 | :------: | :--------------: | :------------: | :-------------------: | :-----------------: | :------------------: | :-----------------------: |
-| `a`      | `g`              | `h`            | no link               | `e`                 | no link              | `b`                       |
+| `a`      | `g`              | `h`            | no link               | `e`                 | no link              | `b`, `c`                  |
 | `b`      | `g`              | `h`            | `a`                   | `e`                 | `a`                  | `c`                       |
 | `c`      | `g`              | `h`            | `a`                   | `e`                 | `b`                  | `d`                       |
-| `d`      | `g`              | `h`            | `a`                   | `e`                 | `c`                  | `e`                       |
+| `d`      | `g`              | `h`            | `a`                   | `e`                 | `b`, `c`             | `e`                       |
 | `e`      | `g`              | `h`            | `a`                   | `g`                 | `d`                  | `f`                       |
 | `f`      | `g`              | `h`            | `e`                   | `g`                 | `e`                  | `g`                       |
 | `g`      | no link          | `h`            | `e`                   | no link             | `f`                  | `h`                       |
-| `h`      | `g`              | no link        | `e`                   | no link             | `g`                  | no link                   |
+| `h`      | `g`              | no link        | `g`                   | no link             | `g`                  | no link                   |
 
 > Note: In this example, `f` has both a `predecessor-version` and
 > `prior-working-copy` link to `e` because `e` was a version but it also was the
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -60,8 +60,11 @@ before the resource was published may have many revisions. Only the published
 revision would be considered a version. The revisions prior to first version
 would have been known as the working copy as each one was created.
 
-If the version is then checked out for further changes (which may or may not be
-published) each new revision becomes the working copy.
+If the version is then checked out for further changes each new revision becomes
+the working copy. When the working copy is checked in and made the default
+revision, this revision becomes the latest version and the version before it is
+its _predecessor version _. If no other revisions come after the latest version,
+it may also be considered the working copy.
 
 Sophisticated servers may support multiple versions of a resource as well as
 multiple working copies of a resource (e.g., to support multiple
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -131,13 +131,6 @@ mechanism, it **MUST** respond with a `400 Bad Request`.
 If a server is able to process the version argument but an appropriate version
 cannot be located, the server **MUST** respond with a `404 Not Found`.
 
-If a server knows a version argument to be invalid for the requested version
-negotiation mechanism, it **MUST** respond with a `400 Bad Request`.
-
-If the server cannot establish a version argument to be invalid for the
-requested version negotiation mechanism, it **MUST** respond with a
-`501 Not Implemented`.
-
 # Version Negotiators
 
 ## ID-Based Version Negotiator
@@ -171,3 +164,11 @@ The `rel` version negotiator has the following valid version argument strings:
   - `latest-version`: requests the latest default revision of a resource.
   - `working-copy`: requests revision of a resource to which changes can be
     made.
+
+If any of the following version arguments is received in a `rel`-based version
+identifier, the server **MUST** respond with a `501 Not Implemented`:
+
+  - `predecessor-version`
+  - `successor-version`
+  - `prior-working-copy`
+  - `subsequent-working-copy`
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -167,8 +167,8 @@ same:
 ```
 
 A server **MAY** provide a `version-history` link in a resource object's links
-object. This link must reference a collection resource providing all revisions
-of the context resource object.
+object if a [version history](version-history) endpoint is available for the context resource
+object.
 
 A server **MAY** provide the following resource object links so that a client may
 navigate a resource object's version history:
@@ -201,7 +201,8 @@ a           e       g
 No other revisions were ever the latest version. In this example, the following
 links could be provided:
 
-| revision | `latest-version` | `working-copy` | `predecessor-version` | `successor-version` | `prior-working-copy` | `subsequent-working-copy` |
+| Revision | `latest-version` | `working-copy` | `predecessor-version` | `successor-version` | `prior-working-copy` | `subsequent-working-copy` |
+| :------: | :--------------: | :------------: | :-------------------: | :-----------------: | :------------------: | :-----------------------: |
 | `a`      | `g`              | `h`            | no link               | `e`                 | no link              | `b`                       |
 | `b`      | `g`              | `h`            | `a`                   | `e`                 | `a`                  | `c`                       |
 | `c`      | `g`              | `h`            | `a`                   | `e`                 | `b`                  | `d`                       |
@@ -218,7 +219,7 @@ links could be provided:
 # Version History
 
 A server **MAY** provide a "version history" endpoint. The primary data
-of a response document from a version history endpoing must be a collection of
+of a response document from a version history endpoint must be a collection of
 resource objects.
 
 Unless an `id` contains version information, the `type` and `id` members of each
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -127,15 +127,14 @@ identifier_.
 ## Server Responsibilities
 
 <a id="bad-version-negotiator"></a>A server **MUST** respond with `400 Bad Request` if a version negotiator is not
-supported. In this case, an error object that includes a `type` link to
-`https://jsonapi.org/profiles/drupal/resource-versioning/#bad-version-negotiator`
-**MUST** be included in the response document.
+supported. An error object detailing the source of the error **SHOULD** include a
+`type` link to
+`https://jsonapi.org/profiles/drupal/resource-versioning/#bad-version-negotiator`.
    
 <a id="bad-version-argument"></a>If a server cannot process the given version argument for the given negotiation
-mechanism, it **MUST** respond with a `400 Bad Request`. In this case, an error
-object that includes a `type` link to
-`https://jsonapi.org/profiles/drupal/resource-versioning/#bad-version-argument`
-**MUST** be included in the response document.
+mechanism, it **MUST** respond with a `400 Bad Request`. An error object detailing the
+source of the error **SHOULD** include a `type` link to
+`https://jsonapi.org/profiles/drupal/resource-versioning/#bad-version-argument`.
 
 If a server is able to process the version argument but an appropriate version
 cannot be located, the server **MUST** respond with a `404 Not Found`.
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -232,6 +232,19 @@ If provided, resource objects' `self` links **MUST NOT** be the same.
 
 # Version Negotiators
 
+This profile defines a number of version negotiators which a server **MAY**
+implement. Additional version negotiator's may be added to this profile at a
+later date.
+
+Version negotiators which are not defined by this profile **MUST** adhere to the
+same constraints as [implementation-specific query parameter names](https://jsonapi.org/format/1.1/#query-parameters-custom).
+
+It is **RECOMMENDED** that any alternative version negotiators be added to this
+profile. New version negotiators may be registered by sending one, joint email
+to the profile editors with the subject line: "New JSON:API Resource Version
+Negotiator: {negotiator name}". This will begin a process of refinement and/or
+result in a determination of fitness for addition to this profile.
+
 ## ID-Based Version Negotiator
 
 This profile establishes the `id` version negotiator. An `id`-based version
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -73,13 +73,13 @@ schemes.
 
 ## Usage
 
-An endpoint **MAY** support a `resource_version` query parameter to allow a
+An endpoint **MAY** support a `resourceVersion` query parameter to allow a
 client to indicate which version(s) of a resource should be returned.
 
-If an endpoint does not support the `resource_version` parameter, it **MUST**
+If an endpoint does not support the `resourceVersion` parameter, it **MUST**
 respond with `400 Bad Request` to any requests that include it.
 
-If an endpoint supports the `resource_version` parameter and a client supplies
+If an endpoint supports the `resourceVersion` parameter and a client supplies
 it:
 
   - The server’s response **MUST** contain the most appropriate version of the
@@ -95,7 +95,7 @@ it:
 
 ## Format
   
-The value of the `resource_version` parameter **MUST** be a colon-separated
+The value of the `resourceVersion` parameter **MUST** be a colon-separated
 (U+003A COLON, “:”) string. The first segment of the string **SHOULD** be
 interpreted as an identifier for a _version negotiation mechanism_. A version
 negotiation mechanism defines how a server will locate an appropriate resource
@@ -113,7 +113,7 @@ parameter value is known as the _version identifier_.
                    version-identifier
                    _______|_________
                   /                \
-?resource_version=rel:latest-version
+?resourceVersion=rel:latest-version
                   \_/ \____________/
                    |        |
          version-negotiator |
@@ -138,13 +138,13 @@ cannot be located, the server **MUST** respond with a `404 Not Found`.
 
 # Links
 
-When a server processes a request with a `resource_version` query parameter and
+When a server processes a request with a `resourceVersion` query parameter and
 a `self` link is provided for a top-level links object, the link's `href`
-**MUST** include the `resource_version` query parameter with the same version
+**MUST** include the `resourceVersion` query parameter with the same version
 identifier that was requested.
 
-When a server processes a request with a `resource_version` query parameter
-all resource object `self` links **SHOULD** contain a `resource_version` query
+When a server processes a request with a `resourceVersion` query parameter
+all resource object `self` links **SHOULD** contain a `resourceVersion` query
 parameter which identifies the specific revision represented by that resource
 object.
 
@@ -157,11 +157,11 @@ same:
     "type": "article",
     "id": 1,
     "links": {
-      "self": "/article/1?resource_version=id:42"
+      "self": "/article/1?resourceVersion=id:42"
     }
   },
   "links": {
-    "self": "/article/1?resource_version=rel:latest-version"
+    "self": "/article/1?resourceVersion=rel:latest-version"
   }
 }
 ```
--- a/_profiles/drupal/resource-versioning/index.md
+++ b/_profiles/drupal/resource-versioning/index.md
@@ -1,0 +1,166 @@
+---
+name: Resource Versioning
+short_description: |
+  Defines a protocol for requesting versioned resources via a JSON:API
+  server.
+
+extended_description: |
+  # Overview
+  JSON:API servers that are capable of tracking a record of changes to a
+  resource object may afford the capability of requesting a resource object
+  as it existed in a prior or successive state. This capability is herein
+  defined as _resource versioning_.
+  
+  This profile establishes a protocol for resource versioning by defining a
+  query parameter and its semantics in order to identify arbitrary revisions
+  of a resource and how a server should interpret this query parameter.
+
+minimum_jsonapi_version: 1.0
+minimum_jsonapi_version_explanation:
+
+discussion_url: https://www.drupal.org/project/issues/jsonapi
+
+author_name: Gabriel SULLICE
+author_email: gabriel@sullice.com
+
+categories:
+  - Resource Versioning
+---
+  
+# Concepts
+Resources on a server may undergo changes and the state of a resource with an
+arbitrary number of changes may be accessible. It is often useful to retrieve
+resources as they existed at the time of their creation or in various states of
+change for editorial or archival purposes. This profile establishes a protocol
+for accessing resources in those various states.
+
+For the purpose of understanding this profile, it is helpful to establish a
+vocabulary for describing various possible states of a resource.
+
+A _revision_ is to be understood as an identifiable state of a resource after
+its creation or after some number of changes.
+
+A _version_ is to be understood as a revision of a resource that is available,
+or was previously available, without any version negotiation. In other words, as
+a revision that is or was the default revision of a resource.
+
+A _working copy_ is the revision to which new changes can be made or to which
+they will be applied. Colloquially, a working copy is often thought of as the
+"tip" of a version history.
+
+For example, a resource which existed as a draft and underwent multiple changes
+before the resource was published may have many revisions. Only the published
+revision would be considered a version. The revisions prior to first version
+would have been known as the working copy as each one was created.
+
+If the version is then checked out for further changes (which may or may not be
+published) each new revision becomes the working copy.
+
+Sophisticated servers may support multiple versions of a resource as well as
+multiple working copies of a resource (e.g., to support multiple
+languages or within a version control system which supports multiple branches).
+
+This profile creates a standard for JSON:API to support these diverse versioning
+schemes.
+# Query Parameter
+
+## Usage
+
+An endpoint **MAY** support a `resource_version` query parameter to allow a
+client to indicate which version(s) of a resource should be returned.
+
+If an endpoint does not support the `resource_version` parameter, it **MUST**
+respond with `400 Bad Request` to any requests that include it.
+
+If an endpoint supports the `resource_version` parameter and a client supplies
+it:
+
+  - The server’s response **MUST** contain the most appropriate version of the
+    resource requested.
+  - The server **MUST NOT** include a version of a resource inappropriate for the
+    requested version.
+  - The server **MUST** respond with `404 Not Found` if an appropriate version of
+    the resource requested cannot be located. 
+    
+> Note: This means that a server should not provide "fallbacks" unless the
+> behavior is well defined by the version negotiation mechanism (see below).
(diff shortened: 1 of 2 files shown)


gabesullice:
Ping @effulgentsia

e0ipso:
One thing that we tried (or at least I certainly kept that in mind) is the ability to negotiate versions based on flexible criteria. Different applications will have a slightly different concept of versioning. Some may want to negotiate a resource based on revision-id, others based on relative position to the published version, others based on a date time authoring, etc.

The proposed solution in the Drupal module will take all that into account. That may make this profile adequate for other JSON:API implementations, other than Drupal's. If other projects adopt this as well, I'm sure clients and tooling will follow.

I think that standard clients and common tooling is key for the success of the ecosystem, like GraphiLQ and Apollo have proven in the GraphQL ecosystem.

wimleers:
Can't wait to get feedback from @ethanresnick , @dgeb and others!

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,176 @@
...
+  This profile establishes a protocol for resource versioning by defining a
+  query parameter and its semantics in order to identify arbitrary revisions
+  of a resource and how a server should interpret this query parameter.
+
+minimum_jsonapi_version: 1.0
+minimum_jsonapi_version_explanation:
+
+discussion_url: https://www.drupal.org/project/issues/jsonapi
+
+editor_name: |
```
ethanresnick: Pls use the format from #1349

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,176 @@
...
+?resource_version=rel:latest-version
+                  \_/ \____________/
+                   |        |
+         version-negotiator |
+                     version-argument
+```
+
+## Server Responsibilities
+
+A server **MUST** respond with `400 Bad Request` if a version negotiator is not
```
ethanresnick: It seems maybe weird to use `400` here and `501` below?

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,176 @@
...
+supported.
+   
+If a server cannot process the given version argument for the given negotiation
```
ethanresnick: It might be worth using error `type` URIs to enable clients to distinguish between these two 400 cases.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,176 @@
...
+mechanism, it **MUST** respond with a `400 Bad Request`.
+
+If a server is able to process the version argument but an appropriate version
+cannot be located, the server **MUST** respond with a `404 Not Found`.
+
+If a server knows a version argument to be invalid for the requested version
```
ethanresnick: How does this case differ from the `400` case two paragraphs up?

_profiles/drupal/resource-versioning/index.md
gabesullice: ea9c13c7d86468f00ac1560100378ca19f048555

_profiles/drupal/resource-versioning/index.md
gabesullice: I agree. I was trying to preserve the ability to add new relative version arguments later because I haven't fully worked them out, but also make this a general rule rather than a rule specific to the relative version negotiator.

I think that was probably mistake. I'm happy with this: 9d9e2634a0c9fdf89361329d684ac5f0d5569a05

_profiles/drupal/resource-versioning/index.md
gabesullice: Ah, great idea! I have yet to completely familiarize myself with all the goodies in 1.1 :)

9368cbc984f5f78ab2c611da8f73c77d56e2ef4d

_profiles/drupal/resource-versioning/index.md
gabesullice: This was related to the `501` thing which has been removed. This isn't relevant any longer.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,179 @@
...
+This profile creates a standard for JSON:API to support these diverse versioning
+schemes.
+# Query Parameter
+
+## Usage
+
+An endpoint **MAY** support a `resource_version` query parameter to allow a
+client to indicate which version(s) of a resource should be returned.
+
+If an endpoint does not support the `resource_version` parameter, it **MUST**
```
ethanresnick: Now that JSON:API is recommending camel casing, it might be better to call this `resourceVersion`, to match the format most likely to be used by other profiles.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,179 @@
...
+  - The server **MUST** respond with `404 Not Found` if an appropriate version of
+    the resource requested cannot be located. 
+    
+> Note: This means that a server should not provide "fallbacks" unless the
+> behavior is well defined by the version negotiation mechanism (see below).
+> These rules apply to individual and collection endpoints alike.
+
+## Format
+  
+The value of the `resource_version` parameter **MUST** be a colon-separated
```
ethanresnick: What happens if the string has only one segment (i.e., no colon in it)? Should that be a distinct error case? Or legal but treated as an empty arguments list?

_profiles/drupal/resource-versioning/index.md
gabesullice: Right... hm. Our implementation will need to change, but that's okay, it hasn't landed yet.

_profiles/drupal/resource-versioning/index.md
gabesullice: I think it should be legal to support a case like `?resource_version=recommended` where `recommended` is a negotiator that uses user preferences to find a recommended product variation (like t-shirt size). What do you think?

gabesullice:
I've just added sections for links and for a version history endpoint. IMO, this is ready for review.

_profiles/drupal/resource-versioning/index.md
gabesullice: c351d9ceee9327bf10e6d48417b90a1792bab44d

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+  - `predecessor-version`: links to the version which immediately preceded the
+    context resource object.
+  - `successor-version`: links to the version which immediately succeeded the
+    context resource object.
+  - `prior-working-copy`: links to the working copy which immediately preceded
+    the context resource object.
+  - `subsequent-working-copy`: links to the working copy which immediately
+    preceded the context resource object.
+
+A server **MAY** provide an array of any of these links to support branching.
```
gabesullice, line 198: @ethanresnick, this is in conflict with the base spec. Not sure what to do or how to accommodate this in any other way. Perhaps you have some ideas?

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+change for editorial or archival purposes. This profile establishes a protocol
+for accessing resources in those various states.
+
+For the purpose of understanding this profile, it is helpful to establish a
+vocabulary for describing various possible states of a resource.
+
+A _revision_ is to be understood as an identifiable state of a resource after
+its creation or after some number of changes.
+
+A _version_ is to be understood as a revision of a resource that is available,
```
e0ipso: What if we switch the order to:

A _version_ is to be understood as a revision that is or was the default revision of a resource. In other words, as a revision of a resource that is available, or was previously available, without any version negotiation.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+or was previously available, without any version negotiation. In other words, as
+a revision that is or was the default revision of a resource.
+
+A _working copy_ is the revision to which new changes can be made or to which
+they will be applied. Colloquially, a working copy is often thought of as the
+"tip" of a version history.
+
+For example, a resource which existed as a draft and underwent multiple changes
+before the resource was published may have many revisions. Only the published
```
e0ipso, line 59: You could create working copies by editing a published resource (v1) before publishing v2. The current language hints that working copies can only exist between v0 and v1.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+multiple working copies of a resource (e.g., to support multiple
+languages or within a version control system which supports multiple branches).
+
+This profile creates a standard for JSON:API to support these diverse versioning
+schemes.
+# Query Parameter
+
+## Usage
+
+An endpoint **MAY** support a `resourceVersion` query parameter to allow a
```
e0ipso, line 79: Is this `resourceVersion` against the naming convention? Maybe it's intentional now that the parameter is part of a profile.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+it:
+
+  - The server’s response **MUST** contain the most appropriate version of the
+    resource requested.
+  - The server **MUST NOT** include a version of a resource inappropriate for the
+    requested version.
+  - The server **MUST** respond with `404 Not Found` if an appropriate version of
+    the resource requested cannot be located. 
+    
+> Note: This means that a server should not provide "fallbacks" unless the
```
e0ipso, line 95: Nice clarification 👍

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+         version-negotiator |
+                     version-argument
+```
+
+## Server Responsibilities
+
+<a id="bad-version-negotiator"></a>A server **MUST** respond with `400 Bad Request` if a version negotiator is not
+supported. In this case, an error object that includes a `type` link to
+`https://jsonapi.org/profiles/drupal/resource-versioning/#bad-version-negotiator`
+**MUST** be included in the response document.
```
e0ipso: Perhaps we could relax this link requirement to `**MAY**`? I see your point, though.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+   
+<a id="bad-version-argument"></a>If a server cannot process the given version argument for the given negotiation
+mechanism, it **MUST** respond with a `400 Bad Request`. In this case, an error
+object that includes a `type` link to
+`https://jsonapi.org/profiles/drupal/resource-versioning/#bad-version-argument`
+**MUST** be included in the response document.
```
e0ipso: Same as above.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+
+When a server processes a request with a `resourceVersion` query parameter and
+a `self` link is provided for a top-level links object, the link's `href`
+**MUST** include the `resourceVersion` query parameter with the same version
+identifier that was requested.
+
+When a server processes a request with a `resourceVersion` query parameter
+all resource object `self` links **SHOULD** contain a `resourceVersion` query
+parameter which identifies the specific revision represented by that resource
+object.
```
e0ipso, line 156: Should we be explicit that the version negotiator in the `self` link should be the same as in the processed request?

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+      "self": "/article/1?resourceVersion=id:42"
+    }
+  },
+  "links": {
+    "self": "/article/1?resourceVersion=rel:latest-version"
+  }
+}
+```
+
+A server **MAY** provide a `version-history` link in a resource object's links
```
e0ipso, line 179: Should `version-history` use the same version negotiator as the request? Should it always be `rel`?

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+| `g`      | no link          | `h`            | `e`                   | no link             | `f`                  | `h`                       |
+| `h`      | `g`              | no link        | `g`                   | no link             | `g`                  | no link                   |
+
+> Note: In this example, `f` has both a `predecessor-version` and
+> `prior-working-copy` link to `e` because `e` was a version and `e` was also the
+> revision to which changes could be applied prior to `f`'s creation.
+
+# Version History
+
+A server **MAY** provide a "version history" endpoint. The primary data
```
e0ipso, line 231: Should we provide a URL template for this endpoint?

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+of a response document from a version history endpoint must be a collection of
+resource objects.
+
+Unless an `id` contains version information, the `type` and `id` members of each
+resource object in the collection **MUST** be the same.
+
+If provided, resource objects' `self` links **MUST NOT** be the same.
+
+# Version Negotiators
```
e0ipso, line 240: I ❤️ the idea of having these basic version negotiators in the spec.

In addition to that we should add links to point to these negotiator definitions. That will allow servers that implement custom version negotiators to surface documentation about those. Servers implementing `id` and `rel` can point to this very page, custom ones will have to live elsewhere.

Maybe we should have links at the top level document like `"version-negotiator:rel": "http:/…/#relative-version-negotiator"` if the `resourceVersion` query string parameter is present.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+
+## ID-Based Version Negotiator
```
e0ipso, line 255: MUST this be implemented in every server supporting versioning?

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+
+The resource version returned for any given version argument in an `id`-based
+version identifier **MUST NOT** change over time.
+
+> Note: This profile is agnostic about the format of the version argument in
+> `id`-based version identifiers. For example, one server may use integers as
+> revision IDs, another may use UUIDs and yet another may use content-based
+> hashes.
+
+## Relative Version Negotiator
```
e0ipso, line 271: MUST this be implemented in every server supporting versioning?

_profiles/drupal/resource-versioning/index.md
gabesullice: @ethanresnick actually suggested that I rename it to this because it's an official recommendation to use camelCasing in profiles.

_profiles/drupal/resource-versioning/index.md
gabesullice: Meet in the middle with **SHOULD** ?

gabesullice: What's your reasoning for relaxing this?

e0ipso: **SHOULD** sounds good.

> What's your reasoning for relaxing this?

It feels that status codes are much more universal & standarized than hypermedia. It feels that we are forcing our passion about hypermedia on server implementations.

_profiles/drupal/resource-versioning/index.md
gabesullice: No, I think this would actually defeat the concept I'm trying to promote by saying "the specific revision" and with the example below.

By putting`?resourceVersion=rel:latest-version` in the top-level links object but `?resourceVersion=id:42` on the resource object (different negotiators), you have enough information to deduce that the latest version is now `42`.

I'm even considering that the `rel`-based version negotiation should explicitly be disallowed from being in a resource object's links. All the other "relative" information can be deduced from other links on the resource object (ie. if it doesn't have a `latest-version` link, it _is_ the `latest-version`).

_profiles/drupal/resource-versioning/index.md
gabesullice: No. It should always be the most specific negotiator provided by the server.

_profiles/drupal/resource-versioning/index.md
gabesullice: No, the spec avoids specifying any URLs and 1.1 even forbids profiles from defining endpoint locations. I don't think it's necessary.

_profiles/drupal/resource-versioning/index.md
gabesullice: I _really_ like the ID of making negotiator IDs into URIs ❤️ good idea!

_profiles/drupal/resource-versioning/index.md
gabesullice: No, a server that versions by updated timestamp might use a `timestamp` negotiator as it's "most specific" negotiator. It might not have revision IDs.

_profiles/drupal/resource-versioning/index.md
gabesullice: No. I see `rel` based version negotiation as not much more than a convenience. You can achieve anything you need to without it if you're willing to make 2 requests.

_profiles/drupal/resource-versioning/index.md
gabesullice: I should add that language instead of assuming it's implied.

_profiles/drupal/resource-versioning/index.md
gabesullice: For example, a server could have a negotiator named `at` as long as it doesn't support branched versioning or forward revisions. That would allow a client to request any timestamp and get the appropriate revision for that time period. The version history would just give `self` links equal to `at:{updated timestamp}`.

_profiles/drupal/resource-versioning/index.md
gabesullice: See: https://github.com/json-api/json-api/pull/1333#discussion_r243379557

ethanresnick:
@gabesullice Thanks for this! 

1. This is an awesome profile for a much-needed feature. 

2. It's a great test of the profile review criteria and timeline. (It already reminded me about [another criterion](https://github.com/json-api/json-api/pull/1351) we needed.)

My inline comments thus far have been a bit scattered/design focused, but I'll go through this soon with an attempt to apply the formal review criteria.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+This profile creates a standard for JSON:API to support these diverse versioning
+schemes.
+# Query Parameter
+
+## Usage
+
+An endpoint **MAY** support a `resourceVersion` query parameter to allow a
+client to indicate which version(s) of a resource should be returned.
+
+If an endpoint does not support the `resourceVersion` parameter, it **MUST**
```
ethanresnick, line 82: [Design feedback] Requiring use of the error object `source` here might be useful for clients. Ditto for the other error cases.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+respond with `400 Bad Request` to any requests that include it.
+
+If an endpoint supports the `resourceVersion` parameter and a client supplies
+it:
+
+  - The server’s response **MUST** contain the most appropriate version of the
+    resource requested.
+  - The server **MUST NOT** include a version of a resource inappropriate for the
+    requested version.
+  - The server **MUST** respond with `404 Not Found` if an appropriate version of
```
ethanresnick, line 92: [Editorial] Bullet 3 seems to imply bullet 2, so bullet two might be redundant (esp with the note).

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+cannot be located, the server **MUST** respond with a `404 Not Found`.
+
+# Links
+
+When a server processes a request with a `resourceVersion` query parameter and
+a `self` link is provided for a top-level links object, the link's `href`
+**MUST** include the `resourceVersion` query parameter with the same version
+identifier that was requested.
+
+When a server processes a request with a `resourceVersion` query parameter
```
ethanresnick, line 153: [Design] My biggest piece of feedback about this profile is that it should probably also consider how to update resources (e.g., how to create a new working copy based of an existing one, or change the default version, or reject a PATCH if the resource's working copy has changed from the working copy that the client is trying to apply its PATCH to, etc). When resource versions have [come up in the past](https://github.com/json-api/json-api/pull/824), these updating use cases were actually the primary motivation, so not addressing them here would seem like a major missed opportunity that might be hard to remedy once this is published.

I've put this comment on this paragraph, though, because I think the considerations about updating bear directly on what the `self` link should be. If the `self` link is URL of a specific revision (as it seems to be here), then a `PATCH` to that link would be interpreted as updating the revision -- which should probably usually be immutable -- and creating a new revision would become a POST (or a patch to the default version's URL). So that's something to think about. 

Also, if each revision is it's own HTTP resource (i.e., has its own URL), then the JSON:API `id` would usually be different for each revision too (e.g., so a PATCH actually _could_ target the revision), though I suppose it wouldn't have to be. Distinct `id`s could cause lots of problems, though, because presumably only the default version's `type`/`id` would want to be used in relationship objects. 

So this is all stuff to think about... I don't have any conclusions at the moment, but I'll mull it over, and hopefully you guys can too. I think, as usual, it goes back to the weird relationship that JSON:API resource objects have with HTTP's concepts of resources + entities.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+> `prior-working-copy` link to `e` because `e` was a version and `e` was also the
+> revision to which changes could be applied prior to `f`'s creation.
+
+# Version History
+
+A server **MAY** provide a "version history" endpoint. The primary data
+of a response document from a version history endpoint must be a collection of
+resource objects.
+
+Unless an `id` contains version information, the `type` and `id` members of each
```
ethanresnick, line 235: [Spec compliance] Having multiple resource objects in `data` with the same `type`/`id` is sort of a spec violation. I say "sort of" because the spec technically says only that:

> A compound document MUST NOT include more than one resource object for each `type` and `id` pair.

So, if this version history response isn't a compound document, you're technically in the clear. But the intent has always been that this restriction should apply to all responses. (I can't find the link for that now, but it's come up in conversations with @dgeb over the years.) 

Note: I've also [proposed](https://github.com/json-api/json-api/issues/1332) removing the restriction altogether.

_profiles/drupal/resource-versioning/index.md
ethanresnick: @gabesullice I'd be happy to just define these links in the base spec as a temporary workaround. Note also that the base spec's current approach is that, if an array of links is allowed, an array should be required (like we did for `profile`) to not further proliferate the number of links cases that clients have to handle. 

So, the base spec language would go under the [resource links](https://jsonapi.org/format/#document-resource-object-links) section, and could say something like:

> If present, this object **MAY** also contain an array of links as the value for any of the following keys: `latest-version`, `working-copy`, `predecessor-version`, `successor-version`, [etc]. Links at any of these keys point to resources that are related to the resource object according to the IANA link relation corresponding to their key name.

_profiles/drupal/resource-versioning/index.md
ethanresnick, line 240: [Design feedback/interoperable implementations review requirement] If I'm understanding correctly, each server implementation can define its own version negotiators, with multiple servers using the same negotiator name in different ways (and no central registry). That seems not great for interoperability. It would also seem to mean that no other version negotiators can ever be standardized as part of this profile, because they could conflict with existing implementations. Is that really what you want?

My gut instinct would be to make this profile's spec the canonical list of all legal version negotiators. Then, you can add more over time to this spec as they're requested. To support people who really need to use a version negotiator that isn't part of your profile, you could have some set of negotiator names/schemes/namespaces that are allowed to have implementation-specific meanings.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,273 @@
...
+The resource version returned for any given version argument in a `rel`-based
+version identifier **MAY** change over time.
+
+The `rel` version negotiator has the following valid version argument strings:
+
+  - `latest-version`: requests the latest default revision of a resource.
+  - `working-copy`: requests revision of a resource to which changes can be
+    made.
+
+If any of the following version arguments is received in a `rel`-based version
```
ethanresnick, line 289: [Interoperable implementations requirement] It's not clear to me if servers are allowed to define their own arguments to version negotiators defined in this profile. E.g., could a server support `rel:xxx`? (Note: if a server can, that also limits your ability to extend these argument lists, except to define the four "reserved" first argument names that you give below.)

ethanresnick (changes requested):
Current review, by the formal criteria:

- [x] follows the profile template;
- [ ] is specified precisely enough to enable interoperable implementations [not quite, see comments];
- [ ] complies with the JSON:API spec and its requirements for profiles [not quite, see comments]; 
- [x] wouldn’t cause any problems were it to become widely adopted.

_profiles/drupal/resource-versioning/index.md
ethanresnick: > I think it should be legal to support a case like `?resource_version=recommended`

I agree. You might want to make that more explicit though, since I don't think it's necessarily clear at the moment.

_profiles/drupal/resource-versioning/index.md
e0ipso: Then we probably want language to clarify this section and say that these are optional version negotiators. Also worth noting that an implementor should have at least one version negotiator defined.

_profiles/drupal/resource-versioning/index.md
e0ipso: > which should probably usually be immutable

I think it makes sense to treat revisions as immutable.

----

I agree with all that was said in the comment above, but I want to highlight another implicit way of creating revisions that we may need to be explicit about.

Many frameworks handle revision creation as part of the store update. In Drupal when you save an entity that supports revisioning by `PATCH`ing it (no `resourceVersion` used here), then a new revision will be created for you. You cannot do much about it in most of the cases. This new revision will be a _working-copy_ or a _version_ depending on some custom business logic (usually a _published: true / false_ flag).

I guess that what I'm saying is that strict regulation of how revisions are created may become hard. However, we can put language on how to create revisions **through JSON:API**.

Where I'm going is, do we want this profile to include revision creation or it can be a separate profile?

_profiles/drupal/resource-versioning/index.md
gabesullice: 3e4b67eac6d9726c43948a6fbf50867cb5bf8f19

_profiles/drupal/resource-versioning/index.md
gabesullice: That'd be great :)

_profiles/drupal/resource-versioning/index.md
gabesullice: 515aba916dd058774b553bc901e90a4f00ba54d6

_profiles/drupal/resource-versioning/index.md
gabesullice: 0feddd6

_profiles/drupal/resource-versioning/index.md
gabesullice: b0a1e3d

_profiles/drupal/resource-versioning/index.md
gabesullice: b0a1e3d

_profiles/drupal/resource-versioning/index.md
gabesullice: f0d73db

_profiles/drupal/resource-versioning/index.md
gabesullice: f0d73db82a26d8ab999b38a582969e112b864351

gabesullice: I did not mention that they should have one defined. That seems unnecessary as the specification can't work without it. No?

gabesullice: See: a4329d5

_profiles/drupal/resource-versioning/index.md
gabesullice: See: f0d73db

_profiles/drupal/resource-versioning/index.md
gabesullice: 7a352dc

_profiles/drupal/resource-versioning/index.md
gabesullice: 7a352dc

_profiles/drupal/resource-versioning/index.md
gabesullice: Let's continue this conversation in this comment https://github.com/json-api/json-api/pull/1333#discussion_r243464998

gabesullice: What I've done is applied the implementation-specific query parameter name constraints as a version negotiator constraint and added a mechanism for adding new negotiators to the profile.

Changes here: f0d73db

Elsewhere, @e0ipso suggest that we use URIs but I think this is sufficient. @e0ipso, WDYT?

gabesullice: cc: @wimleers cause it's his inbox that I suggested too 😛

_profiles/drupal/resource-versioning/index.md
gabesullice: 9129eb2

_profiles/drupal/resource-versioning/index.md
gabesullice: ba9bf0b

_profiles/drupal/resource-versioning/index.md
gabesullice: It is not intended to be a compound document.

I don't think this profile can impose a requirement that `id`s contain version information. I think most systems store revision IDs separately from the entity ID. In our case, we use UUIDs for `id` and revision IDs are auto-incrementing integers. We'd have to make some wonky concatenation of these to put revision IDs into our resource object IDs and I don't think this profile should define how to do that.

I've been considering adding a recommendation that implementations add information to resource object's `meta` object that matches their most-specific version negotiator. For example, in our case (since we implement the `id` negotiator) we'd have:

```
{
  "type": "article",
  "id": "some-long-uuid",
  "meta": {
    "resourceVersionId": 42
  },
  "links": {
    "self": "/articles/some-long-uuid?resourceVersion=id:42"
  }
}
```

The rule would be something like:

>It is **RECOMMENDED** that resource object's `meta` object contains a member whose name is `resourceVersion` appended by the upper-camel-cased name of the version negotiator used in the resource object's `self` link. The value of this member should be the the version identifier used in the resource object's `self` link, excluding its first segment, and it **MUST NOT** begin with a colon.

_profiles/drupal/resource-versioning/index.md
gabesullice: No, a server shouldn't support any custom rel-based negotiators. I've clarified that in: 6bc95be

Good catch!

_profiles/drupal/resource-versioning/index.md
gabesullice: > [Not] addressing [PATCH/DELETE] here would seem like a major missed opportunity that might be hard to remedy once this is published.

I think I agree.

> Where I'm going is, do we want this profile to include revision creation or it can be a separate profile?

As above, I think I _would_ like this profile to define how update/delete should be handled.

<hr/>

> I think the considerations about updating bear directly on what the `self` link should be. If the self link is URL of a specific revision (as it seems to be here), then a PATCH to that link would be interpreted as updating the revision -- which should probably usually be immutable -- and creating a new revision would become a POST (or a patch to the default version's URL).

I think I may be in a minority about this, but I try to deeply appreciate the difference between a top-level `self` link and a resource object `self` link. I feel that once one groks what it means when they're different, the way JSON:API reconciles resource objects with HTTP resources + entities becomes much easier to think about.

Where I'm going with that is pretty simple: the top-level `self` link should be the URL used to mutate a resource object, not the resource object self link. Once you establish that, things become pretty simple.

```
GET /article/1?resourceVersion=rel:working-copy

{
  "data": {
    "type": "article",
    "id": 1,
    "links": {
      "self": "/article/1?resourceVersion=id:42"
    }
  },
  "links": {
    "self": "/article/1?resourceVersion=rel:working-copy"
  }
}
```

That means that a `PATCH` would update the working copy via the top-level `self` link, whatever it may be, like so:

```
PATCH /articles/1?resourceVersion=rel:working-copy

{
  "type": "article",
  "id": 1,
  "attributes": { // some changes }
}
```

The server response would then be:

```
200 OK

{
  "data": {
    "type": "article",
    "id": 1,
    "links": {
      "self": "/article/1?resourceVersion=id:43" // <- incremented
    }
  },
  "links": {
    "self": "/article/1?resourceVersion=rel:working-copy" // <- unchanged
  }
}
```

The `working-copy` link could be used instead of the top-level `self` link if the `GET` request was made directly to a resource object via an over-specific version negotiator (this would be the case if one visited `/articles/1?resourceVersion=id:42` directly for some reason).

>> which should probably usually be immutable
>
> I think it makes sense to treat revisions as immutable.

I think it makes sense too, but I don't think this profile needs to define that. I think it can work either way.

Drupal actually _does_ permit updating a revision in-place (whether that's a good idea or not is beside the point 😛 ). With the scheme above, it would be possible for a server to support both in-place editing of a revision or new revision creation via `PATCH` depending on the `self` link used (top-level vs. resource object).

If a server cannot automatically create new revisions via `PATCH` for some reason (maybe it can't/hasn't implemented the `rel` negotiator), then I think the appropriate URL for creating new revisions would be at the `version-history` URL via `POST`. (this is why it makes sense for it to be a collection resource).

gabesullice: The language about updating resources in the spec seems to jive with what I suggested:

>The URL for a resource can be obtained in the `self` link of the resource object. **Alternatively, when a `GET` request returns a single resource object as primary data, the same request URL can be used for updates**.

We could go a step further in that language by adding:

> the same request URL **or the top-level `self` link** can be used for updates.

_profiles/drupal/resource-versioning/index.md
e0ipso: I'm OK with that. However I still think that doc curies offer a lower barrier of entry. One could document the new crazy negotiator in their blog and have the server implementation point to that, instead of adding it here.

Many will just not go through the trouble to send their idea to some email addresses and be potentially blocked by them.

gabesullice: I guess I wanted to discourage non-standard negotiators being used as a standard and keep them isolated. My feeling was that sending an email saying "hey, I'd like to add a negotiator like this..." doesn't seem like a super high barrier. In some ways, it's even less difficult than opening a PR.

What if we add this:

> Note: Don't be shy! New version negotiators are more than welcome, the editors want to see this profile proliferate and that means accepting versioning strategies like yours!

_profiles/drupal/resource-versioning/index.md
gabesullice: I just read the #824 PR. I like the idea of a `revision` key. This profile could adopt `revision` as a `meta` object member and require that it must be equal to the complete version identifier in a resource object's `self` link (or a valid one if a `self` link is not provided).

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,301 @@
...
+minimum_jsonapi_version_explanation:
+
+discussion_url: https://www.drupal.org/project/issues/jsonapi
+
+editors:
+  - name: Gabriel Sullice
+    email: gabriel@sullice.com
+  - name: Mateu Aguiló Bosch
+    email: mateu.aguilo.bosch@gmail.com
+    website: https://humanbits.es
```
e0ipso: Can we change this to https://mateuaguilo.com?

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,301 @@
...
+minimum_jsonapi_version: 1.0
+minimum_jsonapi_version_explanation:
+
+discussion_url: https://www.drupal.org/project/issues/jsonapi
+
+editors:
+  - name: Gabriel Sullice
+    email: gabriel@sullice.com
+  - name: Mateu Aguiló Bosch
+    email: mateu.aguilo.bosch@gmail.com
```
e0ipso: Can we change this to mateu@mateuaguilo.com?

_profiles/drupal/resource-versioning/index.md
gabesullice: Yes... your SSL cert is expired BTW 😛 (or just invalid, I can't tell)

gabesullice: c82ecbe

_profiles/drupal/resource-versioning/index.md
gabesullice: c82ecbe

_profiles/drupal/resource-versioning/index.md
gabesullice: bump.

_profiles/drupal/resource-versioning/index.md
gabesullice: @e0ipso @wimleers thoughts? ^

_profiles/drupal/resource-versioning/index.md
gabesullice: @e0ipso, would that work?

_profiles/drupal/resource-versioning/index.md
wimleers: I also like a `revision` key (or `version`). This would also address @ethanresnick's "sort of" spec violation concern: *if* present, then it would be a _higher specificity resource identifier_ (or perhaps even an _intra resource identifier_?): just like `[type, id]` uniquely identify a resource, `[type, id, version]` uniquely identifies a _version_ of that resource.

Taking this further, there are multiple potential axes along which a _variant_ of a resource can exist: `version` (or `revision`) is one, `lang` (or `translation`) is another. I think those are the two clearest, strongest use cases. It's possible to think of others though, such as `localization`.

I'd say versioning and translations are the two most common needs. I think we should keep both in mind if we're going to go this direction.

But for the sake for consistency and optionality (to ensure backwards compatibility & evolvability), I think we should consider **not** adding a `version` key, but a `variant` key. Under `variant`, this profile could then reserve the `version` key. A translation profile could reserve the `translation` key. This also opens the door for multiple versioning profiles if there is the need. 

After having written this, I continued reading #824, to make sure I didn't miss anything. I think my `variant` proposal would actually address the "identity" concerns that @ethanresnick raised in 
https://github.com/json-api/json-api/pull/824#issuecomment-126200447? :)

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,301 @@
...
+
+extended_description: |
+  # Overview
+  JSON:API servers that are capable of tracking a record of changes to a
+  resource object may afford the capability of requesting a resource object
+  as it existed in a prior or successive state. This capability is herein
+  defined as _resource versioning_.
+  
+  This profile establishes a protocol for resource versioning by defining a
+  query parameter and its semantics in order to identify arbitrary revisions
```
wimleers, line 15: revisions → versions :)

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,301 @@
...
+  - name: Wim Leers
+    email: work@wimleers.com
+    website: https://wimleers.com
+
+categories:
+  - Resource Versioning
+---
+  
+# Concepts
+Resources on a server may undergo changes and the state of a resource with an
```
wimleers, line 38: I think that by "the state […] may be accessible", you mean that _each state_ may be individually accessible?

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,301 @@
...
+arbitrary number of changes may be accessible. It is often useful to retrieve
+resources as they existed at the time of their creation or in various states of
+change for editorial or archival purposes. This profile establishes a protocol
+for accessing resources in those various states.
```
wimleers, line 42: This is talking about _previous_ states, for archival/history purposes. I think this should also explicitly state _next_ states, i.e. drafts that are actively being worked on, or perhaps need to go through legal review, or are ready but simply aren't yet live because they're _scheduled_ to be published in the future.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,301 @@
...
+
+For the purpose of understanding this profile, it is helpful to establish a
+vocabulary for describing various possible states of a resource.
+
+A _revision_ is to be understood as an identifiable state of a resource after
+its creation or after some number of changes.
+
+A _version_ is to be understood as a revision that is or was the default
+revision of a resource. In other words, as a revision of a resource that is
+available, or was previously available, without any version negotiation.
```
wimleers, line 52: I think something like "that used to be the canonical representation, i.e. available without any version negotiation" would make this even more clear?

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,301 @@
...
+
+A _working copy_ is the revision to which new changes can be made or to which
```
wimleers, line 54: This term is specifically borrowed from 
https://tools.ietf.org/html/rfc5829, because this entire profile was modeled after RFC5829. Linking to that at the top of the "Concepts" section would be helpful.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,301 @@
...
+the working copy. When the working copy is checked in and made the default
+revision, this revision becomes the latest version and the version before it is
+its _predecessor version_. If no other revisions come after the latest version,
+it may also be considered the working copy.
+
+Sophisticated servers may support multiple versions of a resource as well as
+multiple working copies of a resource (e.g., to support multiple
+languages or within a version control system which supports multiple branches).
+
+This profile creates a standard for JSON:API to support these diverse versioning
```
wimleers, line 73: "these diverse" refers to both the "multiple versions" and "multiple working copies" in the preceding paragraph. Therefore I think it should be "diverse revisioning schemes", not "diverse versioning schemes"? (Because "versioning" refers only to _versions_, not to _working copies_.)

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,301 @@
...
+
+If an endpoint does not support the `resourceVersion` parameter, it **MUST**
+respond with `400 Bad Request` to any requests that include it. An error object
+detailing the source of the error **SHOULD** include a `source` member referencing
+the `resourceVersion` query parameter.
+
+If an endpoint supports the `resourceVersion` parameter and a client supplies
+it:
+
+  - The server’s response **MUST** contain the most appropriate version of the
```
wimleers, line 90: I think "the most appropriate" is unnecessarily vague. Can't we change `the most appropiate version` to `the requested version`?

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,301 @@
...
+                    version-argument
+```
+
+## Server Responsibilities
+
+<a id="bad-version-negotiator"></a>A server **MUST** respond with `400 Bad Request` if a version negotiator is not
+supported. An error object detailing the source of the error **SHOULD** include a
+`type` link to
+`https://jsonapi.org/profiles/drupal/resource-versioning/#bad-version-negotiator`
+and **SHOULD** include a `source` member referencing the `resourceVersion` query
```
wimleers, line 133: I think at least one of these two `**SHOULD**`s should be come a `**MUST**`. Otherwise a `400 Bad Request` response without any useful information is allowed, which is bad DX.

_profiles/drupal/resource-versioning/index.md
wimleers: I agree with @ethanresnick that the fact that this profile's narrowness (only defining "read" behavior) is its biggest weakness.

(Similarly, I think #824's biggest weakness _also_ was its narrowness, it just was different: it focused on *optimistic concurrency control*.)

On the other hand, I think solid read-only support and leaving modifications to a separate profile or a future iteration is preferable over having a spec that supports fewer use cases (which I think was true of #824 — _it did not allow a particular revision to be retrieved_). Especially if it's based on another established standard (RFC5829) and therefore likely to be more implementation-agnostic.

wimleers: I'm fascinated by @gabesullice's proposal (and the sample response bodies really help, thanks! 🙏). I'm very curious to find out what @ethanresnick thinks about that :)

_profiles/drupal/resource-versioning/index.md
wimleers: @ethanresnick Rather than calling out those 4 specific link relations, perhaps it'd be better to generalize that to something like:
> Links with a key that matches a [registered link relation type](
https://www.iana.org/assignments/link-relations/link-relations.xhtml) which allows multiple values are allowed to have a value that is not a link, but an array of links.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,301 @@
...
+profile. New version negotiators may be registered by sending one, joint email
+to the profile editors with the subject line: "New JSON:API Resource Version
+Negotiator: {negotiator name}". This will begin a process of refinement and/or
+result in a determination of fitness for addition to this profile.
+
+## ID-Based Version Negotiator
+
+This profile establishes the `id` version negotiator. An `id`-based version
+identifier is composed of two segments—the `id` version negotiator and a single
+version argument. Any colons (U+003A COLON, “:”) present in the version
```
wimleers, line 259: This should say that the version argument is an opaque string.

_profiles/drupal/resource-versioning/index.md
```diff
@@ -0,0 +1,301 @@
...
+identifier after the first occurrence **MUST** be interpreted as part of the
+single version argument and **MUST NOT** be interpreted as a segment delimiter.
+
+The resource version returned for any given version argument in an `id`-based
+version identifier **MUST NOT** change over time.
+
+> Note: This profile is agnostic about the format of the version argument in
+> `id`-based version identifiers. For example, one server may use integers as
+> revision IDs, another may use UUIDs and yet another may use content-based
```
wimleers, line 268: This note could go away by defining it as an opaque string, like I suggested above.

wimleers (changes requested):
I did an initial complete pass. I think it's fascinating how #824 was too narrow in one way (focusing on _optimistic concurrency control_ and hence not even providing the ability to retrieve a particular revision/version) and this is too narrow in another way (not talking about `POST`/`PATCH`/`DELETE` _at all_).

I do see that @ethanresnick's insightful (and constructively critical! 👏 ) comments are resulting in solid improvements. I like where this is going :)

_profiles/drupal/resource-versioning/index.md
gabesullice: While I agree that calling out specific link relations "smells" a bit, I would not like to see the concept of links object keys as link relations furthered in the spec. Links can have more than one link relation and this pattern makes that more challenging to fix in later iterations of the spec.

_profiles/drupal/resource-versioning/index.md
gabesullice: > or reject a PATCH if the resource's working copy has changed from the working copy that the client is trying to apply its PATCH to

I neglected to give a solid answer for this. As I said above, I think the `PATCH` should be sent to the top-level `self` link: `/article/1?resourceVersion=rel:working-copy`.

However, this spec could add something like an `appliesToRevisionId: id:42` under the `meta` key of the request document.

gabesullice:
General comment: I think this profile need to be reviewed to ensure that the `revision`/`version` distinction is well maintained in naming. I'm thinking of "version identifier", which I think should become "revision identifier".

_profiles/drupal/resource-versioning/index.md
gabesullice: Crazy thought... maybe we could we extract a variant profile out of this profile? Re-minting `resource_version` as `resource_variant`. The "negotiation mechanism" would become the variant negotiation mechanism. Then you could have something like `?resource_variant=lang:en-US` or `resource_variant=revision:id:42` or `?resource_variant=revision:rel:working-copy`.

In fact, if this variation scheme were part of the base spec, then revisions, translations, etc could just enhance that mechanism for its needs.

_profiles/drupal/resource-versioning/index.md
gabesullice: Not in this instance, remember: "not all revisions are versions, but all versions are revisions"

_profiles/drupal/resource-versioning/index.md
gabesullice: > This is talking about previous states, for archival/history purposes.

I agree that this could be clarified, but I did not mean previous as you are interpreting it.

Even future revisions were created _previously to now_. They may not be the default version _yet_, but that does not mean that the revision itself was created in the future.

_profiles/drupal/resource-versioning/index.md
gabesullice: I very intentionally avoided the word "canonical" here because the `canonical` link relation is not intended to disambiguate revisions of a resource. It disambiguates _duplicates_ of a resource when they have different URIs.

_profiles/drupal/resource-versioning/index.md
gabesullice: Wow, I can't believe that I didn't already do that.

wimleers:
I agree that that aspect is most confusing.

_profiles/drupal/resource-versioning/index.md
antonkomarev: @wimleers noted very important thing. It's not enough to add version of the resource because each resource could have many translations and each translation could have it's own versions.

antonkomarev: Word `locale` could be considered for translation purposes instead of `lang` abbreviation.

_profiles/drupal/resource-versioning/index.md
jelhan: > reject a PATCH if the resource's working copy has changed from the working copy that the client is trying to apply its PATCH to

Support a conflict detection mechanism as described in initial review post wasn't discussed enough in my opinion. Something similar to AWS Amplify should be easy to achieve if already having well defined revisions. It was already discussed as part of #824 some time ago.

Maybe something like this would be everything needed:

> # Conflict detection when updating resources
>
> A server **SHOULD** include the most specific version negotiator it supports in any resource object's meta information under `revision` key.
> A client **MAY** include the revision an update is based on as under `revision` key of `meta` object on a `PATCH` request.
> A server **MUST** reject `409 Conflict` when processing a `PATCH` request to update a resource if a `revision` is provided as meta information, which is not the latest one.
> A server **MAY** reject a `PATCH` request to update a resource if server has included a revision on retrieving the resource but the `PATCH` request does not include a revision.

Issue #1346, opened 2018-12-13
Add an editor(s) section to the profile template


gabesullice:
I just submitted a profile for registration and this came up:

I was able to choose `drupal` as the namespace of the profile in order to not portray myself as the sole creator of the profile, but the `author_name` and `author_email` items defeat that distinction. I did technically _author_ it, but the ideas all came from a collaborative effort.

I think a good solution would be to substitute an "Editor(s)" section for the author section.

This would also mean that more than one person could be responsible for making changes, which would be a nice relief for authors and could increase the velocity of corrections/additions.

ethanresnick:
This sounds good to me. Can you make a PR?

ethanresnick:
Addressed by #1349

PR #1347, opened 2018-12-13
Add an editor(s) section to the profile template

gabesullice:
See #1346

### DIFF
--- a/profile_template
+++ b/profile_template
@@ -22,11 +22,19 @@ minimum_jsonapi_version_explanation:
 # ask questions/start discussions about your extension.
 discussion_url: http://tets.com
 
-author_name: Ethan Resnick
-author_email: ethan.resnick@gmail.com
+# For multiple editors, use a pipe (|) and place each editor on a new line. Use
+# "n/a" to preserve empty rows. For example:
+#   editor_name: |
+#     Jane Doe
+#     John Doe
+#   editor_website: |
+#     n/a
+#     john.doe@example.com
+editor_name: Ethan Resnick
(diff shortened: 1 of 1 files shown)


profile_template
```diff
@@ -22,11 +22,19 @@ minimum_jsonapi_version_explanation:
 # ask questions/start discussions about your extension.
 discussion_url: http://tets.com
 
-author_name: Ethan Resnick
-author_email: ethan.resnick@gmail.com
+# For multiple editors, use a pipe (|) and place each editor on a new line. Use
```
ethanresnick, line 25: If possible, I think it would be nicer here to use a yaml array of objects; i.e., something like:

```yaml
editors:
  - name: Jane Doe
    email: janedoe@gmail.com
  - name: John Doe
    email: johndoe@gmail.com
    website: example.com
```

That way, people don't have to match up the `n/a`s, and a comment can just indicate that name + email are required for each author.

I say "If possible" because I seem to recall that this structure was really hard to work with in liquid for some reason... but I don't see now why it would be.

ethanresnick (commented):
Looking good. See comment. Also, this PR's gonna need to update the html layout for profiles too to actually show the multiple editors.

profile_template
gabesullice: @ethanresnick, if you can, would you mind making this change? I'm not at all familiar with how the project is setup/built etc.

I think you have access make changes to this PR.

ethanresnick:
@gabesullice I didn't have push access to your branch, actually, but I made a new PR based off of your commit. See #1349. Closing this in favor of that.

PR #1348, opened 2018-12-14
Upgrade links to RFC8288-modelled web links

gabesullice:
I know there has been _much_ discussion of links in this project. This PR attempts to faithfully project [RFC8288](https://tools.ietf.org/html/rfc8288) style links onto the JSON:API specification. It is very intentionally concise and _not_ feature rich—meaning that it completely avoids topics like hyper-schema, URI templating, _inter alia_.

If this proposal is not acceptable without those features, I'd ask that this PR be closed. In other words, let's not let those things derail getting a sensible start to better hypermedia support in JSON:API. Nothing here precludes those features. In essence, "a bird in hand is worth two in the bush".

<hr/>

This approach lifts the `link-param` idea from RFC8288 and serializes them in `application/vnd.api+json` format under a new `params` member<sup>[[1]](#fn1)</sup>.

The established link-params `rel`, `anchor` and `rev` have special handling, this is my reasoning:

* `rel`: inherits from the link object key to maximize backwards compatibility. I used the **SHOULD** key word for this inheritance to guard against that case that a key may not be a valid link relation type.
* `anchor`: Because JSON:API has a very well defined structure, I think anchors can be implicitly defined for the `application/vnd.api+json` media type. In other words, that they don't have to explicitly be added as an object member.

  I considered adding language to define a new `anchor` member that could be placed under any `meta` object so that custom anchors could be used, but I left that out. I think that that choice makes this easier to land and to implement for clients and for servers.

  The biggest problem I see here is that it makes it difficult for implementors to comply with [RFC3986's language about consistency of fragments between different representations](https://tools.ietf.org/html/rfc3986#section-3.5). Frankly, I think that's a really pie-in-the-sky requirement to begin with and I wouldn't feel too bad about ignoring it. The `meta.anchor` idea could
  always be added later if it becomes a real problem.

  @ethanresnick [suggested elsewhere](https://github.com/evert/ketting/issues/109) that a fragment should be defined by `type` and `id`. That would make it easier to meet the consistency requirement, but the problem I faced with that approach is that a fragment also should be able to identify a relationships object within a resource object. A fragment syntax like `{type}:{id}(:{relationship field name})?` is the naive approach, but falls down because there are no rules for valid `id` formats. I.e., it would break if a resource object's `id` contained a colon (`:`). Perhaps there are clever ways around that?
* `rev`: this is deprecated by RFC8288 so I tried to reinforce this with a **SHOULD NOT** key word.

All other valid link-params are acceptable, like `hreflang` for example, in addition to any custom ones.

To preclude any misunderstanding with how [RFC8288 Link header link parameters are parsed](https://tools.ietf.org/html/rfc8288#appendix-B.3), I added the language stipulating that multi-value parameters must be arrays so that implementors do not try to use space-delimited or comma-delimited parameter values like one would for a `Link` header.

<hr/>

<a name="fn1">1</a>: I know I said that this PR should not discuss URI templating, but I just want to step in front of any criticism of the `params` member name by saying, "adding a `vars` member would be a perfectly compatible way to support templating later" ;) And that's all I want to say about that :P

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -545,6 +545,10 @@ Within this object, a link **MUST** be represented as either:
   * `meta`: a meta object containing non-standard meta-information about the
     link.
 
+Link objects **MAY** also contain the members `hreflang`, `title`, and `type`.
+Each of these members **MUST** be used in accordance with their semantics as
+defined by [RFC8288 Section 3.4.1](https://tools.ietf.org/html/rfc8288#section-3.4.1).
+
 If a link is represented as a string, or the `rel` member is not present on a
 link object, the link's relation type **SHOULD** be inferred from the name of
 the link object.
@@ -577,27 +581,23 @@ and `href` members be interchanged.
 > Note: Historically, a `rev` link parameter was used for this purpose but has
 > since been deprecated by [RFC8288 Section 3.3](https://tools.ietf.org/html/rfc8288#section-3.3).
 
-#### <a href="#document-links-target-attributes" id="document-links-target-attributes" class="headerlink"></a> Target attributes
+#### <a href="#document-links-target-attributes" id="document-links-target-attributes" class="headerlink"></a> Extension target attributes
 
-A "link parameter object" is used to represent the [target attributes](https://tools.ietf.org/html/rfc8288#section-2.2)
-of the [link object][link] in which it's defined.
+A "link parameter object" is used to represent additional [target attributes](https://tools.ietf.org/html/rfc8288#section-2.2)
+of the [link object][link] in which it appears.
 
-Link parameters **MAY** contain any valid JSON value. However, target
+Link parameter objects **MUST** only contain members that have been defined by
+their accompanying link relation type. Member names **SHOULD** be valid JSON:API
+[member names][member names] and **MUST** be valid target attribute names as
+defined by [RFC8288 Section 2.2](https://tools.ietf.org/html/rfc8288#section-2.2).
+
+Link parameters objects **MAY** contain any valid JSON value. However, target
 attributes that have a cardinality greater than one **MUST** be represented as
 an array of values.
 
-> Note: this means that a target attribute with multiple values should be an
-array of values, not a whitespace-separated string as might be used in a
-[`Link` header serialization](https://tools.ietf.org/html/rfc8288#section-3.5).
-
-All member names used in a link parameter object and not defined by this
-specification **MUST** be valid JSON:API [member names][member names] and
-**MUST** be valid target attribute names as defined by [RFC8288 Section 2.2](https://tools.ietf.org/html/rfc8288#section-2.2).
-
-In addition to members that conform to the requirements above, link parameter
-objects **MAY** contain the members `hreflang`, `media`, `title`, `title`, and
-`type`. Each of these members **MUST** be used in accordance with their
-semantics meanings as defined by [RFC8288 Section 3.4.1](https://tools.ietf.org/html/rfc8288#section-3.4.1).
+> Note: This means that a target attribute with multiple string values should
+not be represented as a single concatenated string with its values separated by
+whitespace as might be the case in a [`Link` header serialization](https://tools.ietf.org/html/rfc8288#section-3.5).
 
 #### <a href="#profile-links" id="profile-links" class="headerlink"></a> Profile Links
 
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -708,12 +708,6 @@ implements at least version 1.0 of the specification.
 > Note: Because JSON:API is committed to making additive changes only, the
 version string primarily indicates which new features a server may support.
 
-### <a href="#document-fragment-syntax" id="document-fragment-syntax" class="headerlink"></a> URI Fragment Syntax
-
-A JSON:API URI fragment **MUST** be a JSON Pointer [[RFC6901](https://tools.ietf.org/html/rfc6901)] which points to a
-[resource object](resource objects) or [relationships object](relationships) in the response document and it **MUST NOT**
-point to any child members within it.
-
 ### <a href="#document-member-names" id="document-member-names" class="headerlink"></a> Member Names
 
 All member names used in a JSON:API document **MUST** be treated as case sensitive
@@ -2498,7 +2492,6 @@ request as equivalent to one in which the square brackets were percent-encoded.
 [link object]: #document-links-link-object
 [link relations]: #document-links-link-relations
 [link parameter object]: #document-links-link-parameter-object
-[fragments]: #document-fragment-syntax
 [profiles]: #profiles
 [timestamps profile]: #profiles-timestamp-profile
 [profile keywords]: #profile-keywords
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -533,25 +533,20 @@ Within this object, a link **MUST** be represented as either:
 * <a id="document-links-link-object"></a>an object ("link object") which can
   contain the following members:
   * `href`: a URI-reference [[RFC3986 Section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1)] to the link's target.
-  * `rel`: a link relation type or an array of link relation types [[RFC8288 Section 2.1](https://tools.ietf.org/html/rfc8288#section-3.3)].
-    An array of link relationship types establishes multiple links that share
-    the same context, target, and target attributes.
+  * `rel`: a [link relation type][link relation type] defining the semantics of
+    the link.
   * `anchor`: a URI-reference [[RFC3986 Section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1)]
     to the link's context. By default, the context of a link is the
     [top-level object][top level], [resource object][resource objects], or
     [relationship object][relationships] in which the link appears.
-  * `params`: a [link parameter object][target attributes] as described
-    below.
+  * `params`: a [link parameter object][link parameters] describing additional
+    information about the link or its target.
   * `meta`: a meta object containing non-standard meta-information about the
     link.
 
 Link objects **MAY** also contain the members `hreflang`, `title`, and `type`.
 Each of these members **MUST** be used in accordance with their semantics as
 defined by [RFC8288 Section 3.4.1](https://tools.ietf.org/html/rfc8288#section-3.4.1).
-
-If a link is represented as a string, or the `rel` member is not present on a
-link object, the link's relation type **SHOULD** be inferred from the name of
-the link object.
 
 Except for the `profile` key in the top-level links object and the `type` 
 key in an [error object]'s links object, each key present in a links object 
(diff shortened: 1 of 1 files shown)


_format/1.1/index.md
```diff
@@ -534,10 +534,31 @@ Within this object, a link **MUST** be represented as either:
...
+  relation type as defined in [RFC8288 Section 2.1](https://tools.ietf.org/html/rfc8288#section-3.3) to include both registered
+  and extension relation types. When the `rel` member is not present, it **SHOULD** be
+  interpreted to contain the parent link object's key name.
+* `anchor`: a string containing URI reference as specified in
+  [[RFC8288 Section 3.2]](https://tools.ietf.org/html/rfc8288#section-3.2). When the `anchor` member is not present and the link belongs
+  to a resource object or relationship object, it **MUST** be interpreted to contain
+  the URI of the resource that generated the current response document with the
+  addition of the [URI fragment](fragments) identifying the object of which the link is a
+  member.
+* Any other valid target attribute name as explained in [RFC8288 Section 2.2](https://tools.ietf.org/html/rfc8288#section-3).
```
e0ipso: Maybe worth listing them in here for readability?

_format/1.1/index.md
```diff
@@ -613,6 +634,12 @@ implements at least version 1.0 of the specification.
 > Note: Because JSON:API is committed to making additive changes only, the
 version string primarily indicates which new features a server may support.
 
+### <a href="#document-fragment-syntax" id="document-fragment-syntax" class="headerlink"></a> URI Fragment Syntax
+
+A JSON:API URI fragment **MUST** be a JSON Pointer [[RFC6901](https://tools.ietf.org/html/rfc6901)] which points to a
+[resource object](resource objects) or [relationships object](relationships) in the response document and it **MUST NOT**
```
e0ipso: Is there a use case to point to a resource identifier inside of the relationship object?

gabesullice:
I was just digging back into JSON Schema and found some prior art WRT to using JSON Pointers as fragments: https://json-schema.org/latest/json-schema-core.html#rfc.section.5

_format/1.1/index.md
gabesullice: I don't believe so since resource identifier objects can't contain links.

_format/1.1/index.md
gabesullice: I don't think we should. The spec just establishes a few valid target attributes, but any implementation can make their own.

_format/1.1/index.md
```diff
@@ -613,6 +634,12 @@ implements at least version 1.0 of the specification.
 > Note: Because JSON:API is committed to making additive changes only, the
 version string primarily indicates which new features a server may support.
 
+### <a href="#document-fragment-syntax" id="document-fragment-syntax" class="headerlink"></a> URI Fragment Syntax
```
gabesullice: See https://tools.ietf.org/html/rfc7320#section-2.5

_format/1.1/index.md
```diff
@@ -530,13 +530,22 @@ of this member **MUST** be an object (a "links object").
...
+    below.
   * `meta`: a meta object containing non-standard meta-information about the
     link.
-  * Any link-specific target attributes described below.
+
+By default, the context of a link is the [top-level object][top level], [resource object][resource objects] or
+[relationship object][relationships] in which the link appears.
+
+If the `rel` member is not present on a link object, it **SHOULD** be interpreted to
+contain the member name of the link object.
```
gabesullice: This is a **SHOULD** because some existing link member names might not be valid link relation type names.

_format/1.1/index.md
```diff
@@ -585,6 +594,100 @@ that includes a [profile alias][profile aliases]:
...
+  * `remove`: The link's target points to a resource where a user agent can cause
+    a removal of all or part of the link's context.
+  * `relate`: The link's target points to a resource where a user agent can cause
+    a direct or indirect relationship to be established with the link's context.
+
+> Note: These link relation types correspond to operations described by the
+JSON:API specification and it is expected that implementations will use them to
+reference resources that will process requests according to [the requirements for
+creating, updating or deleting resources](https://jsonapi.org/format/1.1/#crud). However, this is not a strict
+limitation on their use.
```
gabesullice: Once this is in, I'd like to officially register these, which explains this note and the term "user agent" vs "client".

If for some reason they can't be registered, we can update the spec to say that these strings resolve to the extension-relation-types:

`https://jsonapi.org/format/#link-relation-add`
`https://jsonapi.org/format/#link-relation-update`
`https://jsonapi.org/format/#link-relation-remove`
`https://jsonapi.org/format/#link-relation-relate`

and add those anchor links.

_format/1.1/index.md
dgeb: Possible alternative wording:

> If the `rel` member is not present on a link object, the associated link's relation type **SHOULD** be interpreted as the name of the link.

_format/1.1/index.md
```diff
@@ -530,13 +530,22 @@ of this member **MUST** be an object (a "links object").
 <a href="#document-links-link" id="document-links-link"></a>
 Within this object, a link **MUST** be represented as either:
 
-* a string containing the link's URI.
+* a URI-reference [[RFC3986 Section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1)] to the link's target.
 * <a id="document-links-link-object"></a>an object ("link object") which can
   contain the following members:
-  * `href`: a string containing the link's URI.
+  * `href`: a URI-reference to the link's target.
```
dgeb: This is the only place that URI-reference is mentioned without also including a link to RFC3986 Section 4.1

_format/1.1/index.md
```diff
@@ -585,6 +594,100 @@ that includes a [profile alias][profile aliases]:
...
+attributes.
+
+A link parameter object **MAY** contain any of the following members: `hreflang`,
+`media`, `title`, and `type`. The values of these members **MUST** conform to their
+meanings as defined by [RFC8288 Section 3.4.1](https://tools.ietf.org/html/rfc8288#section-3.4.1).
+
+A link parameter object **MAY** contain other members as target attributes. Any
+target attribute that is in common usage in other link serialisations **SHOULD NOT**
+be used in a manner contrary to the commonly understood meaning of those
+attributes.
```
dgeb: > Any target attribute that is in common usage in other link serialisations **SHOULD NOT**
	be used in a manner contrary to the commonly understood meaning of those
	attributes.

This guidance feels a bit subjective.

_format/1.1/index.md
```diff
@@ -585,6 +594,100 @@ that includes a [profile alias][profile aliases]:
...
+
+Any additional member names **MUST** be valid target attribute names as defined by
+[RFC8288 Section 2.2](https://tools.ietf.org/html/rfc8288#section-2.2).
```
dgeb: It's unclear to me how this paragraph interacts with the previous paragraph.

dgeb (changes requested):
Just a few comments. Looks good overall @gabesullice

_format/1.1/index.md
gabesullice: I like that. I just tweaked your working a tiny bit.

_format/1.1/index.md
gabesullice: Fixed.

_format/1.1/index.md
gabesullice: You're right. I just removed it.

_format/1.1/index.md
gabesullice: I combined it with the previous one. Let me know it still feels disjointed.

_format/1.1/index.md
gabesullice: I also gave RFC8288 another read. I missed some nuance about multiple link relation types:

A link can only have _one_ link relation type, but one can add multiple link relation types to a serialization to establish >1 link in less "space". I copied some of the wording from [RFC8288 3.3](https://tools.ietf.org/html/rfc8288#section-3.3) for that.

_format/1.1/index.md
```diff
@@ -530,13 +530,24 @@ of this member **MUST** be an object (a "links object").
...
+  * `href`: a URI-reference [[RFC3986 Section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1)] to the link's target.
+  * `rel`: an array of link relation types [[RFC8288 Section 2.1](https://tools.ietf.org/html/rfc8288#section-3.3)]
+  * `anchor`: a URI-reference [[RFC3986 Section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1)] to the link's context.
+  * `params`: a [link parameter object][link parameter object] as described
+    below.
   * `meta`: a meta object containing non-standard meta-information about the
     link.
-  * Any link-specific target attributes described below.
+
+By default, the context of a link is the [top-level object][top level], [resource object][resource objects] or
```
dgeb: Minor nit - oxford comma

```suggestion
By default, the context of a link is the [top-level object][top level], [resource object][resource objects], or
```

dgeb: Maybe mention `anchor` here. Could be as simple as saying:

> By default, if `anchor` is not specified, the context of a link ....

Or follow the pattern used below:

> If the `anchor` member is not present, the context of a link ...

_format/1.1/index.md
```diff
@@ -585,6 +596,96 @@ that includes a [profile alias][profile aliases]:
...
+    an update the link's context.
+  * `remove`: The link's target points to a resource where a user agent can cause
+    a removal of all or part of the link's context.
+  * `relate`: The link's target points to a resource where a user agent can cause
+    a direct or indirect relationship to be established with the link's context.
+
+> Note: These link relation types correspond to operations described by the
+JSON:API specification and it is expected that implementations will use them to
+reference resources that will process requests according to [the requirements for
+creating, updating or deleting resources](https://jsonapi.org/format/1.1/#crud). However, this is not a strict
```
dgeb: Use a relative link here:

```suggestion
creating, updating or deleting resources](./#crud). However, this is not a strict
```

_format/1.1/index.md
```diff
@@ -585,6 +596,96 @@ that includes a [profile alias][profile aliases]:
...
+parameter object member:
+
+  * `add`: The link's target points to a resource where a user agent can cause
+    an addition to the link's context.
+  * `update`: The link's target points to a resource where a user agent can cause
+    an update the link's context.
+  * `remove`: The link's target points to a resource where a user agent can cause
+    a removal of all or part of the link's context.
+  * `relate`: The link's target points to a resource where a user agent can cause
+    a direct or indirect relationship to be established with the link's context.
```
dgeb: I'd like to discuss `relate` a bit.

_format/1.1/index.md
dgeb: It does seem better, but still not crystal clear. How about something like: 

> A link parameter object **MAY** contain other members, all of which **MUST** be considered target attributes. The names of target attributes **MUST** conform with [RFC8288 Section 2.2](https://tools.ietf.org/html/rfc8288#section-2.2).

dgeb: I suppose this section limits the ability to ever reserve the meaning of any members beyond `hreflang`, `media`, `title`, and `type`. This makes me a bit wary.

dgeb (commented):
Getting close! A few minor nits and a couple things to discuss.

_format/1.1/index.md
wimleers: I understand the proposed `add`, `update` and `remove` extension relation types. I too would like to understand the proposed `relate` extension relation type.

_format/1.1/index.md
```diff
@@ -585,6 +596,96 @@ that includes a [profile alias][profile aliases]:
...
+#### <a href="#document-links-link-relations" id="document-links-link-relations" class="headerlink"></a> Link relation types
+
+The following link relation types **MAY** be used as the value of the `rel` link
+parameter object member:
+
+  * `add`: The link's target points to a resource where a user agent can cause
+    an addition to the link's context.
+  * `update`: The link's target points to a resource where a user agent can cause
+    an update the link's context.
+  * `remove`: The link's target points to a resource where a user agent can cause
```
wimleers: https://jsonapi.org/format/#crud says "Creating, updating and deleting". Here, we say `add`, `update` and `remove`. The first and the last deviate from the terminology used by the spec today. Am I right in assuming this is to avoid these words from matching HTTP method names?

_format/1.1/index.md
```diff
@@ -585,6 +596,96 @@ that includes a [profile alias][profile aliases]:
...
+the client can create a new comment at the target URL and that requests to the
+target `href` can create a relationship to the article's `comments` field.
+
+```json
+"links": {
+  "self": "http://example.com/articles/1/relationships/comments",
+  "related": "http://example.com/articles/1/comments",
+  "comment": {
+    "href": "http://example.com/articles/1/comments",
+    "rel": ["add", "relate"]
```
wimleers: So this is where the `relate` extension link relation type comes into play.

It's not crystal clear to me. I think this is to convey that the act of adding a new resource object to the collection would also result in a new relationship being established?

_format/1.1/index.md
wimleers: My reading of this is slightly different. I think this is trying to specify that:
- *all* target attributes must conform to RFC8288 section 2.2
- the `hreflang`, `media`, `title` and `type` target attributes have a superset of requirements: they must *also* conform to RFC8288 section 3.4.1 (in addition to conforming to RFC8288 section 2.2)

If my interpretation is correct, I agree that this should be made more clear.

_format/1.1/index.md
```diff
@@ -585,6 +596,96 @@ that includes a [profile alias][profile aliases]:
...
+
+A link parameter object **MAY** contain other members as target attributes. Any
+additional member names **MUST** be valid target attribute names as defined by
+[RFC8288 Section 2.2](https://tools.ietf.org/html/rfc8288#section-2.2).
+
+Link parameter object members with multiple values **MUST** be represented with an
+array (i.e. a target attribute with multiple values should be an array of
+values, not a whitespace-separated string).
+
+A link parameter object **SHOULD NOT** contain a `rev` member.
```
wimleers: This seems oddly specific compared to everything else. But @dgeb didn't comment on this, so probably this is my being insufficiently informed to understand this without additional context.

_format/1.1/index.md
```diff
@@ -585,6 +596,96 @@ that includes a [profile alias][profile aliases]:
...
+
+In order to represent a link with reversed semantics, it is **RECOMMENDED** that an
+alternate link relation type be used or, less preferably, that the `anchor`
+and `href` members be interchanged.
+
+> Note: The `rev` link parameter was deprecated by [RFC8288 Section 3.3](https://tools.ietf.org/html/rfc8288#section-3.3)
```
wimleers: Aha, this explains it. To me, `rev` triggers `revision`, not `reverse`. I think it'd be clearer to first have the `In order to represent […] reversed […]` paragraph, followed by the `SHOULD NOT […] rev`. Then the `Note: […]` would also be positioned more closely and the reader would be left wondering for less time.

_format/1.1/index.md
gabesullice: I significantly cleaned up the section on target attributes. Very curious to know what you think.

_format/1.1/index.md
gabesullice: Removed. I've amended the rules for member names such that they must conform to the rules for member names in any other place in the document. Therefore, `rev` is not a valid member name by definition.

I've left in the guidance on how to represent reversed links, but taken out this over-specific normative statement.

_format/1.1/index.md
```diff
@@ -558,6 +569,48 @@ related resource collection:
...
+In order to represent a link with reversed semantics, it is **RECOMMENDED**
+that an alternate link relation type be used or, less preferably, that the
+`anchor` and `href` members be interchanged.
+
+> Note: Historically, a `rev` link parameter was used for this purpose but has
+> since been deprecated by [RFC8288 Section 3.3](https://tools.ietf.org/html/rfc8288#section-3.3).
+
+##### <a href="#document-links-link-parameters" id="document-links-link-parameters" class="headerlink"></a> Link parameters
+
+The value of the `params` member **MUST** be an object (an “link parameter
```
```suggestion
The value of the `params` member **MUST** be an object (a “link parameter
```

dgeb (approved):

e0ipso:
🎉

PR #1349, opened 2018-12-18
Add an editor(s) section to the profile template

ethanresnick:
See #1346


e0ipso:
This looks good to me. 👍

In `package.json` we usually have an author (individual or organization) + contributors (in this case editors). Maybe worth considering. However the organization will be derived from the description and profile name, so not critical.

wimleers:
LGTM

PR #1352, opened 2018-12-20
 feat(profiles): add naming conventions req for review process

ethanresnick:
While we didn’t go so far as requiring all profiles use camelCasing, I think it’s probably a good idea to require that all _registered_ profiles do, to make sure they look nice when used together.

Although I can also see an argument that this requirement will annoy authors during review, so I'm not wedded to it; seems more like a "nice to have".

### DIFF
--- a/extensions/index.md
+++ b/extensions/index.md
@@ -84,8 +84,8 @@ Once submitted, one of JSON:API's editors will review your profile to check that
 1. follows the template above; 
 2. is specified precisely enough to enable interoperable implementations;
 3. complies with the JSON:API spec and its [requirements for profiles](/format/1.1/#profiles-authoring);
-3. follows JSON:API's [recommended naming conventions](https://jsonapi.org/recommendations/#naming); and
-4. wouldn't cause any problems were it to become widely adopted. 
+4. follows JSON:API's [recommended naming conventions](https://jsonapi.org/recommendations/#naming); and
+5. wouldn't cause any problems were it to become widely adopted. 
 
 If your profile meets these criteria, it will generally be **approved within a week**.
 
--- a/extensions/index.md
+++ b/extensions/index.md
@@ -87,7 +87,7 @@ Once submitted, one of JSON:API's editors will review your profile to check that
 3. follows JSON:API's [recommended naming conventions](https://jsonapi.org/recommendations/#naming); and
 4. wouldn't cause any problems were it to become widely adopted. 
 
-If your profile meets these three criteria, it will generally be **approved within a week**.
+If your profile meets these criteria, it will generally be **approved within a week**.
 
 In limited cases (e.g., if your profile defines a new, fundamental mechanism for
 doing something "architectural" that other profiles may need to do too), it might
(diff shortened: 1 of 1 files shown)


gabesullice:
Even though it wasn't a problem for us, I think this would be a problem for existing implementations that already have profile-like behavior that they want to standardize. It essentially limits registered profiles to green-field development unless the implementation just _happened_ to use camel-casing to begin with.

Maybe you think there are not many actual profile-like custom behaviors already out there?

wimleers:
> It essentially limits registered profiles to green-field development unless the implementation just happened to use camel-casing to begin with.

I think it may be reasonable to have those existing implementations treat `?foo_bar` and `?fooBar` as equivalents. One extra if-statement, or one bit of extra URL normalization isn't going to be a showstopper for them IMHO.

>  I think it’s probably a good idea to require that all registered profiles do, *to make sure they look nice when used together.*

I think this is a reasonable argument to do this.

extensions/index.md
```diff
@@ -83,7 +83,8 @@ Once submitted, one of JSON:API's editors will review your profile to check that
 
 1. follows the template above; 
 2. is specified precisely enough to enable interoperable implementations;
-3. complies with the JSON:API spec and its [requirements for profiles](/format/1.1/#profiles-authoring); and
+3. complies with the JSON:API spec and its [requirements for profiles](/format/1.1/#profiles-authoring);
+3. follows JSON:API's [recommended naming conventions](https://jsonapi.org/recommendations/#naming); and
```
wimleers: s/3/4/

extensions/index.md
```diff
@@ -83,7 +83,8 @@ Once submitted, one of JSON:API's editors will review your profile to check that
...
 4. wouldn't cause any problems were it to become widely adopted.
```
wimleers: s/4/5/

wimleers (changes requested):
I feel ridiculous pointing this out, but I guess somebody should? :P

extensions/index.md
gabesullice: Maybe these should all just be `1`. Markdown (at least Github's flavor) should auto-increment the numbers.

wimleers: I guess that could be why Ethan did it this way. In that case, ignore this :)

ethanresnick: Haha, @wimleers you're giving me too much credit :P I just messed up the numbering. I'm fine with using all `1`s, but we should probably be consistent about that throughout. Since I think we use proper numbering elsewhere, I've just fixed the numbering here for now.

dgeb (approved):

dgeb:
I think that the naming convention requirement is important for consistency among formally registered profiles. 

There are a couple non-breaking options for implementations that use different naming conventions and want to use profiles:

* Don't register the profile here; instead, create a URI at another domain.
* Register a modernized version of the profile here and use it together with [aliases](https://jsonapi.org/format/1.1/#profile-keywords-and-aliases) to support an alternative naming convention.

PR #1354, opened 2018-12-21
feat(profiles): server-required profiles

ethanresnick:
This is one final [section from #1195](https://github.com/json-api/json-api/pull/1195/files?utf8=%E2%9C%93&diff=unified#diff-571f6cee9a492e0e177801c21a4b2b2aR1967) that I realized never made it over to the current spec.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -2226,6 +2226,28 @@ because the value for the `updated` key is not recognized under the profile's
 requirement that the `updated` key hold a string of the form produced by 
 `JSON.stringify`.
 
+### <a href="#profiles-required" id="profiles-required" class="headerlink"></a> Server-Required Profiles 
+
+Generally, servers should not require that clients apply specific profiles to 
+their requests, as such requirements reduce interoperability.
+
+However, in many common cases, a server does require that clients provide 
+certain data (e.g., authentication credentials), and it **MAY** require that 
+this data be provided by applying a specific profile (or any one profile from 
+a set of supported profiles) to the request. 
+
+In the event that the server requires the client to apply a profile to its 
+request, and the client fails to do so, the server **MUST** send a
+`415 Unsupported Media Type` error. 
+
+In the error's response body, there **MUST** be an [error object] for each 
+required profile that is missing, and this error object **MUST**:
+
+1. have a `type` who's URI is `http://jsonapi.org/errors/missing-required-profile`
+2. include the URI of the missing profile in the `extension` member of the 
+   `missing` object, with the `alternatives` key provided if the client could've 
+   used other profiles instead to satisfy the server's requirements.
+
 ### <a href="#profiles-authoring" id="profiles-authoring" class="headerlink"></a> Authoring Profiles
 
 A profile **MAY** assign meaning to elements of the document structure whose use
@@ -2371,6 +2393,23 @@ An error object **MAY** have the following members:
     exists; if it doesn't, the client **SHOULD** simply ignore the pointer.
   * `parameter`: a string indicating which URI query parameter caused
     the error.
+  * <a id="error-objects-source-missing"></a>`missing`: an object indicating
+    that the error was caused by a missing profile, JSON object key (or array 
+    index), or URI query parameter. If present, this object **MUST** contain 
+    exactly one of the following keys:
+      * `extension`: the URI of a missing profile. This key **SHOULD** be
+        used with the `pointer` member to indicate which value in the request
+        document is missing the profile.
+      * `key`: the name of a missing key in a JSON object, or a missing index in 
+        a JSON array, expressed as a string. This key **SHOULD** be used with 
+        the `pointer` member to indicate which object/array in the request 
+        document is missing the key.
+      * `parameter`: the name of a missing URI query parameter.
+    
+      This object **MAY** also contain an `alternatives` key, whose value is an
+      array listing alternate profile URIs, key names, or query parameter names
+      (depending on which type of value is missing) that can be added to resolve
+      this error.
 * `meta`: a [meta object][meta] containing non-standard meta-information about the
   error.
 


wimleers:
#1999 does not exist :) Not sure which issue/PR you're referring to?

ethanresnick:
Whoops. I've updated the OP.

_format/1.1/index.md
```diff
@@ -2226,6 +2226,28 @@ because the value for the `updated` key is not recognized under the profile's
 requirement that the `updated` key hold a string of the form produced by 
 `JSON.stringify`.
 
+### <a href="#profiles-required" id="profiles-required" class="headerlink"></a> Server-Required Profiles 
+
+Generally, servers should not require that clients apply specific profiles to 
+their requests, as such requirements reduce interoperability.
```
wimleers, line 2232: This is a _very_ important sentence for understanding how profiles are meant to be implemented for maximum interoperability! I'm glad we're adding it after all :)

_format/1.1/index.md
```diff
@@ -2226,6 +2226,28 @@ because the value for the `updated` key is not recognized under the profile's
...
+
+However, in many common cases, a server does require that clients provide 
+certain data (e.g., authentication credentials), and it **MAY** require that 
+this data be provided by applying a specific profile (or any one profile from
```
wimleers, line 2236: I find this example not very convincing. I cannot imagine authentication is often embedded into JSON:API documents. Authentication is an orthogonal concern.

Which leaves me wondering: what _is_ a good example of this? Why is this "common"?

_format/1.1/index.md
```diff
@@ -2226,6 +2226,28 @@ because the value for the `updated` key is not recognized under the profile's
...
+
+In the event that the server requires the client to apply a profile to its 
+request, and the client fails to do so, the server **MUST** send a
+`415 Unsupported Media Type` error. 
+
+In the error's response body, there **MUST** be an [error object] for each 
+required profile that is missing, and this error object **MUST**:
+
+1. have a `type` who's URI is `http://jsonapi.org/errors/missing-required-profile`
+2. include the URI of the missing profile in the `extension` member of the
```
wimleers, line 2247: s/extension/profile/

_format/1.1/index.md
```diff
@@ -2371,6 +2393,23 @@ An error object **MAY** have the following members:
     exists; if it doesn't, the client **SHOULD** simply ignore the pointer.
   * `parameter`: a string indicating which URI query parameter caused
     the error.
+  * <a id="error-objects-source-missing"></a>`missing`: an object indicating
+    that the error was caused by a missing profile, JSON object key (or array 
+    index), or URI query parameter. If present, this object **MUST** contain 
+    exactly one of the following keys:
+      * `extension`: the URI of a missing profile. This key **SHOULD** be
```
wimleers, line 2400: s/extension/profile/

_format/1.1/index.md
```diff
@@ -2371,6 +2393,23 @@ An error object **MAY** have the following members:
     exists; if it doesn't, the client **SHOULD** simply ignore the pointer.
   * `parameter`: a string indicating which URI query parameter caused
     the error.
+  * <a id="error-objects-source-missing"></a>`missing`: an object indicating
+    that the error was caused by a missing profile, JSON object key (or array 
+    index), or URI query parameter. If present, this object **MUST** contain
```
wimleers, line 2398: I can't think of a use case for the missing array index. What am I missing?

wimleers (changes requested):
Looking good overall! Two nits, two questions.

_format/1.1/index.md
gabesullice: I think it's strange that this statement broadly contradicts the one before it by saying "_Generally_ servers should not require profiles" and then "in many common cases, servers can require profiles".

I think it's hard to think of an example where extra data must be provided (a `revision` key is the best I could come up with), but I can think of requirements about _how_ that data is provided (i.e. different profiles for different date/time formats).

ethanresnick: > I find this example not very convincing. I cannot imagine authentication is often embedded into JSON:API documents. Authentication is an orthogonal concern.

Authentication isn't a great example; I was imagining a case where maybe a request signature could show up in the body (e.g., in top-level `meta`), but Gabe's examples are better tbh. I can update the PR and the wording.

> I think it's strange that this statement broadly contradicts the one before it by saying "Generally servers should not require profiles" and then "in many common cases, servers can require profiles".

How would you suggest wording it?

_format/1.1/index.md
ethanresnick: E.g., a profile that takes data in an `[x,y,z]` coordinate pair is given an array with fewer entries. More importantly, though, it just seems good to general here

_format/1.1/index.md
gabesullice: > A server **MAY** require that a client applies a specific profile to its request.
>
> However, it is **NOT RECOMMENDED** for servers to require profiles which require additional query parameters or additional request document members, as such requirements reduce interoperability between clients and servers.
>
> >Note: Profiles with optional document members, specific value formats (e.g. ISO 8601 date and time format), or specific query parameters within a query parameter family already reserved by this specification should be treated with care, but do not have the same level of interoperability concerns.

gabesullice: My theory here is that:

1. anything optional obviously doesn't have interop concerns.
2. query parameters within query parameter families are usually implementation specific (filter, sort, etc) so clients will likely not have them hardcoded.
3. like #2, value formats are usually "passed through" a client library and aren't hardcoded; they typically come from a frontend developer working against a specific backend.

dgeb:
I think this PR brings to a head some concerns I've had about requirements for clients. 

In https://jsonapi.org/format/1.1/#profiles-sending we place the same MUST requirement on clients as on servers:

> Clients and servers MUST include the profile media type parameter in conjunction with the JSON:API media type in a Content-Type header to indicate that they have applied one or more profiles to a JSON:API document.

Since profiles are additive to the base spec, I believe that this MUST could be relaxed to a SHOULD.

Similarly, if a server _requires_ that a profile be applied, then it could operate as if the document it receives has applied that profile without it being explicitly stated. Of course, it would still need to respond with an error if a conflicting profile has been applied OR if the required profile has been applied incorrectly.

By making servers more forgiving about accepting documents without explicitly declared profiles we could simplify the experience for clients without harm. This seems inline with [allowing clients to omit the `profile` query parameter](https://jsonapi.org/format/1.1/#profile-query-parameter-omitting).

_format/1.1/index.md
wimleers: The coordinate example helps, thanks!

But I thought profiles only could add elements to the document structure (per 
https://jsonapi.org/format/1.1/#profile-keywords-and-aliases), which means they must have a name, and hence a JSON object key? Wouldn't the coordinate example be a *value* of an object key, and therefore `z` being absent from `[x,y,z]` would not result in a `missing` error since the key would be present, the value itself would just be invalid. In other words: `invalid`, not `missing`?

I'm probably still missing something, sorry.

wimleers:
> Since profiles are additive to the base spec, I believe that this MUST could be relaxed to a SHOULD.

If the client is already updated to specifically send a JSON:API document with a profile applied, then why is requiring that profile to be specified in the media type as well unreasonable?

_format/1.1/index.md
ethanresnick: > Wouldn't the coordinate example be a _value_ of an object key, and therefore z being absent from `[x,y,z]` would not result in a missing error since the key would be present, the value itself would just be invalid.

It's both. I.e., the value is invalid because a key (index 2) is missing.

In other words, if a client client sends: 

```
{ "meta": { "homeBase": [x,y] }
```

then the server's error object would ideally contain `"pointer": "/meta/homeBase"`, whether it contains `"missing"` or not. But that just tells you that the value at `/meta/homeBase` had an error. Being able to supplement the pointer with `"missing": { "key": "2" }` just makes it a bit more specific about what the error was.

In theory, you never _need_ `"missing": { "key": "... " }`, because you can always reinterpret your error as being not that some key is missing, but that the value that's supposed to contain the key (pointed to by "pointer") is invalid. But `missing` is a way to be more specific about "invalid how".

auvipy (approved):
re base please

_format/1.1/index.md
```diff
@@ -2371,6 +2393,23 @@ An error object **MAY** have the following members:
...
+    that the error was caused by a missing profile, JSON object key (or array 
+    index), or URI query parameter. If present, this object **MUST** contain 
+    exactly one of the following keys:
+      * `extension`: the URI of a missing profile. This key **SHOULD** be
+        used with the `pointer` member to indicate which value in the request
+        document is missing the profile.
+      * `key`: the name of a missing key in a JSON object, or a missing index in 
+        a JSON array, expressed as a string. This key **SHOULD** be used with 
+        the `pointer` member to indicate which object/array in the request 
+        document is missing the key.
```
bradjones1, line 2406: Even if server-required profiles doesn't make it into 1.1, I think this is a worthwhile addition, paired with the language added in https://github.com/json-api/json-api/commit/e75cec15f0e663a9676fcc746b0801a773394647 to clarify that the pointer must reference an element in the original request. In the case of a missing key, this should be paired with some other data point (e.g., this element) to tell the client exactly which data was missing.

bradjones1:
To @wimleers' question in https://github.com/json-api/json-api/pull/1354#issuecomment-453464835

> If the client is already updated to specifically send a JSON:API document with a profile applied, then why is requiring that profile to be specified in the media type as well unreasonable?

I think this is precisely because profile functionality has predated the official release of json:api 1.1. E.g., I have a client that is implementing Drupal's fancy filtering without sending a content-type parameter indicating so. I don't think we can require clients to upgrade their request parameters simultaneous with the server.

However, that doesn't mean this improvement is for nothing, in so far as it aids in compatible profile negotiation; as @dgeb points out:

> Of course, it would still need to respond with an error if a conflicting profile has been applied OR if the required profile has been applied incorrectly.

Again, in Drupal's case, this could mean we detect some sort of problem in the fancy-filtering request parameters, and can throw and translate exceptions to the json:api implementation to indicate that we should tell the client precisely which profile they must implement (correctly) to remedy the problem.

dgeb:
This PR was written at a time when we were considering "profiles" to provide a single answer to the separate roles that profiles and extensions now play in v1.1. As such it is outdated, although the work has definitely informed the current state of the spec. Many thanks to @ethanresnick for his work here!

> Again, in Drupal's case, this could mean we detect some sort of problem in the fancy-filtering request parameters, and can throw and translate exceptions to the json:api implementation to indicate that we should tell the client precisely which profile they must implement (correctly) to remedy the problem.

@bradjones1 Yes, I think this is reasonable. And hopefully clients / servers will soon start explicitly negotiating profiles.

watnoii24121997 (commented):
ควยงง

PR #1283, opened 2018-07-09
Add Drupal's `jsonapi` module, plus its ecosystem modules: `jsonapi_e…

wimleers:
…xtras` + `openapi`

Adding these to `jsonapi.org` was *long* overdue!

PR #1355, opened 2018-12-21
Drupal's JSON:API implementation was wrongly listed under "client lib…

wimleers:
…raries"

… but it's a _server_. So "server libraries" would be a better fit, but even that would be wrong actually. It's _not_ just a library. As soon as you install it, it _works_. It exposes Drupal's entire data model.

So … a new "servers" section would be more accurate. For now, just moving it to "server libraries".

Can't believe none of us noticed this before!


gabesullice:
https://github.com/json-api/json-api/pull/1283#discussion_r201040521

I did... I just completely forgot about this!

wimleers:
Hah!

ethanresnick:
👍

PR #1357, opened 2018-12-22
feat(errors): define some common error URIs

ethanresnick:
This is inspired by my comment [here](https://github.com/json-api/json-api/pull/1330#issuecomment-449505662):

> I’d love to create a set of error URIs on jsonapi.org that show canonical docs when dereferenced. Then, all the server implementations can point to those and get docs for free.

This also allows client implementations to recognize common errors automatically, if server implementations use these URIs.

@wimleers Like I mentioned in the other issue, you should have push access to this branch, so feel free to add other errors you think are relevant.

@dgeb Thoughts on having URIs like this on jsonapi.org?


Panman82:
FWIW, I already point to specific scheme headers for some errors, for cases where the spec says something MUST be this way and the request isn't...

PR #1359, opened 2018-12-22
profiles: handling mismatch between `content-type` and `links.profile` list of profiles

ethanresnick:
As @wimleers [pointed out](https://github.com/json-api/json-api/pull/1268#discussion_r192717689) in his original review of #1268, we don't currently say what happens when a different list of profiles is specified in the document (in `links.profile`) than in the `Content-Type` header.

I've been thinking about a bunch of options and have concluded that the solution I offered in my original response to Wim's comment — i.e., making any mismatch trigger a 400 — is the wrong way to go. One issue with that solution is that it only covers documents sent to the server; i.e., it doesn't cover how the client should process documents sent from the server that have the same mismatch. (We could say that the client must entirely discard such responses... but I'm not sure clients would actually do that.) Another problem with the 400 approach — and actually any approach that requires the recipient to read both lists of profiles — is that it introduces a preprocessing step (e.g., to check that the lists match), and we have to pay the cost of that preprocessing step on every request, even though a mismatch should be a very rare event. That seems like a bad idea.

So, my preferred approach at this point, which is embodied in this PR, is simply to treat `links.profile` as the definitive list of applied profiles in the event of any mismatch. By treating one list as definitive, recipients can simply ignore the other (no comparison step required), but we can keep both in the spec, which is nice because both bring something of value. (Specifically, if we removed the parameter on `Content-Type`, profile negotiation would look pretty asymmetrical; i.e., the client would use the `profile` parameter in `Accept`, but not get it back in the response. Meanwhile, `links.profile` is offering an elegant place to list profile aliases, which have to go in the document somewhere.)

I chose to treat `links.profile` as the definitive source because recipients have to look there anyway before processing the document (to check for aliases), so making that the only place they have to look to determine the applied profiles seemed to make the most sense.

Note: one small issue with making `links.profile` definitive is that the "missing a server-required profile error" from #1354 would probably have to be a 400 rather than a 415, because the required profile actually _could_ be listed in `Content-Type` but still be "missing" if its not listed in `links.profile`. I think that's fine, but it means #1354 will have to be tweaked after this PR is merged.

Note 2: Wim also pointed out that we don't say what should happen when the client lists a profile in the `profile` query parameter that isn't in its `Accept` header. But I think that case is already covered implicitly, and the answer is that the server should pay attention to the query parameter. That's because JSON:API says that the `profile` query parameter lists _required_ profiles -- while both HTTP and JSON:API say that a profile can be applied to a response document even if it isn't in the `Accept` header.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1948,6 +1948,11 @@ include a [top-level][top level] [`links` object][links] with a `profile` key,
 and that `profile` key **MUST** include a [link] to the URI of each profile
 that has been applied.
 
+> Beware: while senders are required to list the same profiles in the 
+> `Content-Type` header and the document's `profile` links, recipients must 
+> properly handle the case that these sets of profiles are different (because 
+> the request is malformed). See ["Identifying a Document's Profiles"](#profiles-processing-identifying-applied-profiles).
+
 When an older JSON:API server that doesn't support the `profile` media type
 parameter receives a document with one or more profiles, it will respond with a
 `415 Unsupported Media Type` error.
@@ -1958,9 +1963,8 @@ it has applied to the document and retry its request without the `profile` media
 type parameter. If this resolves the error, the client **SHOULD NOT** attempt to
 use profile extensions in subsequent interactions with the same API.
 
-> The most likely other causes of a 415 error are that the server doesn't
-support JSON:API at all or that the client has failed to provide a required
-profile.
+> The most likely other cause of a 415 error is that the server doesn't support 
+> JSON:API at all.
 
 ### <a href="#profile-query-parameter" id="profile-query-parameter" class="headerlink"></a> `profile` Query Parameter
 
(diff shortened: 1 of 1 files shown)


wimleers:
>  have to pay the cost of that preprocessing step on every request

That's indeed a strong reason.

> So, my preferred approach at this point, which is embodied in this PR, is simply to *treat `links.profile` as the definitive list* of applied profiles in the event of any mismatch. *By treating one list as definitive, recipients can simply ignore the other*

This seems like a pragmatic solution.

But also somewhat confusing, because in the ideal case, the same information is present twice in the same response (once in the header, once in the body). In the less-than-ideal case, it'd be different.

Perhaps we can at least ensure the ideal case by making a strong recommendation in the spec: recommending to _compute_ the `Content-Type` response header on every response based on the value in the `links.profile` response document?

ethanresnick:
> Perhaps we can at least ensure the ideal case by making a strong recommendation in the spec: recommending to compute the Content-Type response header on every response based on the value in the links.profile response document?

Given that the spec already [(intends to) _require_](https://github.com/json-api/json-api/pull/1359/files#diff-571f6cee9a492e0e177801c21a4b2b2aR1940) that senders make the two lists of profiles identical, it feels a bit overboard to suggest that implementations meet that requirement in a particular way (e.g., by computing content-type from links), when they could just as reliably use some other strategy (e.g., compute links.profile from content-type).

However, I do think it was a bit unclear that the spec was requiring these lists to be identical, so I've added a commit clarifying that.

wimleers:
It _may_ be valuable to explain the "why" in the new `Identifying a Document's Profiles` section — i.e. why did we pick this one and not the other?  You've explained it in this issue. I'm saying that bringing that rationale into the spec may be valuable.

I think this is a step forward. Let's do this? 🚢

ethanresnick:
> I think this is a step forward. Let's do this? 🚢
👍 Requesting @dgeb for review.

> It may be valuable to explain the "why" in the new Identifying a Document's Profiles section 

I'm feeling a bit swamped to think through the text of that right now, but would be totally open to a PR to that effect (either made directly to this `different-profile-lists` branch, or to `gh-pages` after this is merged).

auvipy (changes requested):
rebase, please

auvipy:
@dgeb should this be closed?

bradjones1:
Trying to help with some issue cleanup to help get 1.1 in release state.

`links.profile` was removed at https://github.com/json-api/json-api/pull/1484/files#diff-a80e2578f8c89d0702e3677e4f40fd400711ffd6be44a637a5e0b61560d378feL135 though I can find no information in the PR comments or related issues as to why. That being said, this isn't in the spec anymore so this is likely stale? There's no mismatch if there's no `links.profile` member to contradict the `content-type`?

dgeb:
Closing in favor of #1607, which addresses effectively the same issue for 1.1 in its current state.

PR #1368, opened 2019-02-10
'Commercial at' mentioned as reserved

lode:
Now @-Members are in v1.1, shouldn't the [reserved characters](https://jsonapi.org/format/1.1/#document-member-names-reserved-characters) be updated? That one still mentions that the 'commercial at' (@) is not allowed anywhere.

I understand it is only for @-Members, and only as the first char, so maybe a small clarification like 'except as first character in @-Members' would be good?


ethanresnick:
Thanks!

PR #1369, opened 2019-02-12
Adds Grivet JSON:API client to TypeScript implementations

danielmenzel:
We open sourced our JSON:API client lib written in TypeScript. It focuses on traversing resources according to HATEOAS principles. It would be great if it is added to the list of client implementations.


ethanresnick:
Thanks!

PR #1374, opened 2019-02-27
Typo in cursor pagination profile

lode:


wimleers:
Nice find! LGTM :)

wimleers:
Oh, hah, you work for https://decorrespondent.nl/ — awesome :) I've been a subscriber for years.

I see you're building https://github.com/lode/jsonapi, your own PHP-based JSON:API implementation. Why build and maintain your own? It looks like you started working on it a long time ago, so probably that's why: few implementations existed back then.

ethanresnick:
Thanks!

lode:
You're welcome!

I build it because I love the idea of JSON:API and because I had a feeling an implementation could be simple to use without knowing the spec, which most others didn't do back then. Further it is just fun to make stuff right? :)

Btw, great you're a member of DC!

PR #1376, opened 2019-02-27
Address spelling mistake in specification

michaelsw95:
"heterogenous" seems to be a medical/biological term. I think this is what was intended: "heterogeneous"

### DIFF
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -1560,7 +1560,7 @@ is an empty array or an array of [resource identifier objects][resource identifi
 If a client makes a `PATCH` request to a URL from a to-many
 [relationship link][relationships], the server **MUST** either completely
 replace every member of the relationship, return an appropriate error response
-if some resources cannot be found or accessed, or return a `403 Forbidden`
+if some resources can not be found or accessed, or return a `403 Forbidden`
 response if complete replacement is not allowed by the server.
 
 For example, the following request replaces every tag for an article:
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -82,7 +82,7 @@ A document **MUST** contain at least one of the following top-level members:
 * `meta`: a [meta object][meta] that contains non-standard
   meta-information.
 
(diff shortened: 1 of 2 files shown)


_format/1.0/index.md
```diff
@@ -1560,7 +1560,7 @@ is an empty array or an array of [resource identifier objects][resource identifi
 If a client makes a `PATCH` request to a URL from a to-many
 [relationship link][relationships], the server **MUST** either completely
 replace every member of the relationship, return an appropriate error response
-if some resources can not be found or accessed, or return a `403 Forbidden`
+if some resources cannot be found or accessed, or return a `403 Forbidden`
```
dgeb: After reviewing a few sources, it appears that `can not` is just as acceptable as `cannot`. Is there a reason you made this change?

dgeb:
Thanks for these corrections. As noted inline, my only concern is regarding "can not" vs. "cannot", and my preference to avoid making unnecessary changes.

_format/1.0/index.md
michaelsw95: Agreed, my mistake. I have reverted this.

_format/1.0/index.md
```diff
@@ -82,7 +82,7 @@ A document **MUST** contain at least one of the following top-level members:
 * `meta`: a [meta object][meta] that contains non-standard
   meta-information.
 
-The members `data` and `errors` **MUST NOT** coexist in the same document.
+The member's `data` and `errors` **MUST NOT** coexist in the same document.
```
tobyzerner: This change is incorrect. The `data` and `errors` do not belong to "the member", they ARE members (plural) themselves.

tobyzerner:
For reference re: heterogenous vs heterogeneous
https://grammarist.com/usage/heterogeneous-heterogenous/

_format/1.0/index.md
michaelsw95: My mistake, I have reverted this change

PR #1379, opened 2019-03-07
Schema should allow link objects as pagination links

gabesullice:
As a rule, our JSON:API implementation uses link objects for every link. This ensures clients written against our implementation are maximally forward compatible (so they don't break if and when a link has metadata added to it).

We've even applied this rule to pagination links. We've recently discovered that [our responses are failing validation for this reason](https://www.drupal.org/project/jsonapi/issues/3037852#comment-13001794).

It seems like this is a small oversight in the schema and is not an intentional validation rule.

I need to validate that this schema change is correct tomorrow. However, I wanted to throw the PR out there for any initial reactions in case it actually _is_ intentional and I've missed something. Therefore, I'm leaving this as a draft (which, BTW, is a nifty new GH feature!).

### DIFF
--- a/schema
+++ b/schema
@@ -292,28 +292,28 @@
         "first": {
           "description": "The first page of data",
           "oneOf": [
-            { "type": "string", "format": "uri-reference" },
+            { "$ref": "#/definitions/linkage" },
             { "type": "null" }
           ]
         },
         "last": {
           "description": "The last page of data",
           "oneOf": [
-            { "type": "string", "format": "uri-reference" },
+            { "$ref": "#/definitions/linkage" },
             { "type": "null" }
           ]
         },
         "prev": {
           "description": "The previous page of data",
           "oneOf": [
-            { "type": "string", "format": "uri-reference" },
+            { "$ref": "#/definitions/linkage" },
             { "type": "null" }
           ]
         },
         "next": {
           "description": "The next page of data",
           "oneOf": [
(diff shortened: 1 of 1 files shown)


schema
```diff
@@ -292,28 +292,28 @@
         "first": {
           "description": "The first page of data",
           "oneOf": [
-            { "type": "string", "format": "uri-reference" },
+            { "$ref": "#/definitions/linkage" },
```
wimleers: s/linkage/link/

I suspect this is a typo caused by IDE autocompletion :)

wimleers:
This appears to be essentially the same problem as #1159 encountered and #1160 fixed.

The spec talks about [pagination links](https://jsonapi.org/format/#fetching-pagination) as being a set of 4 keys that MUST be used for pagination links, and it also talks about how [each link](https://jsonapi.org/format/#document-links) MUST be either a string or a link object.

The schema currently _requires_ it to be a string, and that's what this PR fixes (at least once the typo I mentioned above is fixed).

wimleers:
Created PR to fix my comment above (https://github.com/json-api/json-api/pull/1379#pullrequestreview-212263141)

gabesullice:
Thanks Wim!

wimleers:
LGTM! 🚢 

EDIT: as they say, the proof is in the pudding: https://www.drupal.org/project/jsonapi/issues/3037852#comment-13004970 :)

PR #1380, opened 2019-03-27
Bulk profile

kbairak:
This is a proposal for bulk operations. Contrary to the JSON-patch approach where arbitrary operations can be sent to the API, this proposal limits itself to doing the same type of operation (eg `PATCH`) on the same API resource collection.

My approach for creating this profile was reading the specification for single operations and try to cover every restriction and concern raised there and adapt it to bulk operations.

### DIFF
--- a/_profiles/transifex/bulk/index.md
+++ b/_profiles/transifex/bulk/index.md
@@ -1,0 +1,396 @@
+---
+name: Bulk
+short_description: Allow bulk operations on API endpoints
+
+extended_description: |
+  Allow multiple POST, PATCH and DELETE operation on the same endpoint with the
+  same request. Every bulk operation must be applied on the same API resource
+  collection and must be of the same type.
+
+minimum_jsonapi_version: 1.1
+
+discussion_url: http://github.com/json-api/json-api/pull/1380
+
+editors:
+  - name: Konstantinos Bairaktaris
+    email: kbairak@transifex.com
+
+categories:
+  - Bulk Operations
+---
+
+# Bulk profile
+
+## Introduction
+
+This page specifies a profile for the `application/vnd.api+json` media type,
+as described in the [JSON:API specification](http://jsonapi.org/format/).
+
+This profile allows clients to perform a number of operations with only one
+request. Instead of supporting arbitrary operations on the whole API, this
+profile limits operations of the same type (aka HTTP verb: `POST`, `PATCH`,
+`DELETE`) on the same API resource.
+
+## Creating Resources In Bulk
+
+### Requests
+
+A collection of resources can be created by sending a `POST` request to a URL
+that represents a collection of resources. The request **MUST** include an
+array of resource objects as primary data. All items of the array **MUST** have
+the same `type` member. The same restrictions that apply to a resource object
+when creating a single resource apply to each item of the array. Either all the
+items in the request **MUST** include client-generated IDs, or none **MUST**.
+
+Every bulk create operation **MUST** be atomic. If the server has changed, all
+the requested items **MUST** have been created. Also, no side-effects are to be
+expected to happen asynchronously on the server's side; assuming no further
+interaction with the server, the returned items **SHOULD** remain as they are
+and further attempts to fetch them **SHOULD** return the same resource objects.
+
+For instance, a collection of articles might be created with the following
+request:
+
+```
+POST /articles HTTP/1.1
+Content-Type: application/vnd.api+json;profile="https://github.com/json-api/json-api/_profiles/transifex/bulk/index.md"
+Accept: application/vnd.api+json;profile="https://github.com/json-api/json-api/_profiles/transifex/bulk/index.md",application/vnd.api+json
+
+{
+  "data": [
+    {
+      "type": "articles",
+      "attributes": {"title": "Creating resources in bulk"}
+    },
+    {
+      "type": "articles",
+      "attributes": {"title": "Creating many resources at the same time"}
+      "relationships": {"author": {"data": {"type": "users", "id": "1"}}}
+    }
+  ]
+}
+```
+
+### Responses
+
+#### `201 Created`
+
+If the requested items did not include Client-Generated IDs and the requested
+resources have been created successfully, the server **MUST** return a
+`201 Created` status code.
+
+The server **MUST NOT** include a `Location` header.
+
+The response **MUST** also include a document that contains an array with the
+primary resources created, in the same order as they appear in the request.
+
+For instance, A response when all requested items were created can be:
+
+```
+201 Created
+{
+  "data": [
+    {
+      "type": "articles",
+      "id": "1",
+      "attributes": {"title": "Creating resources in bulk"}
+    },
+    {
+      "type": "articles",
+      "id": "2",
+      "attributes": {"title": "Handling responses in bulk creation"}
+    }
+  ],
+  "links": {
+    "profile": [
+      "https://github.com/json-api/json-api/_profiles/transifex/bulk/index.md"
+    ]
+  }
+}
+```
+
+#### `204 No Content`
+
+If the requested items did include a Client-Generated ID and the requested
+resources have been created successfully, the server **MUST** return either a
+`201 Created` status code and response document (as described above) or a
+`204 No Content` status code with no response document.
+
+
+> Note: If a 204 response is received the client should consider the resources
+> object sent in the request to be accepted by the server, as if the server had
+> returned them back in a 201 response.
+
+#### Failure
+
+In case of failure, a normal JSON:API error response **MUST** be returned as
+described in the specification. The errors of the response **SHOULD** only
+describe the requested items that caused a problem with the server; requested
+items that would have been accepted on their own **SHOULD NOT** be reflected in
+the response.
+
+In case of multiple errors, like with error responses as described in the
+JSON:API specification, the most generally applicable HTTP error code
+**SHOULD** be used in the response.
+
+In order for the client to identify which of the requested items caused the
+failure, the error object **SHOULD** include the `source` member with a
+`pointer`.
+
+For instance, a response when the server encountered a problem while creating
+the first and third item in a bulk request can be:
+
+```
+400 Bad Request
+{
+  "errors": [
+    {
+      "status": "403",
+      "title": "Forbidden",
+      "detail": "You do not have permissions to perform this action",
+      "source": {"pointer": "/data/0"}
+    },
+    {
+      "status": "409",
+      "title": "Conflict",
+      "detail": "An article with the same title already exists",
+      "source": {"pointer": "/data/2/attributes/title"}
+    }
+  ],
+  "links": {
+    "profile": [
+      "https://github.com/json-api/json-api/_profiles/transifex/bulk/index.md"
+    ]
+  }
+}
+```
+
+## Updating Resources In Bulk
+
+### Requests
+
+A collection of resources can be updated by sending a `PATCH` request to a URL
+that represents a collection of resources. The request **MUST** include an
+array of resource objects as primary data. All items of the array **MUST** have
+the same `type` member. An `id` member is also required to identify the
+resource being edited. The same restrictions that apply to a resource object
+when updating a single resource apply to each item of the array.
+
+Every bulk update operation **MUST** be atomic. If the server has changed, all
+the requested updates **MUST** have been applied. Also, no side-effects are to
+be expected to happen asynchronously on the server's side; assuming no further
+interaction with the server, the returned items **SHOULD** remain as they are
+and further attempts to fetch them **SHOULD** return the same resource objects.
+
+For instance, a collection of article updates might be performed with the
+following request:
+
+```
+PATCH /articles HTTP/1.1
+Content-Type: application/vnd.api+json;profile="https://github.com/json-api/json-api/_profiles/transifex/bulk/index.md"
+Accept: application/vnd.api+json;profile="https://github.com/json-api/json-api/_profiles/transifex/bulk/index.md",application/vnd.api+json
+
+{
+  "data": [
+    {
+      "type": "articles",
+      "id": "1",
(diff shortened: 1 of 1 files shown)


sebastienfilion:
Hey @kbairak thanks for creating this profile. I was wondering, how come there's no bulk fetch? Am I missing something?

ckeboss:
Any idea on addressing asynchronous bulk operations? There is a recommendation for singular resources (https://jsonapi.org/recommendations/#asynchronous-processing), but it breaks down at the point of when the job has completed successfully, returning `303 See Other`, as with bulk operations, there are `n` number of results.

Perhaps `300 Multiple Choices` with an array of data objects? Not sure if it fits semantically (https://tools.ietf.org/html/rfc7231#section-6.4.1)

Or perhaps even `link` headers with an `item` relation (https://tools.ietf.org/html/rfc6573)

kbairak:
@sebastienfilion 
> Hey @kbairak thanks for creating this profile. I was wondering, how come there's no bulk fetch? Am I missing something?

FETCH already works on collections. Even if none of the generic filters are good for your needs, the server could implement a `filter[id]`. Examples:

```
GET /articles?filter[id]=1&filter[id]=2&...`
```
or
```
GET /articles?filter[id][any]=1,2,...
```

kbairak:
@ckeboss 
> Any idea on addressing asynchronous bulk operations?

My initial approach was to support them. However, the specification would have to be complicated to support them. My wish was to copy as much of the original JSONAPI specification as possible, just operate on arrays of objects instead of objects. There was this trade-off between making the profile more capable or simpler. I went for the latter.

I suppose you can support asynchronous tasks on an application level, ie create an API resource that represents the bulk operation itself. Example (writing the payloads in yaml for simplicity):

```yaml
# POST /article_uploads
data:
  - type: article_uploads
    attributes:
      title: title of article 1
  - type: article_uploads
    attributes:
      title: title of article 2

# Response: 201 Created
data:
  - type: article_uploads
    id: 1
    attributes:
      status: pending
  - type: article_uploads
    id: 2
    attributes:
      status: done
    relationships:
      article:
        data:
          type: articles
          id: 2
        links:
          related: /articles/2
included:
  - type: articles
    id: 2
    attributes:
      title: title of article 2
links: ...
```

In this case, the `article_uploads` API resources are created atomically and can be supported by this profile, even though the articles themselves are created asynchronously.

ckeboss:
I understand why you excluded them from the profile.

I think bulk operations is one of the things that really lacks in the JSON:API spec. I believe that a robust spec for bulk operations is really needed to avoid "bikesheding" in regards to bulk operations. 

In regards to the workaround you provided, it does work, but the disadvantage is that the client must know extra information.

For example, say you have a `Article` resource with a property of `isPublished`. When creating or updating that resource, changing any other property other than `isPublished`, the resource would respond with 200 or 204, but if `isPublished` is true, it must do some other, possibly longer runing modification, such as syndicate the resource to other resources, etc. So in that case, it would need to respond with 202. 

Would you be interested in me proposing this further?

auvipy:
> I understand why you excluded them from the profile.
> 
> I think bulk operations is one of the things that really lacks in the JSON:API spec. I believe that a robust spec for bulk operations is really needed to avoid "bikesheding" in regards to bulk operations.
> 
> In regards to the workaround you provided, it does work, but the disadvantage is that the client must know extra information.
> 
> For example, say you have a `Article` resource with a property of `isPublished`. When creating or updating that resource, changing any other property other than `isPublished`, the resource would respond with 200 or 204, but if `isPublished` is true, it must do some other, possibly longer runing modification, such as syndicate the resource to other resources, etc. So in that case, it would need to respond with 202.
> 
> Would you be interested in me proposing this further?

how about extend it as an extension "bulk-operations"/"bulk profile"? would love to know more about your idea whatever you feel would be better for handling bulk ops

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+
+categories:
+  - Bulk Operations
+---
+
+# Bulk profile
+
+## Introduction
+
+This page specifies a profile for the `application/vnd.api+json` media type,
```
jelhan, line 26: Profiles are not allowed to enforce additional processing rules. This must be an extension.

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+profile limits operations of the same type (aka HTTP verb: `POST`, `PATCH`,
+`DELETE`) on the same API resource.
+
+## Creating Resources In Bulk
+
+### Requests
+
+A collection of resources can be created by sending a `POST` request to a URL
+that represents a collection of resources. The request **MUST** include an
+array of resource objects as primary data. All items of the array **MUST** have
```
jelhan, line 40: Neither extensions nor profiles are allowed to alter processing rules of the base specification. JSON:API specification enforces that `data` member must be a single resource object for `POST` request to an endpoint, which represents a collection of resources. The same applies to a `PATCH` request to a `self` link of a resource.

As far as I understand the specification, this extension must define its own member for the JSON:API document. It can not alter the meaning of `data`.  

Following the naming rules for members defined by extension, the member may be named `bulk:data`.

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+the same `type` member. The same restrictions that apply to a resource object
+when creating a single resource apply to each item of the array. Either all the
+items in the request **MUST** include client-generated IDs, or none **MUST**.
+
+Every bulk create operation **MUST** be atomic. If the server has changed, all
```
jelhan, line 45: The wording may be confusing. It's not the server, which has changed. One or more resources represented by that server have changed.

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+expected to happen asynchronously on the server's side; assuming no further
+interaction with the server, the returned items **SHOULD** remain as they are
+and further attempts to fetch them **SHOULD** return the same resource objects.
+
+For instance, a collection of articles might be created with the following
+request:
+
+```
+POST /articles HTTP/1.1
+Content-Type: application/vnd.api+json;profile="https://github.com/json-api/json-api/_profiles/transifex/bulk/index.md"
```
jelhan, line 56: See above.

```suggestion
Content-Type: application/vnd.api+json;ext="https://github.com/json-api/json-api/_profiles/transifex/bulk/index.md"
```

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+
+#### `201 Created`
+
+If the requested items did not include Client-Generated IDs and the requested
+resources have been created successfully, the server **MUST** return a
+`201 Created` status code.
+
+The server **MUST NOT** include a `Location` header.
+
+The response **MUST** also include a document that contains an array with the
```
jelhan, line 84: Just for clarity.

```suggestion
The response **MUST** also include a JSON:API document that contains an array with the
```

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+
+#### `204 No Content`
+
+If the requested items did include a Client-Generated ID and the requested
+resources have been created successfully, the server **MUST** return either a
+`201 Created` status code and response document (as described above) or a
+`204 No Content` status code with no response document.
+
+
+> Note: If a 204 response is received the client should consider the resources
```
jelhan, line 120: Why is this not a _must_?

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+> object sent in the request to be accepted by the server, as if the server had
+> returned them back in a 201 response.
+
+#### Failure
+
+In case of failure, a normal JSON:API error response **MUST** be returned as
+described in the specification. The errors of the response **SHOULD** only
+describe the requested items that caused a problem with the server; requested
```
jelhan, line 128: This is confusing. How should `error` represent requested items, which have not caused a problem?

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+items that would have been accepted on their own **SHOULD NOT** be reflected in
+the response.
+
+In case of multiple errors, like with error responses as described in the
+JSON:API specification, the most generally applicable HTTP error code
+**SHOULD** be used in the response.
```
jelhan, line 134: I assume this refers to the HTTP status code used for the response?

```suggestion
**SHOULD** be used as HTTP status code for the response.
```

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+> object sent in the request to be accepted by the server, as if the server had
+> returned them back in a 201 response.
+
+#### Failure
+
+In case of failure, a normal JSON:API error response **MUST** be returned as
+described in the specification. The errors of the response **SHOULD** only
+describe the requested items that caused a problem with the server; requested
+items that would have been accepted on their own **SHOULD NOT** be reflected in
+the response.
```
jelhan, line 130: We may want to specify explicitly how a server should respond if it does _not_ support creating multiple resources for this endpoint.

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+
+## Updating Resources In Bulk
+
+### Requests
+
+A collection of resources can be updated by sending a `PATCH` request to a URL
+that represents a collection of resources. The request **MUST** include an
+array of resource objects as primary data. All items of the array **MUST** have
+the same `type` member. An `id` member is also required to identify the
+resource being edited. The same restrictions that apply to a resource object
```
jelhan, line 176: Is it allowed to have the same resource in the request? I would tend towards forbidding that. It shouldn't add much more complexity to the client if it needs to merge multiple updates to the same resource in one resource object before sending the request.

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+#### `201 Created`
+
+If the requested items did not include Client-Generated IDs and the requested
+resources have been created successfully, the server **MUST** return a
+`201 Created` status code.
+
+The server **MUST NOT** include a `Location` header.
+
+The response **MUST** also include a document that contains an array with the
+primary resources created, in the same order as they appear in the request.
```
jelhan, line 85: Additionally we should specify the order in which the server must process the resources. The result may depend on the order. E.g. if two resources in the request both define a relationship to the same resource. If that relationship is a to-one, only one of them could be associated with that resource. In this case it depends on execution order, which resources are associated after processing the request.

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+}
+```
+
+### Responses
+
+#### `200 OK`
+
+If a server accepts an update but also changes at least one of the resources in
+ways other than those specified by the request (for example, updating the
+`updated-at` attribute or a computed `sha`), it **MUST** return a `200 OK`
```
jelhan, line 215: The attribute name should be updated to follow latest recommendation to use `camelCase`.

```suggestion
`updatedAt` attribute or a computed `sha`), it **MUST** return a `200 OK`
```

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+
+In case of multiple errors, like with error responses as described in the
+JSON:API specification, the most generally applicable HTTP error code
+**SHOULD** be used in the response.
+
+In order for the client to identify which of the requested items caused the
+failure, the error object **SHOULD** include the `source` member with a
+`pointer`.
+
+For instance, a response when the server encountered a problem while creating
```
jelhan, line 366:
```suggestion
For instance, a response when the server encountered a problem while deleting
```

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+      "attributes": {"title": "Creating resources in bulk"}
+    },
+    {
+      "type": "articles",
+      "id": "2",
+      "attributes": {"title": "Handling responses in bulk creation"}
+    }
+  ],
+  "links": {
+    "profile": [
```
jelhan, line 105: Per latest v1.1 release candidate, applied profiles and extensions should be listed in `jsonapi` object. Not in `links`. See https://jsonapi.org/format/1.1/#document-jsonapi-object

_profiles/transifex/bulk/index.md
```diff
@@ -0,0 +1,396 @@
...
+`DELETE`) on the same API resource.
+
+## Creating Resources In Bulk
+
+### Requests
+
+A collection of resources can be created by sending a `POST` request to a URL
+that represents a collection of resources. The request **MUST** include an
+array of resource objects as primary data. All items of the array **MUST** have
+the same `type` member. The same restrictions that apply to a resource object
```
jelhan, line 41: What is the reason to enforce the same `type` at extension level rather than allowing implementations to decide, which resource types they support for bulk creation? I see that it will be only one resource type in most cases. But in some cases - e.g. polymorphic collections - this restriction would make it impossible to use the extension.

jelhan (commented):
Thanks a lot for putting this together. I think it's a great start for a bulk extension. Are you still open to drive the creation of such an extension?

kbairak:
Hey @jelhan . Thank you so much for your feedback. It's been a while since I worked on this but I will try to find some time to get caught up in how {json:api} deals with extensions/profiles and see if I can push this forwards.

jelhan:
I published a JSON:API bulk create extension here: https://github.com/jelhan/json-api-bulk-create-extension I consider it an early draft for now. Looking for feedback. Please use GitHub issues at that repository for feedback, questions and proposing changes.

I'm going to close this PR as I feel it is superseded by that extension. If someone wants to propose an alternative solution, please feel free to do so. I assume that we want for any extension and profile some experience from real-world usage before promoting it to an official extension or profile.

PR #1383, opened 2019-04-05
Library rename + server

f3ath:
I have built a full-fledged client on top of my old library. Pointing this document to the client library now. Also added a link to a server implementation in Dart.


dgeb:
Thanks!

Issue #1382, opened 2019-04-04
Unclear `Accept` header client documentation


TrevorHinesley:
I might be nitpicking here, but the wording seems intentional so I think it's worth a discussion for clarification. In [this section](https://jsonapi.org/format/#content-negotiation-clients), the language in the second sentence is as follows:

```
Clients that include the JSON:API media type in their Accept header MUST specify the media
type there at least once without any media type parameters.
```

The `at least once` piece is a bit confusing to me.

Does this mean:

1. The client is allowed to specify the JSON:API media type more than once in their Accept header, but *at least* one instance of the media type must be without media type parameters (meaning one or more other instances *can* have media type parameters)?

OR

2. The client is allowed to specify the JSON:API media type more than once in their Accept header, but *all* instances of the media type must be without media type parameters?

gabesullice:
@TrevorHinesley, I'm fairly certain that it's #1. Would you mind submitting a PR to that effect?

My reasoning is that we need to be able to support profile negotiation in version 1.1, which require the use of media type parameters.

TrevorHinesley:
@gabesullice done! Let me know if the wording is sufficient or still lacks clarity.

Dormilich:
Is the quality parameter `q` in an Accept header considered a disallowed parameter?

E.g. If I specify different profiles/extensions (version 1.1), then none of that would be used as the preferred MIME type (with the maximum importance) has no parameters.

freddrake:
The `q` parameter in an Accept header should not be considered part of the media type, though it is encoded as if it were.

freddrake:
@gabesullice: I believe all the questions posed here have been answered, but specification 1.1 has moved on and includes a detailed discussion of the interpretation of media type parameters in the **Accept** header.

This issue seems outdated, and should be closed.

gabesullice:
Thanks for the ping @freddrake! Agreed that this issue is out of date.

PR #1390, opened 2019-04-24
Clarified Client Responsibility (#1382)

TrevorHinesley:
Clarified client responsibility regarding the `Accept` header per https://github.com/json-api/json-api/issues/1382

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -49,8 +49,7 @@ The `profile` parameter is used to support [profiles].
 
 ### <a href="#content-negotiation-clients" id="content-negotiation-clients" class="headerlink"></a> Client Responsibilities
 
-Clients that include the JSON:API media type in their `Accept` header **MUST**
-specify the media type there at least once without any media type parameters.
+Clients that include the JSON:API media type in their `Accept` header **MAY** include it more than once, but for at least one instance of it, they **MUST** specify it without any media type parameters.
 
 When processing a JSON:API response document, clients **MUST** ignore any
 parameters other than `profile` in the server's `Content-Type` header.


_format/1.1/index.md
```diff
@@ -49,8 +49,7 @@ The `profile` parameter is used to support [profiles].
 
 ### <a href="#content-negotiation-clients" id="content-negotiation-clients" class="headerlink"></a> Client Responsibilities
 
-Clients that include the JSON:API media type in their `Accept` header **MUST**
-specify the media type there at least once without any media type parameters.
+Clients that include the JSON:API media type in their `Accept` header **MAY** include it more than once, but for at least one instance of it, they **MUST** specify it without any media type parameters.
```
gabesullice, line 52: @TrevorHinesley, thanks for doing this!

1. I think we need to add new lines after 80 characters :)
2. Do you think this is any clearer? I worked myself into a knot and trying different wordings and maybe this suggestion below is actually worse:

```
Clients that include the JSON:API media type in their `Accept` header **MAY**
include it more than once, however, at least one instance of the JSON:API media
type **MUST NOT** specify any media type parameters.
```

gabesullice (changes requested):

_format/1.1/index.md
ahx: Just a drive-by opinion: I think @TrevorHinesley's version reads a bit better.

TrevorHinesley: How about a slight variation of my original:

```
Clients that include the JSON:API media type in their `Accept` header **MAY** 
include it more than once, but at least one of these instances **MUST** be 
without any media type parameters.
```

gabesullice:
This rule has been removed and so this is no longer applicable. Thank you @TrevorHinesley and @ahx ! We sincerely appreciate your suggestions!

PR #1392, opened 2019-05-08
Add jsonapi_parameters gem to Ruby implementations list

Marahin:
[Visuality](https://github.com/visualitypl) has published a [gem](https://github.com/visualitypl/jsonapi_parameters) that allows JSON:API compliant input to be easily handled in a Rails-expected parameters format. It works both for vanilla Ruby and RoR.

It would be nice to list it along other Ruby implementations.


dgeb:
Thanks!

PR #1393, opened 2019-05-13
Add DatX lib to TypeScript implementations list

isBatak:
It would be great to list DatX under TypeScript implementations list.


dgeb:
Thanks!

PR #1396, opened 2019-05-24
Added Hyral to the Javascript client list

mnugter:
Please consider adding Hyral as a JavaScript client library.

Please let me know if there is anything I should do or change.

Thanks in advance!


dgeb:
Thank you!

Issue #1313, opened 2018-10-26
Old implementations


bbrala:
I've clicked around in the implementations in the documentation and it seems there is quite a lot of project that haven't been updated in ages. For the PHP side i see quite a few that one could argue are abandoned. 

Would you mind if i went through and a least do some sorting so active projects are on top? This way when users click through the list of implementations they see relatively active projects first. I'm not sure if assuming a project dead is really feasable since that sometimes is rather hard to know for sure.

ethanresnick:
I'd like to do some sorting too, as the list is much less useful than it could be because it doesn't highlight the most-developed implementations. The trick is finding some general criteria to apply to sort things in a way that people feel is fair. 

I can think of one thing off the top of my head that should be uncontroversial: if an implementation is still based on [RC3](http://jsonapi-rc3.herokuapp.com/) of the spec, it should definitely be removed. That criterion is easy to check -- does the implementation have separate `attributes` and `relationships` keys in its resource objects? --  and would mean that the implementation hasn't been updated in years.

Maybe you can post a list here of all the implementations that are still based on RC3 and we'll start by killing those? Then, if there's still too much cruft in the implementations list, we can explore other means of sorting what's left?

cc @dgeb 
ref #733 #1151

wimleers:
Oh interesting, first time I hear about RC3 and `attributes` <> `relationships` not yet being a thing there!

Sounds like a great criterion for at least a first round of clean-up!

ethanresnick:
Yup. @wimleers, if you have some time to help, maybe one of you could take the PHP server implementations and the other the client implementations and we'll start there?

wimleers:
I'll keep this issue in mind/this browser tab open for when I do have some down time :)

bbrala:
Yeah, same, please mention here should you start and if you are starting on the client or server list.

Personally i think we should (at least for php) also look at how well supported it is, i mean, when i see Laravel 5.2 support as the max, that is like a few years old, and no longer maintained, which would mean the package is kind of worthless right now.

bbrala:
The PHP clients seem fine actually, i just made a small PR to order them by name.

wimleers:
I've been looking at many implementations today, and _many_ of them are pretty simplistic and incomplete. Implementing the absolute basics of the JSON:API spec is easy, but also including support for collection filtering, sorting, pagination for example is usually lacking.

I'm not sure how we can surface this in easy-to-scan format though. Thoughts?

[...]

PR #1398, opened 2019-05-29
remove unmaintained node.js server library

ghost:
Remove https://github.com/SeyZ/jsonapi-serializer implementation because
it is not actively maintained.
Following the discussion in https://github.com/json-api/json-api/issues/733 and https://github.com/json-api/json-api/issues/1313, I think it would be incredibly helpful to have some way for libraries to prove their adherence to a given version of the specification.


ckeboss:
Could we replace it with https://github.com/danivek/json-api-serializer

dgeb:
I'm reluctant to remove this library because there's been commit activity within the past two months: https://github.com/SeyZ/jsonapi-serializer/commits/master

Given this, I'd want confirmation from @SeyZ before doing anything.

> Could we replace it with https://github.com/danivek/json-api-serializer

@ckeboss That seems fairly popular and doesn't appear to be a fork of SeyZ/jsonapi-serializer. Perhaps danivek/json-api-serializer should have an entry of its own?

ghost:
I'd second making https://github.com/danivek/json-api-serializer a new entry. Given both have similar goals (framework agnostic), perhaps we could annotate that in their descriptions?

ckeboss:
@dgeb Sure. Perhaps a little emphasis that the names are very similar.

PR #1404, opened 2019-07-11
Move package to correct section

bednic:
Fix my mistake, this is server-side library, not client-side.


bednic:
Thanks 😉

dgeb:
@bednic whoops, looks like this PR didn't remove the lib from the client section. Could you please handle that in another PR?

bednic:
https://github.com/json-api/json-api/pull/1405

PR #1405, opened 2019-07-11
remove package from wrong section

bednic:


dgeb:
Great - thanks again!

PR #1408, opened 2019-07-16
Fix misleading wording about quotation

vfaronov:
RFC 7231 Section 5.3.2 requires quoting even a single URI
in a media range parameter, because it contains characters that 
are not allowed in an unquoted token (RFC 7230 Section 3.2.6).


gabesullice:
Thanks @vfaronov, good catch! Would you mind updating this PR to the tip of gh-pages?

vfaronov:
@gabesullice Done.

gabesullice:
Thank you!

PR #1409, opened 2019-07-16
Fix example for the 'profile' media type parameter

vfaronov:
Per RFC 7231 Section 5.3.2, Accept is not an ordered list. Relative preferences are expressed with explicit quality values.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1922,10 +1922,11 @@ request is received, a server **SHOULD** attempt to apply the requested profiles
 to its response.
 
 For example, in the following request, the client asks that the server apply the
-`http://example.com/last-modified` profile if it is able to.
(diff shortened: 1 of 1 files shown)


gabesullice:
Thanks for PR @vfaronov, we've just merged a number of refinements to extensions and profiles. We ended up removing lots of profile related language and some of the examples, including this one. Would you mind re-reading the spec and making sure we've gotten our content negotiation rules examples correctly (since you've shown interest/expertise in these things)?

vfaronov:
@gabesullice Upon a quick skim, I couldn’t find examples for multi-entry `Accept` in the current version, just a note that this is possible, which seems OK, but I submitted #1504 to improve it a bit.

gabesullice:
Thank you! 🙏

Issue #1403, opened 2019-07-10
Feedback requested re: v1.1


dgeb:
On behalf of the spec editors, I'd like to solicit general feedback on the [v1.1 release candidate](https://jsonapi.org/format/1.1/) to understand how things are going in the community. 

We are currently in an awkward state, unable to finalize v1.1 due to the caveat described in the [status](https://jsonapi.org/format/1.1/#status) section:

> This version is expected to be finalized and released on January 31, 2019 (provided there are two compliant implementations by that date; if not the release will wait until such implementations exist to prove its viability).

It's inarguably a bad sign that we're now more than five months past this date without yet being aware of multiple compliant server and client implementations of the v1.1 RC. Now it's possible that we're simply out of the loop, and v1.1 has been fully implemented by several libraries that we (the editors) don't happen to be using. However, it's also possible that the new-to-1.1 [profiles](https://jsonapi.org/format/1.1/#profiles) feature has proven too onerous, impractical, or undesirable to be implemented widely (and this is of course the precise reason that we added that caveat in the first place).

With that in mind, we're reaching out with a few questions to JSON:API implementors:

* Have you attempted to implement any part of v1.1? If so, what part?
* Which parts of v1.1 do you find useful?
* Which parts of v1.1 do you not find useful?
* Have any parts of v1.1 proven to be too onerous or impractical to implement?

For reference, the major deltas introduced in 1.1 are listed in the [update history](https://jsonapi.org/#update-history) section of the home page and copied below:

  - New features include: [profiles](https://jsonapi.org/format/1.1/#profiles), ["@-Members"](https://jsonapi.org/format/1.1/#document-member-names-at-members), and [error object `type` links](https://jsonapi.org/format/1.1/#error-objects). 
  - There are editorial clarifications around: [query parameters](https://jsonapi.org/format/1.1/#query-parameters) (their parsing/serialization, and reserved parameter names), the appropriate status code for [specific](https://github.com/json-api/json-api/pull/1036) [responses](https://github.com/json-api/json-api/pull/1029), and the [presence of the `included` key](https://github.com/json-api/json-api/pull/1236) even when its value is an empty array.
  - JSON:API switched to [recommending camelCased names](https://jsonapi.org/recommendations/#naming).

Please feel free to speak openly in comments below. We are trying to revitalize the spec and push forward on some much needed initiatives. However, we're also willing to re-examine and change some aspects of v1.1 first, because it has definitely become a blocker.

Thanks for any feedback you can provide.

[...]

Issue #1362, opened 2019-01-06
Profiles: Consider allowing servers to opt-out of recognizing aliases


dgeb:
In order to allow incremental adoption of profiles in v1.1, I'd like to consider allowing servers to signal that they do not support certain features, while still specifying that they SHOULD support them. For instance, it seems reasonable to me that a server could respond with a 422 error if a document is sent that includes aliases for keywords in a profile. We could create an official URL for this error so that it would be easily distinguished from other 422 errors.

I wanted to raise this as an issue for discussion prior to opening a PR.

wimleers:
I don't think I fully understand what you're referring to.

Are you referring to the case  where the server supports the profile that the aliased keyword is for, but the server does not support resolving aliases?

dgeb:
> Are you referring to the case where the server supports the profile that the aliased keyword is for, but the server does not support resolving aliases?

Yes. 

My thinking is that the primary use case for aliases is to avoid conflicts between semantically conflicting, yet functionally compatible, profiles. More often than not, servers will support only non-conflicting profiles that do not require aliases. For these servers, there's little benefit to supporting some random and unnecessary alias sent by the client.

While it may be ideal if servers can always support aliases and be maximally forgiving, I think it's unlikely to be the reality. It seems realistic to give servers a way to avoid extra processing and clearly communicate to the client why they don't allow it.

gabesullice:
+1, aliases are the most demanding requirement of v1.1 (re: implementation complexity)

gabesullice:
Superseded by #1412

PR #1410, opened 2019-07-25
Simplify profiles

gabesullice:
JSON:API is already burdened by a high number of concepts. JSON:API 1.1 introduces an entirely new concept, profiles, which itself has a number of new concepts.

This is an attempt to simplify the profiles with the intention of lowering the burden of their implementation.

<hr>

I think the two biggest sources of complexity introduced with profiles are:

1. Aliasing
2. Separate and competing negotiation mechanisms (`Accept`-based and query parameter based)

#### Aliasing
The first, aliasing, anticipates a problem that I'm not yet convinced we'll encounter. That is, it attempts to make it possible for a server to support profiles with conflicting keywords. 

Unfortunately, aliasing is not a trivial thing to implement on the client or the server. A global dictionary must be computed for every HTTP message by either the client or the server. There is no guarantee that two implementations will use the same aliases, nor that they will even use the same aliases between requests. In systems that have hierarchical (de)serialization components, that global dictionary must be shared between them and that can be a cumbersome process. Even after deserialization, every processor must also know about the aliases too, since processors will be looking for specific keywords.

I propose we remove aliasing from v1.1 of the spec. If they really do become necessary, it shouldn't be too difficult to add them to a subsequent version of the spec.

#### Competing negotiation mechanisms
The second, competing negotiation mechanisms, is hard to interpret and creates complex interactions when used in conjunction with one another.

I see the strict correctness of using the `Accept` parameter for profiled documents and the `profile` query parameter for profiles which establish new query parameters and/or further define the `sort`, `filter` and `page` query parameters. However, it doesn't seem worth the complexity it adds. However, the spec does not limit which kinds of profiles are allowed in one of the negotiation mechanisms vs. the other (f.e. only profiles that register new query parameters are permitted in the `profiles` query parameter).

The biggest distinction between the two mechanism is that the `profile` query parameter permits a client to _require_ a profile while the `Accept` header does not. That's because, the `Accept` header based profiling mechanism requires that the client sends at least two media type values if it desires a profiled response document. One of these *must* be the bare JSON:API media type. Consequently, the spec forces the client to be to accept and process both responses. That adds complexity to the client implementation that I doubt many clients will adopt.

It seems far more likely that most clients will be written to accept a specific set of profiles which it knows a specific server will support. Even generalized clients that can may be able to support many different profiles, probably won't be written to dynamically negotiate profiles, rather, they'll probably make the profiles to request a configuration option or some kind of middleware.

Furthermore, the fact that the `profiles` query parameter can be omitted and implied by other query parameters suggests that the we already recognize that most clients are written with a specific server implementation in mind.

Given all that, I propose that we simplify the profile negotiation mechanism by completely removing the `profile` query parameter and then strengthening the `Accept` header mechanism by removing the requirement that the bare JSON:API media type be present. I considered removing the `Accept` mechanism instead, but felt that since `Accept` is already the standard for negotiating over HTTP, we ought to use it.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -559,22 +559,15 @@ related resource collection:
 #### <a href="#profile-links" id="profile-links" class="headerlink"></a> Profile Links
 
 Like all [links][link], a link in an array of `profile` links can be represented
-with a [link object]. In that case, the link object **MAY** contain an `aliases` 
-member listing any [profile aliases].
+with a [link object].
 
-Here, the `profile` key specifies an array of `profile` links, including one
-that includes a [profile alias][profile aliases]:
+Here, the `profile` key specifies an array of `profile` links:
 
 ```json
 "links": {
   "profile": [
     "http://example.com/profiles/flexible-pagination",
-    {
-      "href": "http://example.com/profiles/resource-versioning",
-      "aliases": {
-        "version": "v"
-      }
-    }
+    "http://example.com/profiles/resource-versioning"
   ]
 }
 ```
@@ -2025,7 +2018,7 @@ https://example.com/?page[cursor]=xyz&profile=https://example.com/pagination-pro
 ```
 
 
-### <a href="#profile-keywords-and-aliases" id="profile-keywords-and-aliases" class="headerlink"></a> Profile Keywords and Aliases
+### <a href="#profile-keywords" id="profile-keywords" class="headerlink"></a> Profile Keywords
 
 A profile **SHOULD** explicitly declare "keywords" for any elements that it
 introduces to the document structure. If a profile does not explicitly declare a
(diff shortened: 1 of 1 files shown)


dgeb:
We have seen very lackluster support for adoption of profiles as currently spec'd (see the limited responses to #1403), so I think it makes a lot of sense to simplify them before finalizing 1.1.

I opened an issue to consider making aliases optional (#1362) at the beginning of the year, but I am on board with removing them completely in 1.1.

> If they really do become necessary, it shouldn't be too difficult to a subsequent version of the spec.

I agree.

dgeb:
@gabesullice In reviewing all the profile-related portions of the spec, I am finding considerable implications to removing aliases. For instance, we need to reconsider the [Designing profiles to evolve over time](https://jsonapi.org/format/1.1/#profiles-design-for-evolution) section. And we may want to revisit the [processing](https://jsonapi.org/format/1.1/#profiles-processing) section as well.

I'm not sure if we want to continue to add commits to a single "simplification" PR (this one) or open a separate PR for each individual change (i.e. separate aliases from negotiation).

gabesullice:
Closing in favor of #1412 and #1413.

PR #1413, opened 2019-08-09
Unify profile negotiation via the Accept and Content-Type headers in 1.1-rc2

gabesullice:
This supersedes #1410 as it has been split into two distinct PRs—this one, and #1412 

### Motivation
The current 1.1 release candidate contains two, competing profile negotiation mechanisms. One uses the `Accept` header and the other uses a `profile` query parameter. This is hard to interpret and create confusing interactions when one is used in conjunction with the other.

I see the strict correctness of using the `Accept` parameter for profiled documents and the `profile` query parameter for profiles which establish new query parameters and/or further define the `sort`, `filter` and `page` query parameters. However, it doesn't seem worth the complexity it adds. However, the spec does not limit which kinds of profiles are allowed in one of the negotiation mechanisms vs. the other (f.e. only profiles that register new query parameters are permitted in the `profiles` query parameter).

The biggest distinction between the two mechanism is that the `profile` query parameter permits a client to _require_ a profile while the `Accept` header does not. That's because, the `Accept` header based profiling mechanism requires that the client sends at least two media type values if it desires a profiled response document. One of these *must* be the bare JSON:API media type. Consequently, the spec forces the client to be to accept and process both responses. That adds complexity to the client implementation that I doubt many clients will adopt.

It seems far more likely that most clients will be written to accept a specific set of profiles which it knows a specific server will support. Even generalized clients that can may be able to support many different profiles, probably won't be written to dynamically negotiate profiles, rather, they'll probably make the profiles to request a configuration option or some kind of middleware.

Furthermore, the fact that the `profiles` query parameter can be omitted and implied by other query parameters suggests that the we already recognize that most clients are written with a specific server implementation in mind.

Given all that, I propose that we simplify the profile negotiation mechanism by completely removing the `profile` query parameter and then strengthening the `Accept` header mechanism by removing the requirement that the bare JSON:API media type be present. I considered removing the `Accept` mechanism instead, but felt that since `Accept` is already the standard for negotiating over HTTP, we ought to use it.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -67,6 +67,10 @@ its `source` and which has the following URI as (one of) its `type`s:
 https://jsonapi.org/errors/profile-not-supported
 ```
 
+It is **RECOMMENDED** that servers specify the `Vary` header with the `Accept`
+header as one of its header names to ensure that its responses are
+appropriately cached.
+
 > Note: These content negotiation requirements exist to allow future versions
 of this specification to add other media type parameters for extension
 negotiation and versioning.
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -52,16 +52,6 @@ negotiation and versioning.
 
 ### <a href="#content-negotiation-clients" id="content-negotiation-clients" class="headerlink"></a> Client Responsibilities
 
-Clients that include the `profile` parameter at least once in their `Accept`
-header **MUST** specify at it at least once without the `profile` media type
-parameter.
-
-For example:
-
-```http
-Accept: application/vnd.api+json;profile="http://example.com/last-modified", application/vnd.api+json
-```
-
 When processing a JSON:API response document, clients **MUST** ignore any
 parameters other than `profile` in the server's `Content-Type` header.
 
@@ -1911,7 +1901,8 @@ This profile defines the following keywords:
 
 The `profile` media type parameter is used to describe the application of
 one or more profiles to a JSON:API document. The value of the `profile`
-parameter **MUST** equal a space-separated (U+0020 SPACE, " ") list of profile URIs.
+parameter **MUST** equal a space-separated (U+0020 SPACE, " ") list of profile
+URIs.
 
 > Note: When serializing the `profile` media type parameter, the HTTP
 > specification requires that its value be surrounded by quotation marks
@@ -1924,42 +1915,25 @@ requested profiles to its response.
 
 For example, in the following request, the client asks that the server apply
 the `http://example.com/last-modified` profile and the
-`http://example.com/iso-8601` profile if it is able to.
+`http://example.com/timestamps` profile, if it is able to.
 
 ```http
 GET /articles/1 HTTP/1.1
 Accept: application/vnd.api+json; profile="http://example.com/last-modified http://example.com/iso-8601"
 ...
 
-If the server is unable to apply any of the requested profiles, it may send a
-response without any of the profiles applied:
+Servers **MAY** respond with a subset of the requested profiles applied or none
+of the requested profiles applied. Additionally, servers **MAY** respond with
+unrequested profiles applied.
 
-```
-HTTP/1.1 200 OK
-Content-Type: application/vnd.api+json
-```
-
-Additionally, servers **MAY** respond with only a subset of the requested
-profiles applied.
-
-For example, if the server supports the `http://example.com/iso-8601` profile,
-but not the `http://example.com/last-modified` profile, it may send a
-response with only the supported subset of profiles applied:
-
-```
-HTTP/1.1 200 OK
-Content-Type: application/vnd.api+json; profile="http://example.com/iso-8601"
-```
-
-Servers **MAY** respond using one or more profiles even if the client has
-not requested them. The recipient of a document to which an unknown profile
-has been applied **MUST** ignore any document members that it does not understand.
+The recipient of a document to which an unknown profile has been applied
+**MUST** ignore any document members that it does not understand.
 
 #### <a href="#profiles-sending" id="profiles-sending" class="headerlink"></a> Sending Profiled Documents
 
 Clients and servers **MUST** include the `profile` media type parameter in
-conjunction with the JSON:API media type in a `Content-Type` header to indicate
-that they have applied one or more profiles to a JSON:API document.
+conjunction with the JSON:API media type in a `Content-Type` header when they
+have applied one or more profiles to a JSON:API document.
 
 Likewise, clients and servers applying profiles to a JSON:API document **MUST**
 include a [top-level][top level] [`links` object][links] with a `profile` key,
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -2260,8 +2260,6 @@ An error object **MAY** have the following members:
     exists; if it doesn't, the client **SHOULD** simply ignore the pointer.
   * `parameter`: a string indicating which URI query parameter caused
     the error.
-  * `profiles`: an array of the requested profile URIs that the server was not
-    able to support.
 * `meta`: a [meta object][meta] containing non-standard meta-information about the
   error.
 
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -61,9 +61,9 @@ Servers **MUST** respond with a `415 Unsupported Media Type` status code if
 a request specifies the header `Content-Type: application/vnd.api+json` with
 any media type parameters other than `profile`.
 
-Servers **MUST** respond with a `406 Not Acceptable` status code if a request
-specifies an instance of the JSON:API media type in an `Accept` header with any
-media type parameter other than `profile`.
+Servers **MUST** respond with a `406 Not Acceptable` status code if a request's
+`Accept` header contains the JSON:API media type and all instances of that
+media type are modified with a media type parameter other than `profile`.
 
 ## <a href="#document-structure" id="document-structure" class="headerlink"></a> Document Structure
 
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -37,7 +37,7 @@ Clients and servers **MUST** send all JSON:API data using this media type in the
 
 Further, the JSON:API media type **MUST** always be specified with either no
 media type parameters or with only the `profile` parameter. This applies to both
-the `Content-Type` and `Accept` headers.
+the `Content-Type` and `Accept` headers whenever they are present.
 
 > Note: A media type parameter is an extra piece of information that can
 accompany a media type. For example, in the header
@@ -47,9 +47,6 @@ accompany a media type. For example, in the header
 The `profile` parameter is used to support [profiles].
 
 ### <a href="#content-negotiation-clients" id="content-negotiation-clients" class="headerlink"></a> Client Responsibilities
-
-Clients that include the JSON:API media type in their `Accept` header **MUST**
-specify the media type there at least once without any media type parameters.
 
 When processing a JSON:API response document, clients **MUST** ignore any
 parameters other than `profile` in the server's `Content-Type` header.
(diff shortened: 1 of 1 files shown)


_format/1.1/index.md
```diff
@@ -62,8 +59,22 @@ a request specifies the header `Content-Type: application/vnd.api+json`
...
+its `source` and which has the following URI as (one of) its `type`s:
+
+```
+https://jsonapi.org/errors/profile-not-supported
+```
+
+It is **RECOMMENDED** that servers specify the `Vary` header with `Accept`
+as one of its header names. This will ensure that the servers responses are
+appropriately cached and future client/server negotiations will work as
+expected.
```
gabesullice: Should this recommendation be elevated to a **SHOULD**?

_format/1.1/index.md
```diff
@@ -62,8 +59,22 @@ a request specifies the header `Content-Type: application/vnd.api+json`
...
+its `source` and which has the following URI as (one of) its `type`s:
+
+```
+https://jsonapi.org/errors/profile-not-supported
+```
+
+It is **RECOMMENDED** that servers specify the `Vary` header with `Accept`
+as one of its header names. This will ensure that the server's responses are
+appropriately cached and future client/server negotiations will work as
+expected.
```
gabesullice: Perhaps this recommendation ought to be elevated to a **SHOULD**?

Servers that do not support any profiles or only a specific set of profiles probably should not set the `Vary` header because it opens them up to the possibility of a malicious actor sending many variations of the `Accept` header to fill the server's cache.

OTOH, if negotiation is supported, not setting the `Vary` header will break profile negotiation whenever an HTTP cache is used.

dgeb:
@gabesullice As discussed, please add requirements along the lines of:

* The server MUST understand all profiles applied to a document via its `Content-Type` header or else it should respond with a 415.

* A client MUST include all the profiles required for processing a request via the `Accept` header.

_format/1.1/index.md
wimleers: IIRC many (reverse) proxies ignore `Accept: Vary` anyway precisely because of this risk, or even use it as a reason to not cache this response. I think `Vary: Accept` is fine as a communication of intent, just don't expect it to have the desired effect :) 
(If desired, I can do some digging on this.)
*EDIT*: I see that the next few lines already say that intermediaries may ignore this :)

This _does_ signal to (relatively advanced) users that if they add the necessary normalization logic in their reverse proxy, that they can in fact reliably cache it. If all JSON:API server implementations would do this, it'd enable their users to tune infrastructure for maximum performance/scalability.

Devil's advocate: why not change this to a **MUST**?

_format/1.1/index.md
```diff
@@ -1916,32 +1927,38 @@ parameter **MUST** equal a space-separated (U+0020 SPACE, " ") list of profile U
...
 Accept: application/vnd.api+json;profile="http://example.com/last-modified", application/vnd.api+json
 ```
 
-> Note: The second instance of the JSON:API media type in the example above is
-  required under the [client's content negotiation responsibilities](#content-negotiation-clients).
-  It is used to support old servers that don't understand the profile parameter.
+In the following request, the client asks the server to process its request
+using the `http://example.com/boolean-filters` profile. In this case, the client
+does not specify the bare JSON:API media type because it would not like a
+response unless the request was processed using the specified profile.
```
wimleers: Should this also say that if the server is unable to apply the requested profile, that it **MUST** respond with a 406?

_format/1.1/index.md
```diff
@@ -1916,32 +1927,38 @@ parameter **MUST** equal a space-separated (U+0020 SPACE, " ") list of profile U
...
-requested them. The recipient of a document **MUST** ignore any profiles in that
-document that it does not understand. The only exception to this is profiles
-whose support is required using the `profile` query parameter, as described later.
+```http
+Accept: application/vnd.api+json;profile="http://example.com/boolean-filters"
+```
+
+Servers **MAY** process messages using one or more profiles even if the client has
+not requested them. The recipient of a document to which an unknown profile
+has been applied **MUST** ignore any document members that it does not understand.
```
wimleers: Put differently, a server may apply a _superset_ of the requested profiles, but never a _subset_. Correct?

_format/1.1/index.md
```diff
@@ -1916,32 +1927,38 @@ parameter **MUST** equal a space-separated (U+0020 SPACE, " ") list of profile U
...
 
 #### <a href="#profiles-sending" id="profiles-sending" class="headerlink"></a> Sending Profiled Documents
 
 Clients and servers **MUST** include the `profile` media type parameter in
 conjunction with the JSON:API media type in a `Content-Type` header to indicate
-that they have applied one or more profiles to a JSON:API document.
+that they have applied one or more profiles to a JSON:API message.
```
wimleers: Why this change?

- https://jsonapi.org/format/1.0/ has 0 "message" mentions, 49 "document" mentions.
- https://jsonapi.org/format/1.1/ has 1 "message" mentions, 86 "document" mentions.

_format/1.1/index.md
```diff
@@ -2371,6 +2331,8 @@ An error object **MAY** have the following members:
     exists; if it doesn't, the client **SHOULD** simply ignore the pointer.
   * `parameter`: a string indicating which URI query parameter caused
     the error.
+  * `profiles`: an array of the requested profile URIs that the server was not
```
wimleers: First reaction: Clever addition!

Second reaction: I wonder whether the combination of `links.type = https://jsonapi.org/errors/profile-not-supported` (already required per one of the first change hunks) plus a `links.profiles` would not be adequate too? That would avoid adding an error object member.
Put differently: why an error object member and not an error object's `links` object member?

_format/1.1/index.md
```diff
@@ -1916,32 +1927,38 @@ parameter **MUST** equal a space-separated (U+0020 SPACE, " ") list of profile U
 > (U+0022 QUOTATION MARK, "\"") if it contains more than one URI.
 
 A client **MAY** use the `profile` media type parameter in conjunction with the
-JSON:API media type in an `Accept` header to _request_, but not _require_, that
-the server apply one or more profiles to the response document. When such a
-request is received, a server **SHOULD** attempt to apply the requested profiles
-to its response.
+JSON:API media type in an `Accept` header to request that the server apply one
+or more profiles to the response document. When such a request is received, a
+server **SHOULD** attempt to apply the requested profiles to its response.
```
wimleers: I understand why this is saying
> **SHOULD** attempt to apply …

But I think that
> **MUST** either apply or respond with a 406*

would be more explicit, more clear. I am probably overlooking the reason(s) why we cannot do that. But that would definitely make profiles a lot simpler to understand, which is the primary goal here.

Curious to hear your feedback 😀

wimleers (commented):
I like this already, it's definitely a big simplification. Left some questions and suggestions.

_format/1.1/index.md
gabesullice: @wimleers, I think you _were_ right about **MUST**. However, since you made this comment, the direction of profiles has been changed such that a **MUST** would probably be too heavy handed. IOW, it's valid for a client to ask for profile A, but get a cached response with profile B. I'm gonna change this to a **SHOULD**.

A subsequent PR will add extension negotiation, which is not optional, and in that case I think a **MUST** will be appropriate.

_format/1.1/index.md
gabesullice: Because I wanted the phrasing to apply to query parameters too, not just the document payload. However, `profiles` no longer apply to query parameters. Reverted.

_format/1.1/index.md
gabesullice: Not anymore ;)

_format/1.1/index.md
gabesullice: This has been removed.

_format/1.1/index.md
```diff
@@ -37,36 +37,29 @@ Clients and servers **MUST** send all JSON:API data using this media type in the
...
 
 ### <a href="#content-negotiation-servers" id="content-negotiation-servers" class="headerlink"></a> Server Responsibilities
 
 Servers **MUST** respond with a `415 Unsupported Media Type` status code if
-a request specifies the header `Content-Type: application/vnd.api+json`
-with any media type parameters other than `profile`.
-
-Servers **MUST** respond with a `406 Not Acceptable` status code if a
-request's `Accept` header contains the JSON:API media type and all instances
-of that media type are modified with media type parameters.
```
dgeb, line 65: I think the 406 case is still valid as applies to `Accept`, and just needs to be modified to account for the `profile` param.

_format/1.1/index.md
```diff
@@ -1847,8 +1840,8 @@ parameter from this specification, it **MUST** return `400 Bad Request`.
 ## <a href="#profiles" id="profiles" class="headerlink"></a> Profiles
 
 JSON:API supports the use of "profiles" as a way to indicate additional
-semantics that apply to a JSON:API request/document, without altering the
-basic semantics described in this specification.
+semantics that apply to a JSON:API document, without altering the basic
+semantics described in this specification.
```
dgeb, line 1848: Good adjustment 👍

_format/1.1/index.md
```diff
@@ -1901,39 +1894,39 @@ This profile defines the following keywords:
...
 
-A client **MAY** use the `profile` media type parameter in conjunction with the
-JSON:API media type in an `Accept` header to _request_, but not _require_, that
-the server apply one or more profiles to the response document. When such a
-request is received, a server **SHOULD** attempt to apply the requested profiles
-to its response.
+A client **MAY** use the `profile` media type parameter in an `Accept` header
+to request that the server apply one or more profiles to the response document.
+When such a request is received, a server **SHOULD** attempt to apply the
+requested profiles to its response.
```
```suggestion
requested profile(s) to its response.
```

dgeb (changes requested):
@gabesullice looks really good! Just requested a couple tweaks.

_format/1.1/index.md
gabesullice: Restored with modification.

_format/1.1/index.md
```diff
@@ -37,36 +37,33 @@ Clients and servers **MUST** send all JSON:API data using this media type in the
...
-Servers **MUST** respond with a `406 Not Acceptable` status code if a
-request's `Accept` header contains the JSON:API media type and all instances
-of that media type are modified with media type parameters.
-
-> Note: These content negotiation requirements exist to allow future versions
-of this specification to add other media type parameters for extension
-negotiation and versioning.
+Servers **MUST** respond with a `406 Not Acceptable` status code if a request
+specifies an instance of the JSON:API media type in an `Accept` header with any
+media type parameter other than `profile`.
```
dgeb: I think there's a little nuance lost here. Perhaps something like:

> Servers **MUST** respond with a `406 Not Acceptable` status code if a
> request's `Accept` header contains the JSON:API media type and all instances
> of that media type are modified with a media type parameter other than `profile`.

gabesullice: Your suggestion leaves the door open to _some_ instances of the media type using parameters other than `profile`, was that your intention? If so, why?

gabesullice: Ah, I see that the original allowed the same thing. My change broke those semantics. It still seems like a problem to me though. There must be something I'm missing.

gabesullice: Ah, I see that the original allowed that too, and my suggestion is breaking that. It still seems like a problem to me though.

dgeb: I don't think including an extra parameter as an option that some servers may understand should be rejected. Some future client (say that supports v1.2) may include a future (unknown to v1.1) parameter, and the server could choose to respond to only the v1.1 compatible parameter (`profile`).

gabesullice: Thanks for the clarification. That makes sense.

dgeb (approved):

PR #1428, opened 2019-08-30
Add jsonapi-instance-store to JavaScript implementations.

yoshi12u:
Added [jsonapi-instance-store](https://github.com/uesaiso/jsonapi-instance-store) to JavaScript implementations.
Inserted it near other libraries with similar names.


dgeb:
@uesaiso would you mind moving your library to the bottom of the list (for JavaScript)? Currently the list is chronological, but we should consider whether to make it alphabetical.

yoshi12u:
@dgeb Thank you for your reply.
I don't mind it!

dgeb:
@uesaiso I'm very sorry for not coming back to this earlier. This PR got lost in the queue after my initial comment. 

I see that your library no longer exists at the URL you provided. If you've moved it and want to re-add it to the list here, please open a new PR with the new URL.

PR #1430, opened 2019-09-13
Add missing comma

nogates:
Just noticed there is a missing comma in this example


dgeb:
@nogates good catch - thanks!

PR #1437, opened 2019-10-08
Introduce Atomic Operations extension

dgeb:
This is a proposal for an official extension to the JSON:API spec, as described in #1435. This proposal is based upon #1254, and supersedes that PR.

This extension provides a means to perform multiple "operations" in a linear and atomic manner. Operations are a serialized form of the mutations allowed in the base JSON:API specification. It uses the namespace `atomic` to emphasize the atomicity guarantee it provides.

### DIFF
--- a/ext/atomic/index.md
+++ b/ext/atomic/index.md
@@ -221,9 +221,10 @@ those specified by the request (for example, updating the `updatedAt` attribute
 or a computed `sha`), then the server **MUST** return a result that includes a
 representation of the updated resource as `data`.
 
-If a server accepts an update and doesn’t update any attributes besides those
-provided, the server **MUST** return a result with no `data` or, if all results
-are empty, the server **MAY** respond with `204 No Content` and no document.
+If a server accepts an update and doesn’t update any fields besides those
+provided, the server **MUST** return a result that includes either no `data` or
+a representation of the resource as `data` or, if all results are empty, the
+server **MAY** respond with `204 No Content` and no document.
 
 If a server is unable to update the resource, an appropriate error response
 **MUST** be returned and a response document **SHOULD** be returned that
--- a/ext/atomic/index.md
+++ b/ext/atomic/index.md
@@ -24,9 +24,10 @@ members allowed by the base specification with the exception of `data` and
 In addition, such a document **MAY** include either of the following members,
 but not both:
 
-- `atomic:operations` - an array of [operation objects](#operation-objects).
+- `atomic:operations` - an array of one or more [operation
+  objects](#operation-objects).
 
-- `atomic:results` - an array of [result objects](#result-objects).
+- `atomic:results` - an array of one or more [result objects](#result-objects).
 
 If either `atomic:operations` or `atomic:results` is present, the `errors`
 member **MUST NOT** be included in the same document.
--- a/ext/atomic/index.md
+++ b/ext/atomic/index.md
@@ -123,7 +123,7 @@ respond with a `400 Bad Request` and **SHOULD** include a document with a
 top-level `errors` member that contains an error object. The error object
 **SHOULD** include a `source` member with a `pointer` to the invalid operation.
 
-If an operation is properly formed, but can not be processed by a server, the
+If an operation is properly formed, but cannot be processed by a server, the
 server **MUST** respond with a `400 Bad Request` or a more appropriate error
 response (e.g. `409 Conflict` or `422 Unprocessable Entity`) and **SHOULD**
 include a document with a top-level `errors` member that contains one or more
--- a/ext/atomic/index.md
+++ b/ext/atomic/index.md
@@ -1,0 +1,510 @@
+# Atomic Operations Extension
+
+This extension provides a means to perform multiple "operations" in a linear and
+atomic manner. Operations are a serialized form of the mutations allowed in the
+base JSON:API specification.
+
+Clients can send an array of operations in a single request. This extension
+guarantees that those operations will be processed in order and will either
+completely succeed or fail together.
+
+## <a href="#namespace" id="namespace" class="headerlink"></a> Namespace
+
+This extension uses the namespace `atomic`.
+
+> Note: JSON:API extensions can only introduce new document members using a
+> reserved namespace as a prefix.
+
+## <a href="#document-structure" id="document-structure" class="headerlink"></a> Document Structure
+
+A document that supports this extension **MAY** include any of the top-level
+members allowed by the base specification with the exception of `data` and
+`included`, which **MUST NOT** be included.
+
+In addition, such a document **MAY** include either of the following members,
+but not both:
+
+- `atomic:operations` - an array of [operation objects](#operation-objects).
+
+- `atomic:results` - an array of [result objects](#result-objects).
+
+If either `atomic:operations` or `atomic:results` is present, the `errors`
+member **MUST NOT** be included in the same document.
+
+### <a href="#operation-objects" id="operation-objects" class="headerlink"></a> Operation Objects
+
+An operation object **MUST** contain the following member:
+
+- `op`: an operation code, expressed as a string, that indicates the type of
+  operation to perform. Valid values include:
+
+  - `"add"`: creates a new resource or relationship
+  - `"update"`: updates a resource or relationship
+  - `"remove"`: removes a resource or relationship
+
+An operation object **MAY** contain either of the following members, but not
+both, to specify the target of the operation:
+
+- `ref`: an object that **MAY** contain any of the following combinations of
+  members:
+
+  - `type` and `id`: to target an individual resource.
+  - `type`, `id`, and `relationship`: to target an individual resource's relationship.
+
+  Note that `local:id` **MAY** be used instead of `id` when used together with
+  the local identity extension.
+
+- `href`: a string that contains a URI-reference [[RFC3986 Section
+  4.1](https://tools.ietf.org/html/rfc3986#section-4.1)] that identifies the
+  target of the operation.
+
+An operation object **MAY** also contain any of the following members:
+
+- `data`: the operation's "primary data".
+
+- `meta`: a [meta object](#document-meta) that contains non-standard
+  meta-information about the operation.
+
+Different members are required for processing different types of operations, as
+described below.
+
+### <a href="#result-objects" id="result-objects" class="headerlink"></a> Result Objects
+
+An operation result object **MAY** contain any of the following members:
+
+- `data`: the "primary data" resulting from the operation.
+
+- `meta`: a [meta object](#document-meta) that contains non-standard
+  meta-information about the result.
+
+An empty result object (`{}`) is acceptable for operations that are not required
+to return `data`.
+
+## <a href="#processing" id="processing" class="headerlink"></a> Processing
+
+All HTTP requests sent with this extension **MUST** be issued with `POST`.
+
+A server **MUST** perform operations in the order they appear in the
+`atomic:operations` array.
+
+A server **MUST** perform all operations atomically, so that a failure
+to perform any operation **MUST** invalidate any effects of preceding
+operations.
+
+A server **MUST** respond to a successful operations request with `200 OK` if a
+response document is returned. An array of [result objects](#result-objects)
+**MUST** be returned in a top-level `atomic:results` member. The results array
+**MUST** be the same length as the requested operations array and each result
+**MUST** correspond positionally to its associated operation.
+
+A server **MAY** respond to successful requests that include operations with
+`204 No Content` and no response document if no operations are required to
+return `data`.
+
+### <a href="#processing-errors" id="processing-errors" class="headerlink"></a> Processing Errors
+
+If any operation in a request fails, the server **MUST** respond as described
+[in the base specification](/format/#errors-processing). An array of one or
+more [error objects](/format/#error-objects) **SHOULD** be returned, each with
+a `source` member that contains a `pointer` to the source of the problem in the
+request document.
+
+### <a href="#specific-operations" id="specific-operations" class="headerlink"></a> Processing Specific Operations
+
+The following sections describe how to process specific operations.
+
+#### Creating Resources
+
+An operation that creates a resource **MAY** target a resource collection
+through the operation's `ref` member. The operation **MUST** include an `op`
+code of `"add"` as well as a resource object as `data`. The resource object
+**MUST** contain at least a `type` member.
+
+For example:
+
+```json
+POST /bulk HTTP/1.1
+Host: example.org
+Content-Type: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+Accept: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+
+{
+  "atomic:operations": [{
+    "op": "add",
+    "href": "/blogPosts",
+    "data": {
+      "type": "articles",
+      "attributes": {
+        "title": "JSON API paints my bikeshed!"
+      }
+    }
+  }]
+}
+```
+
+Note that `href` is used in this example to target a resource collection,
+`blogPosts` in this case, that is distinct from the `type` of the resource.
+
+##### Responses
+
+If the server is able to create a resource with a client-generated ID and its
+representation would be identical to the resource in the operation, then the
+server **MAY** return a result with no `data` member or it **MAY** return a
+result that contains the created resource as `data`.
+
+In all other cases in which the server is able to successfully create the
+resource, the server **MUST** return a result that contains the created resource
+as `data`.
+
+For example:
+
+```json
+HTTP/1.1 200 OK
+Content-Type: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+
+{
+  "atomic:results": [{
+    "data": {
+      "links": {
+        "self": "http://example.com/blogPosts/13"
+      },
+      "type": "articles",
+      "id": "13",
+      "attributes": {
+        "title": "JSON API paints my bikeshed!"
+      }
+    }
+  }]
+}
+```
+
+If the server is unable to create the resource, an appropriate error response
+**MUST** be returned and a response document **SHOULD** be returned that
+contains a top-level `errors` as [described above](#processing-errors).
+
+#### Updating Resources
+
+An operation that updates a resource **MUST** target that resource through the
+operation's `ref` member. The operation **MUST** include an `op` code of
+`"update"`.
+
+For example:
+
+```json
+POST /bulk HTTP/1.1
+Host: example.org
+Content-Type: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+Accept: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+
+{
+  "atomic:operations": [{
+    "op": "update",
+    "ref": {
+      "type": "articles",
+      "id": "13"
+    },
+    "data": {
+      "type": "articles",
+      "id": "13",
+      "attributes": {
+        "title": "To TDD or Not"
+      }
+    }
+  }]
+}
+```
+
+##### Responses
+
+If a server accepts an update but also changes the resource in ways other than
+those specified by the request (for example, updating the `updatedAt` attribute
+or a computed `sha`), then the server **MUST** return a result that includes a
+representation of the updated resource as `data`.
+
+If a server accepts an update and doesn’t update any attributes besides those
+provided, the server **MUST** return a result with no `data` or, if all results
+are empty, the server **MAY** respond with `204 No Content` and no document.
+
+If a server is unable to update the resource, an appropriate error response
+**MUST** be returned and a response document **SHOULD** be returned that
+contains a top-level `errors` as [described above](#processing-errors).
+
+#### Deleting Resources
+
+An operation that deletes a resource **MUST** target that resource through the
+operation's `ref` member. The operation **MUST** include an `op` code of
+`"remove"`.
+
+For example:
+
+```json
+POST /bulk HTTP/1.1
+Host: example.org
+Content-Type: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+Accept: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+
+{
+  "atomic:operations": [{
+    "op": "remove",
+    "ref": {
+      "type": "articles",
+      "id": "13"
+    }
+  }]
+}
+```
+
+##### Responses
+
+If a server is able to delete the resource, the server **MUST** return a result
+with no `data` or, if all results are empty, the server **MAY** respond with
+`204 No Content` and no document.
+
+If a server is unable to delete the resource, an appropriate error response
+**MUST** be returned and a response document **SHOULD** be returned that
+contains a top-level `errors` as [described above](#processing-errors).
+
+#### Updating To-One Relationships
+
+An operation that updates a resource's to-one relationship **MUST** target that
+relationship through the operation's `ref` member. The operation **MUST**
+include an `op` code of `"update"`.
+
+For example, the following request assigns a to-one relationship:
+
+```json
+POST /bulk HTTP/1.1
+Host: example.org
+Content-Type: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+Accept: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+
+{
+  "atomic:operations": [{
+    "op": "update",
+    "ref": {
+      "type": "articles",
+      "id": "13",
+      "relationship": "author"
+    },
+    "data": {
+      "type": "people",
+      "id": "9"
+    }
+  }]
+}
+```
+
+And the following request clears a to-one relationship:
+
+```json
+POST /bulk HTTP/1.1
+Host: example.org
+Content-Type: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+Accept: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+
+{
+  "atomic:operations": [{
+    "op": "update",
+    "ref": {
+      "type": "articles",
+      "id": "13",
+      "relationship": "author"
+    },
+    "data": null
+  }]
+}
+```
+
+##### Responses
+
+If a server is able to update the relationship, the server **MUST** return a
+result with no `data` or, if all results are empty, the server **MAY** respond
+with `204 No Content` and no document.
+
+If an update is successful, the server **MUST** return a result with no `data`
+or, if all results are empty, the server **MAY** respond with `204 No Content`
+and no document.
+
+#### Updating To-Many Relationships
+
+An operation that updates a resource's to-many relationship **MUST** target that
+relationship through the operation's `ref` member.
+
+To add members to a to-many relationship, the operation **MUST** include an
+`op` code of `"add"`. For example:
+
+```json
+POST /bulk HTTP/1.1
+Host: example.org
+Content-Type: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+Accept: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+
+{
+  "atomic:operations": [{
+    "op": "add",
+    "ref": {
+      "type": "articles",
+      "id": "1",
+      "relationship": "comments"
+    },
+    "data": [
+      { "type": "comments", "id": "123" }
+    ]
+  }]
+}
+```
+
+To replace all the members of a to-many relationship, the operation **MUST**
+include an `op` code of `"update"`. For example:
+
+```json
+POST /bulk HTTP/1.1
+Host: example.org
+Content-Type: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+Accept: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+
+{
+  "atomic:operations": [{
+    "op": "update",
+    "ref": {
+      "type": "articles",
+      "id": "1",
+      "relationship": "tags"
+    },
+    "data": [
+      { "type": "tags", "id": "2" },
+      { "type": "tags", "id": "3" }
+    ]
+  }]
+}
+```
+
+To remove members from a to-many relationship, the operation **MUST** include an
+`op` code of `"remove"`. For example:
+
+```json
+POST /bulk HTTP/1.1
+Host: example.org
+Content-Type: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+Accept: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
+
+{
+  "atomic:operations": [{
+    "op": "remove",
+    "ref": {
+      "type": "articles",
+      "id": "1",
+      "relationship": "comments"
+    },
+    "data": [
+      { "type": "comments", "id": "12" },
+      { "type": "comments", "id": "13" }
+    ]
+  }]
+}
+```
+
+##### Responses
+
+If a server is able to update the relationship, the server **MUST** return a
+result with no `data` or, if all results are empty, the server **MAY** respond
+with `204 No Content` and no document.
+
+If an update is successful, the server **MUST** return a result with no `data`
+or, if all results are empty, the server **MAY** respond with `204 No Content`
+and no document.
+
+### Processing Multiple Operations
+
+The above examples all perform a single operation that aligns with an equivalent
+singular request from the base spec. Yet, the primary value of this extension is
+that unlocks the ability to perform more than one action linearally and
+atomically.
+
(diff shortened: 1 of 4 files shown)


jugaadi:
Doubts:
1. In the base spec, it is given: `A server MAY choose to stop processing as soon as a problem is encountered, or it MAY continue processing and encounter multiple problems.`. How do we map the error objects with its corresponding operation object in the transaction? Or should we stop processing the transaction as soon as we encounter an error?
2. How do we ensure **idempotency** wrt transaction? Can we pass a `atomic:key` to be used along with the operations object for identifying the transaction? 
3. Do we need a `ref` object when we already have a *data* object that clearly specifies the target resource and *op* property to provide the necessary context?
4. In the context of operations,
    1. Does order matter when multiple operations are issued for a single resource? Ex. Relationship members are first deleted and then updated.
    2. Do we need multiple ways to handle an operation? For example.
        1. Add a new relationship or a new member. 
            - To-One:  op : update , data : null
            - To-Many: op: remove, data : [{...}]
        2. Update/Replace a relationship. Additionally, it can be also used to add or delete a  relationship.
            - Through op: `update` wrt Resource 
            - Through op: `update` wrt Resource Relationship
        3. Remove/Clear a relationship
            - To-One:  op : update , data : null
            - To-Many: op: remove, data : [{...}]

Suggestions:
1. As this extension dictates server processing wrt transactional semantics, we can consider using `transaction` as namespace instead of `atomic`.

tobyzerner:
@jugaadi 

Re: point 1 - Error objects may contain a `source` member which could point to the operation in which the error occurred (or a specific path within the operation object). I'm assuming that once an error is encountered in an operation, no further operations should be processed, but I think the extension is ambiguous on this matter?

Re: point 3 - the `data` object alone is not sufficient to identify the target resource, as the `type` and the target resource collection are not necessarily the same. As demonstrated in the example for creating a resource, where the href is "/blogPosts" and the type is "articles". `ref` is also required to reference a specific relationship. 

Re: point 4.i - 

> A server **MUST** perform operations in the order they appear in the
 `atomic:operations` array. 

So the end result would be the same as if multiple separate requests to the API were issued, but with the guarantee of atomicity.

ext/atomic/index.md
```diff
@@ -0,0 +1,510 @@
...
+##### Responses
+
+If a server accepts an update but also changes the resource in ways other than
+those specified by the request (for example, updating the `updatedAt` attribute
+or a computed `sha`), then the server **MUST** return a result that includes a
+representation of the updated resource as `data`.
+
+If a server accepts an update and doesn’t update any attributes besides those
+provided, the server **MUST** return a result with no `data` or, if all results
+are empty, the server **MAY** respond with `204 No Content` and no document.
```
tobyzerner: I recognise that this is the same as for updating a resource in the base spec... but I'll take this opportunity to ask in both contexts: why forbid including the resource representation if no other attributes have changed? It just makes implementation that little bit more difficult... what is the rationale for making it a MUST rather than a SHOULD or MAY?

ext/atomic/index.md
```diff
@@ -0,0 +1,510 @@
...
+
+##### Responses
+
+If a server is able to update the relationship, the server **MUST** return a
+result with no `data` or, if all results are empty, the server **MAY** respond
+with `204 No Content` and no document.
+
+If an update is successful, the server **MUST** return a result with no `data`
+or, if all results are empty, the server **MAY** respond with `204 No Content`
+and no document.
```
tobyzerner: I'm confused about the difference between this paragraph and the paragraph above it?

tobyzerner:
Overall looks good to me, would be very happy to see this as an official extension and hopefully one day in the base spec. I may try implementing this soon so could give more feedback as a result of that. 

Should the extension have anything to say about how an implementation should handle security? e.g. limiting the number of operations per request. I suppose not as it's in the same domain as rate-limiting which the base spec doesn't say anything about.

Doqnach:
Would this mean, this will also cover the case of wanting to `POST` several resources at once?

At first I thought it was fine to `POST` a `{"data":[{...},{...},...]}`, but the spec actually says you can only `POST` one resource at a time...

Personally I would prefer the above notation for such a case though instead of using an extension.

ext/atomic/index.md
```diff
@@ -0,0 +1,510 @@
...
+to return `data`.
+
+## <a href="#processing" id="processing" class="headerlink"></a> Processing
+
+All HTTP requests sent with this extension **MUST** be issued with `POST`.
+
+A server **MUST** perform operations in the order they appear in the
+`atomic:operations` array.
+
+A server **MUST** perform all operations atomically, so that a failure
```
ahx, line 104: Suggested change:
A server **MUST** perform all operations as a single atomic transaction, so that […].

I think we should mention the word "transaction" or "transactional" at least once to make it clear that this is not about processing multiple atomic operations, but about processing multiple operations as one atomic transaction.

ext/atomic/index.md
dgeb: @tobyzerner Thanks for pointing this out. I had copied the wrong boilerplate for the error handling scenario.

gabesullice:
### @jugaadi:

1. I think an error object's `/source/pointer` member is sufficient to identify the operation object as the source of an error. Does that work in your opinion?
1. The spec does not currently have any built-in idempotency mechanisms and I don't think `atomic` should introduce them. I would love to see an extension that explores `Etag` + `If-Match` _inter alia_ to add more idempotency guarantees to the spec as a whole. It seems that it's an orthogonal concern that would benefit the spec in many areas.
1. I think we do need the `ref` object. Consider the case of adding a relationship. The `data` member will be a resource identifier, not a resource object, thus the `ref` is necessary to communicate to _which_ relationship the resource identifier object should be added. Also, consider the case of a `remove` operation (for resource objects _and_ relationships).
1. ...
  i. Order does matter, operations must be processed serially.
  ii. Sorry, I don't understand what you're asking or pointing out with these examples. Would you mind clarifying?

WRT to `transaction` vs `atomic`, I think @dgeb had a reason he preferred the latter. I'll let him respond.

### @tobyzerner:

Thanks for your reply! I already wrote the above, before I saw this. You're right about `/source/pointer`! To answer your question, yes, the extension is intentionally silent about whether a server should continue processing or not. It's an implementation detail that is use-case specific I think. F.e. adding resource objects in bulk would probably benefit from it, but adding a resource object that is a target of a relationship and then adding the resource object that targets it would not.

Re: point 3, spot on!

Re: point 4, exactly :) Good spec reading!

Re: 200 vs 204, I believe the spec is this way for two reasons: 1) so that a client does not have to unnecessarily update its internal representation and 2) to save bytes.

I'm glad to hear you like the spec and are considering implementing it! That's great :)

TBH, I'm ambivalent about a security section. There's nothing normative we ought to say about it, but I know many specs do try to explicitly call out specific security considerations.

### @Doqnach:

Yes. This allows you to create many resource objects in one request. I agree that your syntax would be simpler for the simple use case you describe, but there are many hidden edge cases. F.e. can those added resources reference one another? If so, must they be created in a specific order? If we support serial additions, why not serial mutations? This extension solves these complex cases while also solving the use case you mentioned, albeit with a little bit less syntactic sugar.

ext/atomic/index.md
dgeb: @ahx thanks for bringing this up. I agree that this line could be clearer. What would you think about the following alternative?

> A server **MUST** perform all operations together atomically, so that [...]

I hope this achieves the effect you're going while avoiding the word "transaction". I actually deliberately avoided using the word "transaction" in this extension, because it brings to mind [ACID](https://en.wikipedia.org/wiki/ACID) guarantees, of which atomicity is just one. And really, the guarantees made by this extension just focus on atomicity. There is no more of a "durability" guarantee than provided by the base spec, for instance. 

On the other hand, I think that using lower-case-t "transaction" in this one instance wouldn't be terribly confusing.

jugaadi:
@tobyzerner @gabesullice @dgeb   Thanks for the clarifications.

Below are some of my concerns. Do correct me if Im wrong.
1. Errors: Yes `/source/pointer` would solve the problem. However, I would still prefer `atomic:results` to reflect a successful as well as an error response so that http `status` is factored in the result object. Have included a format in the suggestion section.
2. Idempotency: My assumption was Local IDs wrt resource objects could be reused to solve idempotency issues. However, in the context of operations, idempotency guarantees becomes a necessity because the same transaction if processed twice may result in inconsistent state. For example:
    - T1:
        1. Create b1
        2. Update a1 relationship: a1 - b1
    - T1:
        1. Create b1 # New b2 will be created.
        2. Update a1 relationship: a1 - b2
3. Ref: My understanding was that `href` could be compulsorily used for all operations instead of `ref`. When combined with the basic `data` element and `op` element, the server should be able to derive the reference object.

UPDATING RESOURCES
```
{
  "atomic:operations": [{
    "op": "update",
    "href": "/articles/1",
    "type": "articles",
    "id": "1",
    "attributes": {
        "title": "To TDD or Not"
    }
  }]
}
```

UPDATING RELATIONSHIPS
```
{
  "atomic:operations": [{
    "op": "add",
    "href": "/articles/1/relationships/responses",
    "type": "articles",
    "id": "1",
    "relationship": {
        "responses": { 
            "data": [
                { "type": "comments", "id": "123" }
            ]
        }
    }
  }]
}
```
4. Operations Order: As the operations are processed in the given order, the onus lies on the client to ensure that conflicting operations are ordered properly. I was wondering if the spec can guide on addressing intra request conflicts. For example, all operations must be processed in the following precedence:
    1. Remove
    2. Add
    3. Update

---
Suggestions

1. Though it is understood that asynchronous operations in a single request cannot be supported, it would be better if it is explicitly mentioned in the spec.
2. Unlike a database transaction, all operations are handled through a single HTTP request. As a result, it would be relatively easy for a server to guarantee ACID semantics for a given request. Moreover, from a client perspective, the base spec + this spec implicitly sets ACID as a default expectation. Hence, suggested "transaction" albeit a lightweight one.
3. A transaction must additionally support asynchronous response as a server might relay it to multiple micro-services internally.
4. A result object can follow a common format: 
`{ status: <http status code>, data: <response data> }.` or `{ <errors object which contains status> }.`

gabesullice:
@jugaadi

gabesullice:
Concerns:

1. I think @dgeb was trying very hard to disassociate operations from HTTP requests. An operations requests is not meant to be a substitute for multiple HTTP requests. Rather, an operations-compatible resource is a single REST resource where interdependent entity mutations are made atomically. It's not a tunnel for HTTP requests. This is why status is not a part of the extension. IOW, a single request either succeeds or fails, atomically. If one operation contains a client error, the whole thing has a client error.

I think it's useful to think of cases where operations can be used with a limited scope. For example, consider this operations request to commerce API:

```

**Comment by gabesullice** [comment]
Responding to your concerns @jugaadi (also, thank you for taking the time to think so deeply about this!):

1. I think @dgeb was trying very hard to disassociate operations from HTTP requests. An operations requests is not meant to be a substitute for multiple HTTP requests. Rather, an operations-capable resource can be seen as a single REST resource where interdependent entity mutations are made atomically through a single representation of a transaction. I.e., it's not a tunnel for HTTP requests. This is why status is not a part of the extension. If one operation in the unit transaction has a client error, the entire transaction has a client error. I think it would cause confusion about the atomicity guarantee if some operations had error codes and some didn't.

2. Hmm, my feeling is that idempotence is something that ought to be solved more generally than just the `atomic` case. For example, there could be a extension called `trace` where a new, required member is added to the top-level document, like so:

```json
{
  "trace:requestID": "{client-generated-uuid}",
  "atomic:operations": [{}, {}, "..."]
}
```

The server could trace requests and reject requests with IDs that it has already processed. This would work for atomic operations and also for `POST`ing individual resource objects. Am I missing something that makes `atomic` exceptional in this regard?

3. I think you're understanding it correctly. IIRC, @dgeb wanted to have `ref` to simplify server implementations (because they wouldn't need to dereference a URL). Maybe I'm missing some additional nuance, so don't take that as 100% of the story.

4. Ahh, I see. Yes, you're right, since operations are processed serially, that definitely places a burden on clients to do things in the correct order. I never thought about the fact that `remove`, `add`, `update` would always be correct when dealing with one of each operation type. That's cool! Unfortunately, it still doesn't help a client that _adds_ 3 resources, `a`, `b`, & `c` with relationships between them: `a <- b <- c`. The client would still need to know to create `a`, _then_ `b` _then_ `c`.

We could give explicit guidance for clients like this:

> All `remove` operations **MUST** precede any `add` or `update` operation(s). All `add` operations **MUST** precede any `update` operation(s). Additionally, operations which establish relationships to resource objects that are added in the same request **MUST** be preceded by the operations that added those related resource objects.

Just for context: Order was made a requirement to simplify server implementations. @dgeb and I felt that clients would generally have an easier time ordering operations than servers would since, in most cases, UI interactions will spawn operations. As long as those spawned operations are kept in order, they should automatically be processable in the order that they were spawned (even if it might be an inefficient order) since anything else would have been a nonsensical user flow.

jugaadi:
Thanks @gabesullice  for the explanation.

1. My understanding was based on the same premise that operation is a single resource and each individual operation is just like an attribute of the `atomic:operations` resource. Therefore, I just used `status` in the same vein as the `status` property in the `error` object i.e 
> When a server encounters multiple problems for a single request, the most generally applicable HTTP error code SHOULD be used in the response.
> ...
> An error object MAY have the following members:
> `status` : the HTTP status code applicable to this problem, expressed as a string value.

If it still gives the same impression, we can avoid it. We can go ahead with the current format.

2. If idempotence is solved in a generic way, it would be awesome. The reason it matters in this extension is the following:
    1. As bulk(n) creation of resources is supported in this spec, 'm' similar requests may allow:
        1. Creation of duplicate(m x n) resources of each type in the server.
        2. Creation of incorrect relationships because of multiple requests.
    2. Decide whether there is a need to consider `PUT` for `atomic:operations` request?
    3. Decide whether idempotence should be part of the spec so that this extension can use it.

3. That works for me. Just wanted to avoid another keyword and format. My assumption was that the current json:api server implementations would have already used urls(separate endpoint) to associate resources with related resources.

4.  I agree with your suggestion. Order is definitely important and arrays seem to be the best option when compared to JSON Pointers. The guidance should be good enough to resolve any ambiguity in implementation.

ext/atomic/index.md
```diff
@@ -0,0 +1,510 @@
...
+
+- `meta`: a [meta object](#document-meta) that contains non-standard
+  meta-information about the result.
+
+An empty result object (`{}`) is acceptable for operations that are not required
+to return `data`.
+
+## <a href="#processing" id="processing" class="headerlink"></a> Processing
+
+All HTTP requests sent with this extension **MUST** be issued with `POST`.
```
bart-degreed, line 99: This used to be PATCH (in #1254). I assume the change is intentional?

ext/atomic/index.md
```diff
@@ -0,0 +1,510 @@
...
+      "attributes": {
+        "title": "JSON API paints my bikeshed!"
+      }
+    }
+  }]
+}
+```
+
+Note that `href` is used in this example to target a resource collection,
+`blogPosts` in this case, that is distinct from the `type` of the resource.
```
bart-degreed, line 172: is `href` optional?

ext/atomic/index.md
```diff
@@ -0,0 +1,510 @@
...
+
+An operation that creates a resource **MAY** target a resource collection
+through the operation's `ref` member. The operation **MUST** include an `op`
+code of `"add"` as well as a resource object as `data`. The resource object
+**MUST** contain at least a `type` member.
+
+For example:
+
+```json
+POST /bulk HTTP/1.1
```
bart-degreed: Maybe change `/bulk` to `/operations` to reflect intended usage.

ext/atomic/index.md
gabesullice: It was :) Good eye!

ext/atomic/index.md
gabesullice:
```suggestion
POST /operations HTTP/1.1
```

I think that makes sense. Leaving a suggestion so @dgeb can just commit it if he concurs.

ext/atomic/index.md
gabesullice: It is. Although, this helped me realize that the introduction under `#### Creating Resources` does not mention `href`. I think we ought to do a better job of clarifying how and when to use `ref` or `href`, in general.

ext/atomic/index.md
dgeb: I agree that `/operations` makes for a better example 👍

ext/atomic/index.md
dgeb: >  It just makes implementation that little bit more difficult... what is the rationale for making it a MUST rather than a SHOULD or MAY?

@tobyzerner Indeed, this requirement is carried over from the base spec for consistency. The original motivation was to simplify client-side implementations so that responses could simply be ignored, rather than parsed and compared. Do you think that server-side implementations become more difficult with this requirement?

ext/atomic/index.md
david-kubecka: Maybe I misunderstand something in the base spec (I'm studying JSON API only shortly) but what does it mean if `href` isn't specified? Is the resource somehow created in the collection of the same name as `data.type`? What if that doesn't exist?

david-kubecka: I forgot to explain the reference to base spec: Is there generally any constraint (or guidance) about the URI of the collection API (e.g. `/articles`) and type(s) of resource objects it can contain?

gabesullice: There are no statements about URL structure in the spec, there are some recommendations outside the spec though: https://jsonapi.org/recommendations/#urls-resource-collections

I like to use "reminders" as an example. I.e. a server could have a `/reminders/{id}` for GETing all reminders globally and _also_ `/user/{userID}/reminders` for GETing and POSTing reminders for only a specific user. All of that would be spec-compliant.

In that example, the `href` member exists to allow a client to perform an `add` operation on the `/user/{userID}/reminders` collection

david-kubecka: What is the default value of `href` then? According to https://jsonapi.org/recommendations/#urls-resource-collections having resource collection with the same name as each type is only recommended (even though it makes great sense). So default value can't just be `/articles`, right?

On a related note, using your `reminders` example, is this scenario valid according to spec/recommendations?

- POSTing a reminder to `/user/{userID}/reminders` makes that reminder available also directly under `/reminders` (with the same `id`).

In different words, JSON API allows to model the concept of "views" (`/user/{userID}/reminders` is a view to `/reminders`).

ext/atomic/index.md
```diff
@@ -0,0 +1,510 @@
...
+#### Updating Resources
+
+An operation that updates a resource **MUST** target that resource through the
+operation's `ref` member. The operation **MUST** include an `op` code of
+`"update"`.
+
+For example:
+
+```json
+POST /bulk HTTP/1.1
```
bart-degreed: @dgeb /bulk occurs multiple times; you may want to update remaining references to /operations too.

dgeb: @bart-degreed good call - I've just updated them all.

ext/atomic/index.md
gabesullice: > What is the default value of href then?

I don't think there needs to be a default. I think this should be part of the clarifications we make about `ref` & `href`

gabesullice: > On a related note, using your `reminders` example, is this scenario valid according to spec/recommendations? ...

Yes. That's valid.

david-kubecka: > > What is the default value of href then?
> 
> I don't think there needs to be a default. I think this should be part of the clarifications we make about `ref` & `href`

Earlier you said that the `href` is optional. If it's missing in which collection the article gets created?

gabesullice: I'm sorry. I see how what I said could be confusing. `ref` and `href` are mutually exclusive and it's usually optional which one you use. If you use `ref`, then you can't use `href`. When using `ref`, a newly created resource object ought to go into the collection for that resource object's type.

In certain cases, you can omit _both_ `href` and `ref`. For example, let's say that you have a `/reminders/operations` endpoint. At that endpoint, all operations must modify or create only `reminder` resource objects. The sensible default for that endpoint would be to add new resource objects to the `/reminders` collection.

This extension (and the spec for that matter) tries not to be overly prescriptive. Your server should choose to do what makes sense. The philosophy of JSON:API is generally, "_if_ you implement this like that, then do it like this." I.e., _if_ you support operations on a generic endpoint like `/operations` then you _should_ use `ref` or `href` as necessary. However, if you do not need `ref` or `href` because your operations endpoint is limited in scope, like `/reminders/operations`, then you are not required to use `ref` _nor_ `href`. That is why the language in the extension is:

> An operation object **MAY** contain [either `ref` or `href`]

instead of:

> An operation object **MUST** contain [either `ref` or `href`]

david-kubecka: Ok. I now understand. Thanks for your explanation! Obviously it would be good if this explanation appeared (in the distilled form) in the final documentation.

jugaadi:
Any updates?

ext/atomic/index.md
```diff
@@ -0,0 +1,510 @@
...
+
+In addition, such a document **MAY** include either of the following members,
+but not both:
+
+- `atomic:operations` - an array of [operation objects](#operation-objects).
+
+- `atomic:results` - an array of [result objects](#result-objects).
+
+If either `atomic:operations` or `atomic:results` is present, the `errors`
+member **MUST NOT** be included in the same document.
```
klobuczek, line 40: Can this somehow be relaxed? I'm thinking about streaming large payloads. It would be convenient to process operation as they are streamed and stream back the results until an error occurs at which point the server would revert all operations and send an error. Potentially even continuing processing to provide to the client a full error list for the entire payload. So rather than having `errors` as a top-level element only, how about allowing an `error` as a result type?

auvipy (commented):
looks great!

bart-degreed:
In the base json:api spec, a server MUST return 201 No Content with no response document only for delete and relationship update. In all other cases, it MAY, allowing the server to return the same data as in the request. This can be useful in cases where it's expensive to determine what has changed.

I believe the equivalent for atomic:operations is: "the server MUST return a result with no data or, if all results are empty, the server MAY respond with 204 No Content and no document."

The cases where this rule applies are the same as in the base spec, except for one:

"If a server accepts an update and doesn’t update any attributes besides those provided, the server MUST return a result with no data or, if all results are empty, the server MAY respond with 204 No Content and no document."

Compared to the base spec:

"If an update is successful and the server doesn’t update any attributes besides those provided, the server MUST return either a 200 OK status code and response document (as described above) or a 204 No Content status code with no response document."

My conclusion is that the atomic:operations spec is more strict than the base spec for resource updates. Is this intentional or an oversight?

I would prefer this to be loosened to MAY, as in:
"If a server accepts an update and doesn’t update any attributes besides those provided, the server MAY return a result with no data..."

ext/atomic/index.md
gabesullice: I think that that use case is sufficiently on the edge that we should not alter the design of this extension to support it. Turning errors into a result type would probably be confusing for the majority of users who will not understand why the spec allows individual errors, yet forces the entire request to roll back. For example, many people would probably want to use this spec for bulk creation + retries if that were the case.

All that said, I _do_ really see value in a different extension that better suits your needs.

gabesullice:
@bart-degreed, great attention to detail! I do not see a problem with loosening the extension to allow the server to either respond with an empty result _or_ "complete" result. The only thing that must be preserved is that the order and count of results must match the order and count of the requested operations.

@dgeb, am I missing some nuance that requires the extension to be more strict? The only reason that I can think of to choose the stricter wording in the extension is to allow the client to skip work if the result is empty.

morvans:
While updating my earlier implementation of the Operations proposal to the new Atomic Operations extension, I found that we do not clearly state if the "atomic:operations" array can be empty, in an operations request.

ext/atomic/index.md
```diff
@@ -0,0 +1,510 @@
...
+  - `"update"`: updates a resource or relationship
+  - `"remove"`: removes a resource or relationship
+
+An operation object **MAY** contain either of the following members, but not
+both, to specify the target of the operation:
+
+- `ref`: an object that **MAY** contain any of the following combinations of
+  members:
+
+  - `type` and `id`: to target an individual resource.
```
morvans, line 59: At first sight, I think it may induce that both `type` & `id` were required, but of course it's not in case of an "add" operation. Should we be more specific ?
```suggestion
  - `type` and `id`: to target an individual resource. `id` **MAY** be omitted in case of an "add" operation.
```

dgeb:
@bart-degreed as @gabesullice said, thanks for your attention to detail! I agree with your assessment.

I have loosened the language around updates to support the same level of strictness as the base spec. The section in question now reads:

> If a server accepts an update and doesn’t update any fields besides those
provided, the server **MUST** return a result that includes either no `data` or
a representation of the resource as `data` or, if all results are empty, the
server **MAY** respond with `204 No Content` and no document.

Note that I've changed `attributes` to `fields` here. I think this should make it back into the base spec since "attributes" has a very specific meaning in the spec, which was never intended here.

ext/atomic/index.md
dgeb: I think I agree with this refinement. What do you think @gabesullice ?

dgeb:
> While updating my earlier implementation of the Operations proposal to the new Atomic Operations extension, I found that we do not clearly state if the "atomic:operations" array can be empty, in an operations request.

@morvans Thanks for raising this. It's an edge case that I have not considered, but will discuss with @gabesullice.

Possible update to consider:

```
- `atomic:operations` - an array of one or more [operation
  objects](#operation-objects).

- `atomic:results` - an array of one or more [result objects](#result-objects).
```

ext/atomic/index.md
dgeb: Sorry, I misunderstand the context here. After talking with @gabesullice we realized that `ref` is unnecessary in an `add` operation, which will already contain a `type` in the `data` member that represents the resource to be added.

morvans: Don't we still need a `ref` to represents the "endpoint" to which the resource will be sent ? 
If `ref` is considered unnecessary for `add` (we infer the endpoint from the resource itself), it  also become unnecessary for `update`, right ?

From an implementer POV, I like the idea of just having to look at `ref` (or `href`) to determine to which endpoint pass the data which is treated like an opaque blob at this point.

morvans: I mean, in this particular context of Atomic Operations, if `ref` is not necessary in an `add` operation, why should we have different endpoints when adding new resources through "plain" JSON:API calls ?  A single /jsonapi/add would suffice.
But (in Drupal JSON:API implementation, at least), `add` operations are performed on different endpoints, based on the resource type, and then there's a check that the data's `type` (and also `id`, for in case of updates) matches the endpoint URL.
I agree that this can be perceived as a kind of "redundancy", but this is already the case with existing "regular" JSON:API, right ?
Let's say I build a JSON:API implementation where *every* write calls goes to a `/jsonapi` accepting PATCH or POST requests and inferring everything from the passed data, would it still be compliant with the original JSON:API spec ?

ext/atomic/index.md
```diff
@@ -0,0 +1,512 @@
...
+members allowed by the base specification with the exception of `data` and
+`included`, which **MUST NOT** be included.
+
+In addition, such a document **MAY** include either of the following members,
+but not both:
+
+- `atomic:operations` - an array of one or more [operation
+  objects](#operation-objects).
+
+- `atomic:results` - an array of one or more [result objects](#result-objects).
```
gabesullice, line 37: I've been writing a simple implementation of this and this flexibility is causing a bit of extra work. Would it make sense to heighten this requirement to:

> In addition, such a document **MUST** include one of the following members, but not both:

while changing <q>A document that supports this extension...</q> above to <q>A document using this extension...</q>

I guess it comes down to a decision about whether we want to force servers to flexibly validate documents according to their declared extensions or whether we want clients to be able to be less precise about the extensions they're really expecting the server to process.

gabesullice: For context, here's a snippet of my code:

```php
$supported_extensions = [
  'https://jsonapi.org/extensions/#atomic'
];
$unsupported_extensions = array_diff($extensions, $supported_extensions);
if (!empty($unsupported_extensions)) {
  throw new UnsupportedMediaTypeException(sprintf('The %s JSON:API media type extension is not supported', current($unsupported_extensions)));
}
if (in_array($supported_extensions[0], $extensions, TRUE)) {
  $request_document = Json::decode((string) $request->getContent());
  if (!isset($request_document['atomic:operations'])) {
    throw new BadRequestHttpException('Request documents using the %s JSON:API media type extension must include an `atomic:operations` top-level member.');
  }
}
```

It's that final `throw new BadRequestException()` that I'm concerned about. According to the language above, I think it's technically valid for a client to send a `Content-Type` header with the atomic extension URI, even when it's not including an `atomic:operations` member.

gabesullice: Actually, I guess it's not that bad. All it means is that I have to do:

```php
$operations = isset($request_document['atomic:operations'])
  ? $request_document['atomic:operations']
  : [];
```

ext/atomic/index.md
gabesullice: @morvans:

> Let's say I build a JSON:API implementation where every write calls goes to a /jsonapi accepting PATCH or POST requests and inferring everything from the passed data, would it still be compliant with the original JSON:API spec?

Yes, I think it would be. Imagine something like `calendar.mydomain.com/`. Here, I could `POST` reminders and events. A `GET` request to `/` (root) might return a mixed collection of reminders and events in chronological order. Is it good API design? I think you could argue both yes and no, but the spec doesn't prevent it.

---

One of the use cases that I wanted to support by using the `href` member in an operation object was the concept of creating resource objects via editorialized endpoints. For example: adding both a `tshirt` and a `sandal` resource object to a `/summer-styles` endpoint.

I think that's sort of similar to what you're asking. You might be implementing a `/operations` endpoint where you want to add both`tshirt` and `sandal` resource objects:

```json
[{
  "op": "add",
  "ref": {
    "type": "tshirt"
  },
  "data": {
    "type": "tshirt",
    "attributes": {
      "sex": "male",
      "color": "blue"
    }
  }
}, {
  "op": "add",
  "ref": {
    "type": "sandal"
  },
  "data": {
    "type": "sandal",
    "attributes": {
      "sex": "unisex",
      "size": "gargantuan"
    }
  }
}]
```

That feels very redundant to me, honestly. However, you do say:

> From an implementer POV, I like the idea of just having to look at ref (or href) to determine to which endpoint pass the data which is treated like an opaque blob at this point.

Knowing that you're implementing this within Drupal I'll take some liberties with my example...

I suspect you're talking about routing subrequests and I see why you'd want to treat the `data` member as an opaque blob, but you have to deserialize the entire JSON document anyway. So, it seems like the code for handling an omitted `ref` member below isn't too onerous or inelegant:

```php
  $operation = $deserialized['atomic:operations'][0];
  $url = $operation['href'] ?? Url::fromRoute("jsonapi.{$operation['data']['type']}.collection.post")->toString();
  $subrequest = Request::create($url);
```

Since allowing `ref` to be omitted is elegant from the spec's perspective and there are still relatively elegant solutions for the server to handle that as well. I think it's okay. WDYT?

gabesullice:
Just gave this another light scan. A few thoughts:

1. the extension should say that its own URI is `https://jsonapi.org/ext/atomic`.
1. it's unclear whether or not operations other than `add`, `update`, and `remove` are permitted.
1. the way that the "combinations" of allowed `ref` members is described is somewhat confusing. I think this needs some reorg and `lid` shouldn't be in parentheses (see general note below)
1. `s/can not/cannot`: https://github.com/json-api/json-api/pull/1437/files#diff-3258e77d8c1edf3edfa2dda4b957adceR122

I think the _Operation objects_ section needs to be rewritten. It feels clunky right now, probably because it grew organically as we tweaked and retweaked it. I don't think the actual structure of operation objects need to change, but I think that if we were to start that section with an empty canvas we could come up with a clearer way of describing the document structure.

dgeb:
@gabesullice those are all good points. I've just pushed commits that attempt to address them, although I'm certainly open to discussing and refining them further.

> I think the Operation objects section needs to be rewritten. It feels clunky right now, probably because it grew organically as we tweaked and retweaked it. 

Yes, I'd be open to collaborating on this 👍

gabesullice:
Thanks everyone for all the input on this! I think it's time to get this merged 🎉

MaxChong3188:
> This is a proposal for an official extension to the JSON:API spec, as described in #1435. This proposal is based upon #1254, and supersedes that PR.
> 
> This extension provides a means to perform multiple "operations" in a linear and atomic manner. Operations are a serialized form of the mutations allowed in the base JSON:API specification. It uses the namespace `atomic` to emphasize the atomicity guarantee it provides.

PR #1440, opened 2019-10-29
Make `type` field in resource optional.

vanb:
This feels redundant to the resource path which already contains information about the resource type. While a type field could be useful for an endpoint that mixes resource types (questionable if that should be allowed), a MUST seems too strict here.


vanb:
Wrong target repo

Issue #1281, opened 2018-06-15
Schema Support


remmeier:
Based on the discussions of deficiencies compared to GraphQL in https://github.com/json-api/json-api/issues/1272 I'm openeing this ticket to discuss the schema support.

This issue has also be raised in the Drupal implementation (https://www.drupal.org/project/jsonapi/issues/2822768) or the Crnk implementation (http://www.crnk.io/releases/stable/documentation/#_meta_module) among likely various others. Goal would be to have a standardized way to specify resources, which in turn would opened up many new possiblities:

- generic UIs to explore JSON API endpoints
- generation of client and server stubs in various programming languages to ease development
- more automated testing
- automation within user interfaces
- ...

what could the specification cover:

- definition of a primitive type system (dates, urls, boolean, string, etc.)
- definition of array types
- definition of complex types composed of attributes having a type
- definition of documents, resources, meta, links, relationship types basedon complex types.
- definition of custom types composed of any other type (primitve, array, complex, etc.)
- set of validation constraints (minimum length, regex pattern, not empty, etc.) that can be attache   to types

the schema could be implemented with two formats:

- a low-level schema definition following the JSON:API specification 
- a more lightweight/human-readable schema definition in a custom DSL similar to GraphQL and others that can be translated to the JSON:API-conform specification.

The former target automated processing/discovery, while the later is targered towards developers.

how could an implementation look like:

- JSON:API-conform schema resources could be make available JSON:API endpoint next to application-specific resources. This would be signaled to clients as profile (#1268)
- the schema definitions can be stored in one or more files with a given file suffix
- a reference implementation is provided to translate a DSL-based schema to a JSON:API-based schema.
- an (extensible) reference implementation translates JSON:API-based schema to client stubs for varous languages.
- reference implementations are made available as command line tools distributed as platform-independent docker images
- (later) a language service would provide syntax highlighting/code completing for IDEs
- the compliance test suite (https://github.com/json-api/json-api/issues/1209) would contribute a schema definition of its own together with pairs of requests/responses to check both client and server implementations.

sources of inspiration for the DSL and JSON:API-based schema:

- graphql
- JSON Schema
- openapi/swagger
- protobuf/gRPC
- https://github.com/crnk-project/crnk-framework/tree/master/crnk-meta/src/main/java/io/crnk/meta/model ( for the JSON:API-base schema, but simpler at various places)

[...]

PR #1447, opened 2019-12-03
Introduce describedby link relation

dgeb:
The `describedby` link relation can be used to identify a description document, which may be an [OpenAPI](https://www.openapis.org) definition, a [JSON Schema](https://json-schema.org) document, or another description format.

It's proposed that the `describedby` member can appear in two places:

* A) in the top-level `links` object, in which case it provides a description for the _current_ document.
* B) inside any link object, in which case it provides a description for the target of the link.

We could optionally allow this member inside any `links` object that can also contain a `self` link - namely, resource links and relationship links. This would be consistent with (A) but also somewhat redundant with (B). I'd like to make a decision about this explicitly prior to merging.

Another question is whether there should be a central definition of a "description document" within the JSON:API spec, or whether using parenthetical examples `(e.g. OpenAPI or JSON Schema)` is sufficient.

related: #1281 

/cc @philsturgeon @gabesullice

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -556,13 +556,15 @@ key in an [error object]'s links object, each key present in a links object
 
 In the example below, the `self` link is simply a URI string, whereas the
 `related` link uses the object form to provide meta information about a
-related resource collection:
+related resource collection as well as a schema that serves as a description
+document for that collection:
 
 ```json
 "links": {
-  "self": "http://example.com/articles/1",
+  "self": "http://example.com/articles/1/relationships/comments",
   "related": {
     "href": "http://example.com/articles/1/comments",
+    "describedby": "http://example.com/schemas/article-comments",
     "meta": {
       "count": 10
     }
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -114,6 +114,8 @@ The top-level [links object][links] **MAY** contain the following members:
   resource relationship.
 * `profile`: an array of [links][link], each specifying a [profile][profiles]
   in use in the document.
+* `describedby`: a [link] to a description document (e.g. OpenAPI or JSON
+  Schema) for the current document.
 * [pagination] links for the primary data.
 
 The document's "primary data" is a representation of the resource or collection
@@ -538,6 +540,8 @@ Within this object, a link **MUST** be represented as either:
     [relationship object][relationships] in which the link appears.
   * `params`: a [link parameter object][link parameters] describing additional
     information about the link or its target.
+  * `describedby`: a [link] to a description document (e.g. OpenAPI or JSON
+    Schema) for the link target.
   * `meta`: a meta object containing non-standard meta-information about the
     link.
 


philsturgeon:
🚢 👍

jugaadi:
Thanks @dgeb @philsturgeon @gabesullice  for initiating this effort.

I'd vote for adopting a central default description format to avoid any confusion between API providers and consumers. Any other format can be handled through profiles. 

In this case, I would prefer OpenAPI because:
1. Large community
2. Industry wide support
3. Tooling
4. Alignment with JSON Schema

PS: We are already using OpenAPI for describing JSON:API resources. This spec is used by both the client and server to handle requests and responses.

dgeb:
@jugaadi Thanks for your feedback! I agree that OpenAPI is a solid choice and a good starting point.

While we don't want to restrict the content type of the description documents pointed to by `describedby`, we definitely want to provide some recommendations in the non-normative section of the site that will cover best practices. We can include example usage of OpenAPI and JSON Schema to start, and would appreciate your contributions to the recommendations based on your experience.

dgeb:
@gabesullice I think this is ready to review. Please let me know what you think about my question from the PR description:

> We could optionally allow this member inside any links object that can also contain a self link - namely, resource links and relationship links. This would be consistent with (A) but also somewhat redundant with (B). I'd like to make a decision about this explicitly prior to merging.

gabesullice:
@dgeb and I discussed this today. WRT to the question in the description, we decided against allowing `describedby` _links_ in resource or relationship objects because it may cause confusion about whether the description document is intended to describe the object as it appears in the current response or the description document of the document you expect to receive by following its `self` link. The `describedby` _target attribute_ is a _link hint_ that clearly indicates that the description document it targets describes the target URL. 

I think this PR is ready to merge.

gabesullice (approved):
This looks good to me.

evert:
There is also this format: https://tools.ietf.org/html/draft-nottingham-link-hint

bit different from how this was done, but does the same thing nonetheless

gabesullice:
@evert, afaict, we could integrate that RFC w/ JSON:API fairly easily. They're not mutually exclusive. To do that, we would add `hints` as an allowed member of a [link objects](https://jsonapi.org/format/1.1/#document-links) both as a sibling of—and similar to—the `params` member.

If you're saying that we could have added `describedby` under `hints`, I don't think that works because that RFC is about providing hints related to HTTP headers, statuses, etc. I suppose it would be a proxy for the `Link` header, but for better or for worse the JSON:API spec, talks about links exclusively about links represented in the JSON response body.

Perhaps you're referring to the `links` hint from that RFC since we could have done something like:

```
links: {
  self: {
    href: /articles/1,
    hints: {
      links: {
        describedby: /schemas/article.json
      }
    }
}
```

Instead of what we chose:

```
links: {
  self: {
    href: /articles/1,
    describedby: {
      href: /schemas/article.json
    }
}
```

We considered something like that (i.e. adding a `links` target attribute to link objects themselves) but decided against it to keep nesting/complexity to a minimum.

PS. It's funny you mentioned that RFC, I _just_ mentioned it [in a post](https://www.drupal.org/project/drupal/issues/2984628) within 30 minutes of your reply 😺

evert:
Yea just for context, I do believe that your use-case is one of the things link-hints hopes to solve, but it would indeed be a deeper nested structure.

I believe the equivalent syntax would actually be:

```
links: {
  self: {
    href: /articles/1,
    hints: {
      formats: {
        'application/vnd+api.json': {
           links: {
             describedby: {
               href: "/schemas/article.json",
               type: "application/schema+json"
           }
        }
      }
    }
}
```

This makes the most sense for links that are explicitly not self links. For a self link, I would just take the format information and lift it to its own link:

```
links: {
  self: {
    href: /articles/1,
  },
  describedby: {
    href: "/schemas/article.json",
    type: "application/schema+json"
  }
}
```

gabesullice:
Whoaaa, I had not grokked that bit of the RFC. That's a very thought provoking idea... It's interesting how the format is specialized by a schema/profile.

> For a self link, I would just take the format information and lift it to its own link:

This PR does exactly that, but only for top-level links.

MikeRalphson:
Probably not worth opening a new issue, but just thought I'd note that `link` `describedBy` is generally about describing a resource (so JSON Schema fits well here), but an OpenAPI document generally describes the surface area of a whole or partial API.

The link rel [service-desc](https://tools.ietf.org/html/rfc8631#section-4.2) might be more appropriate for OpenAPI?

gabesullice:
I expect implementors to link to the HTML representations of their OpenAPI spec where they can add a fragment to the URL. For example, https://petstore.swagger.io/#/user

claudenirmf:
Hello, all! I am facing an interesting situation (not unlike the previous comment from @gabesullice) that could provide some insight.

In our implementation, we want to enhance the experience of developers, mainly because they may have to communicate to similar servers implementing different sets of features (e.g., only some implement a certain filter). So, we have decided to respond to errors with an error object, where we wish to use the link `describedby` to point to a resource where the developer may learn the specifics of that server's implementation.

Here are the catches: (i) this is intended to be human-readable; (ii) this usually is not even about resource itself, but the parameters sent in the request.

We still want to use `describedby` in successful responses such that a software can retrieve the JSON Schema for those resources, for instance. But I am unsure about how to differentiate these distinct usages `describedby`.

PR #1449, opened 2019-12-08
Schema: Improve regex for member names

### LINKED ISSUE #1155: Alternative regex for "attributes" patternProperties in schema
The regex used for "attributes" patternProperties in the [schema](http://jsonapi.org/schema) cannot be parsed by [go's regexp library](https://golang.org/pkg/regexp/) due to usage of the `(?!re)` pattern to disallow relationships and links properties.

The current regex is: `^(?!relationships$|links$)\\w[-\\w_]*$`

There may be an alternative way to express this which supports go's regexp library, for example by using the "not" keyword:
```
...
"patternProperties": {
  "^\\w[-\\w_]*$": {
    "description": "Attributes may contain any valid JSON value."
  },
  "^(relationships|links)$": {
    "description": "Reserved for future use.",
    "not": {}
  }
},
...
```

Related comment: https://github.com/xeipuuv/gojsonschema/issues/93#issuecomment-277501413

wamuir:
Fixes #1155 by eliminating the need for negative lookaheads in the schema. These are not universally supported, for instance in Go.

Also, this request improves alignment to the 1.0 JSON:API spec which states that the hypen minus and low line are not allowed as ending (or beginning) characters of a member name.  Current expressions do not follow this spec and allow these characters at the end of a member name.

### DIFF
--- a/schema
+++ b/schema
@@ -201,9 +201,17 @@
       "description": "Members of the attributes object (\"attributes\") represent information about the resource object in which it's defined.",
       "type": "object",
       "patternProperties": {
-        "^(?!relationships$|links$|id$|type$)\\w[-\\w_]*$": {
+        "^[a-zA-Z0-9](?:[-\\w]*[a-zA-Z0-9])?$": {
           "description": "Attributes may contain any valid JSON value."
         }
+      },
+      "not": {
+        "anyOf": [
+          {"required": ["relationships"]},
+          {"required": ["links"]},
+          {"required": ["id"]},
+          {"required": ["type"]}
+        ]
       },
       "additionalProperties": false
     },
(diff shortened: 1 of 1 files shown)


dgeb (approved):
@wamuir thanks for taking this on, and apologies for the delay in reviewing. I ran some tests and this seems to work well and improve on the former version.

@gabesullice would you mind reviewing as well please?

gabesullice (approved):
Thanks @wamuir! This looks really good.

The specification goes even further than this schema WRT to members named `relationships` or `links`. I.e. even objects nested _within_ an attribute are not permitted to have `relationships` or `links` members. I spent a little time playing around with [JSON Schema recursion](https://json-schema.org/understanding-json-schema/structuring.html#recursion) to achieve that, but it got a little too complicated. If you're interested, maybe you can think of a way :)

Regardless, this is a nice improvement. Merging. Thanks!

PR #1456, opened 2020-02-05
Simplify profiles for v1.1

dgeb:
This PR makes progress on the move to simplify profiles for v1.1 of the spec, as outlined in #1435. 

This continues work that's been merged already to unify profile negotiation. PR #1413 changed profiles to be purely opt-in by the server to align with [RFC 6906](https://tools.ietf.org/html/rfc6906).

This PR makes the following additional changes:

* Removes the ability for profiles to assign meaning to reserved query params such as `filter` and `page`. RFC 6906 states that "A profile MUST NOT change the semantics of the resource representation when processed without profile knowledge, so that clients both with and without knowledge of a profiled resource can safely use the same representation.". Since these params are intended to provide processing instructions, they are inappropriate for profiles.

* Moves profile authoring guidelines out of the base spec and into the non-normative recommendations page.

* Removes guarantees about profile registration and publication on jsonapi.org.


gabesullice:
@dgeb and I reviewed this together. This has been open for a week without comment. Merging...

PR #1457, opened 2020-02-05
Introduce extensions for v1.1

dgeb:
This introduces the concept of "extensions" to v1.1 of the spec. Extensions, and how they relate to profiles, are discussed extensively in #1435: 

> Just as profiles can be negotiated via the profile media type parameter, extensions will be negotiable via the ext media type parameter. Unlike profiles, a server will be required to understand and apply every extension requested via Content-Type and Accept headers, or else respond with an error. In this way, clients can be guaranteed that their requests will be processed according to the base spec plus any extensions they've applied.

> Extensions are strictly additive to the base spec, and can never impose rules about members already reserved for users. Any members introduced via extensions MUST be namespaced with a : prefix (e.g. local:id). Namespaces can only be used by extensions. Besides the fact that members are namespaced, most extensions could be merged with the base spec.

> Extensions will become the means to define custom processing directions for query params such as filter and page.

This PR includes the following:

* Introduces the `ext` media type parameter and describes content negotiation using this parameter.

* Provides an overview of extensions in the spec.

* Describes how extensions can introduce new namespaced members to the document structure.

Examples of extensions that are under review include "Local Identities" (#1436) and "Atomic Operations" (#1437). Merging this PR should unblock those PRs.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1927,6 +1927,19 @@ extension.
 Namespaces guarantee that extensions will never conflict with the current or
 future versions of this specification.
 
+In the following example, an extension requiring that each request document
+contains a unique request ID has been applied. It uses the namespace `request`
+and adds a top-level object member named `request:id`: 
+
+```
+{
+  "request:id": "some-long-uuid",
+  "data": {
+    "type": "comments",
+  }
+}
+```
+
 ## <a href="#profiles" id="profiles" class="headerlink"></a> Profiles
 
 JSON:API supports the use of "profiles" as a way to indicate additional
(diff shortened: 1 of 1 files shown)


gabesullice (approved):
This has been open for a week without comment. It opens the door to merging the local ID extension and operations extensions, as well as making the profile simplification patch appear more rational. IOW, LGTM! Thanks!

aimeos:
Please see https://github.com/json-api/json-api/issues/1435#issuecomment-586686184 and https://github.com/json-api/json-api/issues/1435#issuecomment-586687351 for comments

sandstrom:
@gabesullice and @dgeb Awesome! Looking forward to the local-ID extension

sandstrom:
@gabesullice and @dgeb Awesome! Looking forward to the local-ID extension!

PR #1461, opened 2020-02-26
Add server responsibility to reject unsupported, extended requests.

gabesullice:
Thanks to an astute comment by @freddrake in #1436 ([direct link](https://github.com/json-api/json-api/pull/1436#commitcomment-37403394)), @dgeb and I realized that the recent addition of extension negotiation to the 1.1 spec (#1457) was lacking a detail to aid client-server content negotiation. Namely, that the server must respond with a client error status code if the client sends an extended media type parameter with an unsupported extension. This PR adds the needed normative statement:

> If a request specifies the `Content-Type` header with an instance of
the JSON:API media type modified by the `ext` media type parameter and that
parameter contains an unsupported extension URI, the server **MUST** respond
with a `415 Unsupported Media Type` status code.


dgeb (approved):
I think this nails it. Thanks @gabesullice!

PR #1464, opened 2020-03-14
Remove json-api-store from implementations list

haydn:
I'm no longer maintaining the project and have archived the repository. It's probably more harm than good to keep it on this list.


dgeb:
Thanks for letting us know, @haydn!

PR #1468, opened 2020-04-09
Add flapison python server implementation

multimeric:
Further to discussion in https://github.com/miLibris/flask-rest-jsonapi/issues/188. There are concerns that since flask-rest-jsonapi is no longer maintained, previous users need to be aware that there is still a maintained fork.


dgeb:
@TMiguelT Thanks for your PR. I'm certainly willing to add your fork as a separate library, but there seems to be some activity in that discussion thread since you opened this PR. Can you keep us updated and let us know if you're planning to collaborate on the original library?

multimeric:
Yes, it's a good idea to pause this for the moment. Ideally there will be some solution that results in flask-rest-jsonapi being maintained, but I'm not certain that that will happen so I might keep this open in the meantime.

dgeb:
Sounds good, @TMiguelT.

jelhan:
@multimeric Do you have any updates you could share with us? Is the discussion still ongoing? Or should we either merge or close this one?

multimeric:
Many (but not all) of my PRs were merged, so I'm happy to close.

Issue #1384, opened 2019-04-08
Highlight currently viewed section in spec sidebar


danielcompton:
When reading the spec, I found that I wanted to get an idea of where I was in the document, and found it a bit tricky to quickly find out. It could be handy to highlight the current section that you are looking at in the sidebar, changing the section as you scroll down.

A quick mockup showing what this could look like:

<img width="1263" alt="Screenshot of Safari (8-04-19, 8-59-27 PM)" src="https://user-images.githubusercontent.com/811954/55711644-37612c00-5a41-11e9-98e7-ed02d575567a.png">

gabesullice:
I like this idea @danielcompton, would you consider submitting a PR?

danielcompton:
I don't have a lot of time at the moment, but if I do get a spare moment I'll take a look. Glad to know it looks useful!

nrmaridu:
Would you please assign this issue to me. I can submit a PR for this.

PR #1474, opened 2020-05-10
Highlight currently viewed section in spec sidebar #1384

nrmaridu:
When reading json api specification, it's getting difficult to realize in which section we are right now. This pull request highlight the side navigation links when user clicks on them.


auvipy (approved):

nrmaridu:
@auvipy Who can merge this into master?

dgeb:
@NageswaraRaoMaridu thanks for your PR, and sorry to take so long to review it!

I just tried this out and find that it gets us only part of the way to the goal of #1384. While it does highlight sections when they are clicked, it does not change the highlight as you scroll (thus the highlighting feel broken if it's not kept current). Are you up for that challenge as well?

PR #1475, opened 2020-05-11
pieoneers/jsonapi-go added into implementations list

fkhardikov:
I would like to propose our JSON API implementation written in Go. We originally developed it about 2 years ago for internal needs. It’s being improved every time we think it lacks functionality.  I’ve recently added some documentation and a few usage examples. Please take a look:
 https://github.com/pieoneers/jsonapi-go
We would like to contribute to the JSON API specification and keep development according to the community feedback.


dgeb:
@fkhardikov looks great - thanks!

PR #1478, opened 2020-05-20
Added `ts-japi` to the typescript/javascript implementation list

jrandolf:
`ts-japi` is a highly-modular typescript/javascript-framework agnostic library for serializing data to the JSON:API specification. It is full-featured, meaning all parts of the spec can be realized through this library.

Among other things,
-  This is the **only** typescript-compatible library that fully types the JSON:API specification and performs *proper* serialization.
- **Zero dependencies**.
- This is the **only** library with [resource recursion](https://github.com/jun-sheaf/ts-japi#zdg).
- The modular framework laid out here *highly promotes* the specifications intentions:
  - Using links is no longer obfuscated.
  - Meta can truly be placed anywhere with possible dependencies laid out visibly.
- This library is designed to adhere to the specifications "never remove, only add" policy, so we will remain backwards-compatible.


dgeb:
Great, thanks @jun-sheaf!

PR #1481, opened 2020-06-08
Move Graphiti from Ruby-Client to Ruby-Rails-Server

timkrins:
Graphiti is a server library, not a client.


dgeb:
@timkrins thanks for the correction!

PR #1482, opened 2020-06-11
Language cleanups and clarifications for version 1.1

gabesullice:
These changes perform some minor language cleanups and clarifications for version 1.1 of this specification (mostly related to the addition of the `ext` media type param). There are no normative changes.

PR #1483, opened 2020-06-11
More minor cleanups and clarifications for v1.1

dgeb:
Similar to #1482, this PR contains minor cleanups and clarifications for v1.1.

The only (arguably) normative change is to include a `SHOULD` alongside a `SHOULD NOT` to make a recommendation for representing relationships in `relationships` instead of `attributes`. I consider this more of a clarification than a change.


gabesullice (approved):
These are nice adjustments. I agree that the `SHOULD` bit does not constitute a normative change.

Toronto00:
https://travis-ci.org/ismasan/oat.svg?branch=master

Toronto00 (commented):
- [ ]

PR #1484, opened 2020-06-18
Refine extensions and profile for v1.1

dgeb:


auvipy (approved):

_format/1.1/index.md
```diff
@@ -7,7 +7,7 @@ status: wd
 
 JSON:API is a specification for how a client should request that resources be
 fetched or modified, and how a server should respond to those requests. JSON:API
-can also be easily extended with [profiles] and [extensions].
+can also be easily extended with [extensions] and [profiles].
```
gabesullice:
```suggestion
can be easily extended with [extensions] and [profiles].
```

_format/1.1/index.md
```diff
@@ -17,6 +17,17 @@ JSON:API requires use of the JSON:API media type
...
+
+All document members, query parameters, and processing rules defined by
+this specification are collectively called "specification semantics".
+
+Certain document members, query parameters, and processing rules are reserved
+for implementors to define at their discretion. These are called "implementation
+semantics".
+
+All other semantics are reserved for potential future use by this specification.
+
```
gabesullice, line 30: 👍

auvipy (approved):

_format/1.1/index.md
```diff
@@ -17,6 +17,17 @@ JSON:API requires use of the JSON:API media type
 ([`application/vnd.api+json`](http://www.iana.org/assignments/media-types/application/vnd.api+json))
 for exchanging data.
 
+## <a href="#semantics" id="semantics" class="headerlink"></a> Semantics
```
gabesullice, line 20: Should we add a link definition for this section at the bottom of this document?

_format/1.1/index.md
```diff
@@ -27,35 +38,179 @@ document are to be interpreted as described in
 [[RFC8174](https://tools.ietf.org/html/rfc8174)]
 when, and only when, they appear in all capitals, as shown here.
 
-## <a href="#content-negotiation" id="content-negotiation" class="headerlink"></a> Content Negotiation
+## The JSON:API Media Type
```
gabesullice: Needs an anchor.

```suggestion
## <a href="#jsonapi-media-type" id="jsonapi-media-type" class="headerlink"></a> The JSON:API Media Type
```

_format/1.1/index.md
```diff
@@ -27,35 +38,179 @@ document are to be interpreted as described in
...
+
+Extensions provide a means to "extend" the base specification by defining
+additional [specification semantics](#semantics).
+
+Extensions cannot alter or remove specification semantics, nor can they specify
+implementation semantics.
+
+### <a href="#profiles" id="profiles" class="headerlink"></a> Profiles
+
+Profiles offer a means to share a particular usage of the specification among
```
gabesullice:
```suggestion
Profiles provide a means to share a particular usage of the specification among
```

This change makes this sentence match the first sentence of the `#extensions` section. I don't have a preference for "provide" vs. "offer"—only that they match.

_format/1.1/index.md
```diff
@@ -27,35 +38,179 @@ document are to be interpreted as described in
...
+Profiles can specify [implementation semantics](#semantics), but cannot alter,
+add to, or remove specification semantics.
+
+### <a href="#media-type-parameter-rules" id="media-type-parameter-rules" class="headerlink"></a> Rules for Media Type Parameters
+
+The JSON:API media type **MUST NOT** be specified with any media type parameters
+other than `ext` and `profile`.
+
+Extensions and profiles are each uniquely identified by a
+[URI](https://tools.ietf.org/html/rfc3986). Visiting an extension's or profile's
```
gabesullice: Nit:
```suggestion
[URI](https://tools.ietf.org/html/rfc3986). Visiting an extension's or a profile's
```

_format/1.1/index.md
```diff
@@ -27,35 +38,179 @@ document are to be interpreted as described in
...
+```
+
+#### <a href="#profile-rules" id="profile-rules" class="headerlink"></a> Rules for Profiles
+
+The rules for profile usage are dictated by [RFC
+6906](https://tools.ietf.org/html/rfc6906).
+
+A profile **MAY** define document members, query parameters, and processing
+rules that are reserved for implementors.
+
```
gabesullice, line 145:
```suggestion

A profile **MUST NOT** define, alter, or remove document members, query parameters, or processing rules that are defined by or reserved for future use by this specification or by [extensions].

```

_format/1.1/index.md
```diff
@@ -27,35 +38,179 @@ document are to be interpreted as described in
...
+A server implementation that supports profile-defined query parameters **MUST
+NOT** support more than one profile per unique URL (including the entire query
+string).
+
+Unlike extensions, profiles do not need to define a namespace for document
+members and query parameters because profiles cannot define specification
+semantics and thus cannot conflict with the current or future versions of this
```
gabesullice:
```suggestion
semantics and thus cannot conflict with current or future versions of this
```

_format/1.1/index.md
```diff
@@ -27,35 +38,179 @@ document are to be interpreted as described in
...
+specification. Although it is possible for profiles to conflict with other
+profiles, it is the responsibility of implementors to ensure that all applied
+profiles are compatible.
```
gabesullice:
```suggestion
specification. However, it is possible for profiles to conflict with other
profiles. Therefore, it is the responsibility of implementors to ensure that
they do not support conflicting profiles.
```

_format/1.1/index.md
```diff
@@ -27,35 +38,179 @@ document are to be interpreted as described in
...
+
+In the following example, a profile has specified timestamp attributes that are
+required in each resource object. It has also specified that the value of these
+attributes be represented as ISO-8601 compliant date times. These attributes
+might appear as follows:
```
gabesullice:
```suggestion
In the following example, a profile has specified that all timestamps use the
[RFC 3339](https://tools.ietf.org/html/rfc3339) format and that all resource
objects contain `created` and `modified` attributes. With such a profile
applied, a response might appear as follows:
```

_format/1.1/index.md
```diff
@@ -27,35 +38,179 @@ document are to be interpreted as described in
...
+
+```json
+// ...
```
gabesullice:
```suggestion
```json
HTTP/1.1 200 OK
Content-Type: application/vnd.api+json; profile="https://example.com/resource-timestamps"

// ...
```

_format/1.1/index.md
```diff
@@ -27,35 +38,179 @@ document are to be interpreted as described in
...
+
+The rules for profile usage are dictated by [RFC
+6906](https://tools.ietf.org/html/rfc6906).
+
+A profile **MAY** define document members, query parameters, and processing
+rules that are reserved for implementors.
+
+A server implementation that supports profile-defined query parameters **MUST
+NOT** support more than one profile per unique URL (including the entire query
+string).
```
gabesullice: I forget exactly what our reasoning was for this statement although I'm sure we saw value in it. Was the intent for this statement to prevent two profiles from adding conflicting processing rules for the `filter` parameter?

IIUC, this goes too far and would preclude me from supporting this URL:

```
/users?foo=value&bar=value
```

where `foo` and `bar` are defined by two different profiles.

IOW, I _think_ this prevents an implementer from supporting >1 query parameter-defining profile at a time, which I don't believe we intended to do.

I tried and failed to reword this in a straightforward way. I feel like we've gone back and forth on this 1000 times, but perhaps we need a normative statement which says that profiles are not allowed to define query parameters of their own although they _are_ allowed to define processings rules for specification and extension defined query parameters whose interpretation is left to the implementor, like so:

```suggestion
A profile **MUST NOT** define any query parameters.

A profile **MUST NOT** alter or remove processing rules that have been defined
by this specification or by an [extension][extensions]. However, a profile
**MAY** define processing rules for query parameters whose processing rules
have been reserved for implementors to define at their discretion.

For example, a profile could define rules for interpreting [the `filter` query
parameter](#fetching-filtering), but it could not specify that relationship
names in [the `include` query parameter](#fetching-includes) are
space-separated instead of dot-separated.
```

I know that technically implementation-defined query parameters are part of implementation semantics, but I suspect that if we were to create a backwards-incompatible version 2 of JSON:API, we would leave extensions as the only means to define shareable query parameters.

gabesullice (changes requested):
Hey @dgeb , I only got as far as my last comment/suggestion. I'll finish the remainder of the review tomorrow.

_format/1.1/index.md
```diff
@@ -27,35 +38,179 @@ document are to be interpreted as described in
...
+    "title": "Rails is Omakase",
+    "created": "2020-07-21T12:09:00Z",
+    "modified": "2020-07-30T10:19:01Z"
+  }
+}
+// ...
+```
+
+Other profiles could specify rules for interpreting query parameters that are
+reserved for implementors, such as `filter` or `page`.
```
gabesullice: If we include the section I suggested above:

> For example, a profile could define rules for interpreting the `filter` query parameter, but it could not specify that relationship names in the `include` query parameter are space-separated instead of dot-separated.

Then I think we can remove these lines.

_format/1.1/index.md
```diff
@@ -27,35 +38,179 @@ document are to be interpreted as described in
...
+
+## <a href="#content-negotiation" id="content-negotiation" class="headerlink"></a> Content Negotiation
+
+### <a href="#content-negotiation-all" id="content-negotiation-all" class="headerlink"></a> Universal Responsibilities
+
+Clients and servers **MUST** send all JSON:API data using the JSON:API media
```
gabesullice:
```suggestion
Clients and servers **MUST** send all JSON:API payloads using the JSON:API media
```

_format/1.1/index.md
```diff
@@ -79,6 +234,17 @@ than `ext` or `profile`. If every instance of that media type is modified by the
 `ext` parameter and each contains at least one unsupported extension URI, the
 server **MUST** also respond with a `406 Not Acceptable`.
 
+If the `ext` parameter contains at least one unsupported extension URI, the
+server **MUST** also respond with a `406 Not Acceptable`.
```
gabesullice: This sentence appears to repeat the last sentence of the preceding paragraph.

_format/1.1/index.md
```diff
@@ -79,6 +234,17 @@ than `ext` or `profile`. If every instance of that media type is modified by the
...
+
+If the `profile` parameter is received, a server **SHOULD** attempt to apply any
+requested profile(s) to its response. A server **MUST** ignore any profiles
+that it does not recognize.
+
+> Note: The above rules guarantee strict agreement on extensions between the
+  client and server, while the application of profiles is left to the discretion
+  of the server.
```
gabesullice, line 252: 👌

_format/1.1/index.md
```diff
@@ -651,25 +817,26 @@ values.
...
   "profile": [
     "http://example.com/profiles/flexible-pagination",
     "http://example.com/profiles/resource-versioning"
   ]
 }
 ```
 
-> Note: Additional link types, similar to `profile` links, may be specified in
-the future.
-
```
gabesullice: I think this entire section needs to be removed since we no longer have `ext` or `profile` links defined above, especially since the example shows and array of links.

_format/1.1/index.md
dgeb: Good to be consistent. I think that `provide` feels more appropriate.

_format/1.1/index.md
dgeb: Nice clarification!

_format/1.1/index.md
dgeb: Good idea. Done.

_format/1.1/index.md
dgeb: Agreed. Removed.

_format/1.1/index.md
dgeb: Before removing, let's confirm that we really don't want to allow an in-document representation of which profiles / extensions have been applied.

_format/1.1/index.md
dgeb: I think this refinement adds a lot of clarity to what's allowed for profiles vs. extensions with respect to query params. I struggle to think of a case of a new profile-defined query param that would not be better defined via an extension.

I'm going to accept your changes and also drop "query parameters" from:

> A profile **MAY** define document members, query parameters, and processing
> rules that are reserved for implementors.

_format/1.1/index.md
dgeb: Discussed with @gabesullice and decided to move these URI arrays (`ext` and `profile`) as optional members of the `jsonapi` object.

gabesullice:
@dgeb , I pushed a few more changes. IMO, this PR is ready to go. If you agree, let's merge this!

gabesullice (approved):

_format/1.1/index.md
```diff
@@ -162,10 +162,10 @@ possible for profiles to conflict with other profiles. Therefore, it is the
 responsibility of implementors to ensure that they do not support conflicting
 profiles.
 
-In the following example, a profile has specified that all timestamps use the
-[RFC 3339](https://tools.ietf.org/html/rfc3339) format and that all resource
-objects contain `created` and `modified` attributes. With such a profile
-applied, a response might appear as follows:
+In the following example, a profile has defined a `timestamps`
+[attribute][attributes] as an object that must contain both a `created` and
+`modified` member and that they must use the [RFC 3339](https://tools.ietf.org/html/rfc3339)
```
dgeb: @gabesullice this phrasing ("and that") is now awkward and no longer fits with the verb "has defined". Perhaps a second sentence or rephrasing?

PR #1489, opened 2020-07-07
Add starlette-jsonapi Python server implementation

vladmunteanu:
Simpler way of building json:api compliant services in Python, without being tied to any specifc ORM.


dgeb:
@vladmunteanu great, thanks!

PR #1504, opened 2020-08-16
Note the need for quality values in Accept

vfaronov:
Implementors of `Accept` very often get this wrong. The spec’s own example in a previous draft got this wrong (see #1409).


gabesullice:
Thanks Vasiliy! I added a bit more nuance in order to give a little direction to server implementors as well.

dgeb:
Looks good @vfaronov and @gabesullice 👍

PR #1517, opened 2020-10-15
Add spraypaint and mobx-async-store to implementations

dashcraft:
Spraypaint is part of the graphiti jsonapi suite but is separate and is not tightly coupled to the backend service.
Mobx async store is a currently maintained/used in production library by my company @ArtemisAg.


dgeb:
Thanks @dashcraft!

PR #1523, opened 2020-11-21
Remove Reststate implementations

CodingItWrong:
I'm the maintainer of the Reststate libraries, and they will not be maintained going forward. This removes them from the implementations page so readers will be directed to maintained libraries instead.


dgeb:
@CodingItWrong thanks for the update!

PR #1525, opened 2020-12-01
clarify some details regarding error objects

### LINKED ISSUE #1496: Question about members of error objects
The error object, as defined in both 1.0 and the draft 1.1 specifications, MAY contain many members, but there are none that MUST be provided.  It seems insane that a service would return a completely empty error, but is it valid?

I'd like a statement in the specification that clarified that at least one member must be provided, even if it's the **meta** member with non-standard content.

freddrake:
- fixes https://github.com/json-api/json-api/issues/1496
- in json:api 1.1 error objects:
  - require at least one member be set
  - clarify that links, if provided, is not required to have BOTH about & type
  - strengthen status member with a SHOULD

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -2129,8 +2129,7 @@ An error object **MAY** have the following members, and **MUST** contain at
 least one of:
 
 * `id`: a unique identifier for this particular occurrence of the problem.
-* `links`: a [links object][links] that **MAY** contain the following members,
-  and that **MUST** contain at least one of:
+* `links`: a [links object][links] that **MAY** contain the following members:
   * `about`: a [link][link] that leads to further details about this
     particular occurrence of the problem. When derefenced, this URI **SHOULD**
     return a human-readable description of the error.
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -2125,10 +2125,12 @@ Error objects provide additional information about problems encountered while
 performing an operation. Error objects **MUST** be returned as an array
 keyed by `errors` in the top level of a JSON:API document.
 
-An error object **MAY** have the following members:
+An error object **MAY** have the following members, and **MUST** contain at
+least one of:
 
 * `id`: a unique identifier for this particular occurrence of the problem.
-* `links`: a [links object][links] containing the following members:
+* `links`: a [links object][links] that **MAY** contain the following members,
+  and that **MUST** contain at least one of:
   * `about`: a [link][link] that leads to further details about this
     particular occurrence of the problem. When derefenced, this URI **SHOULD**
     return a human-readable description of the error.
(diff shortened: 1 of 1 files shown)


auvipy (approved):

_format/1.1/index.md
```diff
@@ -2125,18 +2125,20 @@ Error objects provide additional information about problems encountered while
...
 keyed by `errors` in the top level of a JSON:API document.
 
-An error object **MAY** have the following members:
+An error object **MAY** have the following members, and **MUST** contain at
+least one of:
 
 * `id`: a unique identifier for this particular occurrence of the problem.
-* `links`: a [links object][links] containing the following members:
+* `links`: a [links object][links] that **MAY** contain the following members,
+  and that **MUST** contain at least one of:
```
dgeb: I discussed this with @gabesullice and we're both uncomfortable including this MUST requirement here. Throughout the spec, `links` objects do not have any particular required members. It seems plausible that custom `links` could be included here without requiring `about` or `type` in addition.

freddrake: Good point; fixed in 0de92c05fb29e4240505b3b94e30bc6ab1fb2114.

dgeb (approved):
Looks good to me now, thanks @freddrake!

@gabesullice can you please do a quick review as well?

gabesullice (approved):
These are great clarifications, thanks for submitting this @freddrake!

_format/1.1/index.md
lode: It would be great to make this more explicit. I always felt uncomfortable using custom link names in error objects because it was the only place where a links object had documented members. Maybe we can add something like 'may contain any member with the following members having a specified meaning'?

freddrake: Not sure what you mean about this being the only place where a links object has documented members; do self, related, and the pagination links not count?

Regarding "custom" links, maybe that's described somewhere in 1.1, but... I'm hesitant to approach that topic since I'm still trying to bring my implementation up to snuff for 1.0.

dgeb: @gabesullice and I were just talking about how custom members in `links` objects throughout the spec are still underspecified. We'd like to rectify this prior to 1.1 but, since it's a general issue, it doesn't need to hold up this PR.

lode: Good points, thanks!

Art4:
Consider the following response:

```json
{
    "errors": [
        {
            "description": "Some description about the error"
        }
    ]
}
```

This response is perfectly fine with 1.0, but is now invalid in 1.1 because of the new requirement. In my opinion, this is a breaking change that belongs to version 2.0 instead of 1.1. 

I propose that 1.1 should use `**SHOULD** contain at least one of` and in 2.0 this should be changed to `**MUST** contain at least one of`.

jelhan:
@Art4 Your example document was invalid for v1.0 as well. A `description` member in errors object is neither allowed in v1.0 not in v1.1.

Art4:
@jelhan Unknown/undefined properties must be ignored, but are not prohibited. This would be against the "never remove, only add" strategy. 

To quote the spec: https://jsonapi.org/format/1.0/#status

> While 1.0 is no longer the [latest version](https://jsonapi.org/format/) of JSON:API, **new versions will remain compatible with this one**, as JSON:API uses a _never remove, only add_ strategy.

and https://jsonapi.org/format/1.0/#document-structure

> Unless otherwise noted, objects defined by this specification MUST NOT contain any additional members. Client and server implementations MUST ignore members not recognized by this specification.

It is possible that a `description` property will be defined in 1.2 or later (like `jsonapi.profile`, `jsonapi.ext` or `error.source.header` where added in 1.1) and I'm building my parser with this in mind.

But to stay on topic, one can also imagine the following response:

```json
{
    "errors": [
        {}
    ]
}
```

PR #1500, opened 2020-08-07
state that the server may send a subset of fields by default

freddrake:
- the subset may be empty
- fields[type]= can be used to request no fields
- see https://github.com/json-api/json-api/issues/1499

PR #1532, opened 2021-01-03
clarify meaning of empty include query param

### LINKED ISSUE #1530: is empty include allowed?
Both the 1.0 and 1.1 versions of the specification state:

> The value of the ``include`` parameter MUST be a comma-separated (U+002C COMMA, “,”)
> list of relationship paths. A relationship path is a dot-separated (U+002E FULL-STOP, “.”)
> list of relationship names.

It is unclear whether an empty value for ``include`` represents an empty list of relationship paths or a single (invalid) relationship path.

jelhan:
The meaning of an empty value for the `include` query parameter was not clear yet. As @freddrake pointed out in #1530 a reader could interpret either as an empty list of relationship paths or as a single (invalid) relationship path.

The meaning of an empty value for `fields` query parameter is explicitly specified:

> The value of the fields parameter MUST be a comma-separated (U+002C COMMA, “,”) list that refers to the name(s) of the fields to be returned. An empty value indicates that no fields should be returned.

I _think_ the same applies to the `include` query parameter even though it's not mentioned explicitly yet. Otherwise a client would not be able to tell a server to not include any related resources - even if that server includes some resources by default.

I _think_ this is not a breaking change. Let's have a look at several scenarios:

1. An existing implementation may consider an empty value as a single relationship path. An empty string is not a valid relationship path. The server would be unable to identify the relationship path and response with `400 Bad Request`.
   > If a server is unable to identify a relationship path or does not support inclusion of resources from a path, it MUST respond with 400 Bad Request.

2. An existing implementation may silently ignore an `include` parameter if it's value is empty.
    1. If it does not include any related resources by default, the existing implementation would be valid.
    2. If it includes related resources by default even if an `include` parameter with an empty value is given, the implementation is not compliant with the specification even today. A server must not ignore a given `include` query parameter. It must response with a 400 if it does not support it.
        > If an endpoint does not support the include parameter, it MUST respond with 400 Bad Request to any requests that include it.
        >
        > If an endpoint supports the include parameter and a client supplies it, the server MUST NOT include unrequested resource objects in the included section of the compound document.
       > [...]
       > If a server is unable to identify a relationship path or does not support inclusion of resources from a path, it MUST respond with 400 Bad Request.

3. An existing implementation may already interpret an empty `ignore` parameter as an empty list of relationship paths and not include any related resources.

The meaning of an empty value for a Sparse Fieldset was also not explicitly specified since a few months ago. It was added in #1500. This was also _not_ considered a breaking change.

Closes #1530

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1238,7 +1238,7 @@ section of the [compound document].
 The value of the `include` parameter **MUST** be a comma-separated (U+002C
 COMMA, ",") list of relationship paths. A relationship path is a dot-separated
 (U+002E FULL-STOP, ".") list of [relationship][relationships] names. An empty
-value indicates that no related <span class="x x-first x-last">resources</span> should be returned.
+value indicates that no related resources should be returned.
 
 If a server is unable to identify a relationship path or does not support
 inclusion of resources from a path, it **MUST** respond with 400 Bad Request.
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1237,7 +1237,8 @@ section of the [compound document].
 
 The value of the `include` parameter **MUST** be a comma-separated (U+002C
 COMMA, ",") list of relationship paths. A relationship path is a dot-separated
-(U+002E FULL-STOP, ".") list of [relationship][relationships] names.
+(U+002E FULL-STOP, ".") list of [relationship][relationships] names. An empty
+value indicates that no related resource should be returned.
 
 If a server is unable to identify a relationship path or does not support
 inclusion of resources from a path, it **MUST** respond with 400 Bad Request.
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1238,7 +1238,7 @@ section of the [compound document].
 The value of the `include` parameter **MUST** be a comma-separated (U+002C
 COMMA, ",") list of relationship paths. A relationship path is a dot-separated
 (U+002E FULL-STOP, ".") list of [relationship][relationships] names. An empty
-value indicates that no related resource should be returned.
+value indicates that no related <span class="x x-first x-last">resources</span> should be returned.
 
 If a server is unable to identify a relationship path or does not support
 inclusion of resources from a path, it **MUST** respond with 400 Bad Request.


freddrake:
FWIW, I'm proceeding with the assumption that something like this is going to land at some point.

adilaid (approved):

_format/1.1/index.md
```diff
@@ -1237,7 +1237,8 @@ section of the [compound document].
 
 The value of the `include` parameter **MUST** be a comma-separated (U+002C
 COMMA, ",") list of relationship paths. A relationship path is a dot-separated
-(U+002E FULL-STOP, ".") list of [relationship][relationships] names.
+(U+002E FULL-STOP, ".") list of [relationship][relationships] names. An empty
+value indicates that no related resource should be returned.
```
prathe: If we wish to add this clarification. 

```suggestion
(U+002E FULL-STOP, ".") list of [relationship][relationships] names. An empty
value (no characters) indicates that no related resource should be returned.
```

auvipy (approved):

_format/1.1/index.md
dgeb: @prathe I'm not sure the clarification (`no characters`) is needed, but I'm not opposed either. If we were to add this, it should also be added to the similar language around [sparse field sets](https://jsonapi.org/format/1.1/#fetching-sparse-fieldsets). Since it affects multiple areas, maybe a small separate PR would be best.

```suggestion
value indicates that no related resources should be returned.
```

This is a really minor nit, but I think this should be plural.

dgeb (changes requested):
Thanks for adding this clarification. This seems good and is consistent with the language around sparse fieldsets. Just one minor nit to fix and then we can merge this.

_format/1.1/index.md
```diff
@@ -1237,7 +1237,8 @@ section of the [compound document].
 
 The value of the `include` parameter **MUST** be a comma-separated (U+002C
 COMMA, ",") list of relationship paths. A relationship path is a dot-separated
-(U+002E FULL-STOP, ".") list of [relationship][relationships] names.
+(U+002E FULL-STOP, ".") list of [relationship][relationships] names. An empty
+value indicates that no related <span class="x x-first x-last">resources</span> should be returned.
```
jelhan:
```suggestion
value indicates that no related resources should be returned.
```

_format/1.1/index.md
jelhan: Applied the suggestion. It included some markup, which is not shown in GitHub preview. :disappointed:  f4fcfad1b8671ab8f9742ea2f84558670d6d11db Removed that one in 9b90b6380964d308d4079c9b3c2f126f8e94c304. Should be good to merge now.

dgeb (approved):

dgeb:
Thanks again @jelhan for the clarifying PR and also to @freddrake for raising the issue in #1530.

PR #1536, opened 2021-01-16
Add new implementations of JSON:API

R3VoLuT1OneR:
* The **pz/doctrine-rest** unfortunately was in the wrong "client" list.
  * New library that helps implementing JSON:API added.


auvipy (approved):

dgeb (approved):
Thanks for the correction and new addition @R3VoLuT1OneR!

Issue #1492, opened 2020-07-21
Any standard way to include the modified resource for relationship operations?


cmeeren:
A great benefit of JSON:API is that the updated resource can be included in the response of `PATCH` requests, thus updating the client's representation (e.g. if an `updatedAt` field is changed).

Unfortunately, I can't find a way to do this for to-many relationship operations (`POST`/`PATCH`/`DELETE` to add/replace/remove members). At best, the related resources can be included, but if these don't have a relationship back to the resource that was modified, it seems that a separate `GET` request must be performed to get the updated resource.

For example, assuming that an article can have several authors, `POST /articles/1/relationships/authors` adds an author to the article, but since a `user` resource (the author) does not have a relationship back to the modified `article`, there is no way to include it (and it may have been modified in other ways, e.g. the article's `updatedAt` field may have changed). The only solution for the client to get the updated `article` seems to be to perform a separate `GET /articles/1` request after the relationship update was successful.

Is this correct, or am I missing something? Is this something that could be addressed in a future JSON:API version?

cmeeren:
Perhaps a leading dot could be used? E.g. `?include=.` to include the article in the example above, or `?include=.comments` to include the article and its comments. Seems like an elegant and backward compatible way of supporting this. It fits with the remarks in https://github.com/json-api/json-api/issues/1491#issuecomment-662276631.

jelhan:
Current v1.1 of the specification seems to forbid including the updated relationship(s) in the response if they haven't changed:

> 200 OK
>
> If a server accepts an update but also changes the targeted relationship(s) in other ways than those specified by the request, it MUST return a 200 OK response. The response document MUST include a representation of the updated relationship(s).
>
> A server MUST return a 200 OK status code if an update is successful, the client’s current data remain up to date, and the server responds only with top-level meta data. In this case the server MUST NOT include a representation of the updated relationship(s).

I'm especially referring to the second paragraph. It could be rephrased as: The server must not include a representation of the updated relationship(s) if the client's current data remain up to date. It's not entirely clear what is referred to as "client's current data" but in the context it seems to talk only about the targeted relationship(s).

Not saying that this is good. I just wanted to point out that this part of the specification is confusing. Especially as the specification has a note that `include` query parameter can be used when creating a relationship:

> Note: This section applies to any endpoint that responds with primary data, regardless of the request type. For instance, a server could support the inclusion of related resources along with a POST request to create a resource or relationship.

cmeeren:
[...]

Issue #652, opened 2015-05-21
Meta alone as a valid top level member


lgebhardt:
I'd like to allow `meta` to be a valid top level member for all responses. This will allow responses, by way of the top level `meta` member, to return debug, server metrics, and other information for all requests. Currently this is prohibited for deletes and updating relationships (and maybe a few others).

In the cases where 204 No Content is returned I'd propose we allow a 200 OK response as well.

I'd also suggest we change the description of `meta` from "non-standard meta-information about the primary data" to "non-standard meta-information about the primary data or request"

tkellen:
:+1: to a PR implementing this.

bintoro:
Top-level meta items can pertain to anything: primary data, included data, or the response as a whole. So how about just "non-standard meta-information"?

lgebhardt:
:+1: on "non-standard meta-information"

I'll put a PR together.

dgeb:
Closed via #657

PR #1540, opened 2021-02-19
Remove confusing sentence forbidding to include something in case it's not included

### LINKED ISSUE #1581: Clarification of PATCH response
> A server MUST return a 200 OK status code if an update is successful, the client’s current fields remain up to date, and the server responds only with top-level meta data. In this case the server MUST NOT include a representation of the updated resource(s).

I'm confused by this wording. In particular...

> the client's current fields remain up to date

So like... if the fields they sent to be updated were in fact updated and nothing else was updated?

> And the server responds only with top-level meta data

Confused by this one... I get the server may respond with meta data, but huh? How does this play a role in the... what?

I'm then further confused by how the following paragraph relates to the above-described situation.
> 204 No Content
If an update is successful and the server doesn’t update any fields besides those provided, the server MUST return either a 200 OK status code and response document (as described above) or a 204 No Content status code with no response document.

I understand this paragraph: If _only_ what we specified was successfully updated, return 200 and return the new representation of the entity. Or return a 204 with no document. But how exactly does it relate to the part I'm having trouble understanding?

jelhan:
The current specification for HTTP status code and response payload for successful relationship updates is confusing.

It specifies that "the server must not include a representation of the updates relationship(s)" if "the server response only with top-level meta data". But this does not make much sense.

A representation of the updates relationship(s) can only be included as primary data. If a server includes primary data, it would not only response with top-level meta data.

Unless I missed something there is no logical possible way to violate that _must not_ requirement. Therefore I don't think this is a breaking change.

The sentence was added as part of #652. The intend of that pull request was to allow response containing top-level meta data only. I don't think it's attend was to forbid including relationship data if it has not changed. Doing so would even limit some use cases as discussed in #1492.

Closes #1581

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1985,8 +1985,7 @@ updated relationship(s).
 
 A server **MUST** return a `200 OK` status code if an update is successful,
 the client's current data remain up to date, and the server responds
-only with top-level [meta] data. In this case the server **MUST NOT**
-include a representation of the updated relationship(s).
+only with top-level [meta] data.
 
 ##### <a href="#crud-updating-relationship-responses-403" id="crud-updating-relationship-responses-403" class="headerlink"></a> 403 Forbidden
 


auvipy (approved):

auvipy:
@gabesullice @dgeb

dgeb:
Closing in favor of #1638.

PR #1544, opened 2021-02-24
Fix a small typo in atomic extension writeup

acorncom:
Thanks for all your work on the atomic extension, looking forward to using it! 👍


dgeb (approved):
Good catch! Thanks @acorncom

PR #1487, opened 2020-06-30
fix(json-schema): Validate document's links according to the JSONAPI specification

guillemcanal:
Hi,

I believe that the current JSON Schema do not reflect the written specification regarding [document's links](https://jsonapi.org/format/#document-links) and [pagination links](https://jsonapi.org/format/#fetching-pagination).

1. It seems that the `allOf` keyword is misused in [`#/definitions/success/properties/links`](https://github.com/json-api/json-api/blob/gh-pages/schema#L38-L48).

According to the JSON Schema spec:
> An instance validates successfully against this keyword if it validates successfully against **all schemas** defined by this keyword's value. ([see in context](https://json-schema.org/draft/2019-09/json-schema-core.html#allOf))

2. Pagination links  (`next`, `prev`, `last`, `first`) **MAY** be either a `#/definitions/link` or `null`

 
I tested the fix against various examples provided on the JSONAPI website such as:

A resource with links:

```json
{
    "data": {
        "type": "articles",
        "id": "1",
        "attributes": {
            "title": "Rails is Omakase"
        },
        "relationships": {
            "author": {
                "links": {
                    "self": "/articles/1/relationships/author",
                    "related": "/articles/1/author"
                },
                "data": {
                    "type": "people",
                    "id": "9"
                }
            }
        }
    }
}
```

A collection with pagination links as `string`

```json
{
  "meta": {
    "totalPages": 13
  },
  "data": [
    {
      "type": "articles",
      "id": "3",
      "attributes": {
        "title": "JSON:API paints my bikeshed!",
        "body": "The shortest article. Ever.",
        "created": "2015-05-22T14:56:29.000Z",
        "updated": "2015-05-22T14:56:28.000Z"
      }
    }
  ],
  "links": {
    "self": "http://example.com/articles?page%5Bnumber%5D=3&page%5Bsize%5D=1",
    "first": "http://example.com/articles?page%5Bnumber%5D=1&page%5Bsize%5D=1",
    "prev": "http://example.com/articles?page%5Bnumber%5D=2&page%5Bsize%5D=1",
    "next": "http://example.com/articles?page%5Bnumber%5D=4&page%5Bsize%5D=1",
    "last": "http://example.com/articles?page%5Bnumber%5D=13&page%5Bsize%5D=1"
  }
}
```

A collection with pagination links as `link` objects:

```json
{
  "meta": {
    "totalPages": 13
  },
  "data": [
    {
      "type": "articles",
      "id": "3",
      "attributes": {
[...]

PR #1555, opened 2021-03-31
Update JSON Schema to be v1.0 compliant

### LINKED ISSUE #1554: JSON Schema for JSON:API v1.0 is incomplete or broken
Hi all. I promised to work on this, so here goes!

The `schema` file in the root of this repo is a little lacking and in some cases inaccurate.
I feel like what makes the most sense here is to make a draft PR and then talk about the changes I've made and why.

I've not fully tested it, but it should be more than enough to get discussions going!

PR to follow!

Relequestual:
These changes fix a number of issues in the schema which make it invalid or incomplete in terms of validating JSON:API response payloads.

Resolves #1554

### DIFF
--- a/schema
+++ b/schema
@@ -1,19 +1,17 @@
 {
-  "$schema": "http://json-schema.org/draft-06/schema#",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "https://jsonapi.org/schemas/spec/v1.0/draft",
+  "$comment": "The $id URI should be modified before releasing. This URI does not need to be network addressable. It is an ID only.",
   "title": "JSON:API Schema",
-  "description": "This is a schema for responses in the JSON:API format. For more, see http://jsonapi.org",
+  "description": "A JSON Schema downloaded from jsonapi.org/schema on 2020-02-22 - modified to align with the specification. Additionaly, this schema only validates RESPONSES from a request. Validating request payloads requires slightly different constraints.",
   "oneOf": [
     {
       "$ref": "#/definitions/success"
     },
     {
       "$ref": "#/definitions/failure"
-    },
-    {
-      "$ref": "#/definitions/info"
     }
   ],
-
   "definitions": {
     "success": {
       "type": "object",
@@ -37,14 +35,7 @@
         },
         "links": {
           "description": "Link members related to the primary data.",
-          "allOf": [
-            {
-              "$ref": "#/definitions/links"
-            },
-            {
-              "$ref": "#/definitions/pagination"
-            }
-          ]
+          "$ref": "#/definitions/links"
         },
         "jsonapi": {
           "$ref": "#/definitions/jsonapi"
@@ -77,25 +68,6 @@
       },
       "additionalProperties": false
     },
-    "info": {
-      "type": "object",
-      "required": [
-        "meta"
-      ],
-      "properties": {
-        "meta": {
-          "$ref": "#/definitions/meta"
-        },
-        "links": {
-          "$ref": "#/definitions/links"
-        },
-        "jsonapi": {
-          "$ref": "#/definitions/jsonapi"
-        }
-      },
-      "additionalProperties": false
-    },
-
     "meta": {
       "description": "Non-standard meta-information that can not be represented as an attribute or relationship.",
       "type": "object",
@@ -142,7 +114,7 @@
           "$ref": "#/definitions/relationships"
         },
         "links": {
-          "$ref": "#/definitions/links"
+          "$ref": "#/definitions/relationshipLinks"
         },
         "meta": {
           "$ref": "#/definitions/meta"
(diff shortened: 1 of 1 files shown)


Relequestual:
@guillemcanal thanks for your work in #1487! =]
I've taken a slightly different approach here to solve the same problem.

I'd suggest closing #1487 in favour of this PR.

guillemcanal:
@Relequestual LGTM. By the way, do you plan to create a test suite to validate every examples given in the spec against the JSONSchema ? It may be safer :)

auvipy (approved):

Relequestual:
> @Relequestual LGTM. By the way, do you plan to create a test suite to validate every examples given in the spec against the JSONSchema ? It may be safer :)

If someone wants to find a way to extract them from the spec, I'm happy to do the validation test work. Otherwise, I don't have the bandwidth immediatly.

Relequestual:
I'll try to make some time this week to double check the schema and then remove the draft status.

dgeb:
> I'll try to make some time this week to double check the schema and then remove the draft status.

@Relequestual Thank you, much appreciated!

claudenirmf:
This is a much appreciated contribution. How would you proceed with testing? I personally prefer to split a schema as large as this one, but I have not found a good solution to bundle split schemas into a single file yet. Also, do you plan to submit your tests to the repository?

Relequestual:
Sorry it has taken me a little while to get back to you here.
I'm very much snowed under in terms of what I want to get done in JSON Schema land.

> How would you proceed with testing? I personally prefer to split a schema as large as this one, but I have not found a good solution to bundle split schemas into a single file yet. Also, do you plan to submit your tests to the repository? - @claudenirmf n

My testing suggestion was related to examples in the specification. If they could be extracted, they could be tested against the schema. What language is up to you. I'd stick to what I know and a library that supports 2019-09 or above.

In terms of split files. This schema is actually pretty basic, but I still sometimes split even small schemas into multiple smaller ones.

@claudenirmf why would you need to bundle them into a single file?

Relequestual:
Realistically I'm not sure I'm going to have time to review this soon.

If someone wants to review and approve it, then please do, but I make no assurances here 😅.

claudenirmf:
> In terms of split files. This schema is actually pretty basic, but I still sometimes split even small schemas into multiple smaller ones.

Thanks for your response. You are correct, it is a basic schema. I admit that my memory of the actual schema tricked me (past me did not find JSON Schema very readable to begin with).

> @claudenirmf why would you need to bundle them into a single file?

Some applications can only handle single file schemas, either because of libraries that cannot handle references, or because you want to deliver the schema as a single file (I'm thinking of [`describedby`](https://github.com/json-api/json-api/pull/1447) here).

schema
```diff
@@ -206,16 +199,22 @@
         }
       },
       "not": {
+        "$comment": "This is what the specification requires, but it seems bad. https://github.com/json-api/json-api/issues/1553",
         "anyOf": [
-          {"required": ["relationships"]},
-          {"required": ["links"]},
-          {"required": ["id"]},
-          {"required": ["type"]}
```
bill-okara, line 213: would this then allow "id" and "type" as `attributes` member, would that violate one of the [restrictions for fields](https://jsonapi.org/format/#document-resource-object-fields)?

> a resource can not have an attribute and relationship with the same name, nor can it have an attribute or relationship named **type** or **id**.

schema
```diff
@@ -241,16 +240,22 @@
...
+                "links"
+              ]
+            }
           ],
-          "not": {
-            "anyOf": [
-              {"required": ["id"]},
-              {"required": ["type"]}
-            ]
-          },
```
bill-okara, line 253: should this be moved up one level instead of removed to satisfy the restriction

> a resource can not have an attribute and relationship with the same name, **nor can it have an attribute or relationship named type or id.**

schema
```diff
@@ -241,16 +240,22 @@
...
+            {
+              "required": [
+                "links"
+              ]
+            }
           ],
-          "not": {
-            "anyOf": [
-              {"required": ["id"]},
-              {"required": ["type"]}
```
bill-okara, line 251: should this be moved up one level instead of removed to satisfy the restriction

> a resource can not have an attribute and relationship with the same name, **nor can it have an attribute or relationship named type or id.**

Relequestual:
I hope to come back to this at some point soon! =D

schema
```diff
@@ -346,13 +366,13 @@
...
     },
-
     "error": {
       "type": "object",
       "properties": {
         "id": {
           "description": "A unique identifier for this particular occurrence of the problem.",
-          "type": "string"
+          "type": "string",
+          "$comment": "The spec doesn't specify that this MUST be a string, so this could be changed to additionally allow numbers"
```
mrevutskyi, line 375: Should we also allow `null` values?

auvipy:
any progress on this? what aboy v1.1?

Relequestual:
Closing in favour of https://github.com/json-api/json-api/pull/1600
Sorry, I don't have enough time to dig into this enough to be helpful.
It's been a while since I read the JSON:API spec. It's very much not fresh.

PR #1559, opened 2021-03-31
Add entry for drupal-jsonapi-params

d34dman:
> That's not even on the jsonapi.org page, so wouldn't have found it had I not asked.

Somebody complained in slack that they didn't find my module in json-api.org's list, so here i am trying to get it included in :)

Decoupled Drupal applications based on JSON:API is getting more popular, so might be helpful.


gabesullice (approved):
Thanks @d34dman!

PR #1560, opened 2021-04-02
Added JSON:API object converter, a Java backend converter for objects

MieskeB:


dgeb:
@MieskeB great, thank you!

PR #1561, opened 2021-04-05
Clarify reserved characters don't conflict extension member rules

lode:
This makes a similar clarification as for `@` with @-members (see further down in the diff).

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -894,7 +894,7 @@ The following characters **MUST NOT** be used in implementation and
 - U+0029 RIGHT PARENTHESIS, ")"
 - U+002A ASTERISK, "&#x2a;"
 - U+002F SOLIDUS, "/"
-- U+003A COLON, ":"
+- U+003A COLON, ":" (except after the namespace in [Extension members](#extension-members))
 - U+003B SEMICOLON, ";"
 - U+003C LESS-THAN SIGN, "<"
 - U+003D EQUALS SIGN, "="


auvipy (approved):

_format/1.1/index.md
```diff
@@ -894,7 +894,7 @@ The following characters **MUST NOT** be used in implementation and
 - U+0029 RIGHT PARENTHESIS, ")"
 - U+002A ASTERISK, "&#x2a;"
 - U+002F SOLIDUS, "/"
-- U+003A COLON, ":"
+- U+003A COLON, ":" (except after the namespace in [Extension members](#extension-members))
```
gabesullice, line 897: I think this isn't necessary because of the wording above the list of reserved characters:

> The following characters **MUST NOT** be used in implementation and profile defined member names:

IOW, this section is "reserving" these characters for use by the specification and by extensions. Therefore, they may not be used by implementations or profiles. We would have to add this exception to every character.

gabesullice (commented):
Thanks for reviewing 1.1 @lode!

_format/1.1/index.md
lode: Ah I see.

This PR was mainly to avoid confusion. Maybe not per se because it is wrong, but because one might get confused that it is allowed in one place of the spec and not in another. So maybe it is correct on that part, but it does raise some questions to me:

- Do we actually want to allow extensions to use these characters anywhere? Or do we want to protect users of crazy extension makers using these chars? I can see the use of some of these, but for example the control chars feel very unneeded.
- Practically `:` is also not allowed for extensions to use where ever they want, since it is not allowed in the namespace and not the in member after the `<namespace>:` part.
- Do we also remove the mention at the `@` char?
- Do we need a section which is also not allowed for extensions?
- What is the goal of this section if members (and namespaces, @-members, etc) all already have a strict definition of what _is_ allowed?

Sorry if I sound harsh, not meaning to do so! Just curious.

gabesullice: You definitely don't sound harsh :) I appreciate the thought you're putting into the spec!

I'm a little confused myself. I think I'm misunderstanding your concerns. We have [rules for extension members](https://jsonapi.org/format/1.1/#extension-members) defined:

> The name of every new member introduced by an extension **MUST** be prefixed with the extension’s namespace followed by a colon (`:`). The remainder of the name **MUST** adhere to the rules for implementation defined member names.

Meaning, aside from a colon immediately after the namespace, every rule for implementation and profile members also applies to extensions.

Perhaps we should move the extension member rules to the top of the members section instead placing it at the of the end of the section.

lode: I think mainly I'm confused with the goal of the [reserved characters section](https://jsonapi.org/format/1.1/#document-member-names-reserved-characters). I have the feeling we can just remove the whole section. Since both normal members, @-members and extension members are specified strictly, anything else is thus not allowed. Am I missing something?

jelhan: I understand the confusion.

Implementation and profile defined member names are limited to a specific subset of allowed characters:

> 7.8 Member names
>
> Implementation and profile defined member names [...] MUST contain only the allowed characters listed below.
>
> 7.8.1  Allowed Characters
>
> The following “globally allowed characters” MAY be used anywhere in a member name:
>
> - U+0061 to U+007A, “a-z”
> - U+0041 to U+005A, “A-Z”
> - U+0030 to U+0039, “0-9”
> - U+0080 and above (non-ASCII Unicode characters; not recommended, not URL safe)
>
> Additionally, the following characters are allowed in member names, except as the first or last character:
>
> - U+002D HYPHEN-MINUS, “-“
> - U+005F LOW LINE, “_”
> - U+0020 SPACE, “ “ (not recommended, not URL safe)

Unless I missed something this list already excludes all characters listed in "7.8.2 Reserved Characters". It seems as if these reserved characters are excluded twice. Once per not being included in the list of allowed characters and once per being included in the list of reserved characters.

@dgeb Are you able to clarify this confusion?

PR #1565, opened 2021-05-24
Fix typo with missing word

julio73:
Providing clarity with switch from "may named" to "may be named"


dgeb:
Thanks for the correction!

PR #1574, opened 2021-07-31
Add Specification Numbering and build misc.

timothystone:
Citing the specification can be cumbersome. Following RFC numbering
formats, the specification is numbered following ISO 2145 at each HEADER
 element, e.g., `h1`, `h2`, &tc. The numbers are CSS "generated" content
 and provide support for the additive focus of the specification.

The paragraph glyph is repositioned absolutely, relative to the header.

The `_config.yml` is updated to support building. `bundle` otherwise
failed on Ruby 2.7.x. Updated the exclusions file for IntelliJ.


timothystone:
Regarding the `_config.yml` change, I followed the README instructions and found that the site would not build with Ruby 2.3 installed on the Mac. I upgraded to Ruby 2.7 via homebrew—required as the Jekyll markdown gem would not compile—and encountered the same problem. Some research, i.e., SO-fu, found numerous posts regarding the vendor gem path with regards to Jekyll test files. 

Adding this line corrected the problem. The move to Ruby 2.7 appears to have generated updates to the `map` file.

timothystone:
For your convenience... what this looks like running locally is below.

image

timothystone:
The PR had incorrectly used the generated CSS file and not Sass. 

There is an argument to be made that the generated CSS files should be in the `.gitignore` file. I had initially edited the generated CSS file.

dgeb:
Thanks very much @timothystone - this should really help improve referencing the spec!

dgeb:
I had to revert this in #1579 because the styles were not scoped to only the specification pages. Numbering was shown even on the hero on the home page.

Let me know if you have the bandwidth to create a separate PR that scopes these styles, @timothystone.

timothystone:
@dgeb On it. I don't recall seeing it on the hero page, but `<homer-simpson>`D'oh!`</homer-simpson>`

dgeb:
@timothystone thank you!

PR #1580, opened 2021-08-31
Add Specification Numbering

timothystone:
Address oversight of specification numbering scope issue in  #1574.

The Liquid template now applies a CSS class to the body element. The Sass
was updated to scope the changes to the class `.spec_page`, supporting 
both 1.0 and 1.1. Confirmed the numbering does not appear on other pages 
outside the scoped class.

From the original PR:

Citing the specification can be cumbersome. Following RFC numbering
formats, the specification is numbered following ISO 2145 at each HEADER
element, e.g., `h1`, `h2`, &tc. The numbers are CSS "generated" content
and provide support for the additive focus of the specification.

The paragraph glyph is repositioned absolutely, relative to the header.

The `_config.yml` is updated to support building. bundle otherwise
failed on Ruby 2.7.x. Updated the exclusions file for IntelliJ.

Regarding the `_config.yml` change, I followed the README instructions and
found that the site would not build with Ruby 2.3 installed on the Mac.
I upgraded to Ruby 2.7 via homebrew—required as the Jekyll markdown gem
would not compile—and encountered the same problem. Some research, i.e.,
SO-fu, found numerous posts regarding the vendor gem path with regards to
Jekyll test files.

Adding this line corrected the problem. The move to Ruby 2.7 appears to
have generated updates to the map file.


timothystone:
@dgeb Thanks for the heads up. This should address the issue.

dgeb:
@timothystone thank you, this looks great now! 🎉

PR #1586, opened 2021-09-26
add jsonapi-unwrapper to Ruby client libraries

Sonberg:
Hi! I needed a small lib to handle JSON:API payloads. Didn't find one that fitted my needs, so i created one. I would like to share it with the community.


dgeb:
Thanks @Sonberg!

PR #1590, opened 2021-11-03
Update history

sandstrom:
- Reorder, since the history part doesn't seem super-relevant today
- Update example library (active model serializers is basically dead)
- Update links to https

### DIFF
--- a/about/index.md
+++ b/about/index.md
@@ -35,31 +35,9 @@ The following emeritus editors are no longer active:
 
 ## <a href="#history" id="history" class="headerlink"></a> History
 
-JSON:API was originally drafted by [Yehuda Katz](http://twitter.com/wycats)
+JSON:API was originally drafted by [Yehuda Katz](https://twitter.com/wycats)
 in May 2013. This first draft was extracted from the JSON transport
-implicitly defined by [Ember](http://emberjs.com/) Data's REST adapter.
-
-In general, Ember Data's goal is to eliminate the need for ad-hoc code
-per application to communicate with servers that communicate in a
-well-defined way.
-
-Some servers, like Firebase, Parse and CouchDB already define strict
-communication protocols for clients, and were good fits for Ember Data.
-In contrast, servers written in Rails, Node, and Django tend to be
-written in a "REST-style" but lack the precision necessary for drop-in
-client code.
-
-The REST Adapter in Ember Data implicitly defined a protocol that
-custom servers could implement to get a drop-in client for all of their
-resources. [ActiveModel::Serializers][1] is a proof-of-concept library
-for Rails that implemented the serialization format expected by Ember
-Data.
-
-[1]: https://github.com/rails-api/active_model_serializers
-
-Record creation, update, and deletion was defined implicitly by the
-Ember Data library and was close to conventions already in wide use by
-Rails, Django and Node developers.
+implicitly defined by [Ember](https://emberjs.com/) Data's REST adapter.
 
 The goals of the media type are to balance:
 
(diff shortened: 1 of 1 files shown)


auvipy (approved):

about/index.md
```diff
@@ -35,13 +35,25 @@ The following emeritus editors are no longer active:
...
-JSON:API was originally drafted by [Yehuda Katz](http://twitter.com/wycats)
+JSON:API was originally drafted by [Yehuda Katz](https://twitter.com/wycats)
 in May 2013. This first draft was extracted from the JSON transport
-implicitly defined by [Ember](http://emberjs.com/) Data's REST adapter.
+implicitly defined by [Ember](https://emberjs.com/) Data's REST adapter.
 
-In general, Ember Data's goal is to eliminate the need for ad-hoc code
-per application to communicate with servers that communicate in a
-well-defined way.
+The goals of the media type are to balance:
```
jelhan: The media type is only an implementation detail in my perspective. I fear it's confusing to introduce it in the history section. I would rephrase them as goals of the specification.

```suggestion
The goals of the JSON:API specification are to balance:
```

about/index.md
```diff
@@ -35,13 +35,25 @@ The following emeritus editors are no longer active:
...
+
+* A generic media type that can work across a broad set of use cases,
+  including the generally used relationship types
+* Similarity to existing server-side framework practices (and human
+  readability for debugging)
+* Ease of implementation on the server side
+* Ease of implementation on the client side
```
jelhan, line 49: I'm not sure if I would phrase the goals like this. It also conflicts, which the marketing statements on the homepage, which focus on

1. avoid bike-shedding about API design,
2. efficient caching, and
3. avoid unnecessary network requests.

> If you’ve ever argued with your team about the way your JSON responses should be formatted, JSON:API can be your anti-[bikeshedding](http://bikeshed.org/) tool.
> 
> By following shared conventions, you can increase productivity, take advantage of generalized tooling, and focus on what matters: your application.
> 
> Clients built around JSON:API are able to take advantage of its features around efficiently caching responses, sometimes eliminating network requests entirely.

Personally I would phrase the goals (looking back) as the following:

1. Standardize REST API design to
    1. avoid bike-shedding about API design, which is not specific to the business requirements, and
    2. enable development of compatible client- and server-side libraries.
2. Built on-top of established REST API patterns to enable incremental adoption and a flat learning curve.
3. Reduce the risk of write conflicts on REST API design level (e.g. `PATCH` for updating resources and partial updates of to-many relationships using relationship links).
4. Support advanced use cases like polymorphic relationships.
5. Allow clients to fetch all needed data in a single request to avoid unnecessary network requests (under-fetching).
6. Allow clients to request a subset of resource fields to avoid sending unnecessary data over the wire (over-fetching).

jelhan (commented):
@sandstrom Thanks a lot for starting to improve the history section. I think it needs to be updated in general to cover the current state more closely. In my perspective JSON:API specification is not that much focused on Ember and Rails anymore as it was in the early days. I think this should be reflected in the history section as well.

@dgeb Would be great to hear your opinion on it.

about/index.md
sandstrom: Sure, I just copied text around. I didn't write any of that.

What you suggest sounds good! I've enabled maintainers to make copies in this fork, so you can make whatever changes you want.

Ryoc45 (approved):

sandstrom:
@jelhan @dgeb Is this still relevant, or should I cancel the PR?

PR #1591, opened 2021-11-03
Mention blueprinter ruby lib

sandstrom:


jelhan:
Are you sure that this library implements the JSON:API specification? Maybe I fully missed it, but I haven't found the specification mentioned at all in the docs. Also the examples do not seem to use JSON:API.

sandstrom:
@jelhan Yes, you're right. 

It was so long ago that I don't remember why I suggested this, but I've looked through the repo and can't find anything about JSON API either.

Issue #1502, opened 2020-08-13
top-level links for collection question


freddrake:
I want to be sure I understand how top-level links should be generated for a request for a collection.

My current understanding:
* the **self** link *should not* include query parameters related to pagination, filtering, sorting, sparse field sets, or compound documents.
* the pagination links *should* include generated parameters for pagination, and pass along parameters for filtering, sorting, sparse field sets and compound documents as received.

Does this correspond with the intent of the specification authors?  Or is the client responsible for mixing in parameters related to any of those aspects for pagination links?

dgeb:
The `self` link at the top-level should include whatever params are needed to return an identical result. Top-level `self` should be considered a link to the current document. So if the current document contains a collection that is filtered or paginated, those params should be included in the `self` link.

Similarly, paginated links (`next`, `prev`, etc.) should step through the _current_ collection and thus should also include whatever params are present on `self`.

freddrake:
Thanks, @dgeb!  This is very helpful.

Was this covered in the specification in some way that I missed, or will clarifications be added (at least for 1.1)?

dgeb:
Glad to help! I'll reopen so we can review / discuss if clarification is needed in the spec.

freddrake:
Reviewing the 1.0 spec, the description of the top-level self link here:

> https://jsonapi.org/format/#document-links

does not spell out that the link refers specifically to the *document*, and should be constructed to reproduce the current rendering.  The [Top Level](https://jsonapi.org/format/#document-top-level) section does indicate the self link is for the generated response; perhaps that should be considered sufficient, though this portion of the spec doesn't include the exact term "self link", which does yield the other description of the top-level self link.

Perhaps the difficulty is that the top-level self link is described in two places?

freddrake:
Two months and no response to my analysis?

@dgeb, when you reopened for discussion, I was hoping not to be talking to myself.

dgeb:
@freddrake in 1.1 isn't the top-level `self` link only specifically described in https://jsonapi.org/format/1.1/#document-top-level ? That looks pretty accurate and sufficient imo.

I'm not clear how it contradicts any of the general language about `links` contained in https://jsonapi.org/format/1.1/#document-links

freddrake:
[...]

PR #1604, opened 2022-01-30
Clarify query parameters for top-level self link

### LINKED ISSUE #1030: Response for filtered request without result
The specs and the recommendations do not mention anything about responses to filtered requests specifically. I would think that they are exactly the same as the responses for a normal `GET` request to an endpoint requesting a collection of resources.

Is this correct? I would like to see a sentence added stating this, as it would make it easier and more specific. This should be possible to be added to `1.0` imo as it does not change behaviour but rather clarifies indirectly implied behaviour.

Why is this good? I think that the less additional thinking you have to do, when reading specs, the better, so a statement about how to deal with this would be very good.

jelhan:
In the past we have seen several questions regarding the top-level `self` link and if it should contain query parameters supplied by the client, which affected the response document. Adding an explicit note about it should help to prevent such confusion in the future.

Closes #1030 #1502

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -319,7 +319,7 @@ The top-level [links object][links] **MAY** contain the following members:
   Schema) for the current document.
 * [pagination] links for the primary data.
 
-> Note: The `self` link should contain the query parameters provided by the client
+> Note: The `self` link in the top-level `links` object must contain the query parameters provided by the client
 > to generate the current response document. This includes but is not limited to
 > query parameters used for [inclusion of related resources][fetching resources],
 > [sparse fieldsets][fetching sparse fieldsets], [sorting][fetching sorting],
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -319,9 +319,12 @@ The top-level [links object][links] **MAY** contain the following members:
   Schema) for the current document.
 * [pagination] links for the primary data.
 
-> Note: The `self` link in the top-level `links` object must contain the query parameters provided by the client
-> to generate the current response document. This includes but is not limited to
-> query parameters used for [inclusion of related resources][fetching resources],
+> Note: The `self` link in the top-level `links` object allows a client to
+> refresh the data represented by the current response document. The client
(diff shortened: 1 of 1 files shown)


auvipy (commented):
looks good to me

_format/1.1/index.md
```diff
@@ -319,6 +319,12 @@ The top-level [links object][links] **MAY** contain the following members:
   Schema) for the current document.
 * [pagination] links for the primary data.
 
+> Note: The `self` link should contain the query parameters provided by the client
```
freddrake: I'd suggest changing

``The `self` link``

to something like

``The `self` link in the top-level `links` object``

because it's already very easy to lose the specific context of a requirement based on headers even just a few paragraphs away.

Also, ``should`` should be ``**MUST**``, because it's a requirement that the self link reflects the URL that generated the document.

jelhan: Thanks a lot for that valuable feedback. Both are very good points.

```suggestion
> Note: The `self` link in the top-level `links` object must contain the query parameters provided by the client
```

auvipy (approved):

_format/1.1/index.md
freddrake: So why `must` instead of `**MUST**`?  The later is specifically how a compliance point is indicated throughout the specification.

jelhan: I don't see this note as part of the specification. It is an additional explanation provided with the spec. Due to this different intent it should use normal language rather than normative specification terms. This is inline with existing notes in the spec.

jelhan:
Reflecting on the valuable feedback from @freddrake, I noticed that my first proposal was too much focused on the _what_. Existing notes in the specification often explain the intent behind a normative statement. I have rewritten the note to focus more on the _why_. Would be great if you could have another look on it.

freddrake (approved):
Nice!  Thanks for working on this.

auvipy (approved):

dgeb:
Thanks @jelhan for putting this together. I think it's a useful clarification of a somewhat nuanced element.

PR #1607, opened 2022-02-04
Clarify processing rules for malformed `ext` and `profile` members of JSON:API object

jelhan:
The applied extension and profiles must be provided as `ext` and `profile` media type parameters of `Content-Type` header. Additionally they may be listed as `jsonapi.ext` and `jsonapi.profile` member of a JSON:API document.

This opens up the possibility of having different values in both places. So far the specification has not defined processing rules for such malformed documents.

This clarifies that a client and server must honor the `Content-Type` header in case of a conflict. Additionally it clarifies that they should not reject the document but ignore the invalid `jsonapi.ext` and/or `jsonapi.profile` members.

Ignoring them in case of a conflict rather than rejecting the document allows server and clients to _not_ process them at all - even if present. This avoids computing overhead just for the sake of validating the document.

Closes #1359

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -825,7 +825,7 @@ The jsonapi object **MAY** contain any of the following members:
 * `profile` - an array of URIs for all applied [profiles].
 * `meta` - a [meta] object that contains non-standard meta-information.
 
-Clients and servers **MUST NOT** use a `ext` or `profile` member for content
+Clients and servers **MUST NOT** use an `ext` or `profile` member for content
 negotiation. Content negotiation **MUST** only happen based on media type
 parameters in `Content-Type` header.
 
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -825,9 +825,9 @@ The jsonapi object **MAY** contain any of the following members:
 * `profile` - an array of URIs for all applied [profiles].
 * `meta` - a [meta] object that contains non-standard meta-information.
 
-Clients and servers **MUST** ignore a `ext` or `profile` member if its value
-conflicts with `ext` or `profile` media type parameter in the `Content-Type`
-header.
+Clients and servers **MUST NOT** use a `ext` or `profile` member for content
+negotiation. Content negotiation **MUST** only happen based on media type
+parameters in `Content-Type` header.
 
 A simple example appears below:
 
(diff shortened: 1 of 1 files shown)


auvipy (approved):

dgeb:
I agree entirely with the sentiment behind this PR, which is that content negotiation should occur using headers (`Content-Type` and `Accept`) rather than the members of the `jsonapi` object. And I would also say that validation of this object is entirely optional. But I'm not sure that the normative change in this PR is quite right. 

I believe that the guarantees we want to provide are:
* These members, _if present_, MUST match the corresponding media type parameters.
* Because these members are entirely optional they MUST NOT be used for content negotiation.

I think this combination of guarantees should prevent unnecessary validation while still providing a reasonable assumption of correctness if these members are present. 

@jelhan what do you think? Would this combination of normative changes address your concerns?

_format/1.1/index.md
```diff
@@ -825,6 +825,10 @@ The jsonapi object **MAY** contain any of the following members:
 * `profile` - an array of URIs for all applied [profiles].
 * `meta` - a [meta] object that contains non-standard meta-information.
 
+Clients and servers **MUST** ignore a `ext` or `profile` member if its value
+conflicts with `ext` or `profile` media type parameter in the `Content-Type`
+header.
```
jelhan:
```suggestion
Clients and servers **MUST NOT** use a `ext` or `profile` member for content
negotiation. Content negotiation **MUST** only happen based on media type
parameters in `Content-Type` header.
```

jelhan (commented):
Yes. I think your proposal makes sense. Enforcing validation was not my intent. I provided a potential rewording as review comment.

To be honest I'm not sure what value `ext` and `profile` members of `jsonapi` provide. They seem to only duplicate information already available in `Content-Type` header. Not sure if I missed a relevant discussion about the use cases they address.

_format/1.1/index.md
```diff
@@ -825,6 +825,10 @@ The jsonapi object **MAY** contain any of the following members:
 * `profile` - an array of URIs for all applied [profiles].
 * `meta` - a [meta] object that contains non-standard meta-information.
 
+Clients and servers **MUST NOT** use a `ext` or `profile` member for content
```
dgeb: Syntax

```suggestion
Clients and servers **MUST NOT** use an `ext` or `profile` member for content
```

dgeb (approved):
Thanks for the adjustments, @jelhan. Another good clarification!

Issue #1521, opened 2020-11-11
Missing step in definition of `relationships`


chrylis:
In both 1.0 and the draft 1.1, the *Relationships* section says:

> The value of the `relationships` key **MUST** be an object (a “relationships object”).

So far, so good; there should be an object listing some relationships. However, the spec then goes on to define the contents of _a relationship_ (not relationships) object, without specifying the relation between the two levels in the hierarchy.

In particular, the spec should define the allowable "relationships object". Presumably the keys are the name of the relationship. Are the values of mixed cardinality? A single "relationship object"? What is the allowable way to handle to-many relationships?

The example document provided, under `relationships.tags`, has a single-valued "relationship object" where the resource linkage is multi-valued, but a "link object" is not permitted to have multi-valued hrefs or equivalent, so is there really an asymmetry built into the spec such that "relationship as link" may only be single-valued but "relationship as data" may be multi-valued?

freddrake:
More clarity could certainly be added in sentence following the one you quote:

> Members of the relationships object (“relationships”) represent references from the resource object in which it’s defined to other resource objects.

There is a note pointing to the [Fields](https://jsonapi.org/format/1.1/#document-resource-object-fields) and [Member Names](https://jsonapi.org/format/1.1/#document-member-names) sections, which are helpful in clarifying the naming of properties in the *relationships object*.

An explicit statement that the value of each property in the *relationships object* is a *relationship object* would probably be sufficient.  Would this have helped?

The first example in [Compound Documents](https://jsonapi.org/format/1.1/#document-compound-documents) shows a to-many relationship for the `comments` field.

jelhan:
Thanks a lot for reporting. This is addressed in #1608.

dgeb:
Closed via #1608. Thanks @jelhan!

PR #1608, opened 2022-02-06
define relationship object explicitly

jelhan:
The specification uses the term "relationship object" without defining it first. It also does not specify that the name of a relationship is given by its key in the relationships object. This change addresses both.

This has been discussed in length in #946. But it seems as if that merge request got stale _after_ a consensus has been reached. I'm implementing the result of that discussion here.

Closes #946 and #1521

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -475,11 +475,14 @@ as attributes. Instead, [relationships] **SHOULD** be used.
 ##### <a href="#document-resource-object-relationships" id="document-resource-object-relationships" class="headerlink"></a> Relationships
 
 The value of the `relationships` key **MUST** be an object (a "relationships
-object"). Members of the relationships object ("relationships") represent
-references from the [resource object][resource objects] in which it's defined to other resource
-objects.
+object"). Each member of the relationships object ("relationships") represents
+a "relationship", i.e., a reference from the [resource object][resource objects]
+in which they are defined to other resource objects.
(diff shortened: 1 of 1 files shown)


_format/1.1/index.md
```diff
@@ -475,12 +475,15 @@ as attributes. Instead, [relationships] **SHOULD** be used.
 ##### <a href="#document-resource-object-relationships" id="document-resource-object-relationships" class="headerlink"></a> Relationships
 
 The value of the `relationships` key **MUST** be an object (a "relationships
-object"). Members of the relationships object ("relationships") represent
-references from the [resource object][resource objects] in which it's defined to other resource
-objects.
+object"). Each member of the relationships object ("relationships") represents
+a "relationship", i.e., a reference from the [resource object][resource objects]
+in which they are defined to other resource objects.
```
dgeb: I'd like to try to define this section a bit more crisply. What do you think of this?

```suggestion
object"). Each member of a relationships object ("relationships") represents
a "relationship" from the [resource object][resource objects]
in which it has been defined to other resource objects.
```

dgeb (commented):
Thanks for seeing this through @jelhan. It seems very close, although I think one section needs a little refinement.

_format/1.1/index.md
jelhan: Thanks a lot. Reads much better.

PR #1611, opened 2022-02-16
Add implementation jsonapi-browser.

masterT:
Add Playground [jsonapi-browser](https://github.com/masterT/jsonapi-browser).

Demo: https://masterT.github.io/jsonapi-browser

### DIFF
--- a/implementations/index.md
+++ b/implementations/index.md
@@ -65,7 +65,7 @@ isomorphic ActiveRecord clone that issues JSON:API requests instead of SQL and i
 * [DatX](https://github.com/infinum/datx) is an opinionated data store for use with the MobX state management library that adds JSON:API support with [datx-jsonapi](https://github.com/infinum/datx/tree/master/packages/datx-jsonapi) mixin.
 * [jsonapi-fractal](https://github.com/andersondanilo/jsonapi-fractal) JSON:API Serializer inspired by Fractal (PHP)
 * [ts-japi](https://github.com/jun-sheaf/ts-japi) - A zero-dependency, highly-modular, js/ts-friendly, recursible, framework-agnostic library for serializing data to the JSON:API specification. Serializes the entire specification.
-* [drupal-jsonapi-params](https://github.com/d34dman/drupal-jsonapi-params) A library for building query parameters when connecting with Drupal CMS's JSON:API. 
+* [drupal-jsonapi-params](https://github.com/d34dman/drupal-jsonapi-params) A library for building query parameters when connecting with Drupal CMS's JSON:API.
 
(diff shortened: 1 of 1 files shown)


masterT:
Let's wait until it's more stable.

jelhan:
I played around with the tool a little bit. I think it has great potential. Do you have a list of features you are planning to add?

It seems that development has stopped on [JSON:API Document Viewer](http://codeme.lt/json-api-document-viewer/). I'm wondering if we could maybe incorporate a visualization of the resources in a response in a similar way. I think a combination of a browser to navigate between resources together with a visualization of the resources would be very nice.

Maybe even a feature to create, update and delete resources could be added. An editor to create and modify a JSON:API document might be very helpful for teaching and testing as well.

masterT:
Thanks for the feedback! 

I don't have a clear plan for this project, but I have some ideas.

## Add HTML representation for the JSON:API document fetched

Reference: 
- https://github.com/masterT/jsonapi-browser/issues/2

Maybe be able to see the JSON:API document as HTML similar to
- [the Ruby gem administrate](https://administrate-demo.herokuapp.com/)
- [the Vue.js application Vandal UI](https://www.graphiti.dev/guides/vandal)

Ex: Representing a JSON:API document with a resource as primary data:

Screen Shot 2022-02-23 at 10 57 23


Ex: Representing a JSON:API document with resource collection as primary data:

Screen Shot 2022-02-23 at 10 59 57

## Support setting custom HTTP headers

Set custom HTTP headers (such as [Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization)). 

---

> Maybe even a feature to create, update and delete resources could be added. An editor to create and modify a JSON:API document might be very helpful for teaching and testing as well.

I think it could be nice to edit and _PATCH_ a fetched document using an HTML form. I might be possible because the "shape" of the document is already known.

For creating resources I think that without a proper specification of the resource it's not going to be easy to implement, nor pleasant for the user. The only solution I see is a form with a textarea to capture the JSON to send. 😬

That being said I'm working on creating a specification for API implementing JSON:API. The goal is to have a machine-readable specification of an API (something similar to [RAML](https://raml.org/), [swagger/OAS](https://swagger.io/specification/), [API blueprint](https://apiblueprint.org/)). With such a specification I could be possible to implement a tool to perform all the JSON:API operations. 🙂 My specification is at a very early stage, but if you want to sneak peek I created a [gist](https://gist.github.com/masterT/f808382e030b8bd7f7675608a6792d7c).

---

I also have the idea to create an [electron application](https://www.electronjs.org/) to "bypass" the _CORS_, this way it could work on any API. This way it would be possible to set cookies (cross domain).

PR #1615, opened 2022-02-21
only client-requested sparse fieldsets are exception to full linkage requirement

### LINKED ISSUE #1614: Clarify exception to full linkage requirement and server-initiated sparse fieldsets
I'm running into what I think is an unresolved area of ambiguity as it relates to json:api's spec constraints around:

* Full linkage in compound documents, and
* Server-initiated sparse fieldsets.

Full linkage between a response's primary data and included documents is required in all cases, except:

> ...when relationship fields that would otherwise contain linkage data are excluded via [sparse fieldsets](https://jsonapi.org/format/#fetching-sparse-fieldsets).

In the sparse fieldsets section, we learn servers may respond by applying its own sparse fielset rule, in the absence of one being sent in the request query string:

> If a client does not specify the set of [fields](https://jsonapi.org/format/#document-resource-object-fields) for a given resource type, the server MAY send all fields, a subset of fields, or no fields for that resource type.

Does this mean, then, that the server may forego full linkage if it has opted to respond with a subset of fields?

If not, I think we need to add some sort of requirement stating that the obligation to send full linkage cannot be foregone by server-initiated sparse fielsets. (My invented term, but I think we need to call it something for purposes of this discussion.)

jelhan:
The current wording could be misunderstood that also fields excluded by the server if client does _not_ request a sparse fieldset are an exception for full linkage requirement. This changes clarifies that only exception is if a client _explicilty_ requested a sparse fieldset, which does not include a relationship required for full linkage.

Closes #1614

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -623,7 +623,8 @@ included without a direct or indirect relationship to the document's primary
 data.
 
 The only exception to the full linkage requirement is when relationship fields
-that would otherwise contain linkage data are excluded via [sparse fieldsets](#fetching-sparse-fieldsets).
+that would otherwise contain linkage data are excluded due to
+[sparse fieldsets](#fetching-sparse-fieldsets) requested by the client.
 
 A complete example document with multiple included relationships:
 


bradjones1:
I think this does address the question I had in #1614. I might recommend using "specified by" instead of "requested by" the client, or even "specified in the client request"? But I think it's OK as written, too, if you have a strong feeling about it.

jelhan:
> I might recommend using "specified by" instead of "requested by" the client, or even "specified in the client request"? But I think it's OK as written, too, if you have a strong feeling about it.

I don't have strong feelings about. Personally I feel it could be confusing to use the term "specify" in a specification for something else than the specification itself.

But I'm not a native speaker. Will leave this one for editors to decide.

dgeb:
I think this wording is fine. The important clarification here is that the client is in control of the request.

auvipy:
can we expect a RC 4 soon? it's been 16 months a lot of PR merged in the mean time

jonatanlr98 (commented):
links": { "self": "http://example.com/posts" }

jonatanlr98:
links": { "self": "http://example.com/posts" }

jonatanlr98 (changes requested):
links": { "related": { "href": "http://example.com/articles/1/comments", "meta": { "count": 10 } } }

bradjones1:
@jonatanlr98 This is merged and your comments are without context. Open a new issue or PR if you have issues with this.

PR #1616, opened 2022-02-23
Clarify that server should response with 400 Bad Request in both cases

### LINKED ISSUE #1545: Should this query parameter 400 response be OR as opposed to AND?
> If a server encounters a query parameter that does not follow the naming conventions above, **and** the server does not know how to process it as a query parameter from this specification, it MUST return 400 Bad Request.

Should that be OR rather than AND?

Otherwise, I read it as, if the query parameter doesn't follow naming conventions but the server DOES know how to process it as a query parameter from this specification, then the behaviour is undefined.

I find this paragraph a little bit cryptic.

Is there a situation where the query parameter may not follow the naming convention but the server might still know how to process it as a query parameter from the spec?

What was the intent here?

jelhan:
This has been discussed in #1545. The change was already approved by two editors of the spec: @dgeb and @gabesullice But seems as no one has opened a PR yet.

Thanks a lot to @Relequestual for spotting this one!

Closes #1545

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -2105,7 +2105,7 @@ It is **RECOMMENDED** that a capital letter (e.g. camelCasing) be used to
 satisfy the above requirement.
 
 If a server encounters a query parameter that does not follow the naming
-conventions above, and the server does not know how to process it as a query
+conventions above, or the server does not know how to process it as a query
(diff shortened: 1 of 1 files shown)


auvipy (approved):

dgeb (approved):
Thanks for addressing this, @jelhan!

PR #1620, opened 2022-03-25
Update "Value is too short" detail description

isBatak:
Let's make this world a better place for everyone :D
https://twitter.com/sqiouyilu/status/1495971335578832898


dgeb:
Haha, nice @isBatak! This is definitely more appropriate. Perhaps _one_ character would be an even better minimum? https://en.wikipedia.org/wiki/A_Mart%C3%ADnez_(broadcast_journalist)

PR #1621, opened 2022-03-26
Server should respond with 409 even if only type or id does not match endpoint

### LINKED ISSUE #1549: Update MUST 409 clause should be OR rather than AND
> A server MUST return 409 Conflict when processing a PATCH request in which the resource object’s type **and** id do not match the server’s endpoint.
https://jsonapi.org/format/1.1/#crud-updating-responses-409

This should read...

> A server MUST return 409 Conflict when processing a PATCH request in which the resource object’s type **or** id do not match the server’s endpoint.

Otherwise, the implication reads as only if both are true must you return a 409, whereas I expect the intent is if either are true, you must return a 409.

jelhan:
The server should respond with `409 Conflict` also of only `type` or `id` does not match the used. Using _and_ wasn't precise therefore.

General speaking the server should respond with `409 Conflict` if resource object included in a `PATCH` request is _not_ the resource represented by the used endpoint.

This issue was reported by @Relequestual in #1549. He also proposed the fix implemented here.

Closes  #1549

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1787,7 +1787,7 @@ update a resource if that update would violate other server-enforced
 constraints (such as a uniqueness constraint on a property other than `id`).
 
 A server **MUST** return `409 Conflict` when processing a `PATCH` request in
-which the resource object's `type` and `id` do not match the server's endpoint.
+which the resource object's `type` or `id` do not match the server's endpoint.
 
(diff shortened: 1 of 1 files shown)


dgeb (approved):
Looks like a good refinement. Thanks @jelhan!

jonatanlr98 (approved):

PR #1624, opened 2022-05-03
Add implementation kt.jsonapi

freddrake:
This is a server-side Python library for response generation using the Zope Component Architecture to configure adaptation from application objects to JSON:API serialization adapters.


dgeb:
Thanks @freddrake!

PR #1629, opened 2022-06-13
Port "Better multiple related resources request example" to v1.1

jelhan:
This ports the improvements to multiple related resources request example, which were merged to v1.0 in #1227 over to v1.1 as well.


dgeb (approved):
Thank you @jelhan!

Issue #1584, opened 2021-09-23
@-members question: treatment in objects nested under attributes?


freddrake:
Version 1.1 of the specification is clear about the treatment of @-members contained in the JSON:API-defined objects of a document, but it's unclear how they should be handled in objects nested under `attributes` (which are not JSON:API-defined, but are defined entirely by the types of the application.

For example, in this document:

~~~~~~ json
{"data": {
   "type": "person",
   "id": "42",
   "attributes": {
     "@at-thing": "ignore this",
     "address": {
       "@comment": "Would be nice to live here!",
       "street": ["123 Main St.", "Suite 24"],
       "city": "Someplace Warm",
       "state": "CA",
       "postcode": "90120"
   },
   "links": {
     "self": "/api/people/42"
   }
 },
 "links": {
   "self": "/api/people/42?include[person]=address"
 }
}
~~~~~~

the `@at-thing` member should clearly be ignored, but should `@comment` be dropped as non-JSON:API data, or should it be passed along to the application as part of the `address` field?

lode:
Just chiming in here maybe without a clear answer.

What do you mean with 'should clearly be ignored'? As I understand at-members can be added so some consumer of the jsonapi document can use them. Without anyone using them, we can just as well not document their usage.

I would say that middleware that doesn't want to consume the at-members should ignore it where ever it is, but do pass it to a deeper layer which might decide to do something with that data.

But maybe that is a too much simplified way of seeing it? I don't know you use-case.

freddrake:
I'd certainly agree that middleware should ignore-but-pass.

What I'm working on right now is input validation (intended for the server side), which has no need for anything other than the JSON:API content.  For every object defined by the JSON:API specification, I'm pulling out the @-members, validating the rest, and then adding the @-members back in (with those machinations applied because of how the lower-level validation library works).

But the specification is unclear regarding how @-members should be treated down inside the values in `attributes`.

jelhan:
That's a great question. I think it comes down to how the spec treats values of attributes. Should it be agnostic about the value - even if it's an object or an array? Or should it enforce constraints on them?

I would strongly argue that the spec should leave the value of an attribute alone. But sadly it does not:

> Complex data structures involving JSON objects and arrays are allowed as attribute values. However, any object that constitutes or is contained in an attribute MUST NOT contain a relationships or links member, as those members are reserved by this specification for future use.

[...]

Issue #1553, opened 2021-03-31
Spec constraints entities attributes


Relequestual:
I'm creating an updated JSON Schema for the 1.0 spec, and re-read this section of the spec, which is similar / the same in v1.1 in relation to `attributes`...

> Complex data structures involving JSON objects and arrays are allowed as attribute values. However, any object that constitutes or is contained in an attribute **MUST NOT contain a `relationships` or `links` member, as those members are reserved by this specification for future use.**

Bluntly, this is wrong. It's unreasonable to place constraints on the attributes of entities that you have no control over.

Say, for example, my system represents patient data, and those patients have relationships which I express in an object under a `relationships` key. According to the spec, that entity can't be legally represented in the JSON:API spec format.

> reserved by this specification for future use

Making use of or applying semantics to specific attributes in data you have no control over, is asking for trouble.

Could someone please explain to me the motivation behind this specific section and language, please?

jelhan:
It was added as #653 without further explaining or discussing that limitation. I fully agree that the specification should be agnostic about the value of an attribute.

jelhan:
The constraints has been removed in #1630. v1.1 will be agnostic about attribute values and do not put any limitations on them. 🎉

Issue #1367, opened 2019-02-10
Are @-members really allowed everywhere?


lode:
Right now the spec states:

> @-Members **MAY** appear anywhere in a JSON:API document.

I can understand they can appear in places like these:

```json
{
	"@ignore": "foo",
	"data": {
		"@ignore": "foo",
		"type": "user",
		"id": "42",
		"attributes": {
			"@ignore": "foo",
			"foo": "bar"
		},
		"relationships": {
			"friend": {
				"@ignore": "foo",
				"data": {"type": "user", "id": "24"}
			}
		}
	}
}
```

Some others I would find them weird though:

```json
{
	"links": {
		"@ignore": "foo",
		"self": "https://..."
	},
	"data": {
		"type": "user",
		"id": "42",
		"relationships": {
			"@ignore": "foo",
			"friend": {
				"data": {"type": "user", "id": "24"}
			}
		}
	},
	"included": [
		{"@ignore": "foo"},
		{"type": "user", "id": "24", "attributes": {"foo": "bar"}}
	]
}
```

Or:

```json
{
	"errors": [
		{"@ignore": "foo"},
		{
			"links": {
				"@ignore": "foo",
				"about": "https://..."
			},
			"source": {
				"@ignore": "foo",
				"parameter": "title"
			}
		}
	]
}
```

Is this anyway the case? Does the spec leave no restriction to where they may appear at all?

I read the discussion in https://github.com/json-api/json-api/pull/995 about where it should be allowed in. What I don't read there is some concrete examples though, only a generic discussion about that JSON-LD needs it in deep-down places. Now I'm not an expert in JSON-LD, but the places where I find it weird do have something in common which makes me feel it is weird to allow them there: it isn't clear what they describe.

So for example `@ignore` (at root level) describes the document, `data.@ignore` describes the resource, `data.relationships.*.@ignore` describe the relationship.

But `links.@ignore` describes the next link but not all? Or `data.relationships.@ignore` the next relationship but not all? And `included.#.@ignore` in a further empty resource says me nothing, idem for `errors.#.@ignore` in a further empty error object.

I can imagine the spec being changed to state it is not allowed:

1. as the only member of an object inside an array (`data.#.@ignore`, `included.#.@ignore` and `errors.#.@ignore`)
2. at root level of objects with a specific value (`links.@ignore`, `relationships.@ignore` and `errors.#.source.@ignore`), but do allow there were the values have a more free format (`data.attributes.@ignore`)

[...]

PR #1630, opened 2022-06-13
drop constraints on attribute values

jelhan:
This drops any constraints on attribute values as proposed in #1553. Most noticeable the requirement hat an object used as attribute value (even if nested) must not have a key `links` or `relationships`.

There is no need anymore to restrict these for future usage in the specification itself. All potential use cases could be solved by using `links` or `relationships` members of the JSON:API resource object. Or by an extension registering additional members on the JSON:API resource object.

Dropping this requirement simplifies the teaching of JSON:API. We could state that the JSON:API specification is agnostic about attribute values, which wasn't true before.

I will create a complementary merge request to drop the requirement to ignore keys starting with `@` in objects used as attribute values as proposed in #1367 and #1584. It is closely related to this one. But I feel that it is easier to discuss them separately. Especially if there are discussion needs in regards to JSON-LD.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -462,12 +462,8 @@ The value of the `attributes` key **MUST** be an object (an "attributes
 object"). Members of the attributes object ("attributes") represent information
 about the [resource object][resource objects] in which it's defined.
 
-Attributes may contain any valid JSON value.
-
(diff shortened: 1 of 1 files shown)


dgeb (approved):
After discussing this in several issues, we've reached consensus on this.

PR #1636, opened 2022-06-25
docs: be more specific about behavior of empty value for sparse fieldsets

ThorstenSuckow:
The wording for the use case that a client submits the `fields[TYPE]`-parameter with an empty value is a little bit fuzzy given the availability of the key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document, indicated by a bold font and uppercase spelling.

If an empty value is understood as advising the endpoint to not include any fields at all for the resource object with the given type, I'd propose to use the key word **MUST NOT** in this section. 
Or is the lowercased `should not` to be interpreted as the keyword **SHOULD NOT**, which would mean that - in the end - this is just a recommendation for the endpoint's behavior and equally to omitting the `fields[TYPE]` parameter at all?

### DIFF
--- a/_format/1.0/index.md
+++ b/_format/1.0/index.md
@@ -1000,7 +1000,8 @@ response on a per-type basis by including a `fields[TYPE]` parameter.
 
 The value of the `fields` parameter **MUST** be a comma-separated (U+002C
 COMMA, ",") list that refers to the name(s) of the fields to be returned.
-An empty value indicates that no fields should be returned.
+An empty value indicates that an endpoint **MUST NOT** return any [fields] 
+in resource objects of that type in its response.
 
 If a client requests a restricted set of [fields] for a given resource type,
 an endpoint **MUST NOT** include additional [fields] in resource objects of
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1318,7 +1318,8 @@ response on a per-type basis by including a `fields[TYPE]` query parameter.
 
 The value of any `fields[TYPE]` parameter **MUST** be a comma-separated (U+002C
 COMMA, ",") list that refers to the name(s) of the fields to be returned.
-An empty value indicates that no fields should be returned.
+An empty value indicates that an endpoint **MUST NOT** return any [fields] 
+in resource objects of that type in its response.
 
 If a client requests a restricted set of [fields] for a given resource type,
 an endpoint **MUST NOT** include additional [fields] in resource objects of


jelhan:
Thanks a lot for this contribution.

> Or is the lowercased should not to be interpreted as the keyword SHOULD NOT, which would mean that - in the end - this is just a recommendation for the endpoint's behavior and equally to omitting the fields[TYPE] parameter at all?

As I read the specification, an empty fieldset means that the server must not include and field. The specification says:

> If a client requests a restricted set of fields for a given resource type, an endpoint **MUST NOT** include additional [fields] in resource objects of that type in its response.

If client includes `fields[TYPE]` query parameter in a request, the client requests a restricted set of fields. The server must only include those fields, which have been requested by the client. If the client requested to include no fields at all (empty fieldset), the server must not include any field.

In my reading the first paragraph specifies the value of `fields[TYPE]` parameter:

> The value of the `fields` parameter **MUST** be a comma-separated (U+002C COMMA, ",") list that refers to the name(s) of the fields to be returned. An empty value indicates that no fields should be returned.

The second paragraph defines processing rules for the server if it is present:

> If a client requests a restricted set of [fields] for a given resource type, an endpoint **MUST NOT** include additional [fields] in resource objects of that type in its response.

Your changes mixes both as it defines processing rules for the server in the first paragraph as well. Not sure if that's helpful.

Maybe the clarification how a server must process an empty fieldset could be added at the end of the second paragraph? Some thing like this:

> The value of the `fields` parameter **MUST** be a comma-separated (U+002C
> COMMA, ",") list that refers to the name(s) of the fields to be returned.
> The list **MAY** be empty, indicating that no fields should be returned.
> 
> If a client requests a restricted set of fields for a given resource type,
> an endpoint **MUST NOT** include additional fields in resource objects of
> that type in its response. If a client requests an empty set of fields, an
> endpoint **MUST NOT** include any field in the resource object of that
> type in the response.

Not perfect yet. I especially do not like the duplication in last paragraph. But I hope it gives an idea what I mean anyways.

ThorstenSuckow:
> > If a client requests an empty set of fields, an
> > endpoint **MUST NOT** include any field in the resource object of that
> > type in the response.
> 
> Not perfect yet. I especially do not like the duplication in last paragraph. But I hope it gives an idea what I mean anyways.

Hey @jelhan, this gives a better indication about how the specification needs to be understood, thanks for the reply.

I added the **MUST NOT** to the first paragraph since the paragraph itself along with the subsequent paragraphs define the use cases available with sparse fieldsets:

1. `fields[TYPE]=` (empty)
2. `fields[TYPE]={list of fields}` 
3. (not used at all)

Sending `nothing` (as in  1.) to the server and the fact that each server may understand `nothing` differently, along with the rather confusing `should not`, might leave too much room for interpretation.

What if in the end it turns out that the `should not`  was actually meant as **SHOULD NOT** and it just slipped through the editor's eyes... 👀

If you want to close the hasty PR, I can create a ticket for it instead.

dgeb:
@ThorstenSuckow thanks for your contribution! 

I will try to go into the thinking behind the normative statements in this section.

This statement establishes the client's responsibility to use the `fields` param:

> The value of the `fields` parameter **MUST** be a comma-separated (U+002C
 COMMA, ",") list that refers to the name(s) of the fields to be returned.
 An empty value indicates that no fields should be returned.

The reason that normative language is not used within the second sentence is that it is a clarification to help interpret the normative statement in the first sentence. It is not a normative statement itself.

Finally, this normative statement, just below, covers the server's responsibility:

> If a client requests a restricted set of [fields] for a given resource type,
 an endpoint **MUST NOT** include additional [fields] in resource objects of
that type in its response.

As @jelhan mentioned, we should avoid mixing processing rules for the server in with usage rules for the client.

> If you want to close the hasty PR, I can create a ticket for it instead.

Please feel free to create an issue if this is still unclear.

Issue #1524, opened 2020-11-30
Please clarify: 200 OK response when updating relationships


maurei:
Please clarify the motive behind the following [bit of the spec](https://jsonapi.org/format/#crud-updating-relationship-responses-200)

> 200 OK
> If a server accepts an update but also changes the targeted relationship(s) in other ways than those specified by the request, it MUST return a 200 OK response. The response document MUST include a representation of the updated relationship(s).

Specifically: what are some use-cases for this?

We have been working hard on improving [JSON:API .NET Core](www.jsonapi.net) and making it fully spec compliant. We understand from a technical perspective how such side-effects could arise, but we are lacking knowledge of any clear use-cases for it, and because introducing it could come with a significant performance cost we have decided to hold off implementing this element of the specification.

We came up with some examples to illustrate this.

1. Given a resource `a` of type `A` with relationships `has-one B` and `has-one C`, we can imagine how setting `B=b1` and `C=c1`  would trigger some database procedure that e.g. would effectively yield `B=b1` and `c=null`.
However, our concern with this is that
	- we don't know of any relevant real-life examples for this, and
	- the performance implications of checking for such effects could be disastrous in the case of a `has-many` relationship when the size of the collection is very large. 
2. The next case is similar to the previous, but this one *does* come with a real-life use-case. Consider a resource type `Article` with a **required** relationship to `User` (`has one Author`). Consider an API user that creates an article (`POST /articles`) without providing a reference for `Author`, and let's assume that in the context of this API it is guaranteed that user who creates the article is always its author. The server could then take care of setting this required relationship by deducing the author from the request context, freeing the API user from the responsibility of passing a (trivial) reference to himself. The problem with this is however:
	- the spec specifically states that side-effects should only be checked for *targeted relationship(s)*, i.e. for the current scenario a 201 No Content would suffice. This makes sense, because performance implications of having to check for every non-targeted relationship of a resource would quickly become unmanageable for any non-trivial model. 

Some explanation would be greatly appreciated!

Thank you

gabesullice:
Caveat: I'm speaking as an implementer here, not as a spec editor because I don't know exactly what the original intent was.

---
Two thoughts:

[...]

PR #1638, opened 2022-06-27
Standardize and clarify success (2xx) responses

dgeb:
This PR applies consistent language and ordering across success (2xx) responses for creating, updating, and deleting resources and relationships.

It clarifies that it is always ok to return a 200 (or 201 in the case of creation) response with a document that does not include primary data when the server does not change the targeted resource(s) / relationship(s) in any way different from what was requested. This allows the server to include other top-level members, such as `meta`, in responses without primary data.

This also removes a logical inconsistency in which MUST was used for 204 responses which contradicted a MAY allowance for 200/201 responses with documents that include `meta`.

Closes #1540 #1581

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1584,10 +1584,10 @@ server **MUST** return a `202 Accepted` status code.
 ##### <a href="#crud-creating-responses-204" id="crud-creating-responses-204" class="headerlink"></a> 204 No Content
 
 If the requested resource has been created successfully and the server does not
-change the resource in any way (for example, if the resource includes a
-[Client-Generated ID](#crud-creating-client-ids)), the server **MUST** return
-either a `201 Created` status code and response document (as described above) or
-a `204 No Content` status code with no response document.
+change the resource in any way (for example, by assigning an `id` or `createdAt`
+attribute), the server **MUST** return either a `201 Created` status code and
+response document (as described above) or a `204 No Content` status code with no
+response document.
 
 ##### <a href="#crud-creating-responses-403" id="crud-creating-responses-403" class="headerlink"></a> 403 Forbidden
 
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1533,16 +1533,14 @@ to create a resource with a client-generated ID.
 
 ##### <a href="#crud-creating-responses-201" id="crud-creating-responses-201" class="headerlink"></a> 201 Created
 
-If a `POST` request did not include a [Client-Generated
-ID](#crud-creating-client-ids) and the requested resource has been created
-successfully, the server **MUST** return a `201 Created` status code.
+If the requested resource has been created successfully and the server changes
+the resource in any way (for example, by assigning an `id`), the server **MUST**
+return a `201 Created` response and a document that contains the resource as
+primary data.
 
 The response **SHOULD** include a `Location` header identifying the location
 of the newly created resource, in order to comply with [RFC
 7231](http://tools.ietf.org/html/rfc7231#section-6.3.2).
-
-The response **MUST** also include a document that contains the primary
-resource created.
 
 If the [resource object][resource objects] returned by the response contains a `self` key in its
 `links` member and a `Location` header is provided, the value of the `self`
@@ -1568,6 +1566,12 @@ Content-Type: application/vnd.api+json
 }
 ```
 
+A server **MAY** return a `201 Created` response with a document that contains
+no primary data if the requested resource has been created successfully and the
+server does not change the resource in any way (for example, if the resource
+includes a [Client-Generated ID](#crud-creating-client-ids)). Other top-level
+members, such as [meta], could be included in the response document.
+
 ##### <a href="#crud-creating-responses-202" id="crud-creating-responses-202" class="headerlink"></a> 202 Accepted
 
 If a request to create a resource has been accepted for processing, but the
@@ -1576,15 +1580,11 @@ server **MUST** return a `202 Accepted` status code.
 
 ##### <a href="#crud-creating-responses-204" id="crud-creating-responses-204" class="headerlink"></a> 204 No Content
 
-If a `POST` request *did* include a [Client-Generated
-ID](#crud-creating-client-ids) and the requested resource has been created
-successfully, the server **MUST** return either a `201 Created` status code
-and response document (as described above) or a `204 No Content` status code
-with no response document.
-
-> Note: If a `204` response is received the client should consider the resource
-object sent in the request to be accepted by the server, as if the server
-had returned it back in a `201` response.
+If the requested resource has been created successfully and the server does not
+change the resource in any way (for example, if the resource includes a
+[Client-Generated ID](#crud-creating-client-ids)), the server **MUST** return
+either a `201 Created` status code and response document (as described above) or
+a `204 No Content` status code with no response document.
 
 ##### <a href="#crud-creating-responses-403" id="crud-creating-responses-403" class="headerlink"></a> 403 Forbidden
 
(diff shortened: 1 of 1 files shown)


_format/1.1/index.md
```diff
@@ -1568,6 +1566,12 @@ Content-Type: application/vnd.api+json
 }
 ```
 
+A server **MAY** return a `201 Created` response with a document that contains
+no primary data if the requested resource has been created successfully and the
+server does not change the resource in any way (for example, if the resource
+includes a [Client-Generated ID](#crud-creating-client-ids)). Other top-level
+members, such as [meta], could be included in the response document.
```
jelhan: Having a _client-generated ID_ in the request is a requirement for this case, isn't it? Naming it as an example may be confusing. Wondering if calling this requirement out in a separate note would be better from teaching perspective.

_format/1.1/index.md
```diff
@@ -1576,15 +1580,11 @@ server **MUST** return a `202 Accepted` status code.
...
-with no response document.
-
-> Note: If a `204` response is received the client should consider the resource
-object sent in the request to be accepted by the server, as if the server
-had returned it back in a `201` response.
+If the requested resource has been created successfully and the server does not
+change the resource in any way (for example, if the resource includes a
+[Client-Generated ID](#crud-creating-client-ids)), the server **MUST** return
+either a `201 Created` status code and response document (as described above) or
+a `204 No Content` status code with no response document.
```
jelhan: As above, I would move the _client-generated ID_ out to separate note and naming it as a technical requirement for clarity.

_format/1.1/index.md
```diff
@@ -1576,15 +1580,11 @@ server **MUST** return a `202 Accepted` status code.
...
 
-If a `POST` request *did* include a [Client-Generated
-ID](#crud-creating-client-ids) and the requested resource has been created
-successfully, the server **MUST** return either a `201 Created` status code
-and response document (as described above) or a `204 No Content` status code
-with no response document.
-
-> Note: If a `204` response is received the client should consider the resource
-object sent in the request to be accepted by the server, as if the server
-had returned it back in a `201` response.
```
jelhan, line 1587: Any specific reason to drop this note? It seem to help implementing client-side libraries and makes it more clear.

_format/1.1/index.md
```diff
@@ -1747,31 +1747,30 @@ does not want to allow deletion of records the client has not seen.
 
 #### <a href="#crud-updating-responses" id="crud-updating-responses" class="headerlink"></a> Responses
 
+##### <a href="#crud-updating-responses-200" id="crud-updating-responses-200" class="headerlink"></a> 200 OK
+
+If a server accepts an update but also changes the targeted resource(s) in ways
+other than those specified by the request (for example, updating the
+`updated-at` attribute or a computed `sha`), it **MUST** return a `200 OK`
```
jelhan: The example used here should follow latest recommendation for naming convention.

```suggestion
`updatedAt` attribute or a computed `sha`), it **MUST** return a `200 OK`
```

_format/1.1/index.md
jelhan: > Other top-level members, such as [meta], could be included in the response document.

This is a little bit confusing as `meta` _must_ be included in this case due to this requirement:

> A document **MUST** contain at least one of the following top-level members:
> 
> * `data`: the document's "primary data"
> * `errors`: an array of [error objects](#errors)
> * `meta`: a [meta object][meta] that contains non-standard
>   meta-information.

_format/1.1/index.md
```diff
@@ -1970,6 +1969,19 @@ server, and we are defining its semantics for JSON:API.
 
 #### <a href="#crud-updating-relationship-responses" id="crud-updating-relationship-responses" class="headerlink"></a> Responses
 
+##### <a href="#crud-updating-relationship-responses-200" id="crud-updating-relationship-responses-200" class="headerlink"></a> 200 OK
+
+If a server accepts an update but also changes the targeted relationship(s)
```
jelhan: Relationships in plural is confusing to me. A request to update a relationship using a relationship link can only target one relationship. It may affect more relationships, if changing one relationship has a side-effect on another. But that wouldn't be a "targeted relationship".

Maybe this is not talking about a relationship but instead of the related resources identified by their resource identifier objects?

I raised that point recently in https://github.com/json-api/json-api/issues/1524#issuecomment-1166288191.

_format/1.1/index.md
```diff
@@ -1970,6 +1969,19 @@ server, and we are defining its semantics for JSON:API.
...
+in other ways than those specified by the request, it **MUST** return a `200
+OK` response and a document that includes the updated relationship(s) in its
+primary data.
```
jelhan: What does "include the updated relationship(s)" mean? How does an example response payload look like?

Assuming we have a relationship `tags` of an `articles` resource. Let's assume that currently two resources are referenced by that relationship: `{ "tags": "1", "tags": 2 }`

A client sends a request to add `{ "tags": "3" }` additionally. Due to some domain-specific constraints the same article can not be related to `{ "tags": "3" }` and `{ "tags": "1" }`. The server removes `{ "tags": "1" }` from the relationship automatically.

What does a response payload look like in this case?

The wording above (and in the specification right now) could be interpreted in two ways:

1. Response with all resources currently associated to that relationship.
2. Response with the resources, which are added or removed from that relationship due to the update.

The first one would be easy. The server would response with:

```json
{
  "data": [
    { "tags": "2" },
    { "tags": "3" }
  ]
}
```

The second one is not supported by the base spec. As far as I know the server can not inform the client that `{ "tags": "1" }` has been removed from the relationship.

Not sure if you want to address this one in this pull request as well. Or keep it for a follow-up PR.

jelhan (commented):
Thanks a lot for addressing this one.

Wondering if you want to address #1524 in this one as well or keep it for a follow-up PR. In that case please ignore the two comments related to it.

Beside that one the comments are only about minor points.

_format/1.1/index.md
dgeb: > Having a client-generated ID in the request is a requirement for this case, isn't it? Naming it as an example may be confusing.

The whole decision about whether a server must include primary data in all of these sections comes down to whether the server changes the resource(s) beyond what's been requested. The spec can not fully control what that means once we factor in implementation semantics such as `meta` data and user-defined fields, not to mention extensions.

> Wondering if calling this requirement out in a separate note would be better from teaching perspective.

Let me experiment with this option to keep the normative text tighter.

dgeb: This has been adjusted to allow for extensions that may allow other top-level members.

_format/1.1/index.md
dgeb: My thinking was that including this note for creation, but not updating or deleting, is inconsistent. And all the sections about `204` are quite clear that it's only an appropriate response when the request has been processed successfully. I'm really not sure how this note adds any clarity beyond that.

_format/1.1/index.md
jelhan: > The whole decision about whether a server must include primary data in all of these sections comes down to whether the server changes the resource(s) beyond what's been requested. The spec can not fully control what that means once we factor in implementation semantics such as `meta` data and user-defined fields, not to mention extensions.

Fully agree. But the specification enforces changing the resource object by adding a server-generated ID to it unless request included a client-generated ID. That makes the example problematic in my opinion because in that case it's mandatory not optional. Maybe it could use `createdAt` as an example to avoid that confusion and explicitly call out the client-generated ID in a note? Because it's such a very likely case?

_format/1.1/index.md
dgeb: > Relationships in plural is confusing to me.

I used `relationship(s)` here because a single request can impact multiple one-to-many relationships between unique resources. Which I think aligns with what you're getting at here:

> Maybe this is not talking about a relationship but instead of the related resources identified by their resource identifier objects?

_format/1.1/index.md
jelhan: I fear it did not landed or got lost:

https://github.com/json-api/json-api/blob/25c69821f0cfb2fef1fb14aaff6f6bd9add5f91c/_format/1.0/index.md?plain=1#L78-L83

https://github.com/json-api/json-api/blob/25c69821f0cfb2fef1fb14aaff6f6bd9add5f91c/_format/1.1/index.md?plain=1#L293-L298

_format/1.1/index.md
jelhan: Got it. But I think "relationships" and "relationship" is defined in another way in the specification:

> The value of the `relationships` key **MUST** be an object (a "relationships object"). Each member of a relationships object ("relationships") represents a "relationship" from the resource object in which it has been defined to other resource objects.

In my reading this defines the term "relationship" as a reference from one resource to one or many other resources. That's also the common definition of "relationship" in IT, which I'm aware of. The term "to-many relationship" only makes sense, when a relationship could be a collection of related resources.

What you mean name relationship here, seems to be called "related resources" at other places of the specification. For example in the term "related resource link".

> A "related resource link" provides access to resource objects linked in a relationship.

_format/1.1/index.md
dgeb: I will create a new issue / PR to address this prior to v1.1. This wording should probably be updated to have the same meaning within the base spec, but also clearly allow for extensions to expand possible top-level members.

_format/1.1/index.md
dgeb: After review, I'm changing from `relationship(s)` to `relationship`, since a request can in fact only affect one relationship but potentially multiple relationship members. Thanks for calling this out, @jelhan.

jelhan (approved):
Thanks a lot for addressing all open topics. Looks good to me now.

_format/1.1/index.md
bkoelman: As the maintainer of [JsonApiDotNetCore](https://github.com/json-api-dotnet/JsonApiDotNetCore), the leading server-side JSON:API implementation in ASP.NET, I can say that we're never going to implement option 1. Because it's potentially very expensive to determine and useless to clients, except for theoretical cases. Consider an audit log, where old entries are purged as a side-effect. So POSTing to this to-many relationship would require returning thousands of entries in the response (there's no pagination). Option 2 is so extremely unlikely to ever occur that we can't justify doing extra queries to detect this.

I can see the elegance in using similar statements throughout the spec, but it's pretty impractical when it comes to relationships. Another case is using rowversion/xmin/timestamp columns for optimistic concurrency. They change all the time, but depending on at which side the foreign key is defined, we cannot consistently return that information, so honestly I don't see the point.

From a spec perspective, I think the only sensible solution is to change MUST to MAY in:
> If a server accepts an update but also changes the targeted relationship in other ways than those specified by the request, it MUST return a 200 OK response and a document that includes the updated relationship data as its primary data.

jelhan: > Because it's (...) useless to clients, except for theoretical cases.

I disagree with this. It eases handling of client-side caches as the client can be sure that local cache is still up-to-date. Otherwise a client either needs to refetch the data or hope that no side-effects happened to manage client-side cache.

> Another case is using rowversion/xmin/timestamp columns for optimistic concurrency. They change all the time, but depending on at which side the foreign key is defined, we cannot consistently return that information, so honestly I don't see the point.

The specification is only talking about changes to the relationship in the sense of additional resources being added to or removed from that relationship than those explicitly requested by the client. Additional meta data a server may store with a relationship (or even attributes of a related resource) are not relevant to determine if "the targeted relationship (changes) in other ways than those specified by the request". At least in my reading of the specification.

I see your point though that for some edge cases returning all resource identifier objects for related resources may be too expensive. This may be the case for 1) to-many relationships, 2) having many related resources, which cannot be queried in an efficient way, and 3) adding or removing a related resource causes side-effects on other resources being added.

bkoelman: If the goal is to handle client-side caches, then the response structure should be much more flexible. It should be possible to return type/id entries of any resource type, as well as full resource objects of any type. The choice to _only_ return type/id entries that occurred in the request seems arbitrary and not very practical.

Please prove me wrong and show that it's not useless, by providing a real-world compelling example where this makes sense. I haven't found any.

PR #1642, opened 2022-07-27
only top-level member may be defined by extension

jelhan:
An extension may introduce it's own top-level member. E.g. the [atomic operations extension]() defines the additional top-level members `atomic:operations` and `atomic:results`.

So far the specification enforced that a JSON:API document has either `data`, `errors` or `meta` top-level member. Even if there is another top-level member introduced by an applied extension.

This is not desirable. And it was not intended. It is likely that some implementations would ignore that requirement. Even the examples given in official atomic operations extension are ignoring it:

> ```http
> POST /operations HTTP/1.1
> Host: example.org
> Content-Type: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
> Accept: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
> 
> {
>   "atomic:operations": [{
>     "op": "add",
>     "href": "/blogPosts",
>     "data": {
>       "type": "articles",
>       "attributes": {
>         "title": "JSON API paints my bikeshed!"
>       }
>     }
>   }]
> }
> ```

A `meta` top-level member with an empty object could be added to fulfill that requirement. The example payload from atomic operations extensions would be changed like the following:

> ```patch
>   POST /operations HTTP/1.1
>   Host: example.org
>   Content-Type: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
>   Accept: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
> 
>   {
>     "atomic:operations": [{
>       "op": "add",
>       "href": "/blogPosts",
>       "data": {
>         "type": "articles",
>         "attributes": {
>           "title": "JSON API paints my bikeshed!"
>         }
>       }
> -   }]
> +   }],
> +   "meta": {}
>   }
> ```

But there doesn't seem to be any value in doing so. Instead we should relax the constraint of the base spec to better support this use case of extensions.

I see two potential options to address the issue:

1. We could relax the base spec to not enforce _any_ top-level member. To do so, we would replace _MUST_ with _MAY_ in the sentence.
   ```patch
   - A document **MUST** contain at least one of the following top-level members:
   + A document **MAY** contain at least one of the following top-level members:
   ```
2. We could keep the requirement but add top-level members defined by an applied extension as a forth option.
    ```patch
      A document **MUST** contain at least one of the following top-level members:

      * `data`: the document's "primary data".
      * `errors`: an array of [error objects](#errors).
      * `meta`: a [meta object][meta] that contains non-standard meta-information.
    + * a member defined by an applied extension.
    ```

I tend toward the second approach:

- A JSON:API document without any top-level member would not have any value. I think we should disallow this case in the base spec.
- Allowing a JSON:API document without `data`, `errors` or `meta` top-level member even if no extension is applied, could be considered a breaking change.
- As extension did not exist in v1.0 changing semantics only in case an extension is applied, is _not_ a breaking change.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -296,7 +296,7 @@ A document **MUST** contain at least one of the following top-level members:
 * `errors`: an array of [error objects](#errors).
 * `meta`: a [meta object][meta] that contains non-standard
   meta-information.
-* a member defined by an applied extension.
+* a member defined by an applied [extension](#extensions).
 
 The members `data` and `errors` **MUST NOT** coexist in the same document.
 
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -292,16 +292,17 @@ document containing data. This object defines a document's "top level".
 
 A document **MUST** contain at least one of the following top-level members:
 
-* `data`: the document's "primary data"
-* `errors`: an array of [error objects](#errors)
+* `data`: the document's "primary data".
+* `errors`: an array of [error objects](#errors).
 * `meta`: a [meta object][meta] that contains non-standard
   meta-information.
+* a member defined by an applied extension.
 
 The members `data` and `errors` **MUST NOT** coexist in the same document.
 
 A document **MAY** contain any of these top-level members:
 
-* `jsonapi`: an object describing the server's implementation
+* `jsonapi`: an object describing the server's implementation.
 * `links`: a [links object][links] related to the primary data.
 * `included`: an array of [resource objects] that are related to the primary
   data and/or each other ("included resources").


_format/1.1/index.md
```diff
@@ -292,16 +292,17 @@ document containing data. This object defines a document's "top level".
 
 A document **MUST** contain at least one of the following top-level members:
 
-* `data`: the document's "primary data"
-* `errors`: an array of [error objects](#errors)
+* `data`: the document's "primary data".
+* `errors`: an array of [error objects](#errors).
 * `meta`: a [meta object][meta] that contains non-standard
   meta-information.
+* a member defined by an applied extension.
```
```suggestion
* a member defined by an applied [extension](#extensions).
```

dgeb (commented):
Thanks for addressing this, @jelhan. I agree with this approach and the clear wording you've used. I think it just needs a link back to the extensions section.

dgeb (approved):

Art4:
@jelhan I'm sorry for not following this topic earlier, but allowing extensions to define the only member in top-level document is a breaking change.

> * Allowing a JSON:API document without data, errors or meta top-level member even if no extension is applied, could be considered a breaking change.

This is really the case here: A `1.0` parser ignores the `atomic:operations` member and is left with a top-level member without any member at all, which is invalid.

`1.0` requires at least `data`, `errors` or `meta` as a top-level member, so your example with an empty `meta` member is a valid workaround.

```diff
  POST /operations HTTP/1.1
  Host: example.org
  Content-Type: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"
  Accept: application/vnd.api+json;ext="https://jsonapi.org/ext/atomic"

  {
    "atomic:operations": [{
      "op": "add",
      "href": "/blogPosts",
      "data": {
        "type": "articles",
        "attributes": {
          "title": "JSON API paints my bikeshed!"
        }
      }
-   }]
+   }],
+   "meta": {}
  }
```

Sure, this workaround is a bit annoying, but would keep BC with `1.0`. The same goes for relationship objects, see #1644.

The changes in this PR should be part of `2.0` instead of `1.1`. And the examples in the atomic extension should mention this temporary workaround until `2.0` is released.

jelhan:
@Art4 [Content negotiation](https://jsonapi.org/format/#content-negotiation) between client and server prevents that issue. A client requests that a server applies an extension via `Accept` request header. If the server does not support v1.1 (or if it does not support that specific extension), the server must reject the request with a `406 Not Acceptable` response.

Art4:
@jelhan Good point, thank you very much. Sadly this way version 1.0 is not forward compatible with 1.1, but at least 1.1 is backward compatible to 1.0.

My mistake was that after reading e.g. [this comment](https://github.com/json-api/json-api/issues/867#issuecomment-143004973) I assumed that JSON:API would be forward compatible.

I will adjust my parser with this in mind. Thank you for your patient.

jelhan:
It's forward compatible. A server implementing v1.0 handles every request to apply an extension as not supporting that specific extension.

But I see that it gets problematic if a library only implements parts of the processing rules. In your case the library seems to be limited to parsing a JSON:API document. But not implementing the content negotiation. I would recommend to document in that case that it only supports JSON:API documents without any extension applied. As extensions can define additional processing rules, a JSON:API document cannot be parsed ignoring applied extensions.

PR #1647, opened 2022-07-31
Improve explanation for reserving brackets in member names

jelhan:
Brackets are used in all query parameter families. Not only in sparse fieldsets.

Looks as if we missed to update this explanation when _query parameter families_ have been added as a concept.


dgeb (approved):
Great refinement, thanks @jelhan!

PR #1648, opened 2022-08-01
update can only target a single resource

jelhan:
An update can only _target_ a single resource. Speaking of "targeted resource(s)" may confuse the reader therefore.

An update may _affect_ multiple resources if it has side-effects. E.g. updating a `rating` resource may indirectly update `averageRating` attribute of a related `product` resource. But nevertheless the update only _targets_ the `rating` resource.

The difference between _targeted_ and _affected_ resources is important. HTTP response status codes required by the specification only depends on _targeted_ resources. E.g. a server must include the _targeted_ resource in response if it has been changed in any other way than specified. But a server may or may not include another, non-targeted resource in response even if it has been changed due to a side-effect.

A similar issue for "targeted relationship(s)" has been fixed in https://github.com/json-api/json-api/pull/1638#discussion_r907483986.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -1752,14 +1752,14 @@ does not want to allow deletion of records the client has not seen.
 
 ##### <a href="#crud-updating-responses-200" id="crud-updating-responses-200" class="headerlink"></a> 200 OK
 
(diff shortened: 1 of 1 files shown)


dgeb (approved):
Agreed. Another good refinement, @jelhan!

PR #1658, opened 2022-08-22
allow extension defined member as only member in links object of relationship object

jelhan:
This allows an extension defined member as the only member in a `links` object of a relationship object.

This is not a breaking change compared. Extensions are not part of v1.0. So there could not be an applied extension if implementing v1.0.

Similar changes have been landed in #1642 and #1644 for other cases.

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -495,6 +495,7 @@ A "relationship object" **MUST** contain at least one of the following:
     for the related resources as its primary data.
     (See [Fetching Relationships](#fetching-relationships).)
   * `related`: a [related resource link]
+  * a member defined by an applied [extension](#extensions).
 * `data`: [resource linkage]
 * `meta`: a [meta object][meta] that contains non-standard meta-information about the
   relationship.


VGirol:
Perhaps one should also allow a member defined by an extension as the only member in:
- the top level links object
- the links object of a resource object.

jelhan:
> Perhaps one should also allow a member defined by an extension as the only member in:
> 
>     * the top level links object
> 
>     * the links object of a resource object.

We assume that it is not needed in this cases. In both cases it only lists allowed members. But does not forbid any other members:

> The top-level [links object](https://jsonapi.org/format/1.1/#document-links) MAY contain the following members:

> If present, this links object MAY contain a self [link](https://jsonapi.org/format/1.1/#document-links) that identifies the resource represented by the resource object.

Extensions are allowed to define new members anywhere in the document:

> An extension MAY define new members within the document structure defined by this specification. 

The only problematic case is if current wording requires another member of a fixed list to be present as well. I have included more details in #1642.

jelhan:
Maybe the wording used for the `links` object in the different contexts should be aligned:

https://github.com/json-api/json-api/blob/f9e4e0e13d8d1005228878e7d03f833adbea4042/_format/1.1/index.md?plain=1#L313-L323

https://github.com/json-api/json-api/blob/f9e4e0e13d8d1005228878e7d03f833adbea4042/_format/1.1/index.md?plain=1#L489-L497

https://github.com/json-api/json-api/blob/f9e4e0e13d8d1005228878e7d03f833adbea4042/_format/1.1/index.md?plain=1#L578-L579

I don't see good reasons for having a different wording in this three cases. In all cases a list of potential members is specified. Using a different wording may only cause confusion.

There is one important difference: a `links` object in context of a relationship object must have at least one member. The others are allowed to be empty as well. I don't think that these difference is by intend.

VGirol:
I had in mind that : 
https://github.com/json-api/json-api/blob/f9e4e0e13d8d1005228878e7d03f833adbea4042/_format/1.1/index.md?plain=1#L281-L283

So, i thought that, because of this, extension defined members could **not** be the only member of a links object in the others contexts (top-level and resource).
But i forgot this : 
https://github.com/json-api/json-api/blob/f9e4e0e13d8d1005228878e7d03f833adbea4042/_format/1.1/index.md?plain=1#L96-L98

So, you're right.

> The only problematic case is if current wording requires another member of a fixed list to be present as well.

Sorry for the inconvenience...

jelhan:
> Sorry for the inconvenience...

Not at all! I'm happy to see other people having a close look at the details at well. It will help us to not miss anything important before releasing v1.1.

dgeb (approved):

dgeb:
Thanks for the discussion @jelhan and @VGirol.

> Maybe the wording used for the links object in the different contexts should be aligned

Sounds good. As we discussed separately, it's fine with me to handle this in a subsequent PR.

PR #1662, opened 2022-08-29
Update jsonapi.svg

bradjones1:
Fix viewBox


dgeb:
@bradjones1 could you take a look at #1663? I updated both the sketch file and the svg to match.

PR #1663, opened 2022-08-29
Convert fonts in logo to outlines

dgeb:
Converts fonts in logo to outlines and trim border in svg.

Closes #1662


bradjones1:
LGTM

PR #1676, opened 2022-10-27
Remove unnecessary and potential confusing alias "relationships"

jelhan:
Introducing the term _"relationships"_ additionally to _"relationships object"_ and _"relationship"_ is unnecessary and potentially confusing. How it was introduced made it seen as an alias for "relationships object". But it is later used just as plural of "relationship". I think we could simplify by not introducing it as a special term.


dgeb (approved):
This is a nice simplification, thanks @jelhan 💯

PR #1677, opened 2022-10-28
Clarify usage of parallel relationships

### LINKED ISSUE #1361: Support for parallel relationships
The spec is silent about parallel relationships (that is, more than one resource identifier object with the same `type` and `id` combo) in all but [one case](https://jsonapi.org/format/1.0/#crud-updating-to-many-relationships):

> If a client makes a **POST** request to a URL from a relationship link, the server **MUST** add the specified members to the relationship unless they are already present. If a given `type` and `id` is already in the relationship, the server **MUST NOT** add it again.

I think it would be useful for this spec to provide explicit support for parallel relationships. This would mean relaxing the above statement and then clarifying their treatment and serialization.

Let's think about a real-world example:

- I have a resource object with the `type`: `organization`
- I have a relationship field named `members`.
- I have a resource object with the `type`: `user`
- I have a `user` who is both a "founder" of the organization and the "treasurer" & these roles have start and end dates.

A request to `/organizations/1/relationships/members` should return:

```
HTTP/1.1 200 Okay

{
  "data": [{
    "type": "user",
    "id": 42,
    "meta": {
      "role": "founder",
      "since": "2012-09-01T15:53:00+06:00",
      "until": "heat death of the universe",
    }
  }, {
    "type": "user",
    "id": 42,
    "meta": {
      "role": "treasurer",
      "since": "2017-01-01T00:00:00+06:00",
      "until": "2021-01-01T00:00:00+06:00"
    }
  }]
}
```

Obviously, one could model these memberships as standalone resource objects, however this can pose a significant technical challenge on systems which do not natively model relationships this way (my scenario). Moreover, in a traditional relational database, modeling relationships in that way can lead to application performance issues due to an increase in the average number of `JOIN`s required for its typical queries.

We currently work around this shortcoming by adding a new member to the `meta` object of resource objects that we call `arity`:

```
{
    "type": "user",
    "id": 42,
    "meta": {
      "arity": 0
    }
  }, {
    "type": "user",
    "id": 42,
    "meta": {
      "arity": 1
    }
  }
```

We use this member to help a client disambiguate parallel relationships, to identify the relationship to delete on relationship `DELETE` requests and to provide a stable mechanism to _add_ a parallel relationship via `POST` (this is the part that's in conflict with the spec).

I found this issue while attempting to author a profile for this mechanism, but realized it was in conflict with the spec. This made me think that it might actually be more valuable to add support for parallel relationships in the base spec and not in a profile since it's so closely related to core functionality.

So, what's my proposal?

Add an `arity` member to a resource identifier object with the following language:

> A “resource identifier object” **MUST** contain an `arity` member when it is in an array that includes another resource identifier object with the same `type` and `id` values. The `arity` member **MUST NOT** be present if the resource identifier object is unique within the array.
>   * The value of the `arity` member **MUST** be a non-negative integer.
>   * The value of the first occurrence of an `arity` member **MUST** be zero for each set of resource identifier objects in the array with shared `type` and `id` values. For each subsequent resource identifier object within that set, the value **MUST** be incremented by one.
>   * The `arity` member **MUST NOT** be used to indicate an order for resource identifier objects in an array.
>   * When an array of resource identifier objects is reordered, the value of each `arity` member **MUST** be reindexed.
>
> For example, it is possible for two resource identifier objects to have the same `arity` as long as they do not share the same `type` and `id` value; unrepeated resource identifier objects will not have an `arity` member;  and `arity` does not correspond to array order:
>
> ```json
> {
>   "data": [
>     {"type": "color", "id": "red", "arity": 0},
>     {"type": "color", "id": "blue", "arity": 0},
>     {"type": "color", "id": "green"},
>     {"type": "color", "id": "red", "arity": 1},
>     {"type": "color", "id": "red", "arity": 2},
>     {"type": "color", "id": "blue", "arity": 1}
>   ]
> }
> ```

This is proposal is exactly what we have been doing successfully in our own implementation (except we put `arity` under the `meta` member). In doing so, we've worked all the gritty details out for complete CRUD support too. More information here: https://www.drupal.org/node/2992674

jelhan:
This PR _tries_ to implement the agreement we came up in #1361.

I struggled with the wording to be honest. I found it difficult to distinguish between the different kinds of parallel relationships. For sake of clarity let me show them in code:

1. **Different relationships**
    ```json
    {
      "data": {
        "type": "articles",
        "id": "1",
        "relationships": {
          "author": {
            "data": {
              "type": "users",
              "id": "12"
            }
          },
         "editors": {
           "data": [
             {
               "type": "users",
               "id": "12"
             }
           ]
         }
      }
    }
    ```

2. **Intermediate resource**
    ```json
    {
      "data": {
        "type": "teams",
        "id": "1",
        "relationships": {
          "members": {
            "data": {
              "type": "teamMembers",
              "id": "12"
            }
          },
         "editors": {
           "data": [
             {
               "type": "teamMembers",
               "id": "13"
             }
           ]
         }
      },
      "included": [
        {
          "type": "teamMembers",
          "id": "12",
          "attributes": {
            "startDate": "2021-01-01",
            "endDate": "2021-12-31",
          },
          "relationships": {
            "user": {
              "data": {
                "type": "users",
                "id": "23"
              }
            }
          }
        },
        {
          "type": "teamMembers",
          "id": "13",
          "attributes": {
            "startDate": "2022-06-01",
            "endDate": "2022-12-31",
          },
          "relationships": {
            "user": {
              "data": {
                "type": "users",
                "id": "23"
              }
            }
          }
        }
      ]
    }
    ```

3. **Multiple, direct reference on same relationship**
    ```json
    {
      "data": {
        "type": "teams",
        "id": "1",
        "relationships": {
          "members": {
            "data": [
              {
                "type": "users",
                "id": "23",
                "meta": {
                  "startDate": "2021-01-01",
                  "endDate": "2021-12-31",
                }
              },
             {
                "type": "users",
                "id": "23",
                "meta": {
                  "startDate": "2022-06-01",
                  "endDate": "2022-12-31",
                }
              }
            ]
          }
        }
      }
    }
    ```

We only aim to discourage using the third type of parallel relationships. We recommend to use the second type instead. I'm not sure how well this is expressed in the current wording.

Open questions:

1. Do we want to backport this recommendation to v1.1?
2. Do we need a better wording to distinguish between relationships of a resource (`author`, `editors`) and references to resource in such relationships?

Closes #1361

### DIFF
--- a/_format/1.2/index.md
+++ b/_format/1.2/index.md
@@ -508,6 +508,25 @@ data, not the related resources.
 
 > Note: See [fields] and [member names] for more restrictions on this container.
 
+A to-many relationship **SHOULD NOT** reference the same resource object more than
+once.
+
+> Note: This recommendation does _not_ prevent parallel relationships:
+> 
+> 1. A resource object may reference another resource object on different
+> relationships. For example, an `article` resource may reference the same `user`
+> resource both on its `author` and `editors` relationship.
+> 2. A resource object may reference another resource object more than once
+> on the same relationship through an intermediate resource object. For example,
+> a `team` resource may reference the same `user` resource more than once on its
+> `members` relationship _through_ a `teamMember` resource. The `teamMember`
+> resource could present additional information as fields such as `startDate`,
+> `endDate`, and `invitedBy`.
+> 
+> Only multiple references to the same resource on a single relationship _without_
+> an intermediate resource are discouraged as it often leads to presenting additional
+> information about that relationship as [meta-information](meta).
+
 ##### <a href="#document-resource-object-related-resource-links" id="document-resource-object-related-resource-links" class="headerlink"></a> Related Resource Links
 
 A "related resource link" provides access to [resource objects][resource objects] [linked][links]
--- a/_format/1.2/index.md
+++ b/_format/1.2/index.md
@@ -514,11 +514,11 @@ once.
 > Note: This recommendation does _not_ prevent parallel relationships:
 > 
 > 1. A resource object may reference another resource object on different
-> relationships. For example, an `article` resource may reference the same `user`
+> relationships. For example, an `article` resource may reference the same `people`
 > resource both on its `author` and `editors` relationship.
 > 2. A resource object may reference another resource object more than once
 > on the same relationship through an intermediate resource object. For example,
-> a `team` resource may reference the same `user` resource more than once on its
+> a `team` resource may reference the same `people` resource more than once on its
 > `members` relationship _through_ a `teamMember` resource. The `teamMember`
 > resource could present additional information as fields such as `startDate`,
 > `endDate`, and `invitedBy`.


stefanmoro:
Hey. We're using the JSON:API spec for some REST APIs at TIDAL, and have discussed the topic of parallell relationships. We have weighed some pros and cons for a specific case, and we will probably go for a solution where we support parallell relationships. The **SHOULD NOT** wording seems to strike a good balance in the specification.

In addition to adding the **SHOULD NOT** clarification on to-many relationships, would it make sense to also update the wording in the [Updating to-many relationships](https://jsonapi.org/format/1.2/#crud-updating-to-many-relationships) section? Currently it reads, for POSTs:
> If a given type and id is already in the relationship, the server **MUST NOT** add it again.

Changing **MUST NOT** to **SHOULD NOT** also for CRUD operations wrt parallell relationships, would keep the recommendation not to support parallell relationships, but also clarify that it is a valid implementation.

Thanks!

jelhan:
> In addition to adding the **SHOULD NOT** clarification on to-many relationships, would it make sense to also update the wording in the [Updating to-many relationships](https://jsonapi.org/format/1.2/#crud-updating-to-many-relationships) section? Currently it reads, for POSTs:
> 
> > If a given type and id is already in the relationship, the server **MUST NOT** add it again.
> 
> Changing **MUST NOT** to **SHOULD NOT** also for CRUD operations wrt parallell relationships, would keep the recommendation not to support parallell relationships, but also clarify that it is a valid implementation.

I feel we should keep the **MUST NOT** in this case. One of the design goals of JSON:API is ensuring that a client can change the server's state without needing to know it. That eases consumers a lot. E.g. in case a request to add a relationship fails, the client can send the request again without needing to check before if the server has processed the request or not.

Maybe we should add a similar clarification that a `DELETE` request for a specific member must delete all references to that member on the relationship. And not only a subset of it if the resource is referenced multiple times.

I get that this requires using full replacement to manage parallel relationships. But in my perspective that's only one of the many drawbacks you get when using that discouraged pattern. The benefits for the recommended usage of relationships strongly outweigh that trade-off in my perspective.

Please note that the primarily reason for using **SHOULD NOT** instead of **MUST NOT** for parallel relationships is not introducing breaking changes. So far I have not seen any use case, which is better solved by parallel relationships than intermediate resources.

stefanmoro:
Thank you for the swift and clear reply @jelhan, much appreciated.

First of all, I do agree that from a modelling perspective, using intermediate resources most likely represents the best solution. The reason we are leaning towards supporting parallell relationships are strictly practical. We are in essence retrofitting JSON:API on top of a previously existing system/API. 

The way our data is currently structured, and the way we handle includes, would mean either quite a bit of work for us, or quite a compromise on performance and flexibility for clients, if we relied on intermediate resources. Also, each resource in the relationship will have a unique id in `meta`, making sure we have a way to uniquely address each relationship in the relationship object.

A few follow ups, just to make sure I understand correctly:

1. Does using **SHOULD NOT** wording in general for to-many relationships, and **MUST NOT** wording for mutation operations for to-many relationships, imply that - strictly speaking - it is valid JSON:API for a server to expose a read-only to-many relationship containing parallell resources, but not a mutable one?

2. Refs: 
> One of the design goals of JSON:API is ensuring that a client can change the server's state without needing to know it.

Got it, makes sense. What is your take on how compatible this goal is with the note in [section 7.2.2.4](https://jsonapi.org/format/1.2/#document-resource-object-linkage), essentially saying that an implementation may choose to treat a to-many relationship as ordered (something we are also doing for several to-many relationships!). In essence, if an implementation chooses to treat a to-many relationship as ordered, doesn't the ordering of the to-many relationship represent a state that a client may need to relate to before mutating the relationship?

Thank you.

jelhan:
> First of all, I do agree that from a modelling perspective, using intermediate resources most likely represents the best solution. The reason we are leaning towards supporting parallell relationships are strictly practical. We are in essence retrofitting JSON:API on top of a previously existing system/API.
> 
> The way our data is currently structured, and the way we handle includes, would mean either quite a bit of work for us, or quite a compromise on performance and flexibility for clients, if we relied on intermediate resources. Also, each resource in the relationship will have a unique id in `meta`, making sure we have a way to uniquely address each relationship in the relationship object.

I struggle imagining such a scenario. Would be great if you could elaborate. Maybe I'm missing relevant use cases for parallel relationships.

I assume it's just a matter of the serialization if representing it as an intermediate resource or not. If I get you right, you already have a unique identifier for each reference to a resource in that parallel relationship. In that case you already have an `id` for the intermediate resource. The `type` could be derived from the types of the related resources and the names of the relationship.

I expect you have a serialization such as the following today:

```Json
{
  "data": {
    "type": "teams",
    "id": "1",
    "relationships": {
      "members": {
        "data": [
          {
            "type": "teams_members",
            "id": "23",
            "meta": {
              "id": "101"
            }
          },
         {
            "type": "users",
            "id": "23",
            "meta": {
              "id": "102"
            }
          }
        ]
      }
    }
  }
}
```

What additional challenges due you face serializing it as the following:

```json
{
  "data": {
    "type": "teams",
    "id": "1",
    "relationships": {
      "members": {
        "data": [
          {
            "type": "teamsMembers",
            "id": "101"
          },
         {
            "type": "teamsMembers",
            "id": "102"
          }
        ]
      }
    }
  },
  "included": [
    {
      "type": "teamsMembers",
      "id": "101",
      "relationships": {
        "user": {
          "type": "users",
          "id": "23"
        }
      }
    },
    {
      "type": "teamsMembers",
      "id": "102",
      "relationships": {
        "user": {
          "type": "users",
          "id": "23"
        }
      }
    }
  ]
}
```

Unless I missed something, I have just moved existing information around and made up a `type` for the intermediate resource. I know it is more complex if a unique identifier for the reference in the relationship does not exist yet. But that doesn't seem to be the case for you.

> 1. Does using **SHOULD NOT** wording in general for to-many relationships, and **MUST NOT** wording for mutation operations for to-many relationships, imply that - strictly speaking - it is valid JSON:API for a server to expose a read-only to-many relationship containing parallell resources, but not a mutable one?

No. Mutation of parallel relationships are possible. But with the limitation that full replacement must be used in some cases:

- Adding more than one reference to a resource to the same relationship requires full replacement. `POST` to a relationship link does not support adding a second (or more) references to the same resource to a relationship. 
- Removing only some references to a resource on the same relationship requires full replacement. `DELETE` to a relationship link will delete _all_ references to the resources referenced by the provided resource identifier objects.

> > One of the design goals of JSON:API is ensuring that a client can change the server's state without needing to know it.
> 
> Got it, makes sense. What is your take on how compatible this goal is with the note in [section 7.2.2.4](https://jsonapi.org/format/1.2/#document-resource-object-linkage), essentially saying that an implementation may choose to treat a to-many relationship as ordered (something we are also doing for several to-many relationships!). In essence, if an implementation chooses to treat a to-many relationship as ordered, doesn't the ordering of the to-many relationship represent a state that a client may need to relate to before mutating the relationship?

The JSON:API specification does not define how a client can change the order of an ordered to-many relationship. If sending the same request multiple times could lead to unexpected results depends on the specific implementation. I expect that most implementations does not have such a problem. E.g. if the order is mutated by a full-replacement of the relationship, the client describes the desired state in the request independently of the server state. The result is the same regardless if the request is successfully processed once or multiple times. The same is true if the implementation provides some way for a client to specify adding a reference in a specific position of a relationship. It's only an issue if an implementation allows mutation depending on current state (e.g. "move one up/down").

stefanmoro:
> Unless I missed something

You didn't miss anything, and your example illustrates exactly how we would model it using intermediate resources. 

As I mentioned, the reasons we're leaning towards not using intermediate resources are strictly practical and a result of some design decisions/trade offs, and our current implementation. One concrete example is our logic that generates compound documents (`included` section). It currently doesn't support inclusion of nested paths. So, either we would have to rebuild our inclusion logic to support nested paths and generate full linkage (which, with the current setup we have, would introduce a performance decrease), or clients would have to make two consecutive calls to our backend in order to retrieve the same information. Basically, 1/ GET /teams/<id>?include=members, and 2/ GET /teamMembers/<id>?include=user.

> No. Mutation of parallel relationships are possible. But with the limitation that ...

Thanks

> The JSON:API specification does not define how a client can change the order of an ordered to-many relationship. If sending the same request multiple times could lead to unexpected results depends on the specific implementation. I expect that most implementations does not have such a problem. E.g. if the order is mutated by a full-replacement of the relationship, the client describes the desired state in the request independently of the server state. The result is the same regardless if the request is successfully processed once or multiple times. The same is true if the implementation provides some way for a client to specify adding a reference in a specific position of a relationship. It's only an issue if an implementation allows mutation depending on current state (e.g. "move one up/down").

Thanks, and yes - it is the "_It's only an issue if an implementation allows mutation depending on current state (e.g. "move one up/down")._" case I had in mind. "Move up/down" could be one scenario. "Insert after" could be another.

Issue #1678, opened 2022-11-01
Allow custom resource links


toedter:
Maybe I understand something wrong here, but the spec says:

> Unless otherwise noted, objects defined by this specification or any applied extensions MUST NOT contain any additional members. Client and server implementations MUST ignore non-compliant members.

Furthermore the spec says:

> The optional links member within each [resource object](https://jsonapi.org/format/1.1/#document-resource-objects) contains [links](https://jsonapi.org/format/1.1/#document-links) related to the resource.
> 
> If present, this links object MAY contain a self [link](https://jsonapi.org/format/1.1/#document-links) that identifies the resource represented by the resource object.
> 

I interpret this language so that the following JSON would NOT be compliant with the spec:

``` JSON
{
  "data": {
    "id": "4",
    "type": "movies",
    "attributes": {
      "title": "The Matrix"
    },
    "links": {
      "imdb": "https://www.imdb.com/title/tt0133093/"
    }
  }
}
```

If my interpretation is right, please allow custom links. Otherwise many REST hypermedia use cases could not be implemented in a spec-compliant way.

If my interpretation is wrong, please clarify the language of the spec by explicitly adding language to allow custom resource links.

Just a side note: If you validate the above JSON with https://www.jsonschemavalidator.net/ (and use the provided JSON:API schema), it says it is valid JSON:API.

jelhan:
Your interpretation of the specification is correct. If not stated explicitly implementations (and [profiles](https://jsonapi.org/format/#profiles)) are _not_ allowed to add additional members. This includes [`links`](https://jsonapi.org/format/#document-links) objects.

The names are reserved for potential further usage by the specification itself. If implementations would be allowed to define additional members, adding new members to the specification would be a breaking change.

However this does not prevent usage of custom links. [Extensions](https://jsonapi.org/format/#extensions) are allowed to define additional members. Including additional members of the `links` object.

freddrake:
It's also perfectly valid to have attributes whose values are URLs; URL values are not restricted to being placed in the ``links`` object.

~~~~~~ json
{
  "data": {
    "id": "4",
    "type": "movies",
    "attributes": {
      "imdb": "https://www.imdb.com/title/tt0133093/",
      "title": "The Matrix"
    }
  }
}
~~~~~~

toedter:
Thx for the replies.

@jelhan Could an extension allow any custom links, without specifying them explicitely?

[...]

PR #1691, opened 2023-04-21
Clarify intent of links member

### LINKED ISSUE #1684: Clarify intent of links member
There seems to be confusion about the intent of the `links` member. Readers of the specification assume that it is meant to hold any links. Even many some implementations have followed that misunderstanding. See #1019 and #1656 as examples. Allowing implementation-specific members in the `links` objects has been requested at many places. See #1678 and https://discuss.jsonapi.org/t/links-with-custom-relations/1942, https://discuss.jsonapi.org/t/links-object-clarification-needed/130 and as examples.

The main reason for both the misunderstandings as well as the requests to allow implementation-specific members are based on the assumption that all URLs should go into `links` member. In my understanding of the spec the `links` is only for discovering related JSON:API documents. Other URLs should be an attribute of a resource (e.g. for referencing a file attached to a resource), should go into `meta` or extension-specific members (e.g. for alternative representations of the same resource).

I feel we should communicate that intent more clearly (assuming that my understanding of the specification is correct). Especially the following introduction sentence in _Links_ chapter may be misleading:

> Where specified, a `links` member can be used to represent links.

I feel we should clarify that it is only about representing links to related JSON:API documents. Additionally we may want to give a hint how to handle other URLs / links in a note.

jelhan:
There has been many confusion about the intent of the links member. People tried using it for links not pointing to other JSON:API documents. Such as related files, related resources represented in different formats or other representations of the same resource in other formats. This PR tries to clarify the intent of `links` member and pointing readers to alternative ways solving their use case.

Open questions:

- Should we backport this to v1.1?

Closes #1684 #1678

### DIFF
--- a/_format/1.2/index.md
+++ b/_format/1.2/index.md
@@ -750,8 +750,12 @@ For example:
 
 ### <a href="#document-links" id="document-links" class="headerlink"></a> Links
 
-Where specified, a `links` member can be used to represent links. The value
-of this member **MUST** be an object (a "links object").
+Where specified, a `links` member can be used to represent links to related JSON:API
+documents. The value of this member **MUST** be an object (a "links object").
+
+> Note: The `links` member is intended to reference related JSON:API documents only.
+> Other links can be included as resource attributes, `meta` information or using
+> as members defined by an applied extension.
 
 <a href="#document-links-link" id="document-links-link"></a>
 Within this object, a link **MUST** be represented as either:
--- a/_format/1.2/index.md
+++ b/_format/1.2/index.md
@@ -755,7 +755,7 @@ documents. The value of this member **MUST** be an object (a "links object").
 
 > Note: The `links` member is intended to reference related JSON:API documents only.
 > Other links can be included as resource attributes, `meta` information or using
-> as members defined by an applied extension.
+> members defined by an applied extension.
 
 <a href="#document-links-link" id="document-links-link"></a>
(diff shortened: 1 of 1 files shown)


auvipy (commented):
looks better

lindyhopchris:
@jelhan someone raised a good point in my Laravel JSON:API repo - can `links` be modified via profiles? Assuming that's not the intent, based on what's drafted in this PR.

FYI I'm in support of this clarification - I've always found it confusing as to whether you're allowed to customise the `links` member. So it's good to get this clarified.

jelhan:
> @jelhan someone raised a good point in my Laravel JSON:API repo - can `links` be modified via profiles? Assuming that's not the intent, based on what's drafted in this PR.

No. Profiles can only define document members and processing rules that are reserved for implementors. The `links` object is not reserved for implementors. The JSON:API specification defines an infinite set of members, which are allowed within the `links` object depending on the context.

An extension can define additional members for the `links` object and define processing rules for them. As every extension-defined member, such a member must be namespaced. This change would not limit an extension to define additional members for the `links` object. But it would clarify, which extension-specific members should be added to the `links` object and for which use cases an extension should define another document member.

lindyhopchris:
Ah yes, sorry I think I meant profiles or extensions.

Agree with what you've written.

auvipy (approved):
IMHO it could be back ported in v1.1

PR #1692, opened 2023-04-21
allow relationship paths in query parameter families

### LINKED ISSUE #1689: 1.1 spec disallows e.g. filter[author.name]
The 1.1 spec says:

> A “query parameter family” is the set of all query parameters whose name starts with a “base name”, followed by zero or more instances of empty square brackets (i.e. []) or square-bracketed **legal [member names](https://jsonapi.org/format/1.1/#document-member-names)**. The family is referred to by its base name.

(Emphasis mine.)

It also says (like 1.0 did) that period `.` is a reserved character that may not be used in member names.

Finally, it reserves the `filter` query parameter family for filtering data (which is effectively what 1.0 also did).

This then (hopefully accidentally) disallows e.g. `filter[author.name]`, since `author.name` is not an allowed member name. However, that syntax is a very useful way of filtering on values in related resources (for details, see #1445). Indeed, it is even in line with recommendations for sorting:

> Note: It is recommended that dot-separated (U+002E FULL-STOP, “.”) sort fields be used to request sorting based upon relationship attributes. For example, a sort field of `author.name` could be used to request that the primary data be sorted based upon the `name` attribute of the `author` relationship.

Furthermore, it can also be considered a breaking change. The 1.0 spec was not clear about query parameters vs. families, and there was nothing in it that said that constraints were placed on the values in brackets. (An alternative interpretation is that any non-`a-z` character made it OK to use the query parameter, which would include `[` and `]`, and therefore no query parameter families could be reserved. This seems so wrong I assume that is not a correct interpretation.)

Finally: We use filters like `filter[relationshipName.fieldName]` extensively in our APIs, and there is no chance we can change that (or that the JSON:API framework [Felicity](https://github.com/cmeeren/Felicity) will enforce disallowing `.` in `[]` in query parameters). So if this is actually an intended part of the spec, this implementation and a host of APIs will be indefinitely non-compliant in this regard.

I assume and hope this is just an oversight. Feel free to clarify if not.

jelhan:
This fixes a bug introduced in v1.1 preventing the use of relationship paths in filtering. Please see #1689 for details.

TODO:

- [x] Backport to v1.1

Closes #1689

### DIFF
--- a/_format/1.2/index.md
+++ b/_format/1.2/index.md
@@ -2091,7 +2091,7 @@ family.
 
 A "query parameter family" is the set of all query parameters whose name starts
 with a "base name", followed by zero or more instances of empty square brackets
-(i.e. `[]`), square-bracketed legal [member names] or square-bracketed
+(i.e. `[]`), square-bracketed legal [member names], or square-bracketed
 dot-separated lists of legal member names. The family is referred to
 by its base name.
 
@@ -2101,9 +2101,9 @@ For example, the `filter` query parameter family includes parameters named:
 family, because `_` is not a valid [member name][member names].
 
 > Note: Dot separated lists of legal member names are intended to be used for
-> relationship paths. E.g. this allows filtering strategies using relationship
-> paths as defined for [sorting][fetching sorting] in query parameters such as
-> `GET /posts?sort=author.name&filter[author.status]=active`.
+> relationship paths. For example, this allows filtering strategies using
+> relationship paths as defined for [sorting][fetching sorting] in query parameters
+> such as `GET /posts?sort=author.name&filter[author.status]=active`.
 
 ### <a href="#extension-query-parameters" id="extension-query-parameters" class="headerlink"></a> Extension-Specific Query Parameters
 
--- a/_format/1.2/index.md
+++ b/_format/1.2/index.md
@@ -2091,13 +2091,19 @@ family.
 
 A "query parameter family" is the set of all query parameters whose name starts
 with a "base name", followed by zero or more instances of empty square brackets
-(i.e. `[]`) or square-bracketed legal [member names]. The family is referred to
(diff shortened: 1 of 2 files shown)


cmeeren:
While we're on the topic: What is actually the reason for limiting the values in `[]`? Why not just allow arbitrary values?

_format/1.2/index.md
```diff
@@ -2091,13 +2091,19 @@ family.
...
 
 For example, the `filter` query parameter family includes parameters named:
 `filter`, `filter[x]`, `filter[]`, `filter[x][]`, `filter[][]`, `filter[x][y]`,
-etc. However, `filter[_]` is not a valid parameter name in the family, because
-`_` is not a valid [member name][member names].
+`filter[x.y]`, etc. However, `filter[_]` is not a valid parameter name in the
+family, because `_` is not a valid [member name][member names].
+
+> Note: Dot separated lists of legal member names are intended to be used for
+> relationship paths. E.g. this allows filtering strategies using relationship
```
cmeeren: `E.g.` here should be replaced by `For example,`

_format/1.2/index.md
```diff
@@ -2091,13 +2091,19 @@ family.
 
 A "query parameter family" is the set of all query parameters whose name starts
 with a "base name", followed by zero or more instances of empty square brackets
-(i.e. `[]`) or square-bracketed legal [member names]. The family is referred to
+(i.e. `[]`), square-bracketed legal [member names] or square-bracketed
```
cmeeren: Based on skimming the start of the spec, it's using Oxford comma, so there should be a comma before `or`.

Faih4440:
[allow-relationship-paths-in-query-parameter-families](allow-relationship-paths-in-query-parameter-families) 

> ``

auvipy (approved):
I would love to see this back ported to v1.1

dgeb (approved):

PR #1721, opened 2023-12-01
Create devcontainer.json

kaaid:


jelhan:
Please provide an explanation why this configuration file would be needed. Closing it fit now. May delete it later as spam unless reasoning for this PR is provided.

PR #1723, opened 2023-12-06
allow profiles to define implementation-specific query parameters

### LINKED ISSUE #1718: Profiles should be allowed to define query parameters
Profiles are currently explicitly forbidden to define query parameters:

> A profile MUST NOT define any query parameters.

Implementations are explicitly allowed to define query parameters:

> Implementations MAY support custom query parameters. However, the names of these query parameters MUST come from a family whose base name is a legal member name and also contains at least one non a-z character (i.e., outside U+0061 to U+007A).
> 
> It is RECOMMENDED that a capital letter (e.g. camelCasing) be used to satisfy the above requirement.

This mismatch between capabilities of implementations and profiles conflicts with the general intent of profiles to standardize implementation semantics.

> Profiles provide a means to share a particular usage of the specification among implementations.
>
> Profiles can specify implementation semantics, but cannot alter, add to, or remove specification semantics.

Profiles should be allowed to define query parameters as long as they comply to the same rules as [implementation-specific query parameters](https://jsonapi.org/format/#query-parameters-custom).

I noticed this problem in the following discussion: https://discuss.jsonapi.org/t/dynamic-constraints-a-proposal-for-a-general-specification/1744

jelhan:
Closes #1718

### DIFF
--- a/_format/1.1/index.md
+++ b/_format/1.1/index.md
@@ -142,7 +142,7 @@ The rules for profile usage are dictated by [RFC
 A profile **MAY** define document members and processing rules that are reserved
 for implementors.
 
-A profile **MUST NOT** define any query parameters.
+A profile **MUST NOT** define any query parameters except [implementation-specific query parameters](#query-parameters-custom).
 
 A profile **MUST NOT** alter or remove processing rules that have been defined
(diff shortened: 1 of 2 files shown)


dgeb (approved):
I agree with the change here, good job!

PR #1737, opened 2024-03-27
Docs - Update examples to use lates jsonapi

lb-:
The example output that has a declared 'jsonapi' value is not updated to use the latest 1.1 version.


jelhan (approved):
Thanks a lot.

PR #1739, opened 2024-04-03
Add new link for a Python implementation of JSON:API

Centurix:
I have a new link for PyDANJA, a simple Python JSON:API implementation that uses generic types to control the output from pydantic and ultimately FastAPI. It also contains helper functions for controlling the output of FastAPIs automatic OpenAPI documentation.


jelhan (approved):
Thanks a lot!

Seems to be at a very early state. But I think we don't have any rules how major a project should be to get listed. Or when to remove it if it gets stale.

PR #1751, opened 2024-07-09
Add eygraber/jsonapi-kotlin to implementations

eygraber:


implementations/index.md
```diff
@@ -124,6 +124,10 @@ isomorphic ActiveRecord clone that issues JSON:API requests instead of SQL and i
 * [Izzy JSONAPI parser](https://github.com/undabot/izzy-json-api-android) library for deserializing and serializing resources with support both for Gson and Jackson parsers. It also has simple Retrofit integration.
 * [JsonApiX](https://github.com/infinum/kotlin-jsonapix) annotation processor library for serializing and deserializing Kotlin classes. It's easy to use - setup only requires annotating classes. Retrofit integration is included.
 
+### <a href="#client-libraries-kotlin" id="client-libraries-kotlin" class="headerlink"></a> Kotlin
```
jelhan, line 127: Other Kotlin libraries are listed under _Android_. Do you think we could move this one to that section as well?

I know the existing categorization is far from being perfect...

eygraber: The difference here is that my library is for Kotlin Multiplatform, so Android, JVM, iOS, and web are supported.

Armss9936 (approved):

Issue #1770, opened 2024-10-20
Media Type parameter spec lacks example of multiple values


Roanmh:
[We noticed](https://github.com/jokiefer/drf-json-api-atomic-operations/issues/2) a minor point of confusion around the Media Type parameter formatting for extensions/profiles:

The examples in [section 5.4.1](https://jsonapi.org/format/#extension-rules) show a space between the `;` and `ext`, but the examples in the [atomic operations extension spec](https://jsonapi.org/ext/atomic/) all exclude this space. Are both options valid?

[Section 5.4](https://jsonapi.org/format/#media-type-parameter-rules) talks about spaces in the extension, but doesn't not provide an example of the space separated list of URI (and where the quotes mentioned in the Note belong), so it is hard to interpret the correct syntax, especially for multiple values of a parameter.

I suspect that the post `;` space is optional but that a correct example of the a parameter with multiple values would be `ext="http://uri.ex/1 http://uri.ex/2".

Roanmh:
possible duplicate of #1768 ?

jelhan:
> possible duplicate of #1768 ?

Yes. But as the question came up twice, let's keep this one open and add an example in the spec. Maybe even an example in addition shown an `Accept` header listing JSON API mime type two times with a different set of profiles. A PR is welcome.

jelhan:
@Roanmh I created #1773 to add an example. Would appreciate your review if you have time.

Issue #1768, opened 2024-10-16
Accept header syntax for multiple extensions


bkoelman:
What is the syntax for a JSON:API `Accept` HTTP header with multiple extensions? I can't find it in the spec.

Here are some possibilities for a client that wants to use both the `https://my-ext1` and `https://my-ext2` extensions:

1. `Accept: application/vnd.api+json; ext="https://my-ext1 https://my-ext2"`
2. `Accept: application/vnd.api+json; ext="https://my-ext1, https://my-ext2"`
3. `Accept: application/vnd.api+json; ext="https://my-ext1; https://my-ext2"`
4. `Accept: application/vnd.api+json; ext="https://my-ext1"; "https://my-ext2"`
5. `Accept: application/vnd.api+json; ext="https://my-ext1"; ext="https://my-ext2"`

jelhan:
This is defined by HTTP specification.

> Each parameter is usually delimited by an immediately preceding semicolon.

See [RFC 9110: 5.6.6. Parameters](https://httpwg.org/specs/rfc9110.html#rfc.section.5.6.6) for more details.

The correct representation would be:

```
Accept: application/vnd.api+json;ext="https://my-ext1";ext="https://my-ext2"
```

bkoelman:
Thanks!

bkoelman:
The [JSON:API spec](https://jsonapi.org/format/#media-type-parameter-rules) leads me to believe it's 1 instead of 5:
> The values of the `ext` and `profile` parameters **MUST** equal a space-separated (U+0020 SPACE, “ “) list of extension or profile URIs, respectively.

@jelhan Do you agree this overrules the RFC?

jelhan:
> The [JSON:API spec](https://jsonapi.org/format/#media-type-parameter-rules) leads me to believe it's 1 instead of 5:
> 
> > The values of the `ext` and `profile` parameters **MUST** equal a space-separated (U+0020 SPACE, “ “) list of extension or profile URIs, respectively.
> 
> @jelhan Do you agree this overrules the RFC?

Sorry. You are right. The `ext` parameter should only exist once. Listing all extensions as a space separated list. The HTTP RFC is only relevant if having multiple parameters (`ext` and `profile`).

PR #1771, opened 2024-10-22
media type parameters should be delimited by an immediately preceding semicolony

jelhan:
Ensures that examples given in the specification are inline with HTTP specification:

> Each parameter is usually delimited by an immediately preceding semicolon.

Source: [RFC 9110 - HTTP Semantics: 5.6.6. Parameters](https://httpwg.org/specs/rfc9110.html#rfc.section.5.6.6) [Parameters](https://httpwg.org/specs/rfc9110.html#parameter)

This has been caught in #1768 and #1770.


dgeb (approved):
@jelhan I believe you're correct here. Well done.

Issue #246, opened 2014-07-15
How should we handle file uploads?


ghost:
Part of the API that I'm working on allows for files to be uploaded. Is this something that the spec could / should / does already cover?

steveklabnik:
I don't think there's anything that makes file uploads particularly special. If you can think of some reason why there's something missing, we can talk about it, but we'd need more specifics.

Fivell:
@steveklabnik , media type should be different or not in this case?

NuckChorris:
@steveklabnik What makes file uploads special?

Let's assume you have a profile with an `avatar` field.  In the JSON API it's exposed as a URL in a text field.  You _ideally_ want a symmetry where the client can set that field to a binary blob and they'll get back the new URL.  Some APIs (Stripe, for instance) use an out-of-band upload system, but this brings with it a whole slew of timing and cleanup issues, and removes the symmetry.

This means we basically have three options: we can either submit them as base64 in a JSON body (and eat the 25% overhead plus additional processing), you can submit them as `multipart/form-data` and ditch the whole JSON-API thing, or you can set them up as a sub-resource which can be edited by uploading to it, like `PUT /users/17/avatar` (which means you have to handle it separate from your standard download-modify-upload lifecycle)

I think this qualifies as complex enough and encourages bikeshedding in teams enough that JSON-API ought to tackle it, even if only in an extension.

patrykorwat:
Note that 25% is only about the amount of sent data over the network. Current implementations of web servers (at least in Java) for `multipart/form-data` uses algorithms which just stream data using a single buffer (O(1) memory complexity), whereas in case of JSON files it is required to process whole file in memory (O(n) memory complexity).

NuckChorris:
@meshuga The 25% is for base64 embedded in JSON, not multipart.  Multipart should send it over the wire as plain binary I believe, with just the added overhead of the headers and dividers

You're absolutely right that Multipart is O(1) memory and `data:` is O(n), and this just goes to prove my point: file uploads are friggin confusing and should be within the scope of JSONAPI or an extension

johnnncodes:
+1. Would be great if jsonapi spec would cover this.

ersinakinci:
+1.

gponsu:
+1

saneshark:
+1

shannontan-addepar:
+1

wendelin:
+1

nickreynke:
+1

[...]

PR #1772, opened 2024-10-31
Clarify combination of JSON:API with other standars

jelhan:
JSON:API can and should be combined with other standards. But there has been some confusion about it in the past. E.g. people tried using it for authentication concerns such as login instead of established standards such as OAuth 2.0. Others assumed an API implementing JSON:API cannot support file upload. There seems to be a need clarifying it in the specification to avoid such confusion going forward.

See https://github.com/json-api/json-api/issues/246#issuecomment-2449156935 for recent discussions regarding this.

### DIFF
--- a/_format/1.2/index.md
+++ b/_format/1.2/index.md
@@ -13,6 +13,8 @@ JSON:API is designed to minimize both the number of requests and the amount of
 data transmitted between clients and servers. This efficiency is achieved
 without compromising readability, flexibility, or discoverability.
 
+JSON:API can and should be combined with additional standards addressing other needs of HTTP APIs such as [OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc6749) for authorization and [Open API](https://spec.openapis.org/oas/latest.html) for documentation. A server supporting JSON:API can support other API standards such as [OData](https://www.odata.org/) or [GraphQL](https://graphql.org/) on the same or other endpoints.
+
 JSON:API requires use of the JSON:API media type
 ([`application/vnd.api+json`](http://www.iana.org/assignments/media-types/application/vnd.api+json))
 for exchanging data.
--- a/_format/1.2/index.md
+++ b/_format/1.2/index.md
@@ -17,7 +17,13 @@ JSON:API requires use of the JSON:API media type
 ([`application/vnd.api+json`](http://www.iana.org/assignments/media-types/application/vnd.api+json))
 for exchanging data.
 
-> For additional needs of HTTP APIs, like authentication, documentation, file uploads, and more, JSON:API can and should be combined with additional standards such as [OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc6749) for authorization and [Open API](https://spec.openapis.org/oas/latest.html) for documentation. It is allowed for a server supporting JSON:API to also support other API standards on the same or other endpoints, such as [OData](https://www.odata.org/) or [GraphQL](https://graphql.org/).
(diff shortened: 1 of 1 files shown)


_format/1.2/index.md
```diff
@@ -13,6 +13,8 @@ JSON:API is designed to minimize both the number of requests and the amount of
 data transmitted between clients and servers. This efficiency is achieved
 without compromising readability, flexibility, or discoverability.
 
+JSON:API can and should be combined with additional standards addressing other needs of HTTP APIs such as [OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc6749) for authorization and [Open API](https://spec.openapis.org/oas/latest.html) for documentation. A server supporting JSON:API can support other API standards such as [OData](https://www.odata.org/) or [GraphQL](https://graphql.org/) on the same or other endpoints.
```
lode: One suggestion by putting the needs first and the available standards second. This makes it more clear what needs could be, also when a need doesn't have a clear standard (e.g. file uploads).

```suggestion
For additional needs of HTTP APIs, like authentication, documentation, file uploads, and more, JSON:API can and should be combined with additional standards such as [OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc6749) for authorization and [Open API](https://spec.openapis.org/oas/latest.html) for documentation. It is allowed for a server supporting JSON:API to also support other API standards on the same or other endpoints, such as [OData](https://www.odata.org/) or [GraphQL](https://graphql.org/).
```

lode (approved):
This would be nice! Added one suggestion.

_format/1.2/index.md
jelhan: Thanks a lot. That's a good improvement.

_format/1.2/index.md
```diff
@@ -13,6 +13,8 @@ JSON:API is designed to minimize both the number of requests and the amount of
 data transmitted between clients and servers. This efficiency is achieved
 without compromising readability, flexibility, or discoverability.
 
+For additional needs of HTTP APIs, like authentication, documentation, file uploads, and more, JSON:API can and should be combined with additional standards such as [OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc6749) for authorization and [Open API](https://spec.openapis.org/oas/latest.html) for documentation. It is allowed for a server supporting JSON:API to also support other API standards on the same or other endpoints, such as [OData](https://www.odata.org/) or [GraphQL](https://graphql.org/).
```
dgeb:
```suggestion
JSON:API requires use of the JSON:API media type
 ([`application/vnd.api+json`](http://www.iana.org/assignments/media-types/application/vnd.api+json))
 for exchanging data.

> For additional needs of HTTP APIs, like authentication, documentation, file uploads, and more, JSON:API can and should be combined with additional standards such as [OAuth 2.0](https://datatracker.ietf.org/doc/html/rfc6749) for authorization and [Open API](https://spec.openapis.org/oas/latest.html) for documentation. It is allowed for a server supporting JSON:API to also support other API standards on the same or other endpoints, such as [OData](https://www.odata.org/) or [GraphQL](https://graphql.org/).
```

Although this is quite useful, it is more guidance than anything else. I would favor moving it to the end and preceding it with a blockquote (`>`).

dgeb (changes requested):

_format/1.2/index.md
jelhan: @dgeb Makes sense. I incorporated this feedback.

dgeb (approved):

PR #1773, opened 2024-11-07
add example for multiple extensions applied

### LINKED ISSUE #1770: Media Type parameter spec lacks example of multiple values

jelhan:
The question how a media type looks like if multiple extensions or profiles have been applied came up twice (#1770, #1768). It was already well defined in the specification:

> The values of the `ext` and `profile` parameters **MUST** equal a space-separated (U+0020 SPACE, " ") list of extension or profile URIs, respectively.

This PR adds an additional example similar to the examples we have in other cases for additional clarity.

Closes #1770

### DIFF
--- a/_format/1.2/index.md
+++ b/_format/1.2/index.md
@@ -93,6 +93,13 @@ values of the `ext` and `profile` parameters **MUST** equal a space-separated
 > specification requires that parameter values be surrounded by quotation marks
 > (U+0022 QUOTATION MARK, "\"").
 
+In the following example, two extensions and one profile have been applied:
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/vnd.api+json;ext="https://example.com/version https://example.com/bulk";profile="https://example.com/resource-timestamps"
+```
+
 #### <a href="#extension-rules" id="extension-rules" class="headerlink"></a> Rules for Extensions
 
 An extension **MAY** impose additional processing rules or further restrictions


_format/1.2/index.md
```diff
@@ -93,6 +93,13 @@ values of the `ext` and `profile` parameters **MUST** equal a space-separated
 > specification requires that parameter values be surrounded by quotation marks
 > (U+0022 QUOTATION MARK, "\"").
 
+In the following example, two extensions and one profile have been applied:
+
+```http
+HTTP/1.1 200 OK
+Content-Type: application/vnd.api+json;ext="https://example.com/version https://example.com/bulk";profile="https://example.com/resource-timestamps"
+```
```
jelhan, line 101: I wasn't sure if the example should go here or a little bit later into the _Content Negotiation_ section. The specification addresses the topic at two places. Just before this new paragraph it defines that the value of the media type parameters is a space-separated list. But how and when the media type is used, is defined later in the section on content negotiation.

I'm not opinionated at all where we put this example.

Roanmh:
Looks great, thank you!

dgeb (approved):
Since this is titled "Rules for Media Type Parameters" I think this is a fine placement of this section.

PR #1804, opened 2025-11-05
Add TypeScript client lib @bjornharrtell/json-api implementation link

bjornharrtell:


bjornharrtell:
Might be rather unknown still, but it's being used with success at a few projects I'm connected to and I've today improved the project with full integration tests including the atomic operations support.

PR #1806, opened 2025-11-20
Add Node.js CI workflow for testing and building

der6774:
This workflow installs Node.js dependencies, builds the project, and runs tests across multiple Node.js versions.


der6774:
Node.js Cl workflow for testing and building

der6774 (commented):
gh pr checkout 1806

auvipy (changes requested):
why we need this? I can see already an existing one here https://github.com/json-api/json-api/blob/gh-pages/.github/workflows/node.yml

jelhan:
> why we need this? I can see already an existing one here https://github.com/json-api/json-api/blob/gh-pages/.github/workflows/node.yml

This is spam.

